PR #55, opened 2013-01-20
Fix mouse pointer sprites

bartw72:
The mouse pointer sprites tend to be not sharp and have loose pixels around the sprite. This was also an issue with some vnc (I believe tightvnc). I've changed the implementation to have a configuration switch to use a set of common X cursors which have a fix-up. The cursor is calculated from the used mask which seems to be unique enough to determine the right cursor. I've also made some minor performance optimizations based on not setting the mask if it is the already known value (0) and use some bit shifts instead of calculated offsets.


metalefty:
More explanation needed.  Please reopen with new patch and explanation.

PR #63, opened 2013-02-06
Disable glyph cache

LawrenceK:
With the current xrdp head the display is getting corrupted when scrolling long lines, I suspect the glyph cache is not being updated on the client machine properly. This disables the glyph cache capability when using the freerdp1 plugin. This is not the long term solution as it has a performamce impact.

### DIFF
--- a/freerdp1/xrdp-freerdp.c
+++ b/freerdp1/xrdp-freerdp.c
@@ -1332,8 +1332,10 @@ lfreerdp_pre_connect(freerdp *instance)
 
     // TODO
     //instance->settings->glyph_cache = true;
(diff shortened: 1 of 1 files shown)


itamarjp:
Can you fix the conflicts against the current code ? so we can test / merge it

LawrenceK:
I no longer work for Osirium and do not have access at present to the repo that the patch sits in. I shall contact them and see what the state of play is. It may just have to be closed/not merged. The problem may of gone away if work has been done the glyph cache arena.

Lawrence

jsorg71:
I've been working with Osirium.
Let's close this and we'll open new PRs.

PR #121, opened 2014-04-17
make login screen window title configurable

cocoon:
please review. 

tested: I cloned this fresh modified branch, compiled and run 
then no problems with configured title and with completely commented out like this:
# ls_title=xxx


fuco809:
does this option ls_title allow to dynamically show the hostname on the login window?

PR #128, opened 2014-05-16
support for preferred combo box choice added

ArvidNorr:
Users often use the same RDP module every time they connects. With this fix users can provide information about what combobox they prefer.


itamarjp:
Can you fix the conflicts against the current code ? so we can test / merge it

proski:
The idea of the patch is not good, in my opinion. Why add another convention? Why use numeric indices for the combo choice?

If the client sends a domain name that matches a combobox entry, but the autologin is not possible (wrong or missing password), it would be better just to select that option.

metalefty:
As proski says, the idea of the patch is not good.

PR #142, opened 2014-06-02
Make xrdp init.d work with rhel

bently0602:
Changed init.d script to work with redhat/centos/etc. Tested with centos 6.5. Need to change the compiling with Centos page in the wiki.

### DIFF
--- a/instfiles/init.d/xrdp
+++ b/instfiles/init.d/xrdp
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/sh
 #
 # start/stop xrdp and sesman daemons
 #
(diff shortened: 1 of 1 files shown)


itamarjp:
the options are 

1 - keep initscript into debian.
2 - include debian initscript into xrdp/instfiles/debian.init or something without replacing or changing redhat initscript

I think number 2 is the best option.

moobyfr:
I've tried the current git with a debian jessie, there are some redhat-ish options in the support for systemd: use of /etc/sysconfig
moreover, xrdp is started with --nodaemon which makes it killed after 30s ..

itamarjp:
moobyfr, please send a patch to fix that, closing this request for now.

moobyfr:
the pull request exist:
https://github.com/neutrinolabs/xrdp/pull/267

itamarjp:
merged

PR #215, opened 2015-02-18
sesman: fix so shadow accounts aren't incorrectly classified "inactive"

### LINKED ISSUE #214: xrdp-sesman incorrectly classifies some shadow accounts as "invalid"
As reported by Alex in this [Slackware thread](http://www.linuxquestions.org/questions/slackware-14/xrdp-account-disabled-4175533819/), when determining if a shadow account is disabled because more than `sp_max+sp_inact` days have passed since the last password change, we must exclude cases where the `sp_max` or `sp_inact` fields are empty (value of `-1`) or when `sp_lstchg` equals 19700101 (value of `0`).

Otherwise, valid accounts can get incorrectly classified as *inactive*.

mancha1:
fixes #214

### DIFF
--- a/sesman/verify_user.c
+++ b/sesman/verify_user.c
@@ -321,7 +321,10 @@ auth_account_disabled(struct spwd *stp)
         return 1;
     }
 
-    if (today >= (stp->sp_lstchg + stp->sp_max + stp->sp_inact))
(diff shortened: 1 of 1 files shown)


jsorg71:
Thanks you

mancha1:
Welcome.

PR #235, opened 2015-04-17
instfiles: init.d/xrdp shouldn't be installed on systemd system

metalefty:
On systemd system, at least CentOS 7, `service foo start` command
works as wrapper of `systemctl start foo`.  However, xrdp installs
init script into /etc/init.d/xrdp.  This script is not necessary
for systemd system, and, what is worse,  if init.d/xrdp exists,
`service xrdp start` works as wrapper to run init.d/xrdp. Maybe
this is inconvinient for many users.

### DIFF
--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -4,7 +4,6 @@ xrdp.service
 
 SUBDIRS = \
   pam.d \
-  default \
   pulse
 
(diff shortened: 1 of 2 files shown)


jianhuiz:
it seems `default/xrdp` is not needed for systemd as well

metalefty:
I've pushed it. Thanks.

PR #237, opened 2015-04-19
finish for supporting swiss german keyboard

spali:
adds support for swiss german out of the box with the already existing km file.
would be great to merge it into v0.8 and for sure master and/or devel branch.


jsorg71:
Anything more complicated that this should go into devel first for a little testing.
Thanks.

PR #252, opened 2015-06-21
instfiles: merge rc script from FreeBSD

metalefty:
- merge rc script from FreeBSD
- modify Makefile.am to detect OS and install suitable files

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,27 @@ AC_PROG_CC
 AC_C_CONST
 AC_PROG_LIBTOOL
 PKG_PROG_PKG_CONFIG
+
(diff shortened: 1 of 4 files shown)


metalefty:
That's all for merging FreeBSD local patches that I already have.  It is almost working on FreeBSD without any local patches, with TLS enabled.  Hereafter, I'll test each features such as drive redirection, IPv6 support,  etc,

I have to make some modification to `instfiles/Makefiles.am` to support other operating system.  I asked mailing list how many people want to use xrdp on the OS except Linux and FreeBSD.  At this time,  I care about only Linux and FreeBSD.

Issue #196, opened 2014-12-01
Connection fails on first attempt, works on second


jmkekala:
Hi,

I am using Centos 7 with Mate Desktop and I have a similar problem as described here:
https://bugzilla.redhat.com/show_bug.cgi?id=557510

To solve it I added small delay (g_sleep(1000);) to vnc/vnc.c file to lib_mod_connect -function and it works now.

Dunno is this a right way to fix the bug or submit this way but anyways, this one liner helps.

Issue #182, opened 2014-11-05
Xvnc session doesn't launch properly every other time


anthonyagresta:
I am using the Windows mstsc.exe RDP client on Windows 7.
When I try to connect to my Debian box running xrdp using the default "sesman-Xvnc" option, it doesn't seem to start Xvnc properly the first time, or it maybe doesn't wait long enough for the VNC session to start. I get the error pictured here _every other time_ when I try to connect:

![vnc_error_screencap](https://cloud.githubusercontent.com/assets/8355047/4920993/e4c44948-6506-11e4-827f-e42c2f83d771.png)

The first time, it doesn't seem to have the VNC session open in time for xrdp to use it, but the second time I connect using sesman-Xvnc, it seems to reconnect just fine to the existing Xvnc session. Once it connects the second time, everything works great. 
Then, if I disconnect cleanly and go to reconnect, it fails once and then works again.

OS/configuration information: 
Debian 7, i386. (WM/DE customized, running Openbox and lightdm)
xrdp version 0.9.0, built using X11RDP-o-matic against the "master" branch.
Xvnc provided by tightvncserver 1.3.9-6.4 from the Debian repos. (I have also tried TigerVNC 1.3.1 to provide Xvnc, with the same issue)

sesman.ini is default, xrdp.ini has been modified slightly:

```
max_bpp=24
crypt_level=fips
```

Full configs and logfiles here: http://pastebin.com/5N6xggMV

The default sesman-Xvnc worked okay on 0.6.0, but I need the FIPS encryption option that's been added in more recent versions.

Please let me know if any other information would help.

Thanks!

PR #263, opened 2015-07-15
fix timeout when using Xvnc

itamarjp:
Hello, 

for some reason when using Xvnc, its timeout, please take a look in this screenshoot.

https://itamarjp.fedorapeople.org/tmp/Xvnc-timeout.png

alot of people complained about it in redhat bugzilla

After this patch the error stopped, but I don't know if this is the best solution.

what do you think about ?

do you have a better solution ? how about setting up a socket timeout in g_tcp_connect ?

### DIFF
--- a/instfiles/xrdp-sesman.service
+++ b/instfiles/xrdp-sesman.service
@@ -2,6 +2,7 @@
 Description=xrdp session manager
 After=syslog.target network.target
 StopWhenUnneeded=true
+BindTo=xrdp.service
 
 [Service]
 Type=forking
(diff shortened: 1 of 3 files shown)


metalefty:
I think just increasing the default value of `delay_ms` in xrdp.ini can avoid the issue.

neutrinolabs/xrdp#182
neutrinolabs/xrdp#196

itamarjp:
metalefty, this patch looks better than adjusting delay_ms because it will make it works automatically.

metalefty:
:+1:

metalefty:
How would you like to think about automatic delay adjustment again?

itamarjp:
@jsorg71 what do you think about automatic delay ?

metalefty:
This includes other changes. I created #457.

PR #273, opened 2015-08-19
fix wrong handle of commented section in configuration file.

gschwind:
commented section are read as regular section. for instance:
#[section_name]
is read as :
[section_name]

this patch fix this issu and also :
- add few comments
- improve readability

Best regards


jsorg71:
+1

itamarjp:
@jsorg71 if looks good please merge

Issue #232, opened 2015-03-18
Licence issues after legal review for packaging


mirabilos:
Hi,

I’m in the process of preparing experimental packaging of your “devel” branch for Debian (for our organisation’s own use at first, though may be shared with the Debian project itself, and thus goes by the same guidelines), and have found a few licence issues (all relative to commit 5788133c4d119b9a57f8cc77dd22c3e73600bc62):

• several files have no explicit copyright and author notice; I assume they come under the “umbrella” Apache 2 licence from the top-level COPYING file (please confirm)

• some manual pages carry only the author information in the descriptive text but no copyright notice; I assume the same for all of them; authors involved: Simone Fedele, Vincent Bernat (please confirm)

• in several places, you use a licence that starts off *almost* like the Historical Permission Notice and Disclaimer (except it adds an explicit permission to sell, repositions the “without fee” words, and has a redundant reproduction requirement verbatim from the MIT licence and uses the MIT disclaimer except disclaiming liablility by “The Open Group”); this is not a problem, but you *may* with to change “The Open Group” in the disclaimer to “THE AUTHORS OR COPYRIGHT HOLDERS” like in [the original MIT licence](http://opensource.org/licenses/MIT), and possibly switch the main licence body to a standardised version (either [HPND](http://opensource.org/licenses/HPND) or MIT) or at least remove the redundant second paragraph (not a problem, but please consider)

• sesman/tools/sestest.c is under no explicit licence, (c) 2008 Simone Fedele ⇒ we absolutely *cannot* accept the presence of this file in the source tree, unless an explicit licence is added (please contact Simone and ask whether Apache 2 or MIT or so is OK for them)

• common/d3des.{c,h} are proprietary software under no licence, “Copyright (C) 1999 AT&T Laboratories Cambridge.  All Rights Reserved.” and “Copyright (c) 1988,1989,1990,1991,1992 by Richard Outerbridge.” ⇒ we absolutely *cannot* accept the presence of this file in the source tree (please replace by a version under a Free and OSI certified Open Source Software licence; “Public Domain” is *not* OK in most countries of the earth)

• there are several ELF binaries in the tree: tcutils/tcutils and tests/gtcp_proxy/gtcp-proxy (please “git rm” them)

[...]

PR #274, opened 2015-08-26
remove remaining proven DFSG-unfree files

mirabilos:
This should fix all “must fix” items remaining from https://github.com/neutrinolabs/xrdp/issues/232 with:

yosemite.bmp was replaced by a picture I just shot myself.

sestest.c was removed from the repository and Makefile.am – it was never installed anyway.


mirabilos:
So, I removed the sestest.c removal and added, after historic review, GPLv2+ to sestest (and sesadmin which shares its problems), since they were distributed combined with the remaining scp code under that licence.

Do note that GPLv2+ combined with Apache v2 gives GPLv3+ though.

Issue #277, opened 2015-08-30
Sync list of keymaps to xrdp_keyboard.ini


mirabilos:
Some new keymaps were added, but xrdp/xrdp_keyboard.ini was not updated with it.

So weren't the Makefiles, but I've got a patch for that already, which will be submitted as part of a larger amount of push requests upcoming soon. For now:

--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -14,10 +14,14 @@ startscript_DATA = \
   km-040c.ini \
   km-0410.ini \
   km-0411.ini \
+  km-0414.ini \
   km-0415.ini \
+  km-0416.ini \
   km-0419.ini \
   km-041d.ini \
   km-0807.ini \
+  km-080c.ini \
+  km-0813.ini \
   km-0816.ini \
   km-e0010411.ini \
   km-e0200411.ini \

itamarjp:
fixed and merged, please tell me if looks good.

https://github.com/neutrinolabs/xrdp/pull/278

please reopen if its not good enough

mirabilos:
Sorry, not enough: the xrdp_keyboard.ini must also be adjusted to list them. I don't know which entries are correct for that, though.

PR #278, opened 2015-08-30
add missing .ini files into extra_dist, so they will appear in xrdp t…

itamarjp:
add missing .ini files into extra_dist, so they will appear in xrdp tarball after a make dist

fix for https://github.com/neutrinolabs/xrdp/issues/277


blackcrack:
by the way, OpenMandriva Linux it is brawny, anyone should look for integrate xrdp as rpm into Openmandriva Linux, as well advertising for xrdp  (as an lil hint) 
best regards
Blacky

PR #280, opened 2015-09-04
add swiss french keyboard map

Dartheo:
I have added the swiss french keyboard map to instdir.


itamarjp:
don't forget to edit Makefile.am and add it into startscript_DATA, so when someone runs make dist a tarball will be created including your file.

itamarjp:
please fix Makefile.am as I have described before and I will merge your request.

metalefty:
I did. Please merge #295. It closes this, too.

PR #292, opened 2015-10-24
Conditionally Use size for session allocation policy

### LINKED ISSUE #291: Reconnect policy does not work
The current config have a "Policy" option that allows to specify what should checked to decide if a new session should spawned or an old should reused:
```
## Policy - session allocation policy                                               
# Type: enum [ "Default" | "UBD" | "UBI" | "UBC" | "UBDI" | "UBDC" ]                
# Default: Xrdp:<User,BitPerPixel> and Xvnc:<User,BitPerPixel,DisplaySize>          
# "UBD" session per <User,BitPerPixel,DisplaySize>                                  
# "UBI" session per <User,BitPerPixel,IPAddr>                                       
# "UBC" session per <User,BitPerPixel,Connection>                                   
# "UBDI" session per <User,BitPerPixel,DisplaySize,IPAddr>                          
# "UBDC" session per <User,BitPerPixel,DisplaySize,Connection>                      
```
But the code at https://github.com/neutrinolabs/xrdp/blob/devel/sesman/session.c#L167 forgot to check if the SESMAN_CFG_SESS_POLICY_D flag is set or not. So it was not possible to select a policy that does _not_ account the display size. This is a problem because most rdp clients can only resize through a reconnect.

janLo:
The DisplaySize was unconditionally used to decide if a new session
should be created or a old one should reused. This fixes this behaviour
to match the configured policy.

fixes #291

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -164,7 +164,8 @@ session_get_bydata(char *name, int width, int height, int bpp, int type, char *c
         }
 
         if (g_strncmp(name, tmp->item->name, 255) == 0 &&
-            (tmp->item->width == width && tmp->item->height == height) &&
+            (!(policy & SESMAN_CFG_SESS_POLICY_D) ||
+             (tmp->item->width == width && tmp->item->height == height)) &&
             (!(policy & SESMAN_CFG_SESS_POLICY_I) ||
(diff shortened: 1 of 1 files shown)


janLo:
Already fixed in https://github.com/janLo/xrdp/commit/12092e809840f1d0ad106f7171b197248ef4c351

PR #293, opened 2015-10-26
Fix compile errors for newer libav versions

janLo:
LibAV changed their constant names. This cause a build error, so this
redefine the old names as alias for the new ones.

### DIFF
--- a/xrdpvr/xrdpvr.c
+++ b/xrdpvr/xrdpvr.c
@@ -24,6 +24,27 @@
 #include "xrdpvr.h"
 #include "xrdpvr_internal.h"
 
+/* Support newer versions of ffmpeg and libav */
+#if ! defined(CODEC_TYPE_AUDIO)
+  #define CODEC_TYPE_AUDIO AVMEDIA_TYPE_AUDIO
+#endif
+
+#if ! defined(CODEC_TYPE_VIDEO)
+  #define CODEC_TYPE_VIDEO AVMEDIA_TYPE_VIDEO
+#endif
+
+#if ! defined(PKT_FLAG_KEY)
+  #define PKT_FLAG_KEY AV_PKT_FLAG_KEY
+#endif
+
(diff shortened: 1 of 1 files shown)


xrdpvr/xrdpvr.c
```diff
@@ -24,6 +24,27 @@
 #include "xrdpvr.h"
 #include "xrdpvr_internal.h"
 
+/* Support newer versions of ffmpeg and libav */
```
proski: That's not specific enough. Please specify versions. "Newer" today could be "stone age" 5 years from now.

xrdpvr/xrdpvr.c
```diff
@@ -24,6 +24,27 @@
...
+#if ! defined(CODEC_TYPE_AUDIO)
```
proski: xrdp appears to standardize on "!defined" without a space. Please follow the existing style.

xrdpvr/xrdpvr.c
```diff
@@ -24,6 +24,27 @@
...
+
+#if ! defined(CODEC_TYPE_VIDEO)
+  #define CODEC_TYPE_VIDEO AVMEDIA_TYPE_VIDEO
+#endif
+
+#if ! defined(PKT_FLAG_KEY)
+  #define PKT_FLAG_KEY AV_PKT_FLAG_KEY
+#endif
+
+#ifndef CODEC_ID_H264
```
proski: Please use "!defined" for consistency.

proski:
I tried to make a similar patch and noticed that AVMEDIA_TYPE_VIDEO is not even a preprocessor define. You would need to use API version to decide whether to define the backport value.

shadoxx:
What's the status of this pull request? Are there any outstanding reasons why this can't be merged into master? xrdpvr does not currently work at all from my testing. Thanks!

metalefty:
Some changes required to merge this PR but I'll change them.

PR #294, opened 2015-10-26
Simplify session match logic.

janLo:
The session match logic had two versions - one for the
SESMAN_SESSION_TYPE_XRDP and SESMAN_SESSION_TYPE_XORG sessions and one
for every other type. The only difference was, that different display
sizes where ignored when searching for sessions to reconnect if the
policy does not have the SESMAN_CFG_SESS_POLICY_D flag set and the type
is SESMAN_SESSION_TYPE_XRDP or SESMAN_SESSION_TYPE_XORG.

The reason was that xvnc cannot resize and the others can do. This two
versions where not necessary because we set the
SESMAN_CFG_SESS_POLICY_D flag every time we have a xvnc session a few
lines above. So the two branches for the different types can be reduced
to one.

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -144,27 +144,9 @@ session_get_bydata(char *name, int width, int height, int bpp, int type, char *c
             tmp->item->client_ip);
 #endif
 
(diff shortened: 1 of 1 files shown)


metalefty:
:+1: @jsorg71 can you look at this?

PR #300, opened 2015-12-13
Avoiding code duplicated and directives that break statements.

RomeroMalaquias:


RomeroMalaquias:
Hello?

jsorg71:
I'm having a hard time to see what this fixes.
Can you elaborate?

itamarjp:
please reopen this ticket when you provide more information, thank you.

RomeroMalaquias:
@jsorg71 , @itamarjp , sorry for delay.
this change don't fix anything, it's  only aesthetic, it's a suggestion to avoid the use of conditional directives that break statements i think that kind of concern improve the code maintainability do you agree?

proski:
OK, you are trying to address a very minor style issue here (preprocessor conditional cutting through `if` block), but you are creating another issue (bad variable name, it should be `is_password` or something like that, definitely not `test_access`). In the same time, you are missing bigger issues in that piece of code.

What is ACCESS anyway? It's always defined in xrdp_login_wnd.c and xrdp_mm.c. Why would anybody want to undefine it? If nobody needs to undefine it, let's just remove it.

Why are we using number 255? If that some magic number? Would `g_strcmp` do the job?

Why is the comparison case sensitive? What if the field is named "Password"? Any reason to treat it differently?

Finally, it would be great to allow internationalization of the login dialog. There should be a way to specify the label and a meaning separately. That's a big task, but I think it's worth the trouble.

proski:
Please see #632 for an alternative approach.

RomeroMalaquias:
Ok, thanks

proski:
This issue can be closed now.

PR #309, opened 2016-01-27
Fixes for autotools, compile warnings and MacOS compatibility

proski:
GOT_PREFIX is a direct inconvenience that Fedora and other distros disable by patching. Symlinks for sources are usually unnecessary, and I don't want to rely on bootstrap to adjust the source tree.


itamarjp:
I don't have tested to see if it breaks something, but appears good, merging.

PR #310, opened 2016-01-30
update japanese keyboard "Henkan_Mode" key and "Muhenkan" key

noldor:
There is a slight error on Japanese keyboard for Windows OS.


metalefty:
We're discussing this issue at xrdp user's group of Japan mailing list.
https://groups.google.com/forum/#!msg/xrdp-jp/GoiqZm8sWg4/mDPjbBLkAgAJ

Finally, I can't argue with noldor's change. It looks good. :+1:

PR #314, opened 2016-02-08
sesman: don't set LANG, let initialization scripts take care of it

proski:
Setting LANG to en_US.UTF-8 suppresses language settings from
/etc/locale.conf

See https://bugzilla.redhat.com/show_bug.cgi?id=1290820

### DIFF
--- a/sesman/env.c
+++ b/sesman/env.c
@@ -130,7 +130,6 @@ env_set_user(char *username, char *passwd_file, int display,
             g_set_current_dir(pw_dir);
             g_sprintf(text, ":%d.0", display);
             g_setenv("DISPLAY", text, 1);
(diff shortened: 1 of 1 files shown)


metalefty:
Setting LANG to en_US.UTF-8 is added at 6ad1d7f. Perhaps it was added on purpose
however the reason of the addition is not clear because it is a big commit.

itamarjp:
I agree, it shouldn't overwrite system variables,

proski:
The idea was probably to imitate the behavior of a desktop manager (like gdm). But the difference is that desktop managers usually have UI to change the locale, and the xrdp login screen doesn't. Maybe xrdp will get that functionality one day. Until then, it should keep LANG alone.

PR #316, opened 2016-02-10
Make it possible to use Xvnc with 32 bpp

proski:
The support is not ideal, acceptable. For some clients, 32 bpp is the only option.

scp_session_set_bpp() was essentially setting bpp to 0 if 32 bpp was requested. Check the error now, log errors, don't continue with "0 bpp".

Allow 32 bpp in scp_session_set_bpp() - that's the global switch.

Allow 32 bpp in VNC, use switch statement for readability.

### DIFF
--- a/sesman/libscp/libscp_session.c
+++ b/sesman/libscp/libscp_session.c
@@ -137,6 +137,7 @@ scp_session_set_bpp(struct SCP_SESSION *s, tui8 bpp)
         case 15:
         case 16:
         case 24:
+        case 32:
             s->bpp = bpp;
         default:
             return 1;
--- a/sesman/libscp/libscp_v0.c
+++ b/sesman/libscp/libscp_v0.c
@@ -252,7 +252,14 @@ scp_v0s_accept(struct SCP_CONNECTION *c, struct SCP_SESSION **s, int skipVchk)
         scp_session_set_height(session, sz);
         /* bpp */
         in_uint16_be(c->in_s, sz);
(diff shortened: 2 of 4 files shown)


itamarjp:
@jsorg71 , please take a look

jsorg71:
Yup, this looks good.

jsorg71:
In libscp_v1s.c, it looks like scp_session_set_bpp is called twice.
This file is not used yet, but we might still want to fix it.

```
    in_uint16_be(c->in_s, cmd);
     scp_session_set_height(session, cmd);
     in_uint8(c->in_s, sz);
+    if (0 != scp_session_set_bpp(session, sz))
+    {
+        scp_session_destroy(session);
+        log_message(LOG_LEVEL_WARNING,
+                    "[v1s:%d] connection aborted: unsupported bpp: %d",
+                    __LINE__, sz);
+        return SCP_SERVER_STATE_INTERNAL_ERR;
+    }
     scp_session_set_bpp(session, sz);
     in_uint8(c->in_s, sz);
     scp_session_set_rsr(session, sz);
```

itamarjp:
I am merging it and removing the duplicated call to scp_session_set_bpp

jsorg71:
It looks like there is a missing break here.

```c
          case 15:
         case 16:
         case 24:
+        case 32:
             s->bpp = bpp;
         default:
             return 1;
```

jsorg71:
I manually fixed it.

proski:
Sorry for that, I should have tested the final result. Thank you for taking care of that.

Issue #311, opened 2016-01-30
Release v0.8.1


metalefty:
@jsorg71 
Why don't you tag current v0.8 branch with v0.8.1? Let's release v0.8.1.

As you may know, RC4 cipher used in classic RDP encryption is considered to be no longer safe.  This means xrdp v0.6.x is no longer safe when directly exposed to the internet. For this reason, I want to recommend TLS instead of classic RDP encryption.  However, there's no specific version number for full TLS support. v0.8.0 is lacking some TLS fixes.  Use v0.8 branch? master or devel branch?  No way.  It is not clear for ordinary users.  If v0.8.1 is released, I can simply say "use v0.8.1 instead of v0.6.x."

metalefty:
I said current v0.8 branch. But I have some commits to cherry-pick into v0.8 branch.

metalefty:
My thoughts of  v0.8.1 is to provide full TLS support with x11rdp for all platforms including FreeBSD.
As a result, these are candidates to be cherry-picked. I'm working on metalefty/xrdp v0.8 branch.
Please let me know your opinion.

- legal packaging #232 
- complete TLS support (also on FreeBSD)
 - including FreeBSD build fixes
 - f59c925f8
- fix portability issues (ex. hardcoded paths 870437a)
- fixes for systemd 5f16481f d4791b39 830a93d8
- fix (possible) crashes or deadlocks 
 - in fuse 8261459
 - in x11rdp shmem on FreeBSD 4a95950f
 - if DISPLAY is not set f59c925f8
- simple fixes with logic change
 - fix --kill 418637f7
 - fix -v -h 4a6567c
 - xorgxrdp xrandr 8a644c69
- other quite simple changes without logic change
 - add new keymaps including #310 93aa8ab3 04b729ccd ef20d9d?
 - update mailing list address 89c0ee0
 - adjust permission a25e5e9

itamarjp:
@proski what do you think about ?

metalefty:
Now we're working on v0.9.1 release, closing.

Issue #318, opened 2016-02-12
Add release tags please


barrydegraaff:
Would it be possible to provide release tags for new releases,

https://github.com/neutrinolabs/xrdp/releases

last tags where added in 2014. I like to build this software from source, but I do also like to use a 

   git checkout tag

So I can use a `stable` release point, right now, this is a bit difficult.

Thanks for considering, Barry

itamarjp:
theres no new versions released since 2014, because of that you're not seeing new tags, when we release a new one we will tag it.

Issue #162, opened 2014-08-10
xrdp 0.6.1 vs 0.9.0


hetzbh:
Hi,

Looking at xrdp.org - it seems the last version there is 0.6.1 while here it's 0.9.0, yet I can't find whats the differences and what's changed since 0.6.1. Could someone shed a light please?

Also, in 0.6.1 the X11rdp-sesman was using RDP 4 IIRC, has this been changed? whats better these days? Xvnc or X11rdp?

Thanks

therevoman:
I too would like to know the answer to this question.

metalefty:
0.6.1 is the previous stable release.
0.9.0 is under development, it hasn't been released yet. I call this 0.9.0beta.
0.8.0 is the latest stable release.

I personally don't recommend to use Xvnc. It is obsoleted. Please don't confuse I'm not saying VNC is obsoleted. Using Xvnc with xrdp is obsoleted.

0.9.0beta (github master branch) with X11rdp is the best, IMHO. 0.8.0 with X11rdp is the second. The reason why I recommend 0.9.0beta than 0.8.0 is (brushed-up) TLS support. Classic RDP cipher is no longer considered to be safe.

jsorg71:
Yes, this is right except that there is about to be a v0.8.1 that has all the TLS fixes.
v0.8.1 will be just like master at that point.

Soon Xvnc will be removed from xrdp unless we can get some security patches in their side.
You will still be able to proxy VNC connection but sesman will not manage them.

metalefty:
That's great. 
Do you have a plan merging Japanese Keyboard fixes that I and AkiraPenguin made v0.8.1?

Most Linux distributors only make packages from officially released tarballs. If v0.8.1 included our fixes,  more Japanese xrdp users can live a happy xrdp life :smiley:

metalefty:
I overlooked JP keyboard changes had already merged into v0.8 branch.

geiseri:
I think there is a problem that packagers do not want to build X11rdp in the current state since you need root, or an absolute file path to build it.  I think until the xorg driver is stable VNC is still the best for generic use.

metalefty:
v0.9.1 has been released.  We recommend to use v0.9.1 rather than v0.6.1. Closing.

PR #323, opened 2016-02-17
v0.8.1 release candidate

metalefty:
I described why I'm waiting for release v0.8.1 in #311.

There's some legal packaging works left. For example, removing 3des.
I'm testing my v0.8 branch on CentOS 7.  It is stable for me.

Jay, if you like this, let's release v0.8.1 based on this.


itamarjp:
in configure.ac it says 0.9.0

https://github.com/neutrinolabs/xrdp/blob/devel/configure.ac

metalefty:
Yes, devel branch is for future 0.9.0, right?
I'm talking about v0.8 branch.
https://github.com/metalefty/xrdp/blob/v0.8/configure.ac#L4

itamarjp:
I strong recommend to have only one branch and use tags when I release a new version.
for example after tagging 0.9.0 change devel to 0.9.1

metalefty:
You're correct.  Actually, it was tough for me to make this pull request.  I never want to repeat such a tough work.  The reason why I suggested v0.8.1 is Jay's comment before mentions v0.8.1 in #162 and v0.8.1 is never released for now.  Either v0.8.1 or v0.9.0 is OK for me.  Important point is to give xrdp a new release tag in order to provide full TLS support for ordinary users. 

Some people also want a new release tag these days. For example, #318, #325 just now.

I repeat what I said in #311.

> As you may know, RC4 cipher used in classic RDP encryption is considered to be no longer safe. This means xrdp v0.6.x is no longer safe when directly exposed to the internet. For this reason, I want to recommend TLS instead of classic RDP encryption. However, there's no specific version number for full TLS support. v0.8.0 is lacking some TLS fixes. Use v0.8 branch? master or devel branch? No way. It is not clear for ordinary users. If v0.8.1 is released, I can simply say "use v0.8.1 instead of v0.6.x."

There's one other thing, xrdp have neither explicit branch plan nor release cycle AFAIK.

How would you like to determine them on this opportunity? As itamarjp says, one branch is simple and easy.  +1 to this. Each new release always made from master.  And then, release cycle.  Time-based ex. quarterly, half-yearly or yearly? Or feature-based?

Thanks for your comment.

metalefty:
Let me know your opinion what prevents making a new release.
I strongly think we want a new release tag.

itamarjp:
@proski what do you think about ?

metalefty:
One more suggestion about version in devel/master branch. 

The version written in [readme.txt ](https://github.com/neutrinolabs/xrdp/blob/devel/readme.txt)and [configure.ac](https://github.com/neutrinolabs/xrdp/blob/devel/configure.ac) in devel/master branch is 0.9.0 but not actually 0.9.0 because 0.9.0 has not been released yet.  This is very very confusing. I suggest stopping this. devel/branch should not be called future release version.

After 0.9.0 is released, the version may be 0.10.0 or 1.0.0 but whichever is ok. Assuming 1.0.0 at this time. Don't bump version to 1.0.0 until it actually be released.  Instead, bump version to 0.9.90 or something like this. 1.0.0 will be called 0.9.90 until release. \*.\*.90 is reserved for under development version.


[TigerVNC/tigervnc](https://github.com/TigerVNC/tigervnc) has similar idea.  How would you like this idea?

moobyfr:
IMHO, I would prefer using several branch: the master (or devel) gets all the features. When a set of functionalities are available, create a branch (like the 0.8). Inside the branch use tags to mark releases: 0..8 0.8.1 etc... with only fixes in the 0.8 branch. but the same fixes have to be applied on the master branch too. New features only goes in the master, and when it's fully functional make another branch 0.9 and make tags inside it.
This is a method largely used in OSS projet, but it needs more work because some patchs have to be applied twice, and needs a release manager which follow changes. In the actual state where there is no human ressource to know when to make tags or release tar.gz this whould be more difficult...

Keep it simple, we need at least tags and tar.gz for the packagers, it's a pity to see that some version bundled by default in distribs are 0.5 or 0.6 , it makes a bad idea or xrdp for users which get old bugs.

jsorg71:
Currently the way it's done, I don't think it that confusing.
devel is latest branch, use at your own risk.
At stable points, devel is merged into master.
master is considered fairly stable.
Then the marked branch v0.8 or v1.0(in the future) are the stable branches.
I've been using the odd minors for development versions like v0.7 and v0.9.
Not saying this is the way it has to be.

I'm good for this merge lets get a new release out.

I'll start a new issue for discussion on the release concerns.

proski:
I'm not familiar with the state of the 0.8 branch. Besides, I haven't really touched X11rdp yet. I suggest that we work on stabilizing the devel branch and make a release based on it. Of my changes, I cannot think of any change that absolutely must be cherry-picked. OK, maybe "Fix building outside the source tree", it fixes a compile error after all. Maybe also the fix for the month display.

Regarding versioning, the even/odd convention is useful if there are stable and unstable releases. It was too much overhead even for Linux kernel. I would not mind if we abandon that convention.

I agree that configure.ac in git should have anything resembling a released version, unless it's a commit from which the release was made. That is, the version is 0.9-pre, it changes to 0.9.0 for one commit, from which the 0.9.0 release is made. The next commit changes it to 0.10-pre. The release branch v0.9 could use 0.9.1-pre, change to 0.9.1 for one commit, then go to 0.9.2-pre if needed.

Fedora packages "version" 0.9.0. It's their mistake. We should release something that looks newer than 0.9.0 and prevent such mistakes in the future by improving our release process.

Let's not go to v1.0 yet, we can do 0.10, 0.11 etc. Releasing 1.0 is an important message that the project can only send once in its lifetime. v1.0 should compile with all features without root access and "x-o-matic" scripts. It should compile without warnings and it should run without warnings and memory leaks in Valgrind. It should also have nice clear documentation. The point is, users who try it should be encouraged to join the project, not be put off by poor code and documentation.

metalefty:
@proski

> configure.ac in git should have anything resembling a released version, unless it's a commit from which the release was made

That is what I wanted to say. Thank you for summarizing.

Just FYI. In FreeBSD case, not to make mistakes like Fedora's 0.9.0, packages of development snapshot has version like 0.9.0.bYYYYMMDD at my discretion. "b" stands for beta.  This ensures upcoming 0.9.0 is calculated greater than 0.9.0.bYYYYMMDD.
```
$ pkg version --test-version 0.9.0 0.9.0.b20160302
>
```
This result means 0.9.0 > 0.9.0.b20160302.

> Let's not go to v1.0 yet,

+1

@jsorg71 
> I'm good for this merge lets get a new release out.

Did you mean 0.8.1? I no longer want this to be merged.  I agree with all proski said now.

metalefty:
@jsorg71 
> I'll start a new issue for discussion on the release concerns.
how's it going?

PR #332, opened 2016-02-23
common: add log for g_tcp_connect

metalefty:
in case getaddrinfo(3) might fail.

In FreeBSD, AI_V4MAPPED support for getaddrinfo(3) was very recently
implemented[1].  Most of FreeBSD systems in the world do not have
this implementation yet.  This will be a problem when AI_V4MAPPED
isn't supported and xrdp is built with IPv6 option.  In such a case,
g_tcp_connect always fails.

Of course getaddrinfo(3) might fail in other cases.  The log helps
us to know what's happening.

[1] https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=198092

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -722,31 +722,31 @@ g_tcp_connect(int sck, const char *address, const char *port)
 int APP_CC
 g_tcp_connect(int sck, const char* address, const char* port)
 {
-  struct sockaddr_in s;
-  struct hostent* h;
+    struct sockaddr_in s;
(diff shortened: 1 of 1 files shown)


metalefty:
@itamarjp @jsorg71 Can you look at this? Just adding logs.

jsorg71:
+1

metalefty:
I will make another pull request for V4MAPPED issue on FreeBSD. Whenever you can merge this.

PR #333, opened 2016-02-24
Key install

proski:
The big thing - installing rsakeys.ini by "make install". Not need to remember to do that for manual install. No need to rely on the init scripts.

More Automake cleanups. Using standard Automake features for installing scripts and making data files distributable.

### DIFF
--- a/instfiles/init.d/xrdp
+++ b/instfiles/init.d/xrdp
@@ -23,7 +23,6 @@ SESMAN_START=yes
 #USERID=xrdp
 # the X11rdp backend only works as root at the moment - GH 20/03/2013
 USERID=root
-RSAKEYS=/etc/xrdp/rsakeys.ini
 NAME=xrdp
 DESC="Remote Desktop Protocol server"
 
@@ -67,18 +66,6 @@ if [ "$(id -u)" = "0" ]; then
         mkdir $PIDDIR
     fi
     chown $USERID:$USERID $PIDDIR
-
-    # Check for rsa key 
-    if [ ! -f $RSAKEYS ] ; then
-        log_action_begin_msg "Generating xrdp RSA keys..."
-        (umask 077 ; xrdp-keygen xrdp $RSAKEYS)
-        chown $USERID:$USERID $RSAKEYS
-        if [ ! -f $RSAKEYS ] ; then
-            log_action_end_msg 1 "could not create $RSAKEYS"
-            exit 1
-        fi
-        log_action_end_msg 0 "done"
-    fi
 fi
 
 
--- a/instfiles/rc.d/xrdp
+++ b/instfiles/rc.d/xrdp
@@ -61,10 +61,6 @@ xrdp_cmd() {
     if [ "${rc_arg}" = "stop" ] ; then
 	xrdp_daemons=$(reverse_list ${xrdp_daemons})
     fi
-    # Generate rsakeys.ini on start
-    if [ "${rc_arg}" = "start" -a ! -f %%PREFIX%%/etc/xrdp/rsakeys.ini ] ; then
-        %%PREFIX%%/bin/xrdp-keygen xrdp %%PREFIX%%/etc/xrdp/rsakeys.ini
-    fi
 
     # Apply to all the daemons.
     for name in ${xrdp_daemons}; do
--- a/keygen/Makefile.am
+++ b/keygen/Makefile.am
@@ -13,3 +13,12 @@ xrdp_keygen_SOURCES = keygen.c
 
 xrdp_keygen_LDADD = \
   $(top_builddir)/common/libcommon.la
+
+xrdpsysconfdir = $(sysconfdir)/xrdp
+
+install-data-hook:
+	umask 077 && \
+	  ./xrdp-keygen xrdp $(DESTDIR)$(xrdpsysconfdir)/rsakeys.ini
+
+uninstall-hook:
+	rm -f $(DESTDIR)$(xrdpsysconfdir)/rsakeys.ini
--- a/docs/man/Makefile.am
+++ b/docs/man/Makefile.am
@@ -1,4 +1,4 @@
-man_MANS = \
+dist_man_MANS = \
   xrdp-dis.1 \
   sesman.ini.5 \
   xrdp.ini.5 \
@@ -14,4 +14,4 @@ man_MANS = \
 noinst_man_MANS = \
   xrdp-xcon.8
 
-EXTRA_DIST = $(man_MANS) $(noinst_man_MANS)
+EXTRA_DIST = $(noinst_man_MANS)
--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -36,7 +36,7 @@ if LINUX
 SUBDIRS += \
   pam.d \
   pulse
-dist_startscript_DATA += xrdp.sh
+dist_startscript_SCRIPTS = xrdp.sh
 if HAVE_SYSTEMD
 dist_systemdsystemunit_DATA = \
   xrdp-sesman.service \
@@ -61,9 +61,7 @@ endif
 if LINUX
 # must be tab below
 install-data-hook:
-	chmod 755 $(DESTDIR)$(sysconfdir)/xrdp/xrdp.sh
 	if [ -f $(DESTDIR)$(sysconfdir)/init.d/xrdp ]; then \
-		chmod 755 $(DESTDIR)$(sysconfdir)/init.d/xrdp; \
 		sed -i 's|__BASE__|$(prefix)|' $(DESTDIR)$(sysconfdir)/init.d/xrdp; \
 	fi
 endif
@@ -71,6 +69,5 @@ endif
 if FREEBSD
 # must be tab below
 install-data-hook:
-	chmod 755 $(DESTDIR)$(sysconfdir)/rc.d/xrdp
 	sed -i '' 's|%%PREFIX%%|$(prefix)|g' $(DESTDIR)$(sysconfdir)/rc.d/xrdp
 endif
--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -1,15 +1,14 @@
 EXTRA_DIST = \
    xrdp.sh \
    xrdp-sesman.service \
-   xrdp.service \
-   $(startscript_DATA)
+   xrdp.service
 
 #
 # files for all platforms
 #
 startscriptdir=$(sysconfdir)/xrdp
 
-startscript_DATA = \
+dist_startscript_DATA = \
   km-0407.ini \
   km-0409.ini \
   km-040c.ini \
@@ -37,9 +36,9 @@ if LINUX
 SUBDIRS += \
   pam.d \
   pulse
-startscript_DATA += xrdp.sh
+dist_startscript_DATA += xrdp.sh
 if HAVE_SYSTEMD
-systemdsystemunit_DATA = \
+dist_systemdsystemunit_DATA = \
   xrdp-sesman.service \
   xrdp.service
 else
--- a/instfiles/default/Makefile.am
+++ b/instfiles/default/Makefile.am
@@ -1,3 +1,2 @@
-EXTRA_DIST = xrdp
-startscriptdir=$(sysconfdir)/default
-startscript_DATA = xrdp
+startscriptdir = $(sysconfdir)/default
+dist_startscript_DATA = xrdp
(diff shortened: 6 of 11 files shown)


metalefty:
The author of  `instfiles/rc.d/xrdp` is me.  I'm not going to oppose you but just want to ask you why should keys be generated install-time?

For example, sshd generates host key pair on daemon start.  The only reason why I chose to generate keys on xrdp daemon start is sshd does so.  I just imitated sshd.  I simply believed OpenSSH way is good without thinking.

keygen/Makefile.am
```diff
@@ -13,3 +13,12 @@ xrdp_keygen_SOURCES = keygen.c
 
 xrdp_keygen_LDADD = \
   $(top_builddir)/common/libcommon.la
+
+xrdpsysconfdir = $(sysconfdir)/xrdp
+
+install-data-hook:
+	umask 077 && \
+	  ./xrdp-keygen xrdp $(DESTDIR)$(xrdpsysconfdir)/rsakeys.ini
```
metalefty: why not 2048bit?

itamarjp: I think that's a problem of xrdp-keygen

metalefty:
Too early to merge such this big change.

In FreeBSD pkg, all files which installed by `make install` must be packaged.
Files installed by `make install` but not listed on packing list is treated as *orphaned* files.

```
% make stage check-orphans package
====> Checking for pkg-plist issues (check-plist)
===> Parsing plist
===> Checking for items in STAGEDIR missing from pkg-plist
Error: Orphaned: %%ETCDIR%%/rsakeys.ini
===> Checking for items in pkg-plist which are not in STAGEDIR
===> Error: Plist issues found.
*** Error code 1

Stop.
make: stopped in /usr/home/meta/github/metalefty/freebsd-ports/net/xrdp-head
```

To avoid this, delete rsakeys.ini once generated by `make install` and re-generate it in post-install target in package definition (equivalent to spec file in rpm).  I don't think this is a good idea.

instfiles/init.d/xrdp
```diff
@@ -67,18 +66,6 @@ if [ "$(id -u)" = "0" ]; then
...
-    if [ ! -f $RSAKEYS ] ; then
-        log_action_begin_msg "Generating xrdp RSA keys..."
-        (umask 077 ; xrdp-keygen xrdp $RSAKEYS)
-        chown $USERID:$USERID $RSAKEYS
-        if [ ! -f $RSAKEYS ] ; then
-            log_action_end_msg 1 "could not create $RSAKEYS"
-            exit 1
-        fi
-        log_action_end_msg 0 "done"
-    fi
```
metalefty: Wht not keep this?  This will work only if rsakeys.ini *does not* exist.
This doesn't have a bad effect even if rsakeys.ini is installed by installe-data-hook but ensure existence of rsakeys.ini.

instfiles/rc.d/xrdp
```diff
@@ -61,10 +61,6 @@ xrdp_cmd() {
     if [ "${rc_arg}" = "stop" ] ; then
 	xrdp_daemons=$(reverse_list ${xrdp_daemons})
     fi
-    # Generate rsakeys.ini on start
-    if [ "${rc_arg}" = "start" -a ! -f %%PREFIX%%/etc/xrdp/rsakeys.ini ] ; then
-        %%PREFIX%%/bin/xrdp-keygen xrdp %%PREFIX%%/etc/xrdp/rsakeys.ini
-    fi
```
metalefty: Why not keep this?

jsorg71:
I think I'd have to agree with @metalefty 
Lets slow down on these merges until we can discuss.

jsorg71:
Yes, we should start using 2048 bit keys.
The default is 512.
I can change the default for xrdp-keygen to 2048 or we can pass the parameter.

speidy:
I agree about that.

בתאריך יום ד׳, 24 בפבר׳ 2016 ב-21:19 מאת jsorg71 <notifications@github.com>:

itamarjp:
@jsorg71, let us know what do you think is the better option
change xrdp-keygen to 2048 or  pass the parameter.

itamarjp:
about the rsakeys, 

for the people that's installing from sources it will be automatically generated.

for packages they need to remove them and re-generate at later stage (during post install)

if you're not happy with this way could you please explain the reasons ?

jsorg71:
I changed the default key size to 2048.

It looks like currently a new key is generated with every "make install."
That is not correct.
It should only create it at startup and only if it does not exist.

speidy:
I think a better way to handle rsa key or self sign certificate creation is
during package installation (like apt-get install)
That way you can ask the user for his preference .
בתאריך יום ה׳, 25 בפבר׳ 2016 ב-0:12 מאת jsorg71 <notifications@github.com>:

proski:
I agree, I would be more comfortable if I have time to react to the suggestions and answer questions before the merge. No need to rush with the merges. Anyway, thank you for merging! If anything needs to be adjusted, please let me know.

metalefty:
For rsakey or self-signed certificate generation, 

@itamarjp  
> for the people that's installing from sources it will be automatically generated.

Yes, I agree. It will be convenient for people.

> for packages they need to remove them

I agree this, too. No argue with removing key generated by install-data-hook during postinstall.

> and re-generate at later stage (during post install)

I half agree with this.  Key generation in init scripts should be kept even if keys are generated in postinstall. I believe it ensures key existence but no bad effect. I welcome to fix FreeBSD packaging if it's really needed.

Jay said:
> It looks like currently a new key is generated with every "make install." That is not correct.

I agree.  That is definitely incorrect.

> It should only create it at startup and only if it does not exist.

Whichever startup or postinstall is possible but existing keys should be kept intact.

I personally prefer startup time generation because it is OpenSSH way.  I need to explain the reason why OpenSSH way is better but please give me time.

keygen/Makefile.am
```diff
@@ -13,3 +13,12 @@ xrdp_keygen_SOURCES = keygen.c
...
+
+uninstall-hook:
+	rm -f $(DESTDIR)$(xrdpsysconfdir)/rsakeys.ini
```
metalefty: Do not remove rsakeys.ini if it wasn't generated by install-data-hook.

proski:
Lesson to myself - the changes to the init scripts should have been split to a separate patch.

From the point of view of the users running "make install", the init scripts may or may not be useful. After all, xrdp and xrdp-sesman can be run manually. I don't want users to depend on the init scripts. That was the main motivation, as I had that issue a few times. xrdp would start but then fail to accept connections. Maybe xrdp should complain loudly and exit if the rsakeys.ini file is missing.

I also looked at the init scripts from the packager's perspective, as a reference code creating distro-specific scripts. Packagers would not want manipulations on files in /etc outside the packages (admin edits are OK), so the removed code was just a noise.

Some distros may need a way to create rsakeys.ini if their packages don't support scripts that run upon installation. I don't know if any such distro or an OS is relevant today.

If there are any arguments for reverting changes to the scripts, I won't object. I just want to understand what the reasons are.

It should be easy to keep the original key on "make install". But the users won't get the key upgraded to the 2048 bit version, so there is an option to only upgrade short keys.

Uninstalling the key conditionally is tricky. It's hard to determine who "owns" the key. It should be possible to leave the key always and do some Automake trick to "make disctcheck" pass.

metalefty:
I 90% agree with your idea.

>  I don't want users to depend on the init scripts.

I understand.  You have a point there.

> Uninstalling the key conditionally is tricky. It's hard to determine who "owns" the key. 

That's true. It's not easy.

It might be too much to worry about that  however we always should be careful with deleting files. I think we shouldn't remove files when we cannot be sure if they aren't generated by user hand.

>  But the users won't get the key upgraded to the 2048 bit version, so there is an option to only upgrade short keys.

I think 2048bit keys are always better than 512bit.  However, for the same reason, we should be careful to replace files without user's intention even if replacing short keys with longer.

---
For key generation timing, let's think about "purpose of xrdp" or "what xrdp provides" - remote login service.  xrdp and ssh are same at this point. Both of them provides remote login service even they're different from GUI or CLI.  I thought OpenSSH can be a good example to xrdp. Of course OpenSSH way is not always correct and best.

Example: Fedora's sshd generates host keys on startup
http://pkgs.fedoraproject.org/cgit/rpms/openssh.git/tree/sshd.init?h=f20#n56
Example: Fedora's sshd systemd case
http://pkgs.fedoraproject.org/cgit/rpms/openssh.git/tree/sshd@.service
Example: FreeBSD's sshd generates host keys on startup
https://svnweb.freebsd.org/base/release/10.2.0/etc/rc.d/sshd?revision=286717&view=markup#l56


Actually, I'm leaning towards post-install keys/certificate generation now. Ready to support post-install key generation in FreeBSD package.

itamarjp:
how about ->

make install [dont install or generate key file]
make rsakeys [optional command to call key generation]

xrdp should complain loudly and exit if the rsakeys.ini file is missing.

proski:
Could somebody please give me an idea how valid is the OpnSSH comparison? Is the xrdp key used to mitigate man-in-the-middle attacks? Does any client warn the user if the key changes? If not, is it something that future clients would want to implement?

If the key replacement can create any inconvenience to the end user, let's keep it on "make install". Otherwise, why not replace it?

I don't like the idea of a separate target. Users will forget to run it. If it's a part of the installation, it should be run on "make install". If it's not part of the installation, the key show be created by other means, not relying on the source tree.

metalefty:
> Is the xrdp key used to mitigate man-in-the-middle attacks?

Using classic RDP encryption, no.  Using TLS, yes.

> Does any client warn the user if the key changes?

As same as above.  For example, FreeRDP warn the user if certificate changes but doesn't warn key change using classic RDP encryption.

metalefty:
I talked about TLS but this topic should be focused to RSA keys for classic RDP encryption.
Let's have another discuss on how to prepare certificates when TLS comes default.

Getting back to RSA keys, I don't like the separate target, too. What about "replace but back up existing key"?

proski:
I'm not a big fan of backups. I hate it when my /etc is littered with *.rpmnew and *.rpmsave, and I don't know whether I should keep what I'm using or pick one of those files.

Let's try to follow the principle of least surprise. If the expected behavior makes sense, it should be used.

If I install new software, I expect to get a new key, obviously. If I uninstall software, I expect all files belonging to that software to be removed. That includes any generated keys, as they are useless without the software. Maybe logs could be preserved, as I may be interested who logged in.

If I upgrade the software, I expect the related data to be preserved if possible. That's a tricky issue with the configuration files, but it's not a tricky issue with the keys. If the old key is compatible with the new software, it should be preserved. If it's not, it should be replaced, as there is no point in keeping it.

Finally, if I run software, and come data files are missing, I expect graceful degradation with a clear warning or an outright failure. I don't expect that the software would just sit in memory and reject all connections.

"make install" can be used both to install and to update software, so let's keep both cases in mind.

metalefty:
All what you're saying makes sense for install, upgrade and run.  What about uninstall?  I think config files differ from the original should be kept but removing all files wouldn't be a serious problem for the people who make & install xrdp from the source.  Package system should keep such files but `make [un]install` is not necessarily required to behave that.

Just a suggestion with that in mind, `make purge` target like the difference between `apt-get remove` and `apt-get purge`.  How about this?

```
# make uninstall # keep config files if modified or always keep
# make purge # aggressively remove everything
```

My final opinion about, 

when to generate keys: I still personally prefer startup time generation but postinstall actually makes sense and is acceptable, no objection.  I'll take it rather than personal preference.

removed code for key generation in  startup script: whichever is ok.  resurrect or keep removed.

certificates generation: let's have  another discuss when TLS comes default.

handling of existing keys: TBD.

proski:
I don't think many people use "make uninstall". For packages not using Automake, "make uninstall" is often broken or incomplete, if at all implemented. For packages using Automake, "make uninstall" is a poor man's package manager. If I cannot get a proper package, I can install software with "make install", play with it and remove it with "make uninstall". In that case, I hope that "make uninstall" would remove everything so that I don't need to clean up files manually.

Why would I want to preserve the key file? Only if I run "make install" immediately after "make uninstall". But I can just run "make install" and have the desired effect.

I can only think of one reason to run "make uninstall" before "make install". That would be to uninstall the files that the new code doesn't install anymore. In that case, the old source tree would need to be preserved, so that "make uninstall" is run from the old tree and "make install" from the new one. I suspect that very few people would bother to do that.

And what's the price of losing the key? At worst, the users would be warned. If it's my personal setup, I would actually appreciate that warning as a sign that the update has been successful. If it's a multi-user system, it should be properly administered, and that means using packaged software and backups, not "make install" and especially not "make uninstall".

PR #334, opened 2016-02-25
keygen: Do not overwrite rsakeys.ini

metalefty:
Check the existence of rsakeys.ini before generating it.


proski:
Actually, "exit 1" is not needed. Anyway, I have improved your patch. Please see https://github.com/neutrinolabs/xrdp/pull/336

metalefty:
#336 improved this. Closing.

PR #336, opened 2016-02-28
Preserve rsakeys.ini if it already exists

proski:
"make install" can be used to upgrade an existing installation, in which
case regenerating rsakeys.ini is unnecessary and could lead to warnings
on the client side.

This is an alternative to pull request 334. The differences:

- The whole command is not printed, it obscures more important messages.
- Instead, the user is told when the key is preserved.
- Empty key is replaced.
- Using KEYFILE variable for shorter code.
- "test" is preferred over brackets in makefiles

### DIFF
--- a/keygen/Makefile.am
+++ b/keygen/Makefile.am
@@ -16,9 +16,14 @@ xrdp_keygen_LDADD = \
 
 xrdpsysconfdir = $(sysconfdir)/xrdp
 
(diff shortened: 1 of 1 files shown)


metalefty:
:+1: I'm going to dismiss #334.

itamarjp:
- ./xrdp-keygen xrdp $(DESTDIR)$(xrdpsysconfdir)/rsakeys.ini
+./xrdp-keygen xrdp $(KEYFILE)

PR #341, opened 2016-03-04
Sanity checks

proski:
Group all initial sanity checks together. Check for rsakeys.ini in that function. Show full name of the PID file in error messages.

Enable -Wall and warnings for printf arguments. Use Autoconf Archive macros to check for those features, as they are not universal, but not limited to gcc.

### DIFF
--- a/common/os_calls.h
+++ b/common/os_calls.h
@@ -41,17 +41,26 @@
 #define g_tcp_select g_sck_select
 #define g_close_wait_obj g_delete_wait_obj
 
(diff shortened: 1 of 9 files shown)


jsorg71:
looks good

itamarjp:
@jsorg71 looks good = merge ?

jsorg71:
I think a 2 day wait for complaints is a good rule of thumb, unless it's obvious.

PR #342, opened 2016-03-06
Warning fixes

proski:
It's important to fix compiler warnings. Many warnings indicate real problems with the code. In particular, unused variables often indicate remote date that is read but not used. Those warnings are not fixed. They should be fixed by those who can improve the code.

What is fixed is the easy stuff, the warnings that obscure bigger problems.

In addition to the compiler warnings, a warning from Valgrind has been fixed. Intercepting SIGKILL doesn't work and should not be done.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -219,22 +219,6 @@ g_snprintf(char *dest, int len, const char *format, ...)
 
 /*****************************************************************************/
 void DEFAULT_CC
-g_writeln(const char *format, ...)
-{
-    va_list ap;
-
-    va_start(ap, format);
-    vfprintf(stdout, format, ap);
-    va_end(ap);
-#if defined(_WIN32)
-    g_printf("\r\n");
-#else
-    g_printf("\n");
-#endif
-}
-
-/*****************************************************************************/
-void DEFAULT_CC
 g_write(const char *format, ...)
 {
     va_list ap;
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2717,17 +2717,6 @@ g_signal_user_interrupt(void (*func)(int))
 /*****************************************************************************/
 /* does not work in win32 */
 void APP_CC
-g_signal_kill(void (*func)(int))
-{
-#if defined(_WIN32)
-#else
-    signal(SIGKILL, func);
-#endif
-}
-
-/*****************************************************************************/
-/* does not work in win32 */
-void APP_CC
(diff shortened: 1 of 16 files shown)


jsorg71:
I don't like using g_write("\n") instead of g_writeln("")
The purpose of writeln is so that you don't need to think about EOL.
\r\n in Windows and \r on mac and \n in Linux.

proski:
Good point, I'll redo that part. I'm torn between adding "-Wno-format-zero-length" for compilers that support it (that warning is not that useful anyway) and having a new macro for a newline. Let me see what would look better.

By the way, xrdp would greatly benefit from using glib - it has a lot of stuff xrdp has to implement on its own, including logging: https://developer.gnome.org/glib/stable/glib-Message-Logging.html

Also, the unnamed semaphores don't work in MacOS. glib would fix that easily. Perhaps we need a separate forum to discuss glib. MacOS is a good platform to support, it has Microsoft RDP client, and it's good to have a reference client and a server on the same system.

proski:
I made a version with variadic g_writeln. The problem is, it uses `##__VA_ARGS__` - most compilers support it, but it's not a standard. I don't want to use the alternative, knows as argument counting, as it's ugly and has its limitations (http://stackoverflow.com/questions/5588855/standard-alternative-to-gccs-va-args-trick)

I'll try "argument counting lite", i.e. some simple macro trick to enable g_writeln with no arguments or something like that.

I thought of GCC pragma to suppress the warning, but I'll need to check that it doesn't break any other compilers.

There is always an option to disable the warning in configure.ac if nothing better cam be done.

Please feel free to merge two other patched for now.

jsorg71:
@proski 
so if we do g_writeln("%s", "");
it's ok, right?

proski:
That looks too complex for adding a newline. I want something that is simple, readable, secure, not wasteful, and doesn't set a bad example. Introducing a new function to print the newline would not be readable either - that's one more thing to remember. The gcc warning for empty format is only useful if the format is used as is, not when something is appended. I'm very close to disabling that warning, which should be easy with Autoconf archive macros. But I still want to try some macro magic, just in moderate amounts - two or three lines.

jsorg71:
I've been trying to avoid variable parameter macros.  They are not standard.
I prefer the double (( )) instead like the LLOGLN macro

jsorg71:
I still prefer g_writeln("%s", "");
over changing g_writeln to a variadic macro

PR #359, opened 2016-04-07
sesman: do not hardcode Xorg/X11rdp/Xvnc executables

metalefty:
In RHEL and its clones, /usr/bin/Xorg is set suid-root. To execute
Xorg with user privileges, /etc/pam.d/xserver needs be edited [1],
or suid bit of Xorg binary needs to be dropped.

In order to keep Xorg and /etc/pam.d/xserver untouched, preparing
non-suid version of Xorg as /usr/bin/Xorg.non-suid for example is
the simplest solution. However, Xorg.non-suid cannot be executed
since it is hardcoded to execute Xorg in sesman.

This change makes more flexible to execute Xorg with non-standard
name or not in PATH environment variable.

[1] https://www.centos.org/forums/viewtopic.php?t=21185

### DIFF
--- a/sesman/config.h
+++ b/sesman/config.h
@@ -220,16 +220,14 @@ struct config_sesman
    */
   struct list* rdp_params;
   /**
-   * @var log
-   * @brief Log configuration struct
+   * @var xorg_params
+   * @brief Xorg additional parameter list
    */
-
   struct list* xorg_params;
   /**
    * @var log
    * @brief Log configuration struct
    */
-
   //struct log_config log;
   /**
    * @var sec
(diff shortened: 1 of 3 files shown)


metalefty:
What do you think of my idea?  This change will destroy compatibility of sesman.ini.
sesman.ini needs to contain `param0=X[11rdp|vnc|org]` parameter  after this change.

The motivation of making this change is because I don't want to change anything installed by other packages.

jsorg71:
+1
I like this and I even thought about using a script, like startwm.sh but something like startdisplay.sh
With this configurable, someone can always do that.
I don't think that backward compatible is as important anyway, we can default param0 if it's not in the file.

metalefty:
>  we can default param0 if it's not in the file.

sounds good.

PR #360, opened 2016-04-10
Warning fixes

proski:
Fixing warnings to keep their number manageable. This also establishes a higher standard for the new code.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -221,6 +221,6 @@ list_dump_items(struct list *self)
 
     for (index = 0; index < self->count; index++)
     {
-        g_writeln("%d: %s", index, list_get_item(self, index));
+        g_writeln("%d: 0x%lx", index, list_get_item(self, index));
     }
 }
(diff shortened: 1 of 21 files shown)


jsorg71:
+1
I like it but I think __always_unused seems odd.
I see __ prefix and I think compiler define.
We can always change it later but ALWAYS_UNUSED seems better.
Or using in_uint8s(s, _bytes); /* name */ works too and with all compilers.

proski:
I agree about compiler defines. Indeed, it would be better to use something
that looks "magic", but not "compiler-internals-magic". I had no good idea,
so I went with the Linux kernel choice. ALWAYS_UNUSED would steal attention
for more important things. It's way too big and loud.

I wanted to preserve more information about the discarded data. If that's
not a big deal, let's use a more portable solution as you suggest.

proski:
I dropped most of the "unused" fixes, including __always_unused. I also removed changes to window_id declaration.

jsorg71:
looks good
+1

proski:
Can somebody merge it soon, please? If there are any questions, I can make a separate pull request for "Fix recently introduced warnings in session.c " as it's a serious issue (incorrect pointer types).

PR #366, opened 2016-04-23
Add unnamed semaphore support on Mac OS

proski:
Standard unnamed semaphores (sem_init, sem_destroy) are not functional on Mac OS. The compiler reports that they are deprecated.

So I used the Grand Central Dispatch implementation. It is standard on Mac OS, nothing needed to be installed (I have XCode installed already to have a development environment).

I considered using named semaphores (sem_open), but they can outlive the process and need special care to be cleaned up.

I made libscp_lock.c use semaphores through the thread_calls wrapper. That should work on all operating systems.

### DIFF
--- a/common/thread_calls.c
+++ b/common/thread_calls.c
@@ -20,6 +20,10 @@
 
 #if defined(_WIN32)
 #include <windows.h>
+#elif defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


jsorg71:
+1

PR #370, opened 2016-05-05
Typo fixes and more

proski:
I used a more sophisticated script to run the sources through the spell checker while suppressing uninteresting stuff (hex constants, include file names etc), which allowed me to catch more typos. Some changes are not strictly typo fixes, so they have been split into separate patches.


jsorg71:
+1

PR #372, opened 2016-05-07
Fix warnings reported by gcc 5.3.0 on Mac OS

proski:
This should make xrdp warning-free and set a higher standard for new code contributions.


jsorg71:
+1
Thanks for your work on this.

Issue #387, opened 2016-06-17
Add /sbin to PATH in sesman/env.c


metalefty:
@jsorg71, do you mind if I do this to fix drive redirection in FreeBSD?

```
diff --git a/sesman/env.c b/sesman/env.c
index 39e020f..6bad2c4 100644
--- a/sesman/env.c
+++ b/sesman/env.c
@@ -122,7 +122,7 @@ env_set_user(char *username, char *passwd_file, int display,
         {
             g_clearenv();
             g_setenv("SHELL", pw_shell, 1);
-            g_setenv("PATH", "/bin:/usr/bin:/usr/local/bin", 1);
+            g_setenv("PATH", "/sbin:/bin:/usr/bin:/usr/local/bin", 1);
             g_setenv("USER", username, 1);
             g_sprintf(text, "%d", uid);
             g_setenv("UID", text, 1);

```

The background is the place of mount program.  If `/sbin` is not in the `PATH`, drive redirection does not work because FreeBSD's fuse mount program which is equivalent to `/bin/fusermount` of GNU/Linux is `/sbin/mount_fusefs`.

If `/sbin` is not in the `PATH`, chansrv say: 
```
fuse: failed to exec mount program: No such file or directory
```

This is another option to fix this.
```
diff --git a/sesman/env.c b/sesman/env.c
index 39e020f..4061e42 100644
--- a/sesman/env.c
+++ b/sesman/env.c
@@ -122,7 +122,11 @@ env_set_user(char *username, char *passwd_file, int display,
         {
             g_clearenv();
             g_setenv("SHELL", pw_shell, 1);
+#ifdef __FreeBSD__
+            g_setenv("PATH", "/sbin:/bin:/usr/bin:/usr/local/bin", 1);
+#else
             g_setenv("PATH", "/bin:/usr/bin:/usr/local/bin", 1);
+#endif
             g_setenv("USER", username, 1);
             g_sprintf(text, "%d", uid);
             g_setenv("UID", text, 1);

```

Which do you like?

jsorg71:
I think I like the first one.
Add /sbin to path.

metalefty:
I still have some to work on drive redirection on FreeBSD but one done, closing.

PR #388, opened 2016-06-20
sesman: Add /sbin to PATH for FreeBSD

metalefty:
to enable to execute FUSE mount program.

The place of FUSE mount programs are:

FreeBSD   : /sbin/mount_fusefs
GNU/Linux : /bin/fusermount

See also #387.


jsorg71:
+1

PR #390, opened 2016-06-22
Cleanups and C++ compatibility

proski:
That's a large set of patches. The common topics are hardening the code, improving logging and updating the documentation. No new features are added. No user visible changes are expected apart from the logs and the messages in the login window.

The only significant bugfix it preventing a string literal from being freed on exit in xrdp-chansrv.

prog_std.txt suggests compiling the code with a C++ compiler. I was able to do that, both with gcc and clang. Only the straightforward and obviously good results of that effort are included, the rest remains on my devel branch for further work.

The -Wwrite-strings flag has been added to enable detection of incorrect string literal access even with C compilers.

To avoid excessive casts from void* on malloc results, I introduced macros g_new and g_new0, exactly as in glib, and they do the cast automatically. I think g_new0 is very good at showing that zero-initialized memory is requested. If we start using glib, we'll have one thing less to worry about.

If any patch raises questions, please let me know, I'll take it away for separate discussion.

The code has been tested by logging in from Mac clients. That included running xrdp and xrdp-sesman in Valgrind.


jsorg71:
Big PR, looks good so far.  Still looking.

speidy:
+1
Also, i think the g_new thing should be changed in one place like in os_calls.c, so diff will be smaller. and merges will be easier.

proski:
g_new needs to be a macro. It needs to know the expected type to case the result. So all the invocations flagged by the C++ compilers need to be changed. I cannot make the diff smaller, unless I'm missing something.

speidy:
Sorry, you're right.
I still didn't review the other changes but seems like a good cleanups here.

Why does cpp compiler warn about the old g_malloc ? Is it about lack of
casting?
בתאריך יום ד׳, 29 ביוני 2016 ב-9:23 מאת Pavel Roskin <
notifications@github.com>:

proski:
Yes. C++ requires explicit cast from void pointer to other pointers. See http://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

speidy:
I know that , but we did use explicit casts in xrdp code.
בתאריך יום ד׳, 29 ביוני 2016 ב-10:09 מאת Pavel Roskin <
notifications@github.com>:

jsorg71:
xrdp g_malloc calls should be casting but some might not.
In C void* can be assigned to other pointer without a cast.  In C++ you need to cast or you get warning.

proski:
Is anything needed on my part?

The reason I introduced g_new() is to avoid new casts where they were missing. Eventually, calls to g_malloc() with casts could be converted to g_new(), but it would be a big change. The point of this branch is to fix C++ errors and warnings without negatively affecting readability.

Also, I have discovered that ''Use g_strdup() for xrdp-chansrv program name, it's freed on exit" is not needed. It fixes a warning, but a better fix would be to use const strings for application and log name and fix all invocations. I think it would be easier to merge this branch as is and then do a cleanup on a separate branch. But if anyone insists, I can remove that patch from the branch.

The reason I work on the warnings is because I want any future fixes to be held against higher coding standards. It's easy to introduce bugs if nobody looks at the warnings. I don't want to add more patches if I don't get feedback.

blackcrack:
hi, 
i have fellow the whole changes in the last month.. and want say,
they should first clean up bring a new version out, where it is fixed fully (you can maybe help with)
before you make some big changed for have a clean Version for distriebutioning 
and then it's this maybe not bad, if you change this whole with g_new() and something
for the next version.. xrdp should bring out/up a next version and come to 1.0 or something ;)
 ( i be inquisitive how long it's need for 1.0 :) ) 
best regards
Blacky

proski:
I think it's telling that the whole cleanup did not reveal any user-facing issue, apart from documentation and log messages. That g_strdup() in xrdp-chansrv was supposedly one fix, but even that is not fixing a user visible bug, as there is another g_strdup() in log_start_from_param(). The code quality is reasonable. We can release 0.9.1 now, with or without change on this branch. Version 1.0 is a once-in-a-lifetime event, let's do it after the documentation is improved.

blackcrack:
@proski : clear if there be no stable and the distriebutions do not obstruct this lil small nice proggy for rdp (which also high usable for Reactos :))))  ) so long it's not installed in the whole distro's ..rdp make the whole WinNT Liux Connectivity really usable and more easily insteat the an bit more technical and complicated vnc. And i bet, with this Xrdp will makes the whole much easyer , i looking forward if it's into the whole Distros include, also OpenMandriva. :+1: yepp

proski:
Please don't merge, there is a bug somewhere that affects the login screen. I'll update the branch once I fix it.

proski:
I've updated the patch set. It's OK to merge now.

The patch named "Make xrdp_wm_log_msg() log its output" removed an unrelated call accidentally, which broke the log window on login failure. That has been fixed.

Instead of running g_strdup() on program_name, program_name has been made a constant string, so the unnecessary path to use g_strdup() for xrdp-chansrv program name has been removed.

The C++ port is complete now. I have cleaned up and included the remaining patches. That includes a patch that makes it the code compiled with C++ run correctly and load modules.

PR #396, opened 2016-07-12
Fix rpmlint warning about W: devel-file-in-non-devel-package

kenhys:
It is common way not to embed version for loadable module.
Without this change, rpmlint reports it as a warning.

For example, https://en.opensuse.org/openSUSE:Packaging_checks#devel-file-in-non-devel-package

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -42,3 +42,6 @@ libcommon_la_LIBADD = \
   -lcrypto \
   -lssl \
   -lpthread
+
+libcommon_la_LDFLAGS = \
+  -avoid-version -module
--- a/libxrdp/Makefile.am
+++ b/libxrdp/Makefile.am
@@ -58,7 +58,8 @@ libxrdp_la_SOURCES = \
   xrdp_sec.c
 
 libxrdp_la_LDFLAGS = \
-  $(EXTRA_FLAGS)
+  $(EXTRA_FLAGS) \
+  -avoid-version -module
 
 libxrdp_la_LIBADD = \
   $(top_builddir)/common/libcommon.la \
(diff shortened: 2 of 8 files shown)


proski:
libcommon is not a module, it's actually a library. Others are modules, and could be compiled as modules, but it would not fix the rpmlint warning.

libcommon contains the code shared between the binaries in the package, but it is not intended to be used by other packages.

I think a better solution for libcommon would be to omit libcommon.so symlink from the rpm package.

proski:
The right fix to so separate modules from shared libraries. Modules should not have the initial "lib" and not should not be versioned. Shared libraries should start with "lib" and should be versioned. RPM packages should omit non-versioned links for shared libraries (they belong to -devel packages).

common/Makefile.am
```diff
@@ -42,3 +42,6 @@ libcommon_la_LIBADD = \
   -lcrypto \
   -lssl \
   -lpthread
+
+libcommon_la_LDFLAGS = \
+  -avoid-version -module
```
proski: Not a module!

libxrdp/Makefile.am
```diff
@@ -58,7 +58,8 @@ libxrdp_la_SOURCES = \
   xrdp_sec.c
 
 libxrdp_la_LDFLAGS = \
-  $(EXTRA_FLAGS)
+  $(EXTRA_FLAGS) \
+  -avoid-version -module
```
proski: Not a module!

metalefty:
I added invalid label as merging this breaks xrdp and xrdp-sesman.

proski:
The issue has been fixed in f25659d2

Issue #399, opened 2016-07-19
Python needs updating


geiseri:
Using openssl 1.0.2.h-1causes python to not build anymore.  I see this on Arch linux.  Is it possible to build with the system python2?

metalefty:
I'll give it a try.

metalefty:
I confirmed that x11rdp properly builds and works with python 2.7.11 the latest version of python2. 

I don't try to use system python2. It is still on TODO list.

metalefty:
Jay said:
> Another option is that we just remove the first line from x11_file_list.txt.
> We don't really need to build python these days.

@jsorg71 @speidy 
I concern about that python2 executable has different name depends on system.  It will be python, python2 or python2.7.  One more concern is python3.  Arch Linux for example.

[Arch Wiki](https://wiki.archlinux.org/index.php/Python) says:
> Any program requiring Python 2 needs to point to /usr/bin/python2, instead of /usr/bin/python, which points to Python 3.

speidy:
@metalefty, I'm not sure it's needed at all for Xorg to work.
בתאריך יום ה׳, 4 באוג׳ 2016 ב-10:56 מאת metalefty <notifications@github.com

metalefty:
Python2.7 is indirectly  needed to build Xorg. Not needed to run built Xorg.
Python2.7 is needed to build xcb-proto.  xcb-proto is needed to build Xorg.

Removing the first line of x11_file_list.txt is no problem with Scientific Linux 7.

geiseri:
This looks to be fixed now.  I think we can close the issue as fixed.  Thanks!

metalefty:
We'll close after making change to use system python2.

metalefty:
Tested on Ubuntu 16.04 and Debian 8, no problem.
Now I think we can remove python from x11_file_list.txt.

PR #402, opened 2016-07-20
X11rdp: update python 2.7 -> 2.7.11

metalefty:
Python 2.7 cannot build with OpenSSL 1.0.2h pointed out in #399.
Python-2.7.11.tar.xz needs to be uploaded to server1.xrdp.org.

I've confirmed X11rdp successfully builds with python 2.7.11 and built X11rdp properly works.
Please don't forget to upload tarball to server1.xrdp.org before merge this.

### DIFF
--- a/xorg/X11R7.6/x11_file_list.txt
+++ b/xorg/X11R7.6/x11_file_list.txt
@@ -1,4 +1,4 @@
-Python-2.7.tar.bz2                              :  Python-2.7                               :
+Python-2.7.11.tar.xz                            :  Python-2.7.11                            :
 util-macros-1.11.0.tar.bz2                      :  util-macros-1.11.0                       :
 xf86driproto-2.1.0.tar.bz2                      :  xf86driproto-2.1.0                       :
 dri2proto-2.3.tar.bz2                           :  dri2proto-2.3                            :


jsorg71:
+1
I uploaded Python-2.7.11.tar.xz

jsorg71:
Another option is that we just remove the first line from x11_file_list.txt.
We don't really need to build python these days.

speidy:
I agree
בתאריך יום ה׳, 4 באוג׳ 2016 ב-9:34 מאת jsorg71 <notifications@github.com>:

metalefty:
I also agree. In RHEL, `yum` is written in python2 so python2 must be always installed.

Actually, x11rdp has build time dependency on `python-libxml2` package ([buildx.sh:24](https://github.com/neutrinolabs/xrdp/blob/devel/xorg/X11R7.6/buildx.sh#L24)).  When we install `python-libxml2` from distribution's repository, python2 also should be installed.  Because `python-libxml2` [depends on python2](https://packages.debian.org/jessie/python-libxml2).  In conclusion, python2 is to be installed to meet x11rdp build requirements.

Let's back to #399 and test the build using system python2. I'll close this later.

metalefty:
What about merging this first and then think about removing python later?

jsorg71:
Yes, good idea.

PR #403, opened 2016-07-21
chansrv: avoid chansrv SEGV when xinode is NULL

kenhys:
When xfuse_create_file_in_xrdp_fs is failed, it returns NULL.

Without this fix, xinode->size causes SEGV, so implementation is changed
to return -1 and check the return value in caller.


jsorg71:
+1

speidy:
+1

PR #406, opened 2016-08-12
Include stdint.h, don't redefine MAX/MIN constants if already defined

proski:
Safer version of the original patch. The constants are defined only if they are not defined in stdint.h or elsewhere.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -1,10 +1,10 @@
-
-EXTRA_FILES =
-
 if XRDP_PIXMAN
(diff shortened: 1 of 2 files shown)


jsorg71:
Lets use HAVE_STDINT_H for this.

We can't build without stdio.h, stdlib.h, and string.h, so we don't need HAVE_STDLIB_H, HAVE_STRINGS_H, and HAVE_STDIO_H

jsorg71:
merged as f4f23b0

PR #410, opened 2016-08-18
Fix clipboard when text/filename contains non-ASCII characters

metalefty:
broken by #314. This is compatible with the fix introduced in #314.

To use non-ASCII text/filename in clipboard, chansrv needs to be run
with LC_CTYPE=\*.UTF-8 because the behaviour of mbstowcs(3) function
called in chansrv depends on LC_CTYPE[1].  However #314 made
LC_CTYPE=C in chansrv context.  Even if LANG and LC_\* are set in
.bashrc, /etc/profile, /etc/locale.conf or something like that,
it doesn't affect in chansrv context because chansrv doesn't source
any of them unlike sesman.

So do not set LC_CTYPE to blank or "C" in g_init() in order to get
g_mbstowcs and g_wcstombs to work properly with non-ASCII UTF-8
characters in any context.

Setting LC_CTYPE to *.UTF-8 doesn't obstruct applying system
language in RHEL [2].

[1] Linux man page says:
      The behavior of mbstowcs() depends on the LC_CTYPE category of
      the current locale.

[2] https://bugzilla.redhat.com/show_bug.cgi?id=1290820

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -136,7 +136,21 @@ g_init(const char *app_name)
 
     WSAStartup(2, &wsadata);
 #endif
-    setlocale(LC_CTYPE, "");
+
+    /* In order to get g_mbstowcs and g_wcstombs to work properly with
+       UTF-8 non-ASCII characters, LC_CTYPE cannot be "C" or blank.
+       To select UTF-8 encoding without specifying any countries/languages,
+       "C.UTF-8" is used but provided in few systems.
+
+       See also: https://sourceware.org/glibc/wiki/Proposals/C.UTF-8 */
+    char *lc_ctype;
+    lc_ctype = setlocale(LC_CTYPE, "C.UTF-8");
+    if (lc_ctype == NULL)
+    {
+        /* use en_US.UTF-8 instead if not available */
+        setlocale(LC_CTYPE, "en_US.UTF-8");
+    }
+
     g_mk_temp_dir(app_name);
 }
 


metalefty:
Fixed commit message s/#317/#314/g.

metalefty:
## Bug description

Can't copy&paste text/file via clipboard when text/filename contains non-ASCII UTF-8 characters.

## Environment 
```
$ rpm -qa|grep xrdp
xorg-x11-drv-xrdp-0.9.0.gita0add6c+fix_utf8_clipboard-1.el7.centos.x86_64
xrdp-0.9.0.gitf4f23b0+devel-1.el7.centos.x86_64
```

I'm using mstsc.exe of Windows 10 at this time but client doesn't matter.

## How reproducible
Always for non-ASCII UTF-8 characters, never for ASCII characters.

## Procedure for reproducing 
 
Install latest devel branch f4f23b0a7d893c871092dcb2f53f733d3909e370  by RH-Matic.
```
$ ./X11RDP-RH-Matic.sh --branch devel --with-xorg-driver --nox11rdp
```

Install Xfce4 and Firefox at this time.

Set system desktop environment and locale.
```
# echo PREFERRED=$(which startxfce4) > /etc/sysconfig/desktop
# echo LANG=ja_JP.UTF-8" > /etc/locale.conf"
```

Run xrdp and sesman in foreground.
```
# xrdp -ns
# xrdp-sesman -ns
```

Connect to xrdp and open www.asahi.com in Firefox. 

Copy some Japanese string to clipboard.   As same result as for other languages for example Chinese or Russian.

Paste copied text into notepad.exe.

See the log dumped in sesman terminal. It says "bad string".

~~~
[2522346826]: CLIPBOARD  clipboard_send_data_response_for_text: 819 : ERROR: clipboard_send_data_response_for_text: bad string
~~~

See `sesman/chansrv/clipboard.c` around line 815.

~~~C
    num_chars = g_mbstowcs(0, data, 0);
    if (num_chars < 0)
    {
        log_error("clipboard_send_data_response_for_text: "
             "bad string");
        num_chars = 0;
    }
~~~

"bad string" is dumped as g_mbstowcs() returns negative value. See Linux man page of mbstowcs(3) to know when mbstowcs() returns negative value.

## How to fix
See the first comment and commit.

metalefty:
@proski @itamarjp
It turned out that `g_setenv("LANG", "en_US.UTF-8", 1);` was actually added on purpose.  Removing it in #314 caused another bug.  I made a fix with keeping respecting system language.

proski:
Looks good to me.

metalefty:
Any objections to merge?

jsorg71:
+1

PR #416, opened 2016-08-31
Add UK English keyboard

proski:
Add the keyboard ini file for UK English from http://component-parts.blogspot.ie/2012/08/fixing-uk-keyboard-mapping-for-xrdp.html

Add locale descriptions to the locale files and remove them from xrdp-genkeymap man page and source. It's more maintainable that way.


metalefty:
Two thoughts.  

One is about Japanese keymap file.  km-e0010411.ini, km-e0200411.ini, km-e0210411.ini are all for Japanese.  Please keep them identical to km-0411.ini.  I'm thinkng to replace the with symbolic link to km-0411.ini.


Another one is about locale descriptions.  XXXX of km-XXXX.ini is called "locale identifier" or "keyboardLayout".  Search [[MS-RDPBCGR].pdf](https://msdn.microsoft.com/en-us/library/cc240445.aspx) with the word "locale identifier".  It is 32bit unsigned value expressed in 8-digit hexadecimal number.  The lower 16bit indicates country/language.  The upper 16bit indicates keyboard layout or active input method.  For example, US Dvorak keyboard has the value `0x00010409`. 

See also
* https://technet.microsoft.com/en-us/library/cc766503(WS.10).aspx
* https://technet.microsoft.com/ja-jp/library/cc748926(WS.10).aspx

When creating keymap file, the upper 16 bits filled in 0 should be ommitted. 

~~~
0x00010409 -> km-10409.ini
0xe0010411 -> km-e0010411.ini
~~~

 This is because this code in `xrdp/lang.c:234`.
~~~C
g_snprintf(filename, 255, "%s/km-%4.4x.ini", XRDP_CFG_PATH, keylayout);
~~~

docs/man/xrdp-genkeymap.8
```diff
@@ -22,36 +22,10 @@ The key map information is stored in the file named \fIoutfile\fP.
...
-.B 041d
-se swedish
-.TP
-.B 0809
-en-uk uk english
-.RE
+Files containing the keyboard mapping for the locale \fIXXXX\fP, which is
+a 4-digit hexadecimal number identifying the language and the country.
+The locale list can be found at
+.URL "https://msdn.microsoft.com/en-us/library/ms912047.aspx"
```
metalefty: This is actually 4-8 digit hexadecimal number expressed in %4.4x style.

proski:
I didn't realize all km-*0411.ini files were identical. Rather than make symlinks for them, xrdp could be changed to fall back to the 16-bit version.

I'll add the additional documentation.

metalefty:
> xrdp could be changed to fall back to the 16-bit version.

Good idea.

metalefty:
I did fallback in #418. We can remove km-????0411.ini now.  

km-????xxxx.ini should be put only if it is not identical to km-xxxx.ini.
km-0409.ini (US) and km-00010409.ini (US Dvorak), for example.

PR #364, opened 2016-04-14
Allows use the xkb options that for example allows you to specify a key combination to change the layouts.

laigor:
Example of xrdp_keyboard.ini for russian keyboard (Please add this to wiki) 

[default_rdp_layouts]
rdp_layout_us=0x00000409
rdp_layout_de=0x00000407
rdp_layout_fr=0x0000040C
rdp_layout_it=0x00000410
rdp_layout_jp=0x00000411
rdp_layout_jp2=0xe0010411
rdp_layout_jp3=0xe0200411
rdp_layout_jp4=0xe0210411
rdp_layout_ru=0x00000419
rdp_layout_se=0x0000041D
rdp_layout_pt=0x00000816
rdp_layout_br=0x00000416
rdp_layout_pl=0x00000415

[rdp_keyboard_ru]
keyboard_type=4
keyboard_subtype=1
model=pc105
options=grp:alt_shift_toggle
rdp_layouts=default_rdp_layouts
layouts_map=layouts_map_ru

[layouts_map_ru]
rdp_layout_us=us,ru
rdp_layout_ru=us,ru

PR #417, opened 2016-09-02
Apply xkb options for x11rdp

metalefty:
Setting xkb options is added  in #364 but options are not applied to x11rdp.
This is already implemented for xorgxrdp.

And some more logs.

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -237,6 +237,9 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
     char rdp_layout[256] = { 0 };
 
     LLOGLN(0, ("xrdp_load_keyboard_layout:"));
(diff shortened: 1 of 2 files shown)


proski:
You may want to remove `LLOGLN(0, ("xrdp_load_keyboard_layout:"));` before the keyboard type/subtype is logged. No need to log the same text twice.

Apart from that, the PR looks good to me.

metalefty:
Removed.

proski:
The code change is good. You got "log" twice in the last commit message. I would just squash the extra commit and force-push the branch. But it's not very important. See http://stackoverflow.com/questions/7947322/preferred-github-workflow-for-updating-a-pull-request-after-code-review/23913725

metalefty:
Thanks, I'll do that if force-push is acceptable. I didn't squash as some people don't like force-push.

proski:
I think it's entirely acceptable in the branches created for pull requests. But it shouldn't matter, your changes are good to be applied either way.

metalefty:
Let's merge.

PR #418, opened 2016-09-02
Use km-0000yyyy.ini if km-xxxxyyyy.ini doen't exist

metalefty:
upper 16 bits of keylayout value indicates keyboard layout.
lower 16 bits of keylayout value indicates country/language.

If the keymap file doesn't match exactly to keylayout value,
the typical keyboard of the country/language should be used.

For example [1]:
  0x00000409: US English
  0x00010409: US Dvorak
  0x00020409: US International
  0x00030409: US Dvorak for left hand
  0x00040409: US Dvorak for right hang
  0x0000040e: Hungarian
  0x0001040e: Hungarian 101-key

[1] https://msdn.microsoft.com/windows/hardware/commercialize/manufacture/desktop/windows-language-pack-default-values

### DIFF
--- a/xrdp/lang.c
+++ b/xrdp/lang.c
@@ -226,22 +226,34 @@ int APP_CC
 get_keymaps(int keylayout, struct xrdp_keymap *keymap)
 {
     int fd;
+    int basic_key_layout = keylayout & 0x0000ffff;
     char *filename;
     struct xrdp_keymap *lkeymap;
 
     filename = (char *)g_malloc(256, 0);
-    /* check if there is a keymap file */
+
+    /* check if there is a keymap file e.g. km-e00100411.ini */
     g_snprintf(filename, 255, "%s/km-%08x.ini", XRDP_CFG_PATH, keylayout);
 
-    /* if the file does not exist, try again with 'en-us' as fallback */
+    /* if the file does not exist, use only lower 16 bits instead */
     if (!g_file_exist(filename))
     {
+        log_message(LOG_LEVEL_INFO, "Cannot find keymap file %s", filename);
+        /* e.g. km-00000411.ini */
+        g_snprintf(filename, 255, "%s/km-%08x.ini", XRDP_CFG_PATH, basic_key_layout);
+    }
+
+    /* finally, use 'en-us' */
+    if (!g_file_exist(filename))
+    {
+        log_message(LOG_LEVEL_INFO, "Cannot find keymap file %s", filename);
         g_snprintf(filename, 255, "%s/km-00000409.ini", XRDP_CFG_PATH);
     }
 
     if (g_file_exist(filename))
     {
         fd = g_file_open(filename);
+        log_message(LOG_LEVEL_INFO, "Loading keymap file %s", filename);
 
         if (fd != -1)
         {


metalefty:
> If the keymap file matches exactly to keylayout value,
> the typical keyboard of the country/language should be used.

Oops, DOESN'T here. Commit message edited.

proski:
You can just amend the existing commit and force-push it.

I like switching from %4.4x to %08x format

Please calculate the shorter number once and make a variable for it if needed. Generally, try to reuse as much as possible in the log messages so that they reflect the actual behavior.

If a file cannot be loaded, It's better to report it with the full path. That's more helpful in case of actual problems. No need to report what would be next.

I would want log messages like this:

[INFO] Cannot find keymap file /etc/xrdp/km-00010809.ini
[INFO] Cannot find keymap file /etc/xrdp/km-0809.ini
[INFO] Loading keymap file /etc/xrdp/km-0409.ini

metalefty:
@proski 
Do you mean all km-%4.4x.ini files to be renamed to km-%08x.ini?
If so, I like it.

@jsorg71 
What do you think? I like %08x too.

New keymap file loading would be:
1. When the client reports 0x00010809, find km-00010809.ini
2. If not found, find km-00000809.ini 
3. If not found, find km-00000409.ini

proski:
I actually meant using 8 hex digit format for the numbers that don't fit 4 hex digits. For example, the US Dvorak keyboard is represented as km-10409.ini now, but it would become km-00010409.ini. We don't even have such keymaps yet.

But I agree it makes sense to rename all keymaps to use 8 hex digit. Distinguishing "full" and "fallback" maps doesn't always make sense. For example, the standard US English keyboard is not a good fallback for US Dvorak.

Now I think we should have some way to indicate that XRDP specifically supports some keyboard variant. Fallback to the default variant should be a cause for concern and an invitation to support the new keyboard variant. It should be `[WARNING]`, not `[INFO]`.

Symlinks would be a poor solution. It would be better to make it possible for a keymap to inherit data from another keymap and override a few symbols if needed. By the way, I don't think km-*0411.ini are supposed to be identical.

metalefty:
FYI, I found the official definition of locale identifier (keylayout in xrdp).
>A  32-bit value which consists of the language ID in the low word (bits 0-15) and the sorting ID (bits 16-19) and a reserved value (bits 20-31) in the high word.

https://www.microsoft.com/Language/en-US/Search.aspx?sString=locale&langID=es-es

metalefty:
> By the way, I don't think km-*0411.ini are supposed to be identical.

Yes. They're not supposed to be identical.  It has been confirmed that four km-0411.ini, km-e0010411.ini, km-e0200411.ini and km-e0210411.ini are identical.

In CJK languages, probably, the differences among them are IMEs. Not the keyboard variants.
https://github.com/awakecoding/FreeRDP-Manuals/blob/master/User/FreeRDP-User-Manual.markdown#keyboard-input-method-editors-imes

proski:
Is this patch still relevant? I suggest that we switch to 8 hex digit format for all km*.ini files.

metalefty:
I'll make another pull request just converting to 8 hex digit format.
And then rework this.

metalefty:
As #437 was merged, I updated patch and force pushed.
@proski could you look at this again?

xrdp/lang.c
```diff
@@ -226,22 +226,35 @@ int APP_CC
 get_keymaps(int keylayout, struct xrdp_keymap *keymap)
 {
     int fd;
+    int shortKeylayout; /* lower 16 bits of keylayout */
```
proski: We don't do camelCase for variables. Also, it's better to declare the variable in the block where it's used. I would call it "basic_key_layout". "short" is a C concept, not a problem domain concept.

xrdp/lang.c
```diff
@@ -226,22 +226,35 @@ int APP_CC
 get_keymaps(int keylayout, struct xrdp_keymap *keymap)
 {
     int fd;
+    int basic_key_layout; /* lower 16 bits of keylayout */
```
metalefty: It makes sense. Rebased to devel head and force pushed.  Thanks a lot.

proski:
Looks good to me.

Issue #422, opened 2016-09-08
Log TLS version and cipher when TLS connection established


metalefty:
Inspired by apache log and postfix log,  the log like this will help us to know information about encryption.

Which TLS version the client uses, which cipher suite the client uses...

```
[08/Sep/2016:12:59:46 +0900] crawl-66-249-79-88.googlebot.com TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 270
[08/Sep/2016:18:31:55 +0900] crawl-66-249-79-143.googlebot.com TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 271
[08/Sep/2016:21:28:56 +0900] spider-141-8-142-22.yandex.com TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 331
[08/Sep/2016:22:15:12 +0900] spider-141-8-184-29.yandex.com TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 333
[08/Sep/2016:23:23:38 +0900] crawl-66-249-79-62.googlebot.com TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 269
[09/Sep/2016:00:45:08 +0900] msnbot-157-55-39-140.search.msn.com TLSv1.2 DHE-RSA-AES128-GCM-SHA256 "GET /robots.txt HTTP/1.1" 331
```

```
Sep  9 01:40:53 glory postfix/smtpd[46564]: Anonymous TLS connection established from mx2.freebsd.org[8.8.178.116]: TLSv1.2 with cipher AECDH-AES256-SHA (256/256 bits)
Sep  9 01:44:47 glory postfix/smtpd[49271]: Anonymous TLS connection established from mx2.freebsd.org[8.8.178.116]: TLSv1.2 with cipher AECDH-AES256-SHA (256/256 bits)
```

metalefty:
Resolved by #426.

PR #426, opened 2016-09-21
TLS: log TLS version and cipher

metalefty:
Hi, I've implemented logging of TLS version and cipher suggested at #422.

The purpose to log it is to know which client use which SSL/TLS version and cipher suites and confirm how safe it is.  For example, the feature to disable SSLv3 and choose cipher suite has been implemented. Disabling vulnerable SSL protocols or cipher suites affects some clients.   Logging help us to know which client is affected when vulnerable SSL protocols or cipher suites are disabled.

I know all SSL or encryption relevant codes are written in `common/ssl_calls.c` and `libxrdp/xrdp_sec.c` , however, in order to log client's IP address and TLS information together, there's only option for me to implement it in `libxrdp/xrdp_rdp.c`.  

This is really a draft.  If there's better way to log it, please leave comment and help me.

### DIFF
--- a/common/ssl_calls.c
+++ b/common/ssl_calls.c
@@ -891,3 +891,17 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
     return g_sck_can_recv(sck, millis);
 }
 
+
+/*****************************************************************************/
+const char*
+ssl_get_version(const struct ssl_st *ssl)
+{
+    return SSL_get_version(ssl);
+}
+
+/*****************************************************************************/
+const char*
+ssl_get_cipher_name(const struct ssl_st *ssl)
+{
+    return SSL_get_cipher_name(ssl);
+}
--- a/common/ssl_calls.c
+++ b/common/ssl_calls.c
@@ -893,14 +893,14 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
 
 
 /*****************************************************************************/
-const char*
+const char *
 ssl_get_version(const struct ssl_st *ssl)
 {
     return SSL_get_version(ssl);
 }
 
 /*****************************************************************************/
-const char*
+const char *
 ssl_get_cipher_name(const struct ssl_st *ssl)
 {
     return SSL_get_cipher_name(ssl);
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -830,6 +830,7 @@ xrdp_rdp_incoming(struct xrdp_rdp *self)
 
     /* log TLS version and cipher when TLS is used */
     /* TODO: client_addr, client_port is empty when IPv6 enabled */
+
     if (iso->selectedProtocol > PROTOCOL_RDP)
     {
         log_message(LOG_LEVEL_INFO,
@@ -837,6 +838,13 @@ xrdp_rdp_incoming(struct xrdp_rdp *self)
                     self->client_info.client_addr,
                     iso->trans->ssl_protocol,
                     iso->trans->cipher_name);
+    }
+    else
+    {
+        log_message(LOG_LEVEL_INFO,
+                    "Non-TLS connection established from %s: "
+                    "encrypted with standard RDP security",
+                     self->client_info.client_addr);
     }
 
     return 0;
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -811,6 +811,9 @@ xrdp_rdp_send_data_update_sync(struct xrdp_rdp *self)
 int APP_CC
 xrdp_rdp_incoming(struct xrdp_rdp *self)
 {
+    struct xrdp_iso *iso;
+    iso = self->sec_layer->mcs_layer->iso_layer;
+
     DEBUG(("in xrdp_rdp_incoming"));
 
     if (xrdp_sec_incoming(self->sec_layer) != 0)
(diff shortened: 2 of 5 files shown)


common/ssl_calls.c
```diff
@@ -866,3 +866,17 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
     return g_sck_can_recv(sck, millis);
 }
 
+
+/*****************************************************************************/
+const char
+*ssl_get_version(const struct ssl_st *ssl)
```
proski: Formatting - please move the initial "*"  to the previous line, it's a part of the return type.

common/ssl_calls.c
```diff
@@ -866,3 +866,17 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
...
+{
+    return SSL_get_version(ssl);
+}
+
+/*****************************************************************************/
+const char
+*ssl_get_cipher_name(const struct ssl_st *ssl)
```
proski: Same thing here.

libxrdp/xrdp_rdp.c
```diff
@@ -812,6 +812,9 @@ xrdp_rdp_send_data_update_sync(struct xrdp_rdp *self)
 int APP_CC
 xrdp_rdp_incoming(struct xrdp_rdp *self)
 {
+    struct xrdp_iso *iso;
+    iso = self->sec_layer->mcs_layer->iso_layer;
```
proski: You can use the "iso" variable throughout this function.

libxrdp/xrdp_rdp.c
```diff
@@ -827,6 +830,18 @@ xrdp_rdp_incoming(struct xrdp_rdp *self)
     g_strncpy(self->client_info.client_port,
               self->sec_layer->mcs_layer->iso_layer->trans->port,
               sizeof(self->client_info.client_port) - 1);
+
+    /* log TLS version and cipher when TLS is used */
+    /* TODO: client_addr, client_port is empty when IPv6 enabled */
+    if (iso->selectedProtocol > PROTOCOL_RDP)
```
proski: Even if it's not a TLS connection, you can still log it. Perhaps you should log selectedProtocol.

proski: Yes, unless it's already logged somewhere. And if it's logged, please consider having both log calls in the same place.

libxrdp/xrdp_rdp.c
metalefty: done.

metalefty:
All I wanted to do is done. Standard RDP security can use 3DES or RC4 but I don't log which is in use at this time.

metalefty:
Merging this before #542 makes it easy to test which SSL protocol is actually in use.

common/ssl_calls.c
```diff
@@ -891,3 +891,17 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
     return g_sck_can_recv(sck, millis);
 }
 
+
+/*****************************************************************************/
+const char*
```
proski: Please add space before `*`

common/ssl_calls.c
```diff
@@ -891,3 +891,17 @@ ssl_tls_can_recv(struct ssl_tls *tls, int sck, int millis)
...
+ssl_get_version(const struct ssl_st *ssl)
+{
+    return SSL_get_version(ssl);
+}
+
+/*****************************************************************************/
+const char*
```
proski: Please add space before `*`

libxrdp/xrdp_rdp.c
```diff
@@ -820,12 +823,30 @@ xrdp_rdp_incoming(struct xrdp_rdp *self)
...
+    g_strncpy(self->client_info.client_port, iso->trans->port,
               sizeof(self->client_info.client_port) - 1);
+
+    /* log TLS version and cipher when TLS is used */
+    /* TODO: client_addr, client_port is empty when IPv6 enabled */
+
+    if (iso->selectedProtocol > PROTOCOL_RDP)
+    {
+        log_message(LOG_LEVEL_INFO,
+                    "TLS connection established from %s: %s with cipher %s",
```
proski: How about replacing the initial "TLS" with the actual protocol name? It could be like "TLSv1.2 connection with cipher AES256-GCM-SHA384 established from 127.0.0.1:5678". How about logging the port? It's rarely useful, but it's just a few bytes in the log.

libxrdp/xrdp_rdp.c
```diff
@@ -820,12 +823,30 @@ xrdp_rdp_incoming(struct xrdp_rdp *self)
...
+                    self->client_info.client_addr,
+                    iso->trans->ssl_protocol,
+                    iso->trans->cipher_name);
+    }
+    else
+    {
+        log_message(LOG_LEVEL_INFO,
+                    "Non-TLS connection established from %s: "
```
proski: Likewise, port could be added here.

libxrdp/xrdp_rdp.c
metalefty: As described in #422, I was inspired by postfix. The log format is brought from postfix.  I like it because protocol name is put beside cipher. Yours is not bad but protocol name and cipher are a little bit distant for me.

I agree with logging the port.  The IP address part could be IPv6.  `127.0.0.1 port 5678` `::1 port 5678` like other logs would be fine.

proski: Good point about IPv6.

proski:
Ship it!

metalefty:
@jsorg71 @speidy Please merge if this looks good to you.

speidy:
+1

PR #434, opened 2016-10-08
Fixes for x11rdp build

proski:
I'm trying to add DESTDIR support to buildx.sh, which would allow packaging it without sudo and other dirty magic. I'm not including that change, but I'm including smaller improvements that I made while working on a more ambitious goal.

buildx.sh had implicit dependencies on an existing X11 installation. By reordering the packages, those implicit dependencies are removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.a
 aclocal.m4
 AUTHORS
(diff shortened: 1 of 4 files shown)


metalefty:
Looks good. No problem to build.

PR #437, opened 2016-10-10
Switch to 8 hex digit keymapfile

metalefty:
as the value keyboardLayout is actually 32-bit unsigned integer.
See [MS-RDPBCGR] - v20160714 p.45.


metalefty:
I and proski suggested switching to 8 hex digit format for all km*.ini files.

metalefty:
Chase #453.

metalefty:
@jsorg71 @itamarjp what do you think of?

itamarjp:
for me looks good to keep in sync with https://goo.gl/4rQByJ

proski:
Looks good to me.

metalefty:
Jay, could you merge this if you agree to switch?

PR #441, opened 2016-10-14
Add support for input event unicode

ricardoteixeiraduarte:
Adds support for INPUT_EVENT_UNICODE


ricardoteixeiraduarte:
I'm not very happy with the way I'm handling the "shift" key.
The current code sends a Shift KEY_DOWN (scancode 42) before sending the real key, and then a KBD_FLAG_UP after the key is processed.

jsorg71:
This might be a good start but this is only one keymap.
A better solution might be to use the keymap files.
xrdp/instfiles/km-0409.ini for example, Key10=49:49, the last value, 49, is the unicode, the first value, 10(from key10), is the keycode, the second value which does not matter here is the x11 keysym.
The [noshift] section, just send as they are, the [shift] section send a shift first, even [altgr] should work, just send altgr first.
xrdp/lang.c is were we have been processing these files.

jsorg71:
```
+    if (!s_check_rem(s, 1))
+    {
+        return 1;
+    }
+    in_uint8(s, code); /* keyCode (1 byte) */
```
This is wrong, the unicode value should be 2 bytes

jsorg71:
The caps bits is missing from this patch
We need to set INPUT_FLAG_UNICODE.
Without it, the client should not send unicode input.

jsorg71:
@ricardoteixeiraduarte  can you check if 2c3d4a9 work for you.
I tested with MS client on Android.  I can't get Parallels client working.

ricardoteixeiraduarte:
Working great.

metalefty:
@jsorg71 @ricardoteixeiraduarte  Implemented  in 2c3d4a9, so can this be closed? Or anything else neeed to be done?

jsorg71:
This is done

PR #445, opened 2016-10-17
Avoid using non-standard Automake variables starting with "EXTRA"

proski:
EXTRA_DIST is a documented Automake variable. Other variables stating with "EXTRA" are not. They are hard on the eye and their names are not descriptive (especially EXTRA_FLAGS). In some files, like xrdpvr/Makefile.am, those variables create clutter without providing any value.

Use AM_CPPFLAGS and AM_LDFLAGS directly for preprocessor and linker flags.

For extra libraries, use variables based on the executable or library that uses them, e.g. LIBXRDP_EXTRA_LIBS. (I could avoid even that, but I would have to reorder the variables more than I would like to)


itamarjp:
looks good for me

PR #450, opened 2016-10-21
Fixing memory leaks

proski:
Valgrind has found 4 memory leaks in xrdp-sesman. No leaks have been found in the xrdp binary. No memory leak is major, but it's good to fix them to keep our coding standards high.


metalefty:
Looks good.

metalefty:
Let's merge.

PR #454, opened 2016-10-25
chansrv: change chansrv log path to ${XDG_DATA_HOME}/xrdp

metalefty:
like Xorg's logfile is written to ${XDG_DATA_HOME}/xorg/Xorg.n.log.

If XDG_DATA_HOME is not defined, the log path will be
${HOME}/.local/share/xrdp.


metalefty:
I like Xorg's log path in user directory so I imitated it. What do you think of?

Natureshadow:
>If XDG_DATA_HOME is not defined, the log path will be
>${HOME}/.local/share/xrdp.

And if that doesn't exist?

metalefty:
${XDG_DATA_HOME}/xrdp/ or ${HOME}/.local/share/xrdp will be created if the directory doesn't exists.
If HOME environment variable is not defined, print error and exit (chansrv.c:1518).

metalefty:
If no objections, I'll merge this after a week.

PR #457, opened 2016-10-27
Fix timeout when using Xvnc

metalefty:
Rework of #263.

### DIFF
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -1042,7 +1042,16 @@ lib_mod_connect(struct vnc *v)
     v->trans->si = si;
     v->trans->my_source = XRDP_SOURCE_MOD;
 
-    error = trans_connect(v->trans, v->ip, con_port, 3000);
+    int count;
+    for (count = 0; count < 10; count++)
+    {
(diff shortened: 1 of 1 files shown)


proski:
Could you please provide some context? If using a loop, cannot we use shorter timeouts to minimize unneeded delays? Why do we need g_sleep(1000)? Why not start another connect attempt right away?

metalefty:
Xvnc ignited by xrdp-sesman is not immediately ready to accept connections.  The purpose of #263 and this is to wait and retry until Xvnc become to be ready.

As #263 includes other changes, I just separated them.  We need not to use sleep 1 sec.  We can use shorter delay as you say.

proski:
I looked at the implementation of trans_connect, and I see that it already has a loop. But it waits in 1/5th intervals of the timeout, so increasing the timeout would increase wasted time. I see that all trans_connect calls use 3000 (milliseconds). Maybe we could change the inner polling interval to 500 milliseconds (just a bit less than the current 3000/5=600) and increase the timeout for this call, say, to 15 seconds?

Issue #424, opened 2016-09-17
Issue compiling on OSX 10.11.6


Leesam:
Pastebin log: http://pastebin.com/0sb4uXrB

Cursory googling suggests this may be related to a newer version of Openssl. I have OpenSSL 1.1.0 installed.

Natureshadow:
Probably fixed in #459.

Leesam:
Yes, it did fix my issue

PR #459, opened 2016-10-27
Fix ssl_calls for OpenSSL 1.1.0

Natureshadow:


itamarjp:
some people reported that this fixed the bug https://github.com/neutrinolabs/xrdp/issues/424

itamarjp:
looks good for me.

proski:
I like that you code for the new API and then backport. The new API is always a better target.

But I would prefer if you used inline functions for the backport instead of adding ifdefs throughout the code.  So you can create a function `HMAC_CTX_new` for older OpenSSL and put the backport code there. The backport functions could go to a separate header.

I worked on a project that needed to support multiple versions of dependencies, and this approach proved to be most maintainable in the long run.

Natureshadow:
>But I would prefer if you used inline functions for the backport
>instead of adding ifdefs throughout the code.  So you can create a
>function `HMAC_CTX_new` for older OpenSSL and put the backport code
>there.

I tried that, but it turned out that the real implementations needed more and more inline functions (like EVP_MD_CTX_*) and it turned out that this approach is better in that case.

I expect to drop support for OpenSSL < 1.1.0 some time soon anyway.

Natureshadow:
Also mind that this way, users of old OpenSSL get the same code as before.

proski:
I've made a change to show what I mean. I actually tested that with OpenSSL 1.0.2j and 1.1.0b. Please see commit 8df61b9765ffec32dff68a6f24a5e9baf70d4ed9

Functions are stricter than macros in argument checking, and even macros are better than the conditional code spread all around the place. The functions can only use their arguments, not arbitrary variables from the caller. The arguments have types and descriptive names.

The main code uses the same logic regardless of the OpenSSL version, which makes development safer. When the backport code is all over the place, it's easy to copy some code and break support for a specific version of the dependency.

I don't know what was the "real" implementation you attempted. The backport functions should have the old code, but with the new API.

I actually expect OpenSSL 1.0 to stick around for a while. Sure, it would be patched for vulnerabilities, but its API would not change.

proski:
Actually, let's apply this PR, and I'll send my patch on top of it.

PR #460, opened 2016-10-28
common: Fix client IP address logging when IPv6 is enabled

metalefty:
g_tcp_accept() and g_sck_accept() should use sockadd_in6 when IPv6 is
enabled. The former code logs client IP address always "0.0.0.0" in such
case.

Fixes #412.

PR #461, opened 2016-10-28
common: If IPv6 not supported, fall back to IPv4

### LINKED ISSUE #432: Does not start with IPv6 disabled
Disabling IPv6 completely (using `ipv6.disable=1`) on the Linux command line) leads to xrdp not being able to start.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=839266 for more details and logs!

metalefty:
The system to run xrdp does not necessarily support IPv6 even though
it is compiled with IPv6.

Fixes #432.
 Issue #460 needs to be merged before merging this.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -416,11 +416,27 @@ g_tcp_socket(void)
 
 #if defined(XRDP_ENABLE_IPV6)
     rv = (int)socket(AF_INET6, SOCK_STREAM, 0);
+    if (rv < 0)
+    {
+        log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
+
+        switch (errno)
+        {
+            case EAFNOSUPPORT: /* if IPv6 not supported, retry IPv4 */
+                log_message(LOG_LEVEL_INFO, "IPv6 not supported, falling back to IPv4");
+                rv = (int)socket(AF_INET, SOCK_STREAM, 0);
+                break;
+
+            default:
+                return -1;
+        }
+    }
 #else
     rv = (int)socket(AF_INET, SOCK_STREAM, 0);
 #endif
     if (rv < 0)
     {
+        log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
         return -1;
     }
 #if defined(XRDP_ENABLE_IPV6)


metalefty:
Just fixed indent and force pushed.

common/os_calls.c
```diff
@@ -416,11 +416,27 @@ g_tcp_socket(void)
 
 #if defined(XRDP_ENABLE_IPV6)
     rv = (int)socket(AF_INET6, SOCK_STREAM, 0);
+    if (rv < 0)
+    {
+        log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", strerror(errno));
```
proski: It turns out were are using g_get_strerror() instead on strerror(errno). I was wrong to use the later in my socket patches, I'll fix it shortly. strerror() has issues beyond portability, it it not thread safe. I expect we would switch to g_strerror() from glib (or whatever apr has), so having some uniformity in the pre-exisitng code would help.

common/os_calls.c
```diff
@@ -416,11 +416,27 @@ g_tcp_socket(void)
...
+
+        switch(errno)
+        {
+            case EAFNOSUPPORT: /* if IPv6 not supported, retry IPv4 */
+                log_message(LOG_LEVEL_INFO, "IPv6 not supported, falling back to IPv4");
```
proski: I have some doubts about it. Should the caller of g_tcp_socket() decide whether to use IPv6?

common/os_calls.c
metalefty: Changed to g_get_strerror() and force pushed.

common/os_calls.c
proski: I mean the caller of the socket may have an idea whether it wants an IPv4 socket or and IPv6 socket, based on what it wants to do to the socket (which may depend on the configuration). But I looked closer at the code. It looks like os_calls.c tries to make IP protocol selection transparent to the caller for all calls. It remains to be seen how well it would work, but it's pre-existing code.

I have no objections against your patch. Just please add space after `switch` and after `(int)`

metalefty:
Formatting changes, rebased to current devel, force pushed.

metalefty:
Generally, release versions will be used different environment from it is built.
The issue occurs when IPv6 enabled pre-built xrdp runs on IPv4 only environment.

This should be included to next release.

PR #462, opened 2016-10-28
Place chansrv log in XDG_CACHE_HOME if it exists.

Natureshadow:


Natureshadow:
Oops. Forogt that this is a duplicate of #454 (besides this one uses XDG_CACHE_HOME instead, which, in my opinion, is a better choice).

metalefty:
Need to determine which one to merge.

metalefty:
Merged as #454.  XDG_CACHE_HOME also makes sense but I decided to use ${XDG_DATA_HOME}/xrdp at this time since Xorg uses the similar directory.  I'd welcome further discussion if you strongly recommend XDG_CACHE_HOME. Then please reopen or create a new issue.

PR #464, opened 2016-10-28
Fix permission issues between xrdp and sesman.

Natureshadow:
Attempt to fix connection problem between xrdp-sesman (running as root) and
xrdp (running as xrdp) by creating the socket for the xorgxrdp server as
0660 in a sgid-xrdp directory, since we can’t easily chgrp here; also fix
retval check for listen(2)

This is accompanied by a patch to xorgxrdp (did I mention the two should be
merged as subtrees?).

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -122,7 +122,7 @@ g_mk_temp_dir(const char *app_name)
                 return 1;
             }
         }
-        g_chmod_hex("/tmp/.xrdp", 0x1777);
+        g_chmod_hex("/tmp/.xrdp", 0x3777);
     }
     return 0;
 }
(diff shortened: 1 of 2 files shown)


proski:
Such changes are better left in the distro patches. Different users use xrdp differently. I don't think any sgid or suid bits should be necessary to make two processes communicate. Changing the /tmp/.xrdp is too radical, there are several sockets in that directory. It would be better to change permissions of specific sockets. Whichever process is root should be changing permissions for the other process to have access. If UNIX sockets don't work, TCP sockets should be used.

metalefty:
This should be left in distro patches.

PR #465, opened 2016-10-28
make FUSE mount root hidden from normal ls

Natureshadow:


moobyfr:
This merge has to be reviewed: there is a config variable in sesman.ini for the mountpoint named 'FuseMountName'. The code at line 400 from [chansrv_fuse.c](https://github.com/neutrinolabs/xrdp/blob/devel/sesman/chansrv/chansrv_fuse.c#L400) uses it. So for the spirit of this patch, the dot should be added in sesman.ini
For the idea itself, I don't understand the need to hide the mountpoint for the user, if he needs to access the share, don't hide it IMHO.

Natureshadow:
You are right. I believe this is the remainder of an old patch that was refreshed to a new code base and is now unnecessary.

I will check that.

metalefty:
I also think the mountpoint shouldn't be hidden.  Of course you can if you want but I don't think it should be upstreamed.  IMO, patch the config locally in distro package if you want to do that.

metalefty:
Mount point shouln't be hidden.

PR #467, opened 2016-10-28
Miscelleanous fixes.

Natureshadow:
Build system fixes
Code quality fixes (fixes compile time warnings and undefined behaviour)

Accompanied by a patch to xorgxrdp (did I mention the two should probably be
subtree-merged ;)?).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,6 +24,7 @@ SUBDIRS = \
   mc \
   $(NEUTRINORDPDIR) \
   libxrdp \
+  librfxcodec \
   xrdp \
   sesman \
   keygen \
@@ -31,4 +32,5 @@ SUBDIRS = \
   instfiles \
   genkeymap \
   xrdpapi \
+  xorgxrdp \
   $(XRDPVRDIR)
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -47,7 +47,12 @@ libcommon_la_SOURCES = \
   xrdp_rail.h \
   $(PIXMAN_SOURCES)
 
+if NEED_LIBDL
+EXTRA_LIBDL = -ldl
+endif
+
 libcommon_la_LIBADD = \
+  $(EXTRA_LIBDL) \
   -lcrypto \
   -lssl \
   -lpthread
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,14 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
   [AC_MSG_ERROR([please install libxrandr-dev or libXrandr-devel])],
   [#include <X11/Xlib.h>])
 
+# checking if we need libdl
+save_LIBS=$LIBS
+AC_SEARCH_LIBS([dlopen], [dl dld], [], [
+  AC_MSG_ERROR([unable to find the dlopen() function])
+])
+AM_CONDITIONAL(NEED_LIBDL, [test x"$LIBS" != x"$save_LIBS"])
+LIBS=$save_LIBS
+
 CFLAGS="$save_CFLAGS"
 
 AC_SUBST([moduledir], '${libdir}/xrdp')
--- a/docs/man/xrdp-dis.1
+++ b/docs/man/xrdp-dis.1
@@ -1,4 +1,4 @@
-.TH "xrdp-dis" "8" "0.7.0" "xrdp team"
+.TH "xrdp-dis" "1" "0.7.0" "xrdp team"
 .SH NAME
 xrdp\-dis \- xrdp disconnect utility
 
--- a/libxrdp/xrdp_jpeg_compress.c
+++ b/libxrdp/xrdp_jpeg_compress.c
@@ -213,7 +213,7 @@ xrdp_jpeg_deinit(void *handle)
 
 struct mydata_comp
 {
-    char *cb;
+    JOCTET *cb;
     int cb_bytes;
     int total_done;
     int overwrite;
@@ -265,8 +265,8 @@ my_term_destination(j_compress_ptr cinfo)
 
 /*****************************************************************************/
 static int APP_CC
-jp_do_compress(char *data, int width, int height, int bpp, int quality,
-               char *comp_data, int *comp_data_bytes)
+jp_do_compress(JOCTET *data, int width, int height, int bpp, int quality,
+               JOCTET *comp_data, int *comp_data_bytes)
 {
     struct jpeg_compress_struct cinfo;
     struct jpeg_error_mgr jerr;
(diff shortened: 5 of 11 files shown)


Makefile.am
```diff
@@ -24,11 +24,13 @@ SUBDIRS = \
   mc \
   $(NEUTRINORDPDIR) \
   libxrdp \
+  librfxcodec \
```
proski: Users who checkout xrdp, don't get librfxcodec by default. They would get a broken experience. That's really too big of a topic to be part of "miscellaneous fixes"

Makefile.am
```diff
@@ -24,11 +24,13 @@ SUBDIRS = \
...
   xrdp \
   sesman \
   keygen \
   docs \
   instfiles \
   genkeymap \
   xrdpapi \
+  xorgxrdp \
```
proski: Likewise, we need to decide if we want xorgxrdp to be part of the tree. I believe most users would want to use the `/usr` prefix for xorgxrdp to make it work, yet they make want to use `/usr/local` or other prefix for xrdp. I would rather have xorgxrdp as a completely separate package.

common/Makefile.am
```diff
@@ -47,7 +47,12 @@ libcommon_la_SOURCES = \
   xrdp_rail.h \
   $(PIXMAN_SOURCES)
 
+if NEED_LIBDL
+EXTRA_LIBDL = -ldl
+endif
+
 libcommon_la_LIBADD = \
+  $(EXTRA_LIBDL) \
```
proski: EXTRA_LIBDL is a poor name. I would prefer OPENSSL_LIBS, which would get `-ldl` added to it conditionally. A bigger issue is that we should support pkg-config for OpenSSL and take the settings from there.

configure.ac
```diff
@@ -246,6 +246,14 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
   [AC_MSG_ERROR([please install libxrandr-dev or libXrandr-devel])],
   [#include <X11/Xlib.h>])
 
+# checking if we need libdl
```
proski: That message should be printed for users to show them what configure is doing. Again, this may not be needed if we use pkg-config to discover OpenSSL.

libxrdp/xrdp_jpeg_compress.c
```diff
@@ -265,8 +265,8 @@ my_term_destination(j_compress_ptr cinfo)
 
 /*****************************************************************************/
 static int APP_CC
-jp_do_compress(char *data, int width, int height, int bpp, int quality,
-               char *comp_data, int *comp_data_bytes)
+jp_do_compress(JOCTET *data, int width, int height, int bpp, int quality,
```
proski: If cleaning up this code, you may also want to add `const` to the pointer (unless it leads to a cascade of changes, in which case it would need to be a separate patch).

sesman/chansrv/chansrv_fuse.c
```diff
@@ -265,7 +272,9 @@ static void xfuse_create_file(fuse_req_t req, fuse_ino_t parent,
 #endif
 
 static void xfuse_dump_fs(void);
+#if 0
```
proski: Please don't do it. It the code is not needed, please remove it. Don't let it rot. If the commented out code is needed (e.g. it illustrates something), please also add an explanation to the code.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -122,6 +128,7 @@ void xfuse_devredir_cb_file_close(void *vp)                                  {}
 #define LOG_ERROR   0
 #define LOG_INFO    1
 #define LOG_DEBUG   2
+#undef LOG_LEVEL
```
proski: Why do you need #undef here? Do you get a warning?

sesman/chansrv/chansrv_fuse.c
```diff
@@ -1084,20 +1101,22 @@ static void xfuse_dump_fs()
...
  *****************************************************************************/
 
+#if 0
 static void xfuse_dump_xrdp_inode(struct xrdp_inode *xino)
 {
     log_debug("--- dumping struct xinode ---");
     log_debug("name:          %s", xino->name);
-    log_debug("parent_inode:  %ld", xino->parent_inode);
-    log_debug("inode:         %ld", xino->inode);
+    log_debug("parent_inode:  %ld", (long)xino->parent_inode);
```
proski: Why? Did you get a warning? What's your compiler? Unneeded casts are evil, then can hide issues.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -2850,9 +2873,9 @@ static void xfuse_cb_setattr(fuse_req_t req, fuse_ino_t ino, struct stat *attr,
 
     if (to_set & FUSE_SET_ATTR_SIZE)
     {
-        log_debug("previous file size: %d", attr->st_size);
+        log_debug("previous file size: %lld", (long long)attr->st_size);
```
proski: Why cannot you use the `z` modifier here? Please provide some context for your environment.

sesman/session.c
```diff
@@ -617,7 +617,7 @@ session_start_fork(int width, int height, int bpp, char *username,
                 else
                 {
                     log_message(LOG_LEVEL_ERROR, "another Xserver might "
-                                "already be active on display %d - see log", display);
+                                "already be active on display %d - see log; possibly comment out the two param lines -logfile /dev/null in /etc/xrdp/sesman.ini to debug", display);
```
proski: I would rather have the log enabled by default, so that we don't have to show such instructions.

xrdp/Makefile.am
```diff
@@ -19,7 +19,7 @@ endif
 if XRDP_RFXCODEC
 AM_CPPFLAGS += -DXRDP_RFXCODEC
 AM_CPPFLAGS += -I$(top_srcdir)/librfxcodec/include
-XRDP_EXTRA_LIBS += $(top_srcdir)/librfxcodec/src/.libs/librfxencode.a
+XRDP_EXTRA_LIBS += $(top_srcdir)/librfxcodec/src/.libs/librfxencode.la
```
proski: You don't want `.libs` here if using `*.la`, it was a hack to force static linking. Again, this is not a cleanup material.

metalefty:
Some cleanups in this PR are useful but needs to be more separated.

docs/man/xrdp-dis.1
```diff
@@ -1,4 +1,4 @@
-.TH "xrdp-dis" "8" "0.7.0" "xrdp team"
+.TH "xrdp-dis" "1" "0.7.0" "xrdp team"
```
metalefty: Fixed as #488.

libxrdp/xrdp_jpeg_compress.c
metalefty: Fixed as #503.

metalefty:
The purpose of downstream patches should be disclosed.
Please do not just throw them to upstream. They need to be separated by purpose.

sesman/chansrv/devredir.h
```diff
@@ -101,6 +101,8 @@ int dev_redir_file_open(void *fusep, tui32 device_id, char *path,
 
 int devredir_file_close(void *fusep, tui32 device_id, tui32 file_id);
 
+int APP_CC devredir_rmdir_or_file(void *fusep, tui32 device_id, char *path, int mode);
```
proski: Fixed in #502

PR #470, opened 2016-10-31
Simplify names of xrdp.ini sections, rename "Session Manager" to "Xorg"

proski:
Make section names equal to the names in the "name=" setting to avoid
confusion and the need to renumber sections. Avoid "sesman-" in the
names, it's not helpful to the user. Move "Xorg" just below "X11rdp",
that would give xorgxrdp more visibility.

### DIFF
--- a/xrdp/xrdp.ini
+++ b/xrdp/xrdp.ini
@@ -56,7 +56,7 @@ grey=dedede
 
 # Section name to use for automatic login if the client sends username
 # and password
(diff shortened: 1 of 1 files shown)


proski:
Correction: fixed capitalization in the default section name.

Note to myself: update https://github.com/neutrinolabs/xrdp/wiki/Compiling-rdp-driver once merged.

PR #472, opened 2016-11-01
Use static inline functions for OpenSSL 1.0 backport

proski:
Conditional preprocessor directives spread throughout the code set a bad
example.

The new backport code is located in one place. The compiler checks
argument types. The backport code has no access to the caller variables.
The main code has all advantages of the new, more compact API.


itamarjp:
thank you,

PR #478, opened 2016-11-07
Install headers used by X11rdp and xorgxrdp

proski:
Installing the headers makes it possible to compile xorgxrdp as a
separate package, without xrdp sources.


metalefty:
+1

PR #479, opened 2016-11-08
Use g_get_strerror() instead of strerror(errno) for portability

proski:
g_get_strerror() is simply returning strerror(errno) for now, except Windows. strerror() is not thread safe, we should replace it across the board. This PR simply restores consistency.


jsorg71:
+1

PR #487, opened 2016-11-13
X11rdp build system fixes

proski:
Stripping X11rdp is unhelpful for anyone trying to debug it or run it in Valgrind. Let's keep the uninstalled binary unstripped.

Saying in a comment that X11RDPBASE needs to be set is not enough, let's check that it's set.

Linking static libraries two times is sloppy. It's not hard to reorder them properly.


itamarjp:
looks good for me, please merge.

Issue #428, opened 2016-09-30
New method needed to disable specific SSL/TLS version


metalefty:
The feature disabling SSL 3.0 has been implemented.  We don't have to disable TLS 1.0 right now but usage of TLS 1.1 should also be configurable. Because PCI DSS v3.1 restricts usage of TLS 1.0, for example.

In future, higher TLS version such as TLS 1.1, TLS 1.2 might be prohibited.  In such situation,  expanding ssl calls like this is stupid.

``` ini
xrdp.ini:
disableSSLv3=true
disableTLSv1_0=true
disableTLSv1_1=true
disableTLSv1_2=true
```

``` c
int APP_CC
ssl_tls_accept(struct ssl_tls *self, 
               int disableSSLv3, int disableTLSv1_0, int disableTLSv1_1, disableTLSv1_2,
               const char *tls_ciphers);
```

So we need a setting like ssl_protocols in xrdp.ini.  Only SSL protocols listed in ssl_protocols will be used, the protocol not in list will be prohibited.

``` ini
ssl_protocols=TLSv1.0 TLSv1.1 TLSv1.2 TLSv1.3
tls_ciphers=HIGH
```

metalefty:
I submitted #542.

PR #488, opened 2016-11-15
Update man pages and config files

metalefty:
Lots of changes. Topics are:

- Add missing descriptions
- Update version
- Replace xrdp website links
- Use two different symbols to comment out *.ini files
  - semicolon for descriptions in config files
  - number sign for actual configurations
- Change order of parameters in xrdp.ini
  - put same layer configuration parameters together
    - TCP
    - security layer
    - RDP features (such as bpp, channels, compression, fast path, multimon)
  - bring lower layer earlier
- Unify mixed boolean value expression to true/false

Documents are not still complete but big one step to keep up-to-date.

### DIFF
--- a/docs/man/xrdp-genkeymap.8
+++ b/docs/man/xrdp-genkeymap.8
@@ -64,4 +64,4 @@ Simone Fedele <ilsimo@users.sourceforge.net>
 .BR unicode (7),
 .URL "https://github.com/FreeRDP/FreeRDP/wiki/Keyboard" "Description of Keyboard Input mapping" .
 
-for more info on \fBxrdp\fR see http://xrdp.sf.net
+for more info on \fBxrdp\fR see http://www.xrdp.org/
--- a/docs/man/xrdp-genkeymap.8
+++ b/docs/man/xrdp-genkeymap.8
@@ -26,31 +26,31 @@ Files containing the keyboard mapping for language \fIXXXXXXXX\fP, which is a 8 
 .RS 8
 .TP
 .B 00000405
-cs czech
+cs Czech
 .TP
 .B 00000407
-de german
+de German
 .TP
 .B 00000409
-en-us us english
+en-us US English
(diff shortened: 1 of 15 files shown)


metalefty:
I'm not a native English speaker. Feel free to correct grammar and expressions.

metalefty:
Rebased to current devel.

docs/man/xrdp-genkeymap.8
```diff
@@ -26,31 +26,31 @@ Files containing the keyboard mapping for language \fIXXXXXXXX\fP, which is a 8
...
-en-us us english
+en-us US English
 .TP
 .B 0000040c
-fr french
+fr French
 .TP
 .B 00000410
-it italy
+it Italy
```
proski: Italian

xrdp/xrdp.ini
```diff
@@ -1,43 +1,57 @@
...
+; can be 'none', 'low', 'medium', 'high', 'fips'
+crypt_level=high
+; X.509 certificate and private key
+; openssl req -x509 -newkey rsa:2048 -nodes -keyout key.pem -out cert.pem -days 365
 certificate=
 key_file=
-# disable SSlv3
-#disableSSLv3=yes
-# set TLS cipher suites
+; regulate xrdp if to accept SSLv3 connections
```
proski: "specify whether SSLv3 should be disabled"

Aside, we should avoid all "negative" options, they require a mental negation. "enableSSLv3" would have been better.

metalefty: I agree.  Not only negate the option,  specifying SSL/TLS version should be more flexible as I mentioned it once in #428.

metalefty:
I'll merge this a week later if no more review comments.

proski:
No need to wait that long. It's just documentation. Others may want to improve on top of your changes.

PR #489, opened 2016-11-15
CI: add .travis.yml

metalefty:
Let's introduce continuous integration in order to automatically test build of pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,42 @@
+sudo: false # use new container
+language: c
+
+branches:
+  except: /^(gh-pages|v[0-9]\..*)/
+
+compiler:
+  - clang
+  - gcc
+
+addons:
+  apt:
+    packages:
+      # xrdp
+      - libpam0g-dev
+      - libjpeg-dev
+      - libfuse-dev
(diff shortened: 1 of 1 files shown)


metalefty:
In order to run build tests at xrdp repo, neutrinolabs owner needs to activate repo at travis-ci.org.
Build for this PR succeeds on my fork: https://travis-ci.org/metalefty/xrdp/builds/175988615

speidy:
I have activated it @ travis-ci.org

proski:
Is that our recommended set of options? We should try to enable them if the needed libraries/functions exist.

I have fixes for FUSE warnings in my fuse branch, I'll split them out and submit. xorgxrdp should use silent rules by default to reduce noise.

If would be nice to compile xrdp with g++ as well. But xorgxrdp is not compilable with C++, as the X server headers are heavily abusing `class`, `private` and other C++ keywords. Maybe xorgxrdp should a separate travis file?

metalefty:
The options are my personal recommended set. 


For xorgxrdp, it's better to create a separate travis file in xorgxrdp repo.  It can test pull requests to xorgxrdp repo.

PR #490, opened 2016-11-16
Omit numbers from the "param" keys, they are not needed

proski:
Users assume that they need to renumber the parameters. That makes
parameter editing more involved than it needs to be.


metalefty:
+1

PR #491, opened 2016-11-16
Fix format warnings if FUSE is enabled

proski:
All code changes are in the logs. That fixes the bulk of the warnings in the FUSE code.


metalefty:
I think this can be included to next release.

PR #492, opened 2016-11-16
Do not create SHA-1 certificates anymore

metalefty:
as many systems in the world still default to create SHA-1 certs if
hash algorithm is not specified explicitly.


jsorg71:
+1

PR #497, opened 2016-11-19
Vncpassfile

jsorg71:
This fixes an old VNC security issue were the VNC password file is based on the user password.
Now it is derived from a random GUID that is generated for each session.
This GUID can be used in the future for session reconnect and other things.

### DIFF
--- a/sesman/libscp/libscp_session.c
+++ b/sesman/libscp/libscp_session.c
@@ -416,6 +416,32 @@ scp_session_set_addr(struct SCP_SESSION *s, int type, const void *addr)
 }
 
 /*******************************************************************/
+int
+scp_session_set_guid(struct SCP_SESSION *s, const char *str)
+{
+    if (0 == str)
+    {
+        log_message(LOG_LEVEL_WARNING, "[session:%d] set_guid: null guid", __LINE__);
+        return 1;
+    }
+
+    if (0 != s->guid)
+    {
+        g_free(s->guid);
+    }
+
+    s->guid = g_strdup(str);
+
+    if (0 == s->guid)
+    {
+        log_message(LOG_LEVEL_WARNING, "[session:%d] set_guid: strdup error", __LINE__);
+        return 1;
+    }
+
+    return 0;
+}
+
+/*******************************************************************/
 void
 scp_session_destroy(struct SCP_SESSION *s)
 {
@@ -428,5 +454,6 @@ scp_session_destroy(struct SCP_SESSION *s)
     g_free(s->client_ip);
     g_free(s->errstr);
     g_free(s->mng);
+    g_free(s->guid);
     g_free(s);
 }
--- a/sesman/libscp/libscp_session.c
+++ b/sesman/libscp/libscp_session.c
@@ -417,26 +417,15 @@ scp_session_set_addr(struct SCP_SESSION *s, int type, const void *addr)
 
 /*******************************************************************/
 int
-scp_session_set_guid(struct SCP_SESSION *s, const char *str)
+scp_session_set_guid(struct SCP_SESSION *s, const tui8 *guid)
 {
-    if (0 == str)
+    if (0 == guid)
     {
         log_message(LOG_LEVEL_WARNING, "[session:%d] set_guid: null guid", __LINE__);
         return 1;
     }
 
-    if (0 != s->guid)
-    {
-        g_free(s->guid);
-    }
-
-    s->guid = g_strdup(str);
-
-    if (0 == s->guid)
-    {
-        log_message(LOG_LEVEL_WARNING, "[session:%d] set_guid: strdup error", __LINE__);
-        return 1;
-    }
+    g_memcpy(s->guid, guid, 16);
 
     return 0;
 }
@@ -454,6 +443,5 @@ scp_session_destroy(struct SCP_SESSION *s)
     g_free(s->client_ip);
     g_free(s->errstr);
     g_free(s->mng);
-    g_free(s->guid);
     g_free(s);
 }
--- a/sesman/libscp/libscp_v0.h
+++ b/sesman/libscp/libscp_v0.h
@@ -61,7 +61,7 @@ scp_v0s_accept(struct SCP_CONNECTION* c, struct SCP_SESSION** s, int skipVchk);
  *
  */
 enum SCP_SERVER_STATES_E
-scp_v0s_allow_connection(struct SCP_CONNECTION* c, SCP_DISPLAY d, char *guid);
+scp_v0s_allow_connection(struct SCP_CONNECTION* c, SCP_DISPLAY d, tui8 *guid);
 
 /**
  *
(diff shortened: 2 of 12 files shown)


jsorg71:
Let me manually merge this if it all looks good.

sesman/libscp/libscp_session.c
```diff
@@ -416,6 +416,21 @@ scp_session_set_addr(struct SCP_SESSION *s, int type, const void *addr)
 }
 
 /*******************************************************************/
+int
+scp_session_set_guid(struct SCP_SESSION *s, const tui8 *guid)
```
proski: If tui8 a useful type? uint8_t is a standard type that is also unsigned 8-bit integer.

sesman/libscp/libscp_session.c
```diff
@@ -416,6 +416,21 @@ scp_session_set_addr(struct SCP_SESSION *s, int type, const void *addr)
...
+{
+    if (0 == guid)
```
proski: Let's use NULL for pointers, it's more readable.

sesman/libscp/libscp_v0.h
```diff
@@ -61,7 +61,7 @@ scp_v0s_accept(struct SCP_CONNECTION* c, struct SCP_SESSION** s, int skipVchk);
  *
  */
 enum SCP_SERVER_STATES_E
-scp_v0s_allow_connection(struct SCP_CONNECTION* c, SCP_DISPLAY d);
+scp_v0s_allow_connection(struct SCP_CONNECTION* c, SCP_DISPLAY d, tui8 *guid);
```
proski: I believe it should be a constant pointer.

sesman/scp_v0.c
```diff
@@ -99,6 +99,11 @@ scp_v0_process(struct SCP_CONNECTION *c, struct SCP_SESSION *s)
 
             if (1 == access_login_allowed(s->username))
             {
+                tui8 guid[16];
+
+                g_random((char*)guid, 16);
```
proski: Please use correct spacing.

sesman/session.c
```diff
@@ -410,7 +410,8 @@ wait_for_xserver(int display)
 static int APP_CC
 session_start_fork(int width, int height, int bpp, char *username,
                    char *password, tbus data, tui8 type, char *domain,
-                   char *program, char *directory, char *client_ip)
+                   char *program, char *directory, char *client_ip,
+                   const tui8 *guid)
```
proski: Would not it be easier to send the session pointed instead of so many fields?

vnc/vnc.c
```diff
@@ -53,11 +53,29 @@ void DEFAULT_CC
...
+    des = ssl_des3_encrypt_info_create(key, 0);
+    ssl_des3_encrypt(des, 8, bytes, bytes);
+    ssl_des3_info_delete(des);
+    des = ssl_des3_encrypt_info_create(key, 0);
+    ssl_des3_encrypt(des, 8, bytes + 8, bytes + 8);
+    ssl_des3_info_delete(des);
+}
+
+/******************************************************************************/
+/* taken from vncauth.c */
```
proski: Function comments should primarily describe what the function does, not where the code came from.

vnc/vnc.c
```diff
@@ -53,11 +53,29 @@ void DEFAULT_CC
...
+void DEFAULT_CC
+rfbHashEncryptBytes(char *bytes, char *passwd)
```
proski: Do both arguments need to be non-constant?

xrdp/xrdp_mm.c
```diff
@@ -1195,11 +1199,18 @@ xrdp_mm_process_login_response(struct xrdp_mm *self, struct stream *s)
     int rv;
     char ip[256];
     char port[256];
+    tui8 guid[16];
+    tui8* pguid;
```
proski: Formatting - `*` sticks to the variable, not to the type.

vnc/vnc.c
```diff
@@ -1422,6 +1447,11 @@ lib_mod_set_param(struct vnc *v, const char *name, char *value)
     {
         v->delay_ms = g_atoi(value);
     }
+    else if (g_strcasecmp(name, "guid") == 0)
```
proski: Why is a user configuration needed? Should it be documented?

sesman/libscp/libscp_session.c
jsorg71: Currently we are not using the *_t types, only the t* types in arch.h

sesman/libscp/libscp_session.c
jsorg71: I know what you mean here, the NULL means it's a pointer.  Currently, zeros are used all over the project.  I've never liked that NULL is defined differently sometimes.

sesman/libscp/libscp_v0.h
jsorg71: I agree.

sesman/scp_v0.c
jsorg71: ok

sesman/session.c
jsorg71: I was thinking that too, I'll take a look at it.

vnc/vnc.c
jsorg71: ok

vnc/vnc.c
jsorg71: This is the only API to give the modules general data.  We don't give the modules access to the entire xrdp internals.
It was not meant to be only user data.

xrdp/xrdp_mm.c
jsorg71: yup

sesman/libscp/libscp_session.c
proski: NULL is a special value for every variable that can be compared to it (i.e. every pointer). But 0 is not always a special value for a variable that can be compared to it. A variable can be 0 and be used, say, as an index. When I see comparison to 0, I need to examine it more closely. It can be a pointer check, or it can be something completely different.

vnc/vnc.c
proski: Ouch! This function needs a real comment.

vnc/vnc.c
proski: I don't understand. Should users configure it? If not, who should configure it? If nobody should configure it, why is it read from the config file?

speidy: lib_mod_set_param is not getting its params from config file only,
please look @ https://github.com/jsorg71/xrdp/blob/ed862ea361e1a364f153da2bd22c402f529b547e/xrdp/xrdp_mm.c#L568

This is the api for modules also for internal use.

proski: Oh, I see it now.

proski:
#523 has been merged. I assume this PR can be closed.

carnil:
Additional note for this pull request.

Originally reported to by Dag-Erling Smørgrav and Petter Reinholdtsen:

> When successfully logging in using RDP into a xrdp session, the file
> `~/.vnc/sesman_${username}_passwd` is created.  Its content is the
> equivalent of the users clear text password, DES encrypted with a known
> key.

This issue got assigned back then [CVE-2013-1430](https://security-tracker.debian.org/CVE-2013-1430)

PR #499, opened 2016-11-21
Use PKG_INSTALLDIR to determine pkg-config file install location

proski:
The PKG_INSTALLDIR macro also enables --with-pkgconfigdir for easy
overrides.

PKG_INSTALLDIR is relatively new, it appeared in pkg-config 0.27. To
ensure we have a recent pkg.m4 copy, copy pkg.m4 to the m4 directory.


metalefty:
I confirmed that I can override the directory with `--with-pkgconfigdir` option.

PR #502, opened 2016-11-22
Fix all FUSE warnings and C++ compatibility, add Mac compilation support

proski:
Bring the FUSE code to the same standards as the rest of the code. Specifically:

* Fix all warnings
* Fix all C++ errors and warnings
* Comment out or remove all dead code
* Use flags reported by pkg-config (that fixes Mac OS support)

Since FUSE is enabled in Travis CI checks and in Fedora rpm, it's important to maintain high quality of that code. Fixing pre-existing warnings will make new warnings stand out right away.

I tried to avoid any changes that would go beyond the stated goals. I believe the changes should be good for 0.9.1.


metalefty:
Looks good.

PR #504, opened 2016-11-22
Remove "-ac" from X server parameters, it's insecure and unnecessary

proski:
Fedora patches sesman.ini to remove "-ac", and that doesn't cause any ill
effects.

### DIFF
--- a/sesman/sesman.ini
+++ b/sesman/sesman.ini
@@ -64,7 +64,6 @@ SyslogLevel=DEBUG
 [X11rdp]
 param=X11rdp
 param=-bs
-param=-ac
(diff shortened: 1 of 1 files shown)


metalefty:
+1 I confirmed it works on FreeBSD.

mirabilos:
I can confirm that this one works without `-ac` too:

    [Xorg]
    param=Xorg
    param=-config
    param=xrdp/xorg.conf
    param=-noreset
    param=-nolisten
    param=tcp
    param=-retro

That’s Debian sid’s latest xrdp and xorgxrdp.

PR #513, opened 2016-11-29
Disable privilege elevation by setuid for the X server

proski:
The solution is not universal, but handy anyway, and it improves security. Using PR_SET_NO_NEW_PRIVS ensures that the setuid bit on Xorg would be ignored. Xorg can fail if run with root permissions, but the original user has no console access, as reported by PAM.

Tested on AWS t2.micro with Amazon Linux 2016.09.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -262,6 +262,8 @@ fi
 
 PKG_INSTALLDIR
 
+AC_CHECK_HEADERS([sys/prctl.h])
+
 AC_CONFIG_FILES([
   common/Makefile
   docs/Makefile
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -29,8 +29,21 @@
  *
  */
 
+#if defined(HAVE_CONFIG_H)
+#include "config_ac.h"
+#endif
+
+#ifdef HAVE_SYS_PRCTL_H
+#include <sys/prctl.h>
+#endif
+
 #include "sesman.h"
 #include "libscp_types.h"
+
+#ifndef PR_SET_NO_NEW_PRIVS
+#define PR_SET_NO_NEW_PRIVS 38
+#endif
+
 
 extern unsigned char g_fixedkey[8];
 extern struct config_sesman *g_cfg; /* in sesman.c */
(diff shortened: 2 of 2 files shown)


Natureshadow:
>The solution is not universal, but handy anyway, and it improves
>security.

Umm…

I figure the setuid is there for a reason. It is needed when consolekit is not available and there is no other way to get console access.

Or I get it wrong.

proski:
Root access is needed to access the the GPU. xorgxrdp doesn't need it.

Natureshadow:
>Root access is needed to access the the GPU. xorgxrdp doesn't need it.

OK… That would be great!

metalefty:
In conclusion, merging this PR and remove `-logfile /dev/null` from Xorg option will work?

proski:
No, it will break RHEL 6, where Xorg would try to write the log to `/var/log` regardless of any XDG variables. If we want a simple solution for now, we can do this:
```
param=-logfile
param=.xorgxrdp.%s.log
```
It would create files like `~/.xorgxrdp.10.log` in the home directory. Hidden files in the home directory are not that bad as non-hidden ones. That would give us two benefits:

- setuid Xorg would not object against using absolute log path if OS is not Linux or if Linux kernel is older than 3.5
- The logs would not be discarded, which is important for the first ever release of xorgxrdp

I'll make a PR for that.

proski:
Patch for log path submitted as #521

metalefty:
Awesome, it looks good to me. Let's include this to 0.9.1.

```
$ ls -l /usr/bin/Xorg
-rwsr-xr-x. 1 root root 2397160 Nov 20  2015 /usr/bin/Xorg
$ ps uxaww|grep Xorg
osc      15474  0.2  1.7 221088 32060 ?        Sl   17:11   0:00 Xorg :10 -config xrdp/xorg.conf -noreset -nolisten tcp -logfile .xorgxrdp.%s.log
```

metalefty:
If no objections, I'll merge this in 24 hours.

PR #514, opened 2016-11-30
Find openssl by pkg-config, add -ldl if needed

proski:
I checked OpenSSL as far back as 0.9.8e, and it has the openssl.pc file (generated by Makefile). The oldest supported version is 1.0.1u.

Some systems need `-ldl` to use `dlopen()`. libcommon calls `dlopen()`, it should not rely on other libraries linking `-ldl`.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -52,4 +52,5 @@ libcommon_la_SOURCES = \
 libcommon_la_LIBADD = \
   -lcrypto \
   -lssl \
-  -lpthread
+  -lpthread \
+  $(DLOPEN_LIBS)
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -18,6 +18,8 @@ AM_CPPFLAGS = \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
   -DXRDP_LOG_PATH=\"${localstatedir}/log\"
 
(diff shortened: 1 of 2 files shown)


metalefty:
I didn't notice that FreeBSD base doesn't supply `*.pc` files. Still have to use old style to discover OpenSSL for FreeBSD.  If users intentionally installed non default version of OpenSSL, `*.pc` files are supplied.

I'll create a patch later, or patch to `configure.ac` FreeBSD locally.

This is well known issue of FreeBSD.
https://github.com/libgit2/libgit2/issues/2118
http://empt1e.blogspot.jp/2011/05/dealing-with-pkg-config-detection-of.html

proski:
That's unfortunate. I would have considered delaying this patch. Anyway, it's an issue with FreeBSD that should be reported to FreeBSD maintainers.

Fortunately, it should not be necessary to patch `configure.ac` or any other file. All that's needed is to specify `OPENSSL_CFLAGS` and `OPENSSL_LIBS` on the `configure` command line, and that will skip the pkg-config check in `configure`. The same applies to all pkg-config checks.

Interestingly, the situation on MacOS is the opposite. The `openssl.pc` file is present, but the headers are missing and not available from Apple. The fix is to install another copy of openssl from Homebrew and set `PKG_CONFIG_PATH` to the directory where the Homebrew version of `openssl.pc` is located.

PR #515, opened 2016-11-30
Log client computer name in xrdp log

### LINKED ISSUE #500: Get client name
I'm currently using the 0.9.0 from Debian Jessie Backport and I want to know how I could retrieve the client's name (which defaults to client's hostname) for each active connection.

speidy:
resolves #500

### DIFF
--- a/common/xrdp_constants.h
+++ b/common/xrdp_constants.h
@@ -676,4 +676,10 @@
 #define XR_RDP_SCAN_LSHIFT 42
 #define XR_RDP_SCAN_ALT    56
 
+#define INFO_TEXT_MAX_BYTES             255
+#define INFO_TEXT_MAX_UNICODE_BYTES     512
+
+#define CS_CLIENT_NAME_BYTES            16
+#define CS_CLIENT_NAME_UNICODE_BYTES    32
+
 #endif
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -617,7 +617,7 @@ xrdp_sec_encrypt(struct xrdp_sec *self, char *data, int len)
 
 /*****************************************************************************/
 static int
-unicode_in(struct stream *s, int uni_len, char *dst, int dst_len)
+unicode_in(struct stream *s, int uni_len, int uni_max_bytes, char *dst, int dst_len)
 {
     int dst_index;
     int src_index;
@@ -627,7 +627,7 @@ unicode_in(struct stream *s, int uni_len, char *dst, int dst_len)
 
     while (src_index < uni_len)
     {
-        if (dst_index >= dst_len || src_index > 512)
+        if (dst_index >= dst_len || src_index > uni_max_bytes)
         {
             break;
         }
@@ -782,12 +782,12 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
         return 1;
     }
 
-    if (unicode_in(s, len_domain, self->rdp_layer->client_info.domain, 255) != 0)
+    if (unicode_in(s, len_domain, 512, self->rdp_layer->client_info.domain, 255) != 0)
     {
         return 1;
     }
     DEBUG(("domain %s", self->rdp_layer->client_info.domain));
-    if (unicode_in(s, len_user, self->rdp_layer->client_info.username, 255) != 0)
+    if (unicode_in(s, len_user, 512, self->rdp_layer->client_info.username, 255) != 0)
     {
         return 1;
     }
@@ -795,7 +795,7 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
 
     if (flags & RDP_LOGON_AUTO)
     {
-        if (unicode_in(s, len_password, self->rdp_layer->client_info.password, 255) != 0)
+        if (unicode_in(s, len_password, 512, self->rdp_layer->client_info.password, 255) != 0)
         {
             return 1;
         }
@@ -815,12 +815,12 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
         }
     }
 
-    if (unicode_in(s, len_program, self->rdp_layer->client_info.program, 255) != 0)
+    if (unicode_in(s, len_program, 512, self->rdp_layer->client_info.program, 255) != 0)
     {
         return 1;
     }
     DEBUG(("program %s", self->rdp_layer->client_info.program));
-    if (unicode_in(s, len_directory, self->rdp_layer->client_info.directory, 255) != 0)
+    if (unicode_in(s, len_directory, 512, self->rdp_layer->client_info.directory, 255) != 0)
     {
         return 1;
     }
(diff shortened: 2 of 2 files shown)


libxrdp/xrdp_sec.c
```diff
@@ -778,20 +778,20 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
         return 1;
     }
 
-    if (unicode_in(s, len_domain, self->rdp_layer->client_info.domain, 255) != 0)
+    if (unicode_in(s, len_domain, 512, self->rdp_layer->client_info.domain, 255) != 0)
```
proski: Can we have named constants for the numbers like 512 and 255?

libxrdp/xrdp_sec.c
```diff
@@ -1603,7 +1605,8 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec* self, struct stream* s)
     in_uint8s(s, 2); /* SASSequence */
     in_uint8s(s, 4); /* keyboardLayout */
     in_uint8s(s, 4); /* clientBuild */
-    in_uint8s(s, 32); /* clientName */
+    unicode_in(s, 30, 32, clientName, 15); /* clientName */
+    log_message(LOG_LEVEL_ALWAYS ,"client computer name: %s", clientName);
```
proski: Formatting: space goes after comma, not before. Why LOG_LEVEL_ALWAYS? It's for messages more important than LOG_LEVEL_ERROR (i.e. fatal errors). This condition looks like LOG_LEVEL_INFO to me - more important than debug information, but not something to worry about.

speidy: ok, fixed now.

libxrdp/xrdp_sec.c
metalefty: @speidy How is constification going? Else changes look good to me.

speidy: I don't think there is special meaning to define it, the values changes for every field, that is given in the protocol specs.

proski: It should be clear whether the number comes from the protocol (in which case it would be defined next to other protocol specific constants) or it's an arbitrary limit (in which case it can be defined in the same file as the implementation).

speidy: I don't find any uniform name for a define. the function argument explains the value purpose well enough, even when not defined.
We have around 5 different fields with the same protocol defined length.
I don't think 5 different define will make the code more clear.

proski: The function name explains how the arguments are used. I don't see why the values are 512 and 255 for this field. We can have one define. But even five defines would be better than no defines IMHO.

proski:
OK, thank you for adding constants. But I think sizeof() would be more appropriate if it can be used.

I looked at `unicode_in` source, it's one huge hack. It takes the lower byte and throws away the upper byte. I believe the correct behavior would be to convert from UTF-16 to UTF-8.

Having three different limits as arguments to unicode_in seems inelegant, but I think all three would be meaningful if proper conversion is implemented. Actually, they would be more meaningful than they are now.

speidy:
I agree unicode_in() is a dirty hack.
We can probably use some standard library functions for that.

speidy:
let me fix unicode_in in a seperate patch.

libxrdp/xrdp_sec.c
```diff
@@ -782,20 +782,20 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
         return 1;
     }
 
-    if (unicode_in(s, len_domain, self->rdp_layer->client_info.domain, 255) != 0)
+    if (unicode_in(s, len_domain, INFO_TEXT_MAX_UNICODE_BYTES, self->rdp_layer->client_info.domain, INFO_TEXT_MAX_BYTES) != 0)
```
proski: The limit on the output comes from the output buffer size. I would rather use `sizeof(self->rdp_layer->client_info.domain) - 1`, which would show where the limit comes from. INFO_TEXT_MAX_BYTES (if needed at all) should be where `domain` is defined, i.e. xrdp_client_info.h

libxrdp/xrdp_sec.c
```diff
@@ -1607,7 +1609,9 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec* self, struct stream* s)
     in_uint8s(s, 2); /* SASSequence */
     in_uint8s(s, 4); /* keyboardLayout */
     in_uint8s(s, 4); /* clientBuild */
-    in_uint8s(s, 32); /* clientName */
+    unicode_in(s, CS_CLIENT_NAME_UNICODE_BYTES - sizeof(twchar), CS_CLIENT_NAME_UNICODE_BYTES,
+               clientName, CS_CLIENT_NAME_BYTES - sizeof(char)); /* clientName, represented as 512 bytes non-null terminated unicode string */
```
proski: I suspect the comment is wrong, at least the number 512. "Unicode string" is ambiguous, especially in the context that mention bytes. It should probably be called UTF-16, not Unicode. Subtracting `sizeof(twchar)` looks wrong to me. If we are dealing with UTF-16, we should be reserving two bytes, not four. And if the string we get from the client is not null-terminated, we should allocate a longer buffer for the termination, not discard the last character from the input.

common/xrdp_constants.h
```diff
@@ -676,4 +676,10 @@
 #define XR_RDP_SCAN_LSHIFT 42
 #define XR_RDP_SCAN_ALT    56
 
+#define INFO_TEXT_MAX_BYTES             255
+#define INFO_TEXT_MAX_UNICODE_BYTES     512
+
+#define CS_CLIENT_NAME_BYTES            16
+#define CS_CLIENT_NAME_UNICODE_BYTES    32
```
proski: "CS" is ambiguous. Let's avoid non-common abbreviations, especially in the defines.

proski:
@speidy It would be great to get `unicode_in()` fixed. It may give us a better understanding what arguments it should take.

Is it really ever needed to limit the number of Unicode characters (codepoints) that can be processed? I think it's not a useful limitation. The RDP protocol should be agnostic to such details.

The protocol should specify the encoding and the byte size of the string. For example, the encoding is UTF-16 and the string is 512 bytes long. When we decode UTF-16 into UTF-8, the caller should provide an output buffer.

It is reasonable to stop in the input buffer or the output buffer is exhausted. But is it ever needed to stop after processing some number of Unicode codepoints if both the input and the output buffers are not exhausted?

Admittedly, I don't know how non-terminated strings are processed. If their length specified in bytes or in Unicode codepoints?

speidy:
It's not UTF16 but Unicode, notice that the RDP protocol, was first implemented to Windows platform. where in windows sizeof(wchar_t) == 2 not 4.

Fixing unicode_in will probably clean up that part of code, that was written before the docs were out.

proski:
UTF-16 is an encoding for Unicode. Unicode itself is not an encoding. Another possibility is UCS-2 encoding, it also has 2 bytes on the wire for ASCII characters, but it has no surrogates to represent codepoints above 0xffff.

Unless there is a clear indication that the encoding in USC-2, I think it's safer to assume UTF-16 and handle surrogates.

jsorg71:
I think MS Unicode is UCS-2.  I have seen some clams that it's UTF-16.

speidy:
I made a new PR for this #699 
Closing this one now.

PR #516, opened 2016-11-30
Adjust Japanese keymaps

metalefty:
Whichever v0.9.1 or later is ok.

* unify km-*0411.ini into km-00000411.ini as they're identical and no longer need to duplicate them after
commit 5725200
* apply changes introduced in #394 to dump-keymaps.sh


proski:
The removal of the duplicate files is good. But please make sure to remove them from xrdp_keyboard.ini as well. Make sure there are no other references to those files.

Changes to genkeymap should not have any effect on the end users per se. Does genkeymap generate a better keymap with pc105? Can you update the keymap?

Would any users benefit from the original keymap? Can it be kept as a variant? Maybe that's a veriant we want to keep?

metalefty:
Ah, yes. I just forgot removing them from xrdp_keyboard.ini.

The output between
```
setxkbmap -model jp106 -layout jp -variant OADG109A
./xrdp-genkeymap ../instfiles/km-00000411.ini
```
and
```
setxkbmap -model pc105 -layout jp -variant OADG109A
./xrdp-genkeymap ../instfiles/km-00000411.ini
```
has no difference so I didn't update it.

proski:
Looks good to me now

PR #518, opened 2016-12-02
Provide reasonable hTotal, vTotal and dotClock for the display mode

proski:
gtk3 programs crash when they divide dotClock by hTotal and vTotal to
calculate the refresh rate.


metalefty:
+1

PR #521, opened 2016-12-02
Change default Xorg logfile to .xorgxrdp.%s.log

proski:
The log file is created in the home directory, %s is replaced with the
display number (e.g. 10).

If Xorg is run setuid root, it can refuse to run if the log path is
absolute.


metalefty:
This should be merged together with #513.

PR #523, opened 2016-12-04
vnc: change password file to use guid hash

jsorg71:
This is cleanup of vncpassfile branch

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -804,6 +804,7 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
         temp->item->data = data;
         g_strncpy(temp->item->client_ip, s->client_ip, 255);   /* store client ip data */
         g_strncpy(temp->item->name, s->username, 255);
+        g_memcpy(temp->item->guid, s->guid, 16);
 
         ltime = g_time1();
         localtime_r(&ltime, &stime);
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -408,9 +408,7 @@ wait_for_xserver(int display)
 /******************************************************************************/
 /* called with the main thread */
 static int APP_CC
-session_start_fork(int width, int height, int bpp, char *username,
-                   char *password, tbus data, tui8 type, char *domain,
-                   char *program, char *directory, char *client_ip)
+session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
 {
     int display = 0;
     int pid = 0;
@@ -445,7 +443,7 @@ session_start_fork(int width, int height, int bpp, char *username,
     if (g_session_count >= g_cfg->sess.max_sessions)
     {
         log_message(LOG_LEVEL_INFO, "max concurrent session limit "
-                    "exceeded. login for user %s denied", username);
+                    "exceeded. login for user %s denied", s->username);
         return 0;
     }
 
@@ -454,7 +452,7 @@ session_start_fork(int width, int height, int bpp, char *username,
     if (temp == 0)
     {
         log_message(LOG_LEVEL_ERROR, "cannot create new chain "
-                    "element - user %s", username);
+                    "element - user %s", s->username);
         return 0;
     }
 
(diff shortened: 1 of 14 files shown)


vnc/vnc.c
```diff
@@ -1083,7 +1094,24 @@ lib_mod_connect(struct vnc *v)
...
+                        char guid_str[64];
+                        char *pguid_str;
+                        int index;
+                        pguid_str = guid_str;
+                        for (index = 0; index < 16; index++)
+                        {
+                            g_snprintf(pguid_str, 4, "%2.2x", v->guid[index]);
+                            pguid_str += 2;
+                        }
+                        guid_str[32] = 0;
```
proski: This code is used twice, please refactor

sesman/session.c
```diff
@@ -688,18 +686,31 @@ session_start_fork(int width, int height, int bpp, char *username,
...
+                    g_sprintf(geometry, "%d", s->height);
                     g_setenv("XRDP_START_HEIGHT", geometry, 1);
 
                     /* fire up Xorg */
                     g_execvp(xserver, pp1);
                 }
                 else if (type == SESMAN_SESSION_TYPE_XVNC)
                 {
-                    env_check_password_file(passwd_file, password);
+                    if (s->guid != 0)
```
proski: This line causes a warning in clang:
`session.c:700:28: warning: comparison of array 's->guid' not equal to a null pointer is always true [-Wtautological-pointer-compare]`
Which raised a bigger question. Why so many ifs? Shouldn't the new behavior be required?

proski:
I checked the patch on Fedora 25. I cannot reconnect to an existing VNC session. I use FreeRDP without autologin. I get authentication failure.

jsorg71:
@proski I think the VNC reconnect issue is fixed.

proski:
The reconnect issue is fixed. The warning is fixed. But I have found another issue. I cannot login to the Xvnc session if `--enable-ipv6` was used to configure xrdp. I'm connecting from Mac OSX Sierra to RHEL 6. The clients are freerdp, MS RDP and MS RDP Beta. Both manual login and autologin fail.

metalefty:
It works for me.  FreeRDP to CentOS 7.

```
$ rpm -qa |grep xrdp-
xorg-x11-drv-xrdp-0.9.0.git822e3b8+devel-1.el7.centos.x86_64
xrdp-0.9.0.gitf14e3dc+vpf-1.el7.centos.x86_64
```

```
[20161205-16:51:00] [INFO ] A connection received from ::1 port 57936
[20161205-16:51:01] [INFO ] ++ created session (access granted): username osc, ip 0.0.0.0:45798 - socket: 12
[20161205-16:51:01] [INFO ] starting Xvnc session...
[20161205-16:51:01] [DEBUG] Closed socket 9 (AF_INET6 :::5910)
[20161205-16:51:01] [DEBUG] Closed socket 9 (AF_INET6 :::6010)
[20161205-16:51:01] [DEBUG] Closed socket 9 (AF_INET6 :::6210)
[20161205-16:51:01] [DEBUG] Closed socket 7 (AF_INET6 ::1:3350)
[20161205-16:51:01] [DEBUG] Closed socket 8 (AF_INET6 ::1:3350)
[20161205-16:51:01] [INFO ] Xvnc :10 -geometry 1024x768 -depth 16 -rfbauth /home/osc/.vnc/sesman_osc_passwd -bs -nolisten tcp -localhost -dpi 96
[20161205-16:51:01] [INFO ] starting xrdp-sessvc - xpid=2300 - wmpid=2299
```

proski:
I was able to reproduce the VNC authentication problem without this patch, with the unchanged `devel` branch. I'll create a separate issue for that.

jsorg71:
so we are good for this PR? any objections?

metalefty:
LGTM

proski:
No objections on my part. Let's merge.

jsorg71:
I merged this but the github server crashed at the same time but it looks like it merged.

PR #526, opened 2016-12-04
Make configure show it's looking for "pixman", not "xrdp_pixman"

proski:
The pixman library in not part of xrdp. If users see that "xrdp_pixman"
is not found, they might assume it is some xrdp specific library.

The Automake conditional is still XRDP_PIXMAN, but the pkg-config
variables become PIXMAN_LIBS and PIXMAN_CFLAGS.


metalefty:
+1 I think this can be included to 0.9.1.

jsorg71:
+1

PR #529, opened 2016-12-08
Fix support for 32-bpp clients connecting to 16-bpp VNC

proski:


proski:
The original patch was longer. After careful testing and code reading, I was able to reduce it to a one line change. With this patch, the Xvnc backend supports all combinations of server and client color depth.

jsorg71:
+1

metalefty:
It's a simple fix so I think we can include this to 0.9.1. Add a milestone.

PR #532, opened 2016-12-09
sesman: unbreak FreeBSD build after #523

metalefty:
`username` here should be changed to `s->username`, too.


metalefty:
The change was introduced by 57905d7.

jsorg71:
opps, sorry about that
+1

proski:
+1

metalefty:
No problem. I should have realize it in review.
I'll merge this shortly because it's a small change and fixes build failure.

PR #533, opened 2016-12-09
Don't use colon to separate IPv6 address from the port

proski:
IPv6 addresses can have colons in their names, so the final colon can be confusing.

http uses [ip]:port, but this is just a log, we can be more user friendly.


metalefty:
+1

metalefty:
@itamarjp 
We talked at gitter chat about to prepare 0.9.1 release. Please don't merge any PRs which doesn't have v0.9.1 milestone.  If you think a PR can be included to 0.9.1, please add v0.9.1 milestone before merge.

PR #531, opened 2016-12-09
xrdp_encoder: make RemoteFX work with MSTSC client

speidy:
should resolve #508 #358

PR #535, opened 2016-12-10
Fix for RemoteFX detection, avoid blank screen when RFX is not supported on server

### LINKED ISSUE #534: when xrdp is built without --enable-rfxcodec, client might get blank screen connecting to xorgxrdp
preconditions:
* xrdp built *without* ./configure --enable-rfxcodec
* client supports rfx

result:
then the user might get blank screen, when using xorgxrdp backend

speidy:
o caps: advertise remotefx codec capsets only when supported (RFX mode, proxy mode)
o xrdp_encoder: do not initialze encoder context for RFX when rfxcodec is not supported

should resolve #534


speidy:
Note: I think we might merge this PR before #531. There is a conflicting part in #531.

proski:
Looks good to me. I agree, fixes should be merged before new features. For every feature patch, it's good to have a working baseline.

metalefty:
Looks good. Let's test again after merge.

PR #536, opened 2016-12-10
add unicode support

jsorg71:
This is part on noorders


proski:
Nice, thank you for splitting it up for the review!

Could you please add constants for the magic numbers like 8, 42, 56, 256?

How can I test this functionality? Which clients need it or can make use of it? I remember older MS RDP Beta (around 8.2.10 or so) had trouble with keyboard, maybe it used Unicode? Too bad I cannot find where to download it now.

speidy:
@proski i think Parallels RDP Client is doing unicode by default

jsorg71:
Magics, sure, I'll make constants.
The best way to test is the Microsoft Android client.

jsorg71:
This one might be hard to test until the other part is in.

proski:
I was able to test MS RDP client on my Android phone, and I was able to see that the Unicode events were actually processed by xrdp.

The good news is that the Parallels clients for Mac can connect to xrdp from the noorders branch, and it's working correctly.

Since the Unicode data comes in 2-byte sequences, can we assume UTF-16 encoding? If so, should not surrogates be processed? The issue can be addressed separately, I cannot test it.

The patch looks good to me.

PR #539, opened 2016-12-13
Make librfxcodec a subpackage

proski:
Since libpainter is going to become a subpackage, we can make librfxcodec
a subpackage as well.

With this change, librfxcodec is configured, built and linked to xrdp
automatically if enabled by "--enable-rfxcodec"

librfxcodec is packaged by "make dist" regardless of "--enable-rfxcodec"

librfxcodec/confgure is run unconditionally, it's needed for "make dist"
to work. But librfxcodec is only compiled if enabled.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,6 +16,12 @@ else
 XRDPVRDIR =
 endif
 
+if XRDP_RFXCODEC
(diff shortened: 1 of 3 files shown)


metalefty:
awesome +1

speidy:
+1

PR #541, opened 2016-12-13
Add xauth support

rouxcorp:


sesman/startwm.sh
```diff
@@ -3,6 +3,8 @@
 #start the window manager
 wm_start()
 {
+  mcookie | sed -e 's/^/add '$DISPLAY' . /' | xauth -q
```
metalefty: I don't like to depend on mcookie command because operating systems other than GNU/Linux doesn't have such a command.  Please be careful of portability. 

I worked on similar issue for TigerVNC.  See also TigerVNC/tigervnc#366.

This is a little bit longer but can replace mcookie command.
`dd if=/dev/urandom bs=16 count=1 2>/dev/null| od -x -An | tr -d ' '`

rouxcorp: Yes, I agree with with you :)
I replaced it just now

sesman/session.c
```diff
@@ -676,6 +676,12 @@ session_start_fork(int width, int height, int bpp, char *username,
                     /* these are the must have parameters */
                     list_add_item(xserver_params, (tintptr) g_strdup(xserver));
                     list_add_item(xserver_params, (tintptr) g_strdup(screen));
+                    
+                    char xauth_file[256];
+                    g_snprintf(xauth_file, 255, "%s/.Xauthority", g_getenv("HOME"));
+                    list_add_item(xserver_params, (tintptr) g_strdup("-auth"));
+                    list_add_item(xserver_params, (tintptr) g_strdup(xauth_file));
```
proski: This should be done for all servers: X11rdp, Xorg and Xvnc. I wonder it it can be done in sesman.ini. The X server should already be running in the user's home directory, so the full path should not be needed. Some setuid X servers reject full path for the log; they could reject full path to `.Xauthority` as well.

Generally, I would prefer the configuration like that to be outside the code for more flexibility. Eventually, all the code for starting backends could become a shell script.

sesman/startwm.sh
```diff
@@ -3,6 +3,8 @@
 #start the window manager
 wm_start()
 {
+  dd if=/dev/urandom bs=16 count=1 2>/dev/null | od -x -An | tr -d ' ' | sed -e 's/^/add '$DISPLAY' . /' | xauth -q
```
proski: xauth has an option to generate the cookie. That might be a cleaner approach.

I believe `startwm.sh` should abort in case of failure. Fallback to insecure operation is not a good idea.

moobyfr: `xauth generate` doesn't really generate a cookie. It extracts the existing cookie from the running X, to add it to a file. This is used when the X server is started by a DM which has already a cookie.

sesman/startwm.sh
moobyfr: Does is still work if the user doesn't have a .Xauthority ? IIRC xauth fails

speidy:
We should consider what will happen in case of a local console session connection as well.
One user might use .Xauthority file for some local session using lightdm for example.
Than connect with xrdp and overwrite this .Xauthority file

What will be the implications?

proski:
Nobody suggests overwriting `.Xauthority` file. xauth operates on records. If would not remove records for other displays.

rouxcorp:
#545 will be much better :)

PR #542, opened 2016-12-13
TLS: new method to specify SSL/TLS version

### LINKED ISSUE #428: New method needed to disable specific SSL/TLS version

metalefty:
SSL/TLS protocols only listed in ssl_protocols should be used.
The name "ssl_protocols" comes from nginx.

Resolves #428.

Document hasn't been updated yet.

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -143,7 +143,7 @@ struct xrdp_client_info
   int use_frame_acks;
   int max_unacknowledged_frame_count;
 
-  int disableSSLv3; /* 0 = no, 1 = yes */
+  long ssl_protocols;
   char tls_ciphers[64];
 
   int client_os_major;
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -135,6 +135,10 @@ struct xrdp_client_info
   char variant[16];
   char options[256];
 
+  /* !!!!!!!!!!!!!!!!!!!!!!!!!! */
+  /* NO CHANGES ABOVE THIS LINE */
+  /* !!!!!!!!!!!!!!!!!!!!!!!!!! */
+
   /* codec */
   int h264_codec_id;
   int h264_prop_len;
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -45,6 +45,7 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
     char *item = (char *)NULL;
     char *value = (char *)NULL;
     char cfg_file[256];
+    char *p = (char *)NULL;
     char *tmp = (char *)NULL;
     int tmp_length = 0;
 
@@ -165,31 +166,36 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
         }
         else if (g_strcasecmp(item, "ssl_protocols") == 0)
         {
-            /* put leading/trailing space to properly detect "TLSv1" without regex */
+            /* put leading/trailing comma to properly detect "TLSv1" without regex */
             tmp_length = g_strlen(value) + 3;
             tmp = g_new(char, tmp_length);
-            g_snprintf(tmp, tmp_length, "%s%s%s", " ", value, " ");
+            g_snprintf(tmp, tmp_length, "%s%s%s", ",", value, ",");
+            /* to accept space after comma */
+            while ((p = (char *) g_strchr(tmp, ' ')) != NULL)
+            {
+                *p = ',';
+            }
 
             /* disable all protocols first, enable later */
             client_info->ssl_protocols =
                 SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1 | SSL_OP_NO_TLSv1_2;
 
-            if (g_pos(tmp, " TLSv1.2 ") >= 0)
+            if (g_pos(tmp, ",TLSv1.2,") >= 0)
             {
                 log_message(LOG_LEVEL_DEBUG, "TLSv1.2 enabled");
                 client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_2;
             }
-            if (g_pos(tmp, " TLSv1.1 ") >= 0)
+            if (g_pos(tmp, ",TLSv1.1,") >= 0)
             {
                 log_message(LOG_LEVEL_DEBUG, "TLSv1.1 enabled");
                 client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_1;
             }
-            if (g_pos(tmp, " TLSv1 ") >= 0)
+            if (g_pos(tmp, ",TLSv1,") >= 0)
             {
                 log_message(LOG_LEVEL_DEBUG, "TLSv1 enabled");
                 client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1;
             }
-            if (g_pos(tmp, " SSLv3 ") >= 0)
+            if (g_pos(tmp, ",SSLv3,") >= 0)
             {
                 log_message(LOG_LEVEL_DEBUG, "SSLv3 enabled");
                 client_info->ssl_protocols &= ~SSL_OP_NO_SSLv3;
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -18,6 +18,7 @@
  * rdp layer
  */
 
+#include <openssl/ssl.h>
 #include "libxrdp.h"
 #include "log.h"
 
@@ -44,6 +45,8 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
     char *item = (char *)NULL;
     char *value = (char *)NULL;
     char cfg_file[256];
+    char *tmp = (char *)NULL;
+    int tmp_length = 0;
 
     /* initialize (zero out) local variables: */
     g_memset(cfg_file, 0, sizeof(char) * 256);
@@ -160,9 +163,37 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
                 client_info->use_fast_path = 0;
             }
         }
-        else if (g_strcasecmp(item, "disableSSLv3") == 0)
+        else if (g_strcasecmp(item, "ssl_protocols") == 0)
         {
-            client_info->disableSSLv3 = g_text2bool(value);
+            /* put leading/trailing space to properly detect "TLSv1" without regex */
+            tmp_length = g_strlen(value) + 3;
+            tmp = g_new(char, tmp_length);
+            g_snprintf(tmp, tmp_length, "%s%s%s", " ", value, " ");
+
+	    /* disable all protocols first, enable later */
+            client_info->ssl_protocols =
+                SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1 | SSL_OP_NO_TLSv1_2;
+
+            if (g_pos(tmp, " TLSv1.2 ") >= 0)
+            {
+                log_message(LOG_LEVEL_DEBUG, "TLSv1.2 enabled");
+                client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_2;
+            }
+            if (g_pos(tmp, " TLSv1.1 ") >= 0)
+            {
+                log_message(LOG_LEVEL_DEBUG, "TLSv1.1 enabled");
+                client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_1;
+            }
+            if (g_pos(tmp, " TLSv1 ") >= 0)
+            {
+                log_message(LOG_LEVEL_DEBUG, "TLSv1 enabled");
+                client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1;
+            }
+            if (g_pos(tmp, " SSLv3 ") >= 0)
+            {
+                log_message(LOG_LEVEL_DEBUG, "SSLv3 enabled");
+                client_info->ssl_protocols &= ~SSL_OP_NO_SSLv3;
+            }
         }
         else if (g_strcasecmp(item, "tls_ciphers") == 0)
         {
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -170,7 +170,7 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
             tmp = g_new(char, tmp_length);
             g_snprintf(tmp, tmp_length, "%s%s%s", " ", value, " ");
 
-	    /* disable all protocols first, enable later */
+            /* disable all protocols first, enable later */
             client_info->ssl_protocols =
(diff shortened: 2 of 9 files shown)


metalefty:
I'd appreciate if you review this PR after next release..  Help me to improve the patch.

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +161,36 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
+            char s[strlen(value) + 2];
+            snprintf(s, sizeof(s), "%s%s", value, "x");
+
+	    /* disable all protocols first, enable later */
+            client_info->ssl_protocols =
+                SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1 | SSL_OP_NO_TLSv1_2;
+
+            if (g_strstr(value, "TLSv1.2") != NULL)
+            {
+                log_message(LOG_LEVEL_DEBUG, "TLSv1.2 enabled");
```
metalefty: Log should be brushed up. This is really a draft.

metalefty:
To know which SSL/TLS protocol is finally selected, #426 enables it.

common/os_calls.c
```diff
@@ -2437,6 +2437,13 @@ g_strncasecmp(const char *c1, const char *c2, int len)
 }
 
 /*****************************************************************************/
+const char* APP_CC
+g_strstr(const char* c1, const char* c2)
+{
+    return strstr(c1, c2);
```
speidy: I think you can use g_pos()

proski: Yes, g_pos() already wraps strstr(). I don't see any need to handle Win32 specially here.

common/os_calls.c
```diff
@@ -2437,6 +2437,13 @@ g_strncasecmp(const char *c1, const char *c2, int len)
 }
 
 /*****************************************************************************/
+const char* APP_CC
+g_strstr(const char* c1, const char* c2)
```
proski: `*` sticks to the right, not to the left. Please see coding_style.md. The preexisting code is not always a good example.

xrdp/xrdp.ini
```diff
@@ -25,8 +25,8 @@ crypt_level=high
 ; openssl req -x509 -newkey rsa:2048 -nodes -keyout key.pem -out cert.pem -days 365
 certificate=
 key_file=
-; specify whether SSLv3 should be disabled
-#disableSSLv3=true
+; set SSL protocols (space separated)
```
proski: Please document all allowed values here. Also, the `xrdp.ini.5` manual page needs to be updated. It documents disableSSLv3.

common/xrdp_client_info.h
```diff
@@ -143,7 +143,7 @@ struct xrdp_client_info
   int use_frame_acks;
   int max_unacknowledged_frame_count;
 
-  int disableSSLv3; /* 0 = no, 1 = yes */
+  long ssl_protocols;
```
proski: We are playing with fire here. This would change the structure that is used by xorgxrdp and X11rdp. The first cleanup we should do would be to lock down `struct xrdp_client_info` to whatever external dependencies expect, and more the rest to some other structure. Or at very least put a big comment there and write `NO CHANGES ABOVE THIS LINE`

speidy: +1

metalefty: OK, I see that we can't change size/length of existing variables. I also think tls_ciphers length is not enough because it will often be longer than 63 characters. For example: https://mozilla.github.io/server-side-tls/ssl-config-generator/

I need to rethink approach to expand tls_ciphers length.

metalefty: If I expand tls_ciphers length, is this fine for example?

```diff
diff --git a/common/xrdp_client_info.h b/common/xrdp_client_info.h
index 1fa5aaf4..2db6e5be 100644
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -143,11 +143,18 @@ struct xrdp_client_info
   int use_frame_acks;
   int max_unacknowledged_frame_count;

-  long ssl_protocols;
-  char tls_ciphers[64];
+  int disableSSLv3;      /* not in use anymore */
+  char tls_ciphers_[64]; /* not in use anymore */

   int client_os_major;
   int client_os_minor;
+
+  /* !!!!!!!!!!!!!!!!!!!!!!!!!! */
+  /* NO CHANGES ABOVE THIS LINE */
+  /* !!!!!!!!!!!!!!!!!!!!!!!!!! */
+
+  long ssl_protocols;
+  char tls_ciphers[1024];
 };

 #endif
```

proski: We would need to check which members are actually used by X11rdp and xorgxrdp. X11rdp is important because it can be compiled separately, perhaps from older sources. I feel this discussion is more suitable for a separate issue.

proski: The line would be between `options` (the last field used by xorgxrdp) and `h264_codec_id` (any fields below are not used by X11rdp or xorgxrdp). `disableSSLv3` is in the "safe" zone. We still need some way to pass newly added messages to the backends.

common/os_calls.c
metalefty: Thanks, I switched to g_pos().

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +161,43 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
+                client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_2;
+            }
+            if (g_pos(value, "TLSv1.1") >= 0)
+            {
+                log_message(LOG_LEVEL_DEBUG, "TLSv1.1 enabled");
+                client_info->ssl_protocols &= ~SSL_OP_NO_TLSv1_1;
+            }
+            if (g_pos(value, "TLSv1 ") >= 0 || g_pos(s, "TLSv1x") >= 0)
```
proski: That's not good. If the whole `ssl_protocols` string is `TLSv1` without the trailing space, it won't be matched here. Actually, g_pos() won't be of much use if we allow values that are substrings of other values. Either we should use names like `TLSv1.0` or we should match word boundaries. The later is more correct, but harder to implement. If this patch is not destined for v0.9.1, maybe we could add glib or apr support first. glib and arp both support regular expressions.

common/xrdp_client_info.h
```diff
@@ -143,11 +143,17 @@ struct xrdp_client_info
...
 
-  int disableSSLv3; /* 0 = no, 1 = yes */
+  int disableSSLv3; /* not in use anymore */
   char tls_ciphers[64];
 
   int client_os_major;
   int client_os_minor;
+
+  /* !!!!!!!!!!!!!!!!!!!!!!!!!! */
+  /* NO CHANGES ABOVE THIS LINE */
```
proski: If the goal is to keep xorgxrdp and X11rdp seeing the same fields they expect, the boundary should be after `options`.

libxrdp/xrdp_rdp.c
metalefty: I want to keep these version strings as same as constants defined in `openssl/ssl.h`

```
# define SSL_TXT_SSLV2           "SSLv2"
# define SSL_TXT_SSLV3           "SSLv3"
# define SSL_TXT_TLSV1           "TLSv1"
# define SSL_TXT_TLSV1_1         "TLSv1.1"
# define SSL_TXT_TLSV1_2         "TLSv1.2"
```

I tested it actually matches.  I did a hack.  In case of no trailing space after `TLSv1`,  append "x" to the end and then compare to "TLSv1x" like a technique `test "x$var" = "xyes"` in shell script.  It is hacky but actually working.

I don't intend to merge this before v0.9.1.

common/xrdp_client_info.h
speidy: I prefer switching to a padded struct, if not merging for 0.9.1

libxrdp/xrdp_rdp.c
proski: Actually, it's not that bad. But let's add " " instead of "x". Then we can match all strings with the final space: "SSLv3 ", "TLSv1 ", "TLSv1.1 " and so on.

common/xrdp_client_info.h
metalefty: If the struct is changed by another PR, I'll follow it.

libxrdp/xrdp_rdp.c
metalefty: ok. I think yours is better.

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +161,43 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
                 client_info->use_fast_path = 0;
             }
         }
-        else if (g_strcasecmp(item, "disableSSLv3") == 0)
+        else if (g_strcasecmp(item, "ssl_protocols") == 0)
         {
-            client_info->disableSSLv3 = g_text2bool(value);
+            char s[strlen(value) + 2];
+            snprintf(s, sizeof(s), "%s%s", value, "x");
```
proski: Since you are copying the whole string anyway, you can add another space in the beginning and search for strings with spaces on both sides. That would guarantee that only whole words would match, not substrings.

But to be more user friendly, is would be better to check every word and report those that are unknown to xrdp.

metalefty: +1 to add leading space.

metalefty: I should use g_snprintf() here.

metalefty:
I took your review comments, rebased to current devel head and force pushed.

metalefty:
I've finished this. @speidy @jsorg71, can you look at this, too?

metalefty:
No problem in such case. I designed the feature with old config file compatibility.

If `ssl_protocols` directive is not found in xrdp.ini, xrdp behaves as same as former. Only SSLv2 will be disabled.  If `ssl_protocols` directive is found and set to empty, all SSL protocols will be disabled.

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +163,45 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
+            /* put leading/trailing space to properly detect "TLSv1" without regex */
+            tmp_length = g_strlen(value) + 3;
+            tmp = g_new(char, tmp_length);
+            g_snprintf(tmp, tmp_length, "%s%s%s", " ", value, " ");
+
+            /* disable all protocols first, enable later */
+            client_info->ssl_protocols =
+                SSL_OP_NO_SSLv3 | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1 | SSL_OP_NO_TLSv1_2;
+
+            if (g_pos(tmp, " TLSv1.2 ") >= 0)
```
metalefty: Yes probably we can but I want to use regex to parse comma separated value  as "TLSv1" is substring of "TLSv1.1". The current implementation correctly detects "TLSv1" without regex.

See also discussions in this tree https://github.com/neutrinolabs/xrdp/pull/542#discussion_r92333250 .

speidy: g_pos can look for "," instead of the value to match. then compare.

metalefty: OK, I could. I'll push the fix later.

metalefty: Now whichever these thee list is accepted.

* comma separated list (no space after comma)
* comma separated list (space after comma)
* space separated list

metalefty:
Rebased to devel head again.

libxrdp/xrdp_rdp.c
```diff
@@ -44,6 +45,9 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
     char *item = (char *)NULL;
     char *value = (char *)NULL;
     char cfg_file[256];
+    char *p = (char *)NULL;
+    char *tmp = (char *)NULL;
```
proski: No cast is needed for NULL

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +164,47 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
-        else if (g_strcasecmp(item, "disableSSLv3") == 0)
+        else if (g_strcasecmp(item, "ssl_protocols") == 0)
         {
-            client_info->disableSSLv3 = g_text2bool(value);
+            /* put leading/trailing comma to properly detect "TLSv1" without regex */
+            tmp_length = g_strlen(value) + 3;
+            tmp = g_new(char, tmp_length);
+            g_snprintf(tmp, tmp_length, "%s%s%s", ",", value, ",");
+            /* to accept space after comma */
+            while ((p = (char *)g_strchr(tmp, ' ')) != NULL) *p = ',';
```
proski: Please add space after cast. Please use braces around the code block and follow the coding standard for block formatting.

libxrdp/xrdp_rdp.c
metalefty: Same for existing *item, *value?

proski: I think so. If it's close enough to be visible in the context, perhaps it should be fixed too.

metalefty:
No more change requests?

libxrdp/xrdp_rdp.c
```diff
@@ -160,9 +164,47 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
-        else if (g_strcasecmp(item, "disableSSLv3") == 0)
+        else if (g_strcasecmp(item, "ssl_protocols") == 0)
         {
-            client_info->disableSSLv3 = g_text2bool(value);
+            /* put leading/trailing comma to properly detect "TLSv1" without regex */
+            tmp_length = g_strlen(value) + 3;
+            tmp = g_new(char, tmp_length);
+            g_snprintf(tmp, tmp_length, "%s%s%s", ",", value, ",");
+            /* to accept space after comma */
+            while ((p = (char *) g_strchr(tmp, ' ')) != NULL) { *p = ','; }
```
proski: Please don't write a loop in on line.

metalefty:
OK, I think it's ready to merge.  What do you think @speidy @jsorg71 ?

metalefty:
If no objections in a week, I'll merge by myself.

speidy:
lgtm

metalefty:
A week past, merged.

PR #543, opened 2016-12-14
Add Korean keymap

westporch:
Hello. 
I'm newbie.
I've added Korean keymap.
Thanks.


instfiles/Makefile.am
```diff
@@ -26,7 +26,8 @@ dist_startscript_DATA = \
   km-0000080c.ini \
   km-00000813.ini \
   km-00000816.ini \
-  km-0000100c.ini 
+  km-0000100c.ini \
+  km-00000412.ini
```
proski: Let's keep the keymaps sorted alphabetically in Makefile.am

metalefty: Let's insert this after `km-00000411.ini` rather than appending to the end.

xrdp/xrdp_keyboard.ini
```diff
@@ -110,3 +112,4 @@ rdp_layout_ch=ch
 rdp_layout_pt=pt
 rdp_layout_br=br(abnt2)
 rdp_layout_pl=pl
+rdp_layout_kr=kr
```
metalefty: After jp, too.

westporch:
I sorted alphabetically order.
Thanks :-)

metalefty:
LGTM, thanks. I think we can include this to 0.9.1.

proski:
I agree, this can go to 0.9.1

PR #545, opened 2016-12-14
Fix xauth

### LINKED ISSUE #540: Current Xorg configuration is insecure
The current Xorg configuration is insecure.
Any user on the network, or locally on the server can access Xevents from another user. This include sending them window, getting keyboard events of grabbing the full display itself.
Example:
User A is logged in through xrdp, on the display `:10`. Another user can dump the screen of the user A with `xwd -display :10` . Or get all key pressed with xkey.c or the tool `xrdp` from `https://github.com/sensepost/xrdp`

To secure Xorg, we must follow `Xsecurity` from `Xorg` (`man Xsecurity`).
The best way to handle that is to use the `-auth` parameter from `Xorg` to share a common cookie with the session for the user `~/.Xauthority` is the common file for that

moobyfr:
This pull request fixes #540 . It creates a XAUTHORITY file (`.Xauthority` in the `$HOME` of the user by default, but can be overriden by setting a variable XAUTHORITY in sesman.ini) if it doesn't exist, and add a secret through the `xauth` command. The backend (Xvnc, Xorg or X11rdp) will get as argument the file to share the secret.


sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
 
     return error;
 }
+
+
+/******************************************************************************/
+int DEFAULT_CC
+env_add_xauth_user(int display, char *cookie, char *file)
```
proski: const char for both strings

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+{
+    FILE *dp;
+    char xauth_str[256];
+    int fd;
+
+    if ( file == NULL )
```
proski: No extra spaces inside parentheses, please

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+    {
+        if (!g_file_exist(".Xauthority"))
```
proski: No trailing spaces, please

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+    FILE *dp;
+    char xauth_str[256];
+    int fd;
+
+    if ( file == NULL )
+    {
+        if (!g_file_exist(".Xauthority")) 
+        {
+            fd = g_file_open(".Xauthority");
+            g_file_close(fd);
```
proski: What's the purpose of that? xauth can create .Xauthority on its own.

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+        }
+
+        g_sprintf(xauth_str, "xauth -q add :%d . %s", display, cookie);
```
proski: I would specify the file in either case. No need to duplicate logic.

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+    }
+    else
+    {
+        if (!g_file_exist(file))
+        {
+            fd = g_file_open(file);
+            g_file_close(fd);
+        }
+
+        g_sprintf(xauth_str, "xauth -q -f %s  add :%d . %s",
```
proski: No double space and no final space, please.

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+            fd = g_file_open(file);
+            g_file_close(fd);
+        }
+
+        g_sprintf(xauth_str, "xauth -q -f %s  add :%d . %s", 
+                file, display, cookie);
+    }
+    log_message(LOG_LEVEL_DEBUG, "xauth command: %s", xauth_str);
+
+    if ((dp = popen(xauth_str, "r")) == NULL) {
```
proski: Please don't put assignments inside if statements unnecessarily. The brace goes to the next line.

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+        log_message(LOG_LEVEL_INFO, "xauth failed, no X security");
```
proski: No way it's LOG_LEVEL_INFO. I would use LOG_LEVEL_ERROR; the function even returns an error code. Also, it's not xauth that failed here. It could not be run. Failures of xauth would be in pclose() return code.

sesman/env.c
```diff
@@ -200,3 +201,45 @@ env_set_user(const char *username, char **passwd_file, int display,
...
+        return 1;
+    }
+
+    pclose(dp);
```
proski: Please check the status. If it's not 0, log it and return error.

sesman/env.h
```diff
@@ -53,4 +53,17 @@ int DEFAULT_CC
 env_set_user(const char *username, char **passwd_file, int display,
              const struct list *env_names, const struct list *env_values);
 
+/**
+ *
+ * @brief create the XAUTORITY file for the user according to the display and the cookie
```
proski: xauthority -> xauthority everywhere

sesman/env.h
```diff
@@ -53,4 +53,17 @@ int DEFAULT_CC
...
+ *        xauth uses XAUTORITY if defined, ~/.Xauthority otherwise
+ * @param display The session display
+ * @param cookie The cookie
+ * @param file If not NULL, write the autorization in the file instead of default location
+ * @return 0 if adding the cookie is ok
+ */
+
+int DEFAULT_CC
+env_add_xauth_user(int display, char *cookie, char * file);
```
proski: const char for both, no space after asterisk

sesman/sesman.h
```diff
@@ -30,6 +30,7 @@
 #if defined(HAVE_CONFIG_H)
 #include "config_ac.h"
 #endif
+#include <stdio.h>
```
proski: It should not be here. It should be in the files that use functions from stdio.

sesman/session.c
```diff
@@ -676,6 +679,27 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
...
                 g_setenv("XRDP_SESMAN_KILL_DISCONNECTED", text, 1);
 
+                /* prepare the Xauthority stuff */
+                if (g_getenv("XAUTHORITY") != NULL)
+                {
+                    g_snprintf(authfile, 255, "%s", g_getenv("XAUTHORITY"));
+                }
+                else
+                {
+                    g_snprintf(authfile, 12, "%s", ".Xauthority");
```
proski: It can be 255 was well. No need to count characters. All that matters is the buffer size.

sesman/env.h
```diff
@@ -53,4 +53,17 @@ int DEFAULT_CC
 env_set_user(const char *username, char **passwd_file, int display,
              const struct list *env_names, const struct list *env_values);
 
+/**
+ *
+ * @brief create the XAUTORITY file for the user according to the display and the cookie
+ *        xauth uses XAUTORITY if defined, ~/.Xauthority otherwise
+ * @param display The session display
+ * @param cookie The cookie
+ * @param file If not NULL, write the autorization in the file instead of default location
```
proski: autorization -> authorization

sesman/session.c
```diff
@@ -676,6 +679,27 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
...
+                }
+
+                /* Create the cookie */
+                for (i = 0; i < 32; i++)
+                {
+                    g_random((char *) &cookie_tmpval, 1);
+                    sprintf(&cookie[i], "%02X", cookie_tmpval & 0xff);
+                }
+                cookie[32] = '\0';
```
proski: Calling g_random() 32 times to fill 16 bytes is highly ineffective. Please call g_random() once to get 16 random bytes. Then use g_bytes_to_hexstr() to convert it to text.

proski:
One more thing. It would be great if you could avoid adding too many lines to `session_start_fork`, it's way too long already. Think of the simplest API possible for `session_start_fork` to get the job done. Define a function that takes the parameter list and adds `-auth` and the xauthority file if successful, and keeps the parameter list unchanged otherwise.

You can create a separate file, say `xauth.c` to hold that function. No need to change env.c at all, it's for environment variable.

It's very important that we fix X security. We have just a few days left to get it into 0.9.1. I'll try to help with coding this weekend.

sesman/env.c
moobyfr: I didn't notice it does create the file, the message on stdout was misunderstood. I'll remove all this part

sesman/session.c
moobyfr: I wasn't aware of `g_bytes_to_hexstr()`. I'll switch to it.

sesman/xauth.c
```diff
@@ -0,0 +1,71 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Jay Sorg 2004-2013
```
proski: Please put your name and the current year here.

sesman/xauth.c
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *
+ * @file xauth.c
+ * @brief XAUTHORITY handling code
+ * @author Emmaunel Blindauer
```
proski: The `@author` line is not needed. `git log` and `git blame` are authoritative sources.

sesman/xauth.c
```diff
@@ -0,0 +1,71 @@
...
+#include "os_calls.h"
+
+
+/******************************************************************************/
+int DEFAULT_CC
+add_xauth_cookie(int display, const char *file)
+{
+    FILE *dp;
+    char cookie[33];
+    char char_cookie[16];
```
proski: Those are confusing names. I would use cookie_str and cookie_bin.

sesman/xauth.c
```diff
@@ -0,0 +1,71 @@
...
+    char xauth_str[256];
+    int ret;
+
+    g_random(char_cookie, 16);
+    g_bytes_to_hexstr(char_cookie, 16, cookie, 33);
+    cookie[32] = '\0';
```
proski: Not needed, g_bytes_to_hexstr() already terminated the output.

sesman/xauth.c
```diff
@@ -0,0 +1,71 @@
...
+
+    if (file == NULL)
```
proski: This function is never called with `file` being NULL, no need to implement it.

sesman/session.c
```diff
@@ -676,6 +678,19 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
...
+                {
+                    g_snprintf(authfile, 255, "%s", g_getenv("XAUTHORITY"));
+                }
+                else
+                {
+                    g_snprintf(authfile, 255, "%s", ".Xauthority");
+                }
+
+                /* Add the entry in XAUTHORITY file */
+                add_xauth_cookie(display, authfile);
```
proski: Running with impaired security is a bad choice. Let's exit here on error with `g_exit(1);`

I would rather have users fix their xauth than have them run xrdp insecurely without knowing it.

sesman/xauth.c
```diff
@@ -0,0 +1,63 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Jay Sorg 2004-2017
+ * Copyright (C) Emmanuel Blindauer 2004-2017
```
proski: I think you should leave only your name and use only one year: 2016.

sesman/xauth.h
```diff
@@ -0,0 +1,42 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Jay Sorg 2004-2013
```
proski: Please fix copyright

sesman/xauth.h
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *
+ * @file xauth.c                                                                 
+ * @brief XAUTHORITY handling code                                               
+ * @author Emmaunel Blindauer
```
proski: Again, let's not track authorship in multiple places. The copyright string and `git blame` should be enough.

sesman/xauth.c
```diff
@@ -0,0 +1,62 @@
...
+    if (dp == NULL)
+    {
+        log_message(LOG_LEVEL_ERROR, "Unable to launch xauth");
+        return 1;
+    }
+
+    ret = pclose(dp);
+    if (ret < 0)
+    {
+        log_message(LOG_LEVEL_ERROR, "An error occured while running xauth");
```
proski: occured -> occurred

speidy:
looks good to me.
can we merge?

moobyfr:
I have nothing more to push

PR #549, opened 2016-12-15
fixes for rle bitmap compress limits and raw bitmaps

jsorg71:
this is part of noorders

### DIFF
--- a/libxrdp/xrdp_bitmap32_compress.c
+++ b/libxrdp/xrdp_bitmap32_compress.c
@@ -107,7 +107,7 @@ fsplit3(char *in_data, int start_line, int width, int e,
         }
         start_line--;
         cy++;
-        if (out_index > 64 * 64)
+        if (out_index + width + e > 64 * 64)
         {
             break;
         }
@@ -195,7 +195,7 @@ fsplit4(char *in_data, int start_line, int width, int e,
         }
         start_line--;
         cy++;
-        if (out_index > 64 * 64)
+        if (out_index + width + e > 64 * 64)
         {
             break;
         }
@@ -422,6 +422,7 @@ xrdp_bitmap32_compress(char *in_data, int width, int height,
     char *sr_data;
     char *sg_data;
     char *sb_data;
+    char *hold_p;
     int a_bytes;
     int r_bytes;
     int g_bytes;
@@ -449,6 +450,7 @@ xrdp_bitmap32_compress(char *in_data, int width, int height,
     r_data = a_data + max_bytes;
     g_data = r_data + max_bytes;
     b_data = g_data + max_bytes;
+    hold_p = s->p;
 
     if (header & FLAGS_NOALPHA)
     {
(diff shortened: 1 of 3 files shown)


libxrdp/xrdp_bitmap32_compress.c
```diff
@@ -107,7 +107,7 @@ fsplit3(char *in_data, int start_line, int width, int e,
         }
         start_line--;
         cy++;
-        if (out_index > 64 * 64)
+        if (out_index + width + e > 64 * 64)
```
proski: I know it's pre-existing code, but if you define a constant for `64 * 64`, it would be easier for me to review the code.

libxrdp/xrdp_bitmap32_compress.c
```diff
@@ -459,35 +461,44 @@ xrdp_bitmap32_compress(char *in_data, int width, int height,
...
+                g_bytes = fpack(g_data, cx, cy, s);
+                b_bytes = fpack(b_data, cx, cy, s);
+                max_bytes = cx * cy * 3;
+                total_bytes = r_bytes + g_bytes + b_bytes;
+                if (total_bytes > max_bytes)
+                {
+                    if (2 + max_bytes <= byte_limit)
+                    {
+                        s->p = hold_p;
+                        foutraw3(s, cx * cy, FLAGS_NOALPHA, sr_data, sg_data, sb_data);
```
proski: I would add checks to the functions like `foutraw3` instead of doing those checks in the callers. The function could also print debug information if the data is being dropped. If any caller doesn't been the check, it's always possible to define `foutraw3_unchecked`.

libxrdp/xrdp_bitmap32_compress.c
```diff
@@ -459,35 +461,44 @@ xrdp_bitmap32_compress(char *in_data, int width, int height,
...
-            if (total_bytes > max_bytes)
-            {
-                /* raw is better */
-                LLOGLN(10, ("xrdp_bitmap32_compress: too big, rgb "
-                       "bytes %d %d %d total_bytes %d cx %d cy %d "
-                       "max_bytes %d", r_bytes, g_bytes, b_bytes,
-                       total_bytes, cx, cy, max_bytes));
-                init_stream(s, 0);
-                foutraw3(s, cx * cy, FLAGS_NOALPHA, sr_data, sg_data, sb_data);
+                s->p = hold_p;
```
proski: Why would you want to rewind the stream pointer? Does it discard the data? Cannot we avoid writing that data in the first place?

libxrdp/xrdp_orders.c
```diff
@@ -2235,6 +2235,11 @@ xrdp_orders_send_raw_bitmap(struct xrdp_orders *self,
 
     Bpp = (bpp + 7) / 8;
     bufsize = (width + e) * height * Bpp;
+    while (bufsize + 16 > MAX_ORDERS_SIZE)
```
proski: `bufsize > MAX_ORDERS_SIZE - 16` would be equivalent but easier on the eye.

libxrdp/xrdp_bitmap32_compress.c
jsorg71: If the bitmap does not fit, yes, you have to rewind.  This algorithm is planar.
With the interleaved algorithm, you can just stop when out of room.

libxrdp/xrdp_orders.c
jsorg71: I disagree

libxrdp/xrdp_bitmap32_compress.c
jsorg71: Minor issue, maybe later.

PR #553, opened 2016-12-16
add libpainter for drawing when client does not have minimum orders

jsorg71:
This is last part of noorders

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -148,6 +148,8 @@ struct xrdp_client_info
 
   int client_os_major;
   int client_os_minor;
+
+  int no_orders_supported;
 };
 
 #endif
--- a/configure.ac
+++ b/configure.ac
@@ -117,6 +117,11 @@ AC_ARG_ENABLE(pixman, AS_HELP_STRING([--enable-pixman],
               [Use pixman library (default: no)]),
               [], [enable_pixman=no])
 AM_CONDITIONAL(XRDP_PIXMAN, [test x$enable_pixman = xyes])
+AM_CONDITIONAL(XRDP_PAINTER, [test x$enable_painter = xyes])
+AC_ARG_ENABLE(painter, AS_HELP_STRING([--enable-painter],
+              [Use painter library (default: no)]),
+              [], [enable_painter=no])
+AM_CONDITIONAL(XRDP_PAINTER, [test x$enable_painter = xyes])
 
 # Check if -ldl is needed to use dlopen()
 DLOPEN_LIBS=
--- a/libxrdp/xrdp_caps.c
+++ b/libxrdp/xrdp_caps.c
@@ -159,6 +159,15 @@ xrdp_caps_process_order(struct xrdp_rdp *self, struct stream *s,
     DEBUG(("desktop cache size %d", i));
     in_uint8s(s, 4); /* Unknown */
     in_uint8s(s, 4); /* Unknown */
+
+    /* check if libpainter should be used for drawing, instead of orders */
+    if (!(order_caps[TS_NEG_DSTBLT_INDEX] && order_caps[TS_NEG_PATBLT_INDEX] &&
+          order_caps[TS_NEG_SCRBLT_INDEX] && order_caps[TS_NEG_MEMBLT_INDEX]))
+    {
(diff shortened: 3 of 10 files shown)


metalefty:
I fixed travis build. Could you rebase to current devel head and force push?  Then build test will succeed.

configure.ac
```diff
@@ -117,6 +117,11 @@ AC_ARG_ENABLE(pixman, AS_HELP_STRING([--enable-pixman],
               [Use pixman library (default: no)]),
               [], [enable_pixman=no])
 AM_CONDITIONAL(XRDP_PIXMAN, [test x$enable_pixman = xyes])
+AM_CONDITIONAL(XRDP_PAINTER, [test x$enable_painter = xyes])
+AC_ARG_ENABLE(painter, AS_HELP_STRING([--enable-painter],
+              [Use painter library (default: no)]),
+              [], [enable_painter=no])
+AM_CONDITIONAL(XRDP_PAINTER, [test x$enable_painter = xyes])
```
proski: If there any need to support configurations both with and without libpainter?

Please add `AC_CONFIG_SUBDIRS([libpainter])` to make libpainter a subpackage. I tried that, it works much better than I expected.

Also, the top-level Makefile.am should have `libpainter` added to `SUBDIRS` above `xrdp` to compile `libpainter` before it's used.

libxrdp/xrdp_caps.c
```diff
@@ -159,6 +159,15 @@ xrdp_caps_process_order(struct xrdp_rdp *self, struct stream *s,
     DEBUG(("desktop cache size %d", i));
     in_uint8s(s, 4); /* Unknown */
     in_uint8s(s, 4); /* Unknown */
+
+    /* check if libpainter should be used for drawing, instead of orders */
+    if (!(order_caps[TS_NEG_DSTBLT_INDEX] && order_caps[TS_NEG_PATBLT_INDEX] &&
+          order_caps[TS_NEG_SCRBLT_INDEX] && order_caps[TS_NEG_MEMBLT_INDEX]))
+    {
+        g_writeln("xrdp_caps_process_order: not enough orders supported by client, using painter.");
```
proski: This is important enough to be logged properly at the INFO level. Would not it be useful to print exactly which capability is missing?

common/xrdp_client_info.h
```diff
@@ -148,6 +148,8 @@ struct xrdp_client_info
 
   int client_os_major;
   int client_os_minor;
+
+  int no_orders_supported;
```
proski: I wonder if `painter_needed` or `painter_used` would be a better name. Does it need to be in `struct xrdp_client_info`? That structure is becoming a "god object" that knows everything.

metalefty:
Thanks!

xrdp/xrdp_types.h
```diff
@@ -429,6 +429,9 @@ struct xrdp_painter
   struct xrdp_session* session;
   struct xrdp_wm* wm; /* owner */
   struct xrdp_font* font;
+  void *painter;
+  struct xrdp_region *dirty_region;
+  int begin_end_level;
```
proski: The name looks confusing. begin_final_level? begin_level_end?

xrdp/Makefile.am
```diff
@@ -28,6 +28,12 @@ AM_CPPFLAGS += $(PIXMAN_CFLAGS)
 XRDP_EXTRA_LIBS += $(PIXMAN_LIBS)
 endif
 
+if XRDP_PAINTER
+AM_CPPFLAGS += -DXRDP_PAINTER
+AM_CPPFLAGS += -I$(top_srcdir)/libpainter/include
+XRDP_EXTRA_LIBS += $(top_srcdir)/libpainter/src/.libs/libpainter.a
```
proski: That should be `top_builddir` if libpainter is used or will be used as a subpackage.

configure.ac
proski: The proposed additions have been submitted as #555

jsorg71: ok, great

proski:
Let's merge it as is, and we can always rename variables later. I have a fixup patch that can be applied on top of this patch. It's important to get the code into the final shape and leave some time for testing.

metalefty:
I agree. Merging as is.

PR #555, opened 2016-12-16
Make libpainter a subpackage of xrdp

proski:
This is a part of the noorders task for 0.9.1. With this change, libpainter will be configured and compiled automatically as part of the build.


proski:
If we want libpainter support to be conditional for now, I'll need to change Makefile.am to respect XRDP_PAINTER, but I can only do it after #553 is merged.

speidy:
+1 
That's great!

PR #557, opened 2016-12-17
Fix Parallels Client on Mac

proski:
This is for 0.9.1. This code was in the original noorders branch (d3d2bdc001f951d7b456256bca756d79a4556d6f) but wasn't submitted (yet).

---

self->line_size should be calculated in xrdp_bitmap_create_with_data()

The code was in the original noorders branch but got lost. Without this
fix, the image is garbled.

The client should be configured with compression disabled, or it will
disconnect. That's a known problem.


proski:
@jsorg71 You are welcome to submit this change as a separate PR under your name, and I'll close this one.

PR #562, opened 2016-12-18
Distribute all files except git and github specific data

proski:
Another patch that needs to be in 0.9.1. I don't know in the 0.9.1 release will be made using `make dist`. If yes, it would be missing a lot of files that were included in older releases. Even the sources of X11rdp would not be included.

I believe that it would be better to err on the side of distributing files. If some files are truly useless, we should remove them from git first.


speidy:
+1

PR #563, opened 2016-12-18
Update submodules to the latest revisions

proski:
This is for 0.9.1

For libpainter, empty files `AUTHORS`, `ChangeLog`, `NEWS` and `README` would not be created by `bootstrap`. Those files are distributed if found. Distributing empty `README` would by not nice to the users.

For librfxcodec, it compiles with C++ now, so it brings the code to the level that we ask the contributors to follow. The compilation would also stops complaining of the unknown `NASM` tag.


metalefty:
+1

PR #566, opened 2016-12-20
Compile libpainter only if enabled, fix path to libpainter.a

proski:
This is for 0.9.1 - fixes on top of the last noorders patches.


metalefty:
+1

PR #567, opened 2016-12-20
Don't fail without working nasm if rfxcodec is not enabled

proski:
This is for 0.9.1. Users without nasm should still be able to compile xrdp if they don't specifically request rfxcodec.


metalefty:
+1

PR #568, opened 2016-12-20
Update version to 0.9.1

proski:
This is for 0.9.1, obviously, to be applied immediately before the release. We have many places where the version is used, let's take care of every occurrence. Please note that xrdp-sesadmin.8 uses 0.7.0 now, not 0.9.0.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 # Process this file with autoconf to produce a configure script
 
 AC_PREREQ(2.65)
-AC_INIT([xrdp], [0.9.0], [xrdp-devel@googlegroups.com])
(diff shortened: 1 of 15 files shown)


itamarjp:
why not using auto-tools to replace the version in man files ?

proski:
I like the idea of using autoconf to create manual pages. It would allow us to document paths as configured. But it's obviously not 0.9.1 material.

itamarjp:
I mean using @VERSION@ and AC_CONFIG_FILES()

proski:
I understand

itamarjp:
Can you send a new one, something like 0.9.1.1 ?

proski:
Done in #583, no need to change version at this time.

PR #576, opened 2016-12-23
Remove obsolete --enable-nopam option

proski:


metalefty:
+1 the option was obsoleted more than 2 years ago. we can remove it now.

PR #578, opened 2016-12-23
Warning fixes

proski:
This is a result of compiling xrdp with most warnings enabled and suppressing non-interesting warnings. This PR addressed duplicate and misplaced function declarations. It also fixes logging data that is not guaranteed to be initialized.


metalefty:
+1

PR #579, opened 2016-12-23
Use "void" for empty argument list in declarations

proski:
In C, an empty argument list in a declaration means that the function
can accept any arguments. Use "void" instead, it means "no arguments".

C++ treats void and empty list as "no arguments".


metalefty:
+1

PR #580, opened 2016-12-23
rail: remove unneeded call to XGetWindowAttributes()

proski:
It is already called above "if" with the same arguments.


metalefty:
+1 it's a clear fix. merging.

PR #581, opened 2016-12-23
One way clipboard feature

VMelandri:
This is a one-way clipboard basic implementation.
Discards clipboard selection event so they are not propagated to the client.

Feature needed for companies having to restrict the user ability to take data out from the server.

### DIFF
--- a/sesman/chansrv/Makefile.am
+++ b/sesman/chansrv/Makefile.am
@@ -9,7 +9,9 @@ AM_CPPFLAGS = \
   -DXRDP_SBIN_PATH=\"${sbindir}\" \
   -DXRDP_SHARE_PATH=\"${datadir}/xrdp\" \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
-  -I$(top_srcdir)/common
+  -I$(top_srcdir)/common \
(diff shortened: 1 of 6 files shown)


Natureshadow:
What are the unrelated changes, e.g. to Makefile.am, for?

I think you should fix the commit messages as well, at least the snafu one which does not tell amything useful.

VMelandri:
Well, it's really just a typo that I realized after creating the PR, pasted a string in wrong place :) Will squash.

The changes to the makefile are to include config.h in clipboard.c, needs to read configuration during clipboard init.
My C skills are VERY rusty, haven't used it in >10 years, please point me to a better way to do the same thing, if possible.

VMelandri:
mmmm seems like github does not allow me to rewrite history to squash my commits.. Will redo this
PR

Natureshadow:
>mmmm seems like github does not allow me to rewrite history to squash
>my commits.. Will redo this
>PR

Sure it does…

VMelandri:
let me retry

VMelandri:
It did let me, with --force, now the history in my repo seems fine, but this PR still shows the old commits.. 
¯\_(ツ)_/¯

PR #582, opened 2016-12-23
Outbound-restricted clipboard: initial implementation by skipping xorg selection events

VMelandri:
This is a one-way clipboard basic implementation.
Discards clipboard selection event so they are not propagated to the client.

Feature needed for companies having to restrict the user ability to take data out from the server.

> recreated after fixing commit history


sesman/chansrv/Makefile.am
```diff
@@ -63,7 +65,8 @@ xrdp_chansrv_SOURCES = \
   sound.c \
   sound.h \
   xcommon.c \
-  xcommon.h
+  xcommon.h \
+  config.c
```
moobyfr: This file list is ordered, so config.c should be added in another place. BTW the name config.c could be better renamed to reflext what it does

sesman/chansrv/clipboard.c
```diff
@@ -158,7 +158,7 @@ x-special/gnome-copied-files
 
 
 */
-
+#include <../config.h>
```
moobyfr: `#include <...>` is mostly used for system headers or external libraries, you should better use `#include "..."`. This could also fix the modifications you added in `Makefile.am`

sesman/chansrv/clipboard.c
```diff
@@ -372,6 +378,16 @@ clipboard_init(void)
         return 0;
     }
 
+	if (0 != config_read(&g_cfg)) {
+		g_printf("clipboard: error reading config. quitting.\n");
```
moobyfr: Some formatting issue : new line needed before the brace, and don't use tabs, but spaces. There is a file for coding standard at the base of the project

sesman/chansrv/config.c
```diff
@@ -0,0 +1 @@
+#include "../config.c"
```
moobyfr: That's strange, what is the idea here ? usually only headers are included. Here you duplicate all the file to be recompiled. For your feature, I thing you don't have to create file but just modify the  existing one.

sesman/config.c
```diff
@@ -235,16 +236,22 @@ config_read_security(int file, struct config_security *sc,
             }
         }
         if (0 == g_strcasecmp(buf, SESMAN_CFG_SEC_ALWAYSGROUPCHECK))
-        {
-            sc->ts_always_group_check = g_text2bool((char *)list_get_item(param_v, i));
-        }
+		{
```
speidy: bad indentation, please user spaces, not tabs.
look at xrdp coding style guide in wiki.

sesman/config.h
```diff
@@ -59,6 +59,7 @@
 #define SESMAN_CFG_SEC_USR_GROUP     "TerminalServerUsers"
 #define SESMAN_CFG_SEC_ADM_GROUP     "TerminalServerAdmins"
 #define SESMAN_CFG_SEC_ALWAYSGROUPCHECK "AlwaysGroupCheck"
+#define SESMAN_CFG_RESTRICT_CLIPBOARD "RestrictClipboard"
```
speidy: RestrictClipboard doesn't mention the direction of restriction.
I would prefer RestrictOutboundClipboard which is more clear.

The direction should be expressed in the code as well.

sesman/chansrv/clipboard.c
```diff
@@ -225,12 +225,16 @@ static char g_bmp_image_header[] =
...
 
+struct config_sesman g_cfg; /* config.h */
+
 extern Display *g_display;      /* in xcommon.c */
 extern int g_x_socket;          /* in xcommon.c */
 extern tbus g_x_wait_obj;       /* in xcommon.c */
 extern Screen *g_screen;        /* in xcommon.c */
 extern int g_screen_num;        /* in xcommon.c */
 
+int restricted = 0;
```
speidy: please add g_ prefix for global vars. (like g_x_socket)

sesman/chansrv/clipboard.c
```diff
@@ -372,6 +378,16 @@ clipboard_init(void)
         return 0;
     }
 
+	if (0 != config_read(&g_cfg)) {
+		g_printf("clipboard: error reading config. quitting.\n");
+		return 1;
+	}
+
+	log_debug("clipboard_init: clipboard restricted -> "+g_cfg.sec.restrict_clipboard)
```
speidy: please use spaces around `+`

sesman/chansrv/clipboard.c
```diff
@@ -2490,7 +2508,13 @@ clipboard_xevent(void *xevent)
     switch (lxevent->type)
     {
         case SelectionNotify:
-            clipboard_event_selection_notify(lxevent);
+        	if (restricted == 0) {
```
speidy: bad indentation, tabs to spaces.

sesman/chansrv/config.c
VMelandri: I had problem referencing it from clipboard.c, Don't really get how it should have been done properly. 
My C has not been in use since high school, last century :)

VMelandri:
Please take a look if changes are now up to standard :)

sesman/chansrv/clipboard.c
```diff
@@ -372,6 +378,16 @@ clipboard_init(void)
         return 0;
     }
 
+    if (0 != config_read(&g_cfg)) {
```
proski: I don't think it's a good design to make xrdp-chansrv read the configuration file of xrdp-sesman.

There is an existing mechanism for disabling channels. The configuration is in `xrdp.ini` and it's xrdp that takes care of it. xrdp is talking to xrdp-chansrv using `server_send_to_channel()`

I would rather make xrdp tell xrdp-chansrv not to send clipboard data to the clients. The configuration parsing would be done in xrdp. The configuration would be in `xrdp.ini`.

sesman/sesman.ini
```diff
@@ -13,6 +13,9 @@ TerminalServerAdmins=tsadmins
 ; When AlwaysGroupCheck=false access will be permitted
 ; if the group TerminalServerUsers is not defined.
 AlwaysGroupCheck=false
+; When RestrictOutboundClipboard=true clipboard from the
+; server is not pushed to the client.
+RestrictOutboundClipboard=false
```
proski: A manual page change is needed too. But let's get the design right first.

VMelandri:
Manual page is in.

sesman/chansrv/clipboard.c
```diff
@@ -312,7 +318,7 @@ get_atom_text(Atom atom)
     return g_last_atom_name;
 }
 
-/*****************************************************************************/
+/**************************login_retry***************************************************/
```
VMelandri: Been testing the cut+paste a lot to test the change, guess how that happened... :) will fix.

sesman/chansrv/Makefile.am
```diff
@@ -9,7 +9,9 @@ AM_CPPFLAGS = \
   -DXRDP_SBIN_PATH=\"${sbindir}\" \
   -DXRDP_SHARE_PATH=\"${datadir}/xrdp\" \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
-  -I$(top_srcdir)/common
+  -I$(top_srcdir)/common \
+  -I$(top_srcdir)/sesman \
+  -I$(top_srcdir)/sesman/libscp
```
VMelandri: need to read the sesman configuration, needed to reference config.c from sesman

VMelandri:
merged devel into branch to remove conflict on sesman.ini.5 man file

sesman/chansrv/clipboard.c
VMelandri: Argh, haven't read this comment yet. I see, that is kind of what i was thinking too, but was not sure about architecture and did not know where in xrdp made more sense to stop the data. Or how to tell if data are inbound/outbound. Will try to wrap my head around it and get back to you.

VMelandri: Turned out to be not as easy this in xrdp (at least for me), this solution is clearer and cuts the clipboard flow at the source.
About reading the sesman conf, I would argue that chansrv is part of sesman, it does not seem inappropriate.

VMelandri:
@speidy @proski can we go ahead with this? I don't think there will be more development on this.

proski:
I think the right approach would be to add support for private configuration flags. Xrdp sends message of type 1 (init) for successful login. 3 is used for configuration, but it always comest from RDP clients. We could use another number, say, 7003, for private configuration flags, with the same parameters. That message could be sent by xrdp immediately after the init message.

bolkedebruin:
This would be an awesome feature. How can we get this moving again?

metalefty:
No feedback, no response to our change request. Out of scope now.

PR #583, opened 2016-12-24
Substitute variables in manual pages

proski:
This is valuable for the end user, as the actual paths are used in the manual pages.

The manual pages don't need to be updated with the new version before every release.

The xrdp-xcon manual page had to be removed. Make refused to build it, which is reasonable. Test tools that are not installed don't need manual pages. They can be documented by other means.

Autoconf documentation recommends using substitution in the makefile, not in AC_CONFIG_FILES, as the directories are only guaranteed to be fully expanded in the makefiles.


proski:
Thank you! In the future, please follow the 24 hours rule. A pull request should be at least 24 hours old for merging to let others comment on it. Even the PR author can find that additional improvements are needed.

PR #584, opened 2016-12-25
Add --enable-strict-locations option, describe default in its help text

proski:
Rather than guess user's intention based on whether --prefix was
specified, use an explicit option to enable strict GNU Coding Standards
for installation directories.

The default is to use /etc and /var rather than corresponding directories
under prefix.

Use --enable-strict-locations in "make distcheck", it expects all
installed files to be under prefix.


proski:
The choice of the command line option name is consistent with xorgxrdp, which already has that option.

As it stands now, `--prefix` has unexpected side effects. The PR addresses the issue by

* decoupling `--prefix` from non-standard install directories
* documenting the default directories in `./configure --help`
* providing an option to get standard behavior

metalefty:
+1

PR #585, opened 2016-12-25
.travis.yml: use -Werror, test two configurations - minimal and maximal

proski:
Use separate entries for every command. Travis CI would still run the
commands sequentially, and it would provide better annotation for the
output.


metalefty:
LGTM.

PR #586, opened 2016-12-25
Fix all warnings in TurboJPEG code

proski:
Actually use the error code from tjCompress() by logging the errors.

Make sure width is more than zero before filling the pad with the last
pixel data.


speidy:
Looks good to me.

PR #587, opened 2016-12-25
bootstrap: checkout submodules if they are missing

proski:


Natureshadow:
Maybe we should simply use subtrees instead of tying code to deprecated and fragile Git specifics.

proski:
OK, then please see #593

PR #588, opened 2016-12-25
keygen: add CA extensions to self-signed certificates

speidy:
Self-signed certificates lacks the required CA extensions.
MS RDP client for iOS cannot connect because of it. the app is crashing right after TLS handshake.
I think it worked before we added our openssl.conf file.

### DIFF
--- a/keygen/openssl.conf
+++ b/keygen/openssl.conf
@@ -1,41 +1,36 @@
 [req]
 distinguished_name = req_distinguished_name
-x509_extensions = v3_ca	# The extentions to add to the self signed cert
+# The extensions to add to the self signed cert
+x509_extensions = v3_ca
 
 [req_distinguished_name]
 
-[ v3_ca ]
-
-# Extensions for a typical CA
-
-
-# PKIX recommendation.
-
-subjectKeyIdentifier=hash
-
-authorityKeyIdentifier=keyid:always,issuer
+[v3_ca]
+# Extensions for a typical CA - PKIX recommendation.
+subjectKeyIdentifier = hash
+authorityKeyIdentifier = keyid:always, issuer
 
 # This is what PKIX recommends but some broken software chokes on critical
 # extensions.
-#basicConstraints = critical,CA:true
+#basicConstraints = critical, CA:true
 # So we do this instead.
 basicConstraints = CA:true
 
 # Key usage: this is typical for a CA certificate. However since it will
 # prevent it being used as an test self-signed certificate it is best
 # left out by default.
-# keyUsage = cRLSign, keyCertSign
+#keyUsage = cRLSign, keyCertSign
 
 # Some might want this also
-# nsCertType = sslCA, emailCA
+#nsCertType = sslCA, emailCA
 
 # Include email address in subject alt name: another PKIX recommendation
-# subjectAltName=email:copy
+#subjectAltName = email:copy
 # Copy issuer details
-# issuerAltName=issuer:copy
+#issuerAltName = issuer:copy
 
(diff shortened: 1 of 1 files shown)


keygen/openssl.conf
```diff
@@ -1,4 +1,41 @@
 [req]
 distinguished_name = req_distinguished_name
+x509_extensions = v3_ca	# The extentions to add to the self signed cert
 
 [req_distinguished_name]
+
+[ v3_ca ]
```
proski: I would prefer that we don't use spaces inside section names. It could be just `[v3_ca]`

keygen/openssl.conf
```diff
@@ -1,4 +1,41 @@
...
+
+# Extensions for a typical CA
+
+
+# PKIX recommendation.
+
```
proski: Too many empty lines. You can write: "PKIX recommendation - extensions for a typical CA". That would be one line.

keygen/openssl.conf
```diff
@@ -1,4 +1,41 @@
...
+subjectKeyIdentifier=hash
```
proski: Please use spaces around `=` consistently.

keygen/openssl.conf
```diff
@@ -1,4 +1,41 @@
...
+# This is what PKIX recommends but some broken software chokes on critical
+# extensions.
+#basicConstraints = critical,CA:true
+# So we do this instead.
+basicConstraints = CA:true
+
+# Key usage: this is typical for a CA certificate. However since it will
+# prevent it being used as an test self-signed certificate it is best
+# left out by default.
+# keyUsage = cRLSign, keyCertSign
```
proski: Let's distinguish comments and commented out parameters. For instance, comments could start with two hashes. Or the commented out parameters should not have a space after hash.

keygen/openssl.conf
```diff
@@ -1,4 +1,41 @@
...
+# nsCertType = sslCA, emailCA
+
+# Include email address in subject alt name: another PKIX recommendation
+# subjectAltName=email:copy
+# Copy issuer details
+# issuerAltName=issuer:copy
+
+# DER hex encoding of an extension: beware experts only!
+# obj=DER:02:03
+# Where 'obj' is a standard or added object
```
proski: It would be better to put comments above the parameters they describe.

keygen/openssl.conf
```diff
@@ -1,4 +1,35 @@
...
+
+# Some might want this also
+#nsCertType = sslCA, emailCA
+
+# Include email address in subject alt name: another PKIX recommendation
+# subjectAltName=email:copy
+# Copy issuer details
+#issuerAltName=issuer:copy
+
+# DER hex encoding of an extension: beware experts only!
```
proski: "beware" is not needed, "experts only" is cleat enough

speidy:
no space after hash means an example line. I would like to keep it that way.

keygen/openssl.conf
```diff
@@ -1,4 +1,35 @@
 [req]
 distinguished_name = req_distinguished_name
+x509_extensions = v3_ca	# The extentions to add to the self signed cert
```
proski: extentions -> extensions
Please avoid tabs. The comment should be above that line, not at the end.

proski:
I've tested the patch with several Android clients. I don't see any difference w.r.t. authentication. Parallels Client (legacy and non-legacy) refuses to work. Other clients work. If any client is fixed by this patch, let's merge it.

speidy:
In my testing, only iOS client was crashing. Parallels clients have another issues, not related to TLS.

PR #589, opened 2016-12-25
Use constant strings for input arguments in keygen, kill dead code

proski:
It's hard to read code where input string arguments use type `char *`, especially if it's cryptographic code. Functions should not modify its input arguments. The compiler can produce better code if it knows the constraints.

Remove code for compatibility with OpenSSL prior to 0.9.8. It's hard to find old versions of OpenSSL. Everybody who cares about security should be using OpenSSL 1.0.2 or newer. But let's allow for older OpenSSL for systems that backport security fixes.

### DIFF
--- a/common/ssl_calls.c
+++ b/common/ssl_calls.c
@@ -344,8 +344,8 @@ ssl_reverse_it(char *p, int len)
 
 /*****************************************************************************/
 int APP_CC
-ssl_mod_exp(char *out, int out_len, char *in, int in_len,
-            char *mod, int mod_len, char *exp, int exp_len)
+ssl_mod_exp(char *out, int out_len, const char *in, int in_len,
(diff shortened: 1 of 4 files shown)


metalefty:
That makes sense. OpenSSL 0.9.8 is no longer supported. https://www.openssl.org/policies/releasestrat.html

metalefty:
No objections to drop OpenSSL 0.9.8 support?

proski:
Actually, this PR removes support for OpenSSL versions _before_ 0.9.8. 0.9.8 is still "supported". It should be a no-brainer.

jsorg71:
The only reason I can see for allowing older openssl is if someone is supporting a legacy package on a machine behind a firewall.  In general I don't like to tell people where they can use xrdp.
But I won't protest this PR.
+1

PR #590, opened 2016-12-25
Respect verbosity settings when generating manual pages

proski:


jsorg71:
merged as 7b32fc864c6ad9bb4115d67165e4c31889c2ee01

PR #591, opened 2016-12-25
Suppress versioning for modules, tell libtool they are modules

proski:
Versioning is for libraries. Modules are not libraries; no code is linked
against them.

Libtool makes sure the modules can be opened by dlopen(). That is already
true for ELF format, but other file formats may need special processing.


jsorg71:
+1

jsorg71:
merged as f25659d2eeb32efb519a96e1f1ab4894de115807

PR #593, opened 2016-12-25
Convert git submodules to git subtrees

proski:


Natureshadow:
I'd not squash the histories. Keeping full history is a good thing (and makes finding things and splitting easier for others).

proski:
OK, the history is preserved now.

proski:
I don't know if this patch can be merged safely. I suggest using this gist on the clean checkout and submitting the result. https://gist.github.com/proski/c1a240426d2fa58fca57116987b85d61

jsorg71:
Do we really want to switch to subtrees?
I'm not convinced they are better.

jsorg71:
I did read up on subtrees but the goal is to move the submodules out entirely.
I want to move them away not integrate them more tightly.
I can live with either too so I think it's up to you guys.

proski:
Oh, I see. That makes sense. Then maybe you could merge #587 for now? It can always be reverted as the submodules are devolved. I just want to close this topic one way or another.

jsorg71:
I'll merge that one.
@Natureshadow is pro subtree
@proski  you sound neutral, so am I.
If you other guys want subtrees lets do it but I think we should do a new PR.

proski:
Closing this PR. If anybody wants to continue this effort, you can use my gist for conversion.

PR #596, opened 2016-12-29
RFX: fix frame acks handling

speidy:
This should make RemoteFX usable with Parallels Client. they always want zero unacked frames on the wire.

### DIFF
--- a/libxrdp/xrdp_caps.c
+++ b/libxrdp/xrdp_caps.c
@@ -541,6 +541,8 @@ xrdp_caps_process_frame_ack(struct xrdp_rdp *self, struct stream *s, int len)
     in_uint32_le(s, self->client_info.max_unacknowledged_frame_count);
     if (self->client_info.max_unacknowledged_frame_count < 0)
     {
+        g_writeln("  invalid max_unacknowledged_frame_count value (%d), setting to 0",
+                  self->client_info.max_unacknowledged_frame_count);
         self->client_info.max_unacknowledged_frame_count = 0;
     }
     g_writeln("  max_unacknowledged_frame_count %d", self->client_info.max_unacknowledged_frame_count);
--- a/libxrdp/xrdp_caps.c
+++ b/libxrdp/xrdp_caps.c
@@ -539,6 +539,10 @@ xrdp_caps_process_frame_ack(struct xrdp_rdp *self, struct stream *s, int len)
     g_writeln("xrdp_caps_process_frame_ack:");
     self->client_info.use_frame_acks = 1;
     in_uint32_le(s, self->client_info.max_unacknowledged_frame_count);
+    if (self->client_info.max_unacknowledged_frame_count < 0)
+    {
+        self->client_info.max_unacknowledged_frame_count = 0;
+    }
     g_writeln("  max_unacknowledged_frame_count %d", self->client_info.max_unacknowledged_frame_count);
     return 0;
 }
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -2258,21 +2258,6 @@ xrdp_mm_check_wait_objs(struct xrdp_mm *self)
                                                  enc_done->enc->flags,
                                                  enc_done->enc->frame_id);
                     }
-                    else
-                    {
-#if 1
-                        ex = self->wm->client_info->max_unacknowledged_frame_count;
-                        if (self->encoder->frame_id_client + ex > self->encoder->frame_id_server)
-                        {
-                            if (self->encoder->frame_id_server > self->encoder->frame_id_server_sent)
-                            {
-                                LLOGLN(10, ("xrdp_mm_check_wait_objs: 1 -- %d", self->encoder->frame_id_server));
-                                self->encoder->frame_id_server_sent = self->encoder->frame_id_server;
-                                self->mod->mod_frame_ack(self->mod, 0, self->encoder->frame_id_server);
-                            }
-                        }
-#endif
-                    }
                     g_free(enc_done->enc->drects);
                     g_free(enc_done->enc->crects);
                     g_free(enc_done->enc);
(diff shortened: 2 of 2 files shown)


libxrdp/xrdp_caps.c
```diff
@@ -539,6 +539,10 @@ xrdp_caps_process_frame_ack(struct xrdp_rdp *self, struct stream *s, int len)
     g_writeln("xrdp_caps_process_frame_ack:");
     self->client_info.use_frame_acks = 1;
     in_uint32_le(s, self->client_info.max_unacknowledged_frame_count);
+    if (self->client_info.max_unacknowledged_frame_count < 0)
+    {
+        self->client_info.max_unacknowledged_frame_count = 0;
+    }
     g_writeln("  max_unacknowledged_frame_count %d", self->client_info.max_unacknowledged_frame_count);
```
proski: I would prefer to log the raw data. If it's -1 on the wire, let's log -1 and threat is as 0.

Does it happen often? Could it indicate some error?

jsorg71:
```
                                                  enc_done->enc->flags,
                                                  enc_done->enc->frame_id);
                     }
-                    else
-                    {
-#if 1
-                        ex = self->wm->client_info->max_unacknowledged_frame_count;
-                        if (self->encoder->frame_id_client + ex > self->encoder->frame_id_server)
-                        {
-                            if (self->encoder->frame_id_server > self->encoder->frame_id_server_sent)
-                            {
-                                LLOGLN(10, ("xrdp_mm_check_wait_objs: 1 -- %d", self->encoder->frame_id_server));
-                                self->encoder->frame_id_server_sent = self->encoder->frame_id_server;
-                                self->mod->mod_frame_ack(self->mod, 0, self->encoder->frame_id_server);
-                            }
-                        }
-#endif
-                    }
                     g_free(enc_done->enc->drects);
                     g_free(enc_done->enc->crects);
                     g_free(enc_done->enc);
```
why remote this else, it's needed in case client does not support frame acks.

xrdp/xrdp_mm.c
```diff
@@ -2258,21 +2258,6 @@ xrdp_mm_check_wait_objs(struct xrdp_mm *self)
...
-                    {
-#if 1
-                        ex = self->wm->client_info->max_unacknowledged_frame_count;
-                        if (self->encoder->frame_id_client + ex > self->encoder->frame_id_server)
-                        {
-                            if (self->encoder->frame_id_server > self->encoder->frame_id_server_sent)
-                            {
-                                LLOGLN(10, ("xrdp_mm_check_wait_objs: 1 -- %d", self->encoder->frame_id_server));
-                                self->encoder->frame_id_server_sent = self->encoder->frame_id_server;
-                                self->mod->mod_frame_ack(self->mod, 0, self->encoder->frame_id_server);
```
proski: Why is this code removed?

xrdp/xrdp_mm.c
```diff
@@ -2303,7 +2288,7 @@ xrdp_mm_frame_ack(struct xrdp_mm *self, int frame_id)
         return 1;
     }
     ex = self->wm->client_info->max_unacknowledged_frame_count;
-    if (self->encoder->frame_id_client + ex > self->encoder->frame_id_server)
+    if (self->encoder->frame_id_client + ex >= self->encoder->frame_id_server)
```
proski: Could you add some a short comment here? I cannot figure out why this change is needed.

libxrdp/xrdp_caps.c
speidy: It's added for safety. It shouldn't happen. 
I'll add a log.

xrdp/xrdp_mm.c
speidy: This part it not needed anymore since it is kind of workarounds the issue I solved in xrdp_mm_frame_ack().

speidy: @jsorg71 regarding clients doesn't frame acks, its handeled correctly here
https://github.com/speidy/xrdp/blob/1f930f252ff1ccac30bae24740b0a2a7fd6f6bac/xrdp/xrdp_mm.c#L2255

xrdp/xrdp_mm.c
speidy: That change is needed in the case when max_unacknowledged_frame_count is 0. that happens with Parallels Client.
In that case the client would send a frame ack on frame 1, where the current server frame is also 1.

speidy: That change is needed in the case when max_unacknowledged_frame_count is 0. that happens with Parallels Client.
In that case the client would send a frame ack on frame 1, where the current server frame is also 1.

xrdp/xrdp_mm.c
speidy: @jsorg71 regarding clients doesn't support frame acks, its handeled correctly here
https://github.com/speidy/xrdp/blob/1f930f252ff1ccac30bae24740b0a2a7fd6f6bac/xrdp/xrdp_mm.c#L2255

libxrdp/xrdp_caps.c
```diff
@@ -539,6 +539,12 @@ xrdp_caps_process_frame_ack(struct xrdp_rdp *self, struct stream *s, int len)
     g_writeln("xrdp_caps_process_frame_ack:");
     self->client_info.use_frame_acks = 1;
     in_uint32_le(s, self->client_info.max_unacknowledged_frame_count);
+    if (self->client_info.max_unacknowledged_frame_count < 0)
+    {
+        g_writeln("  invalid max_unacknowledged_frame_count value (%d), setting to 0",
```
proski: I think we should avoid g_writeln() in the new and modified code. That would be `log_message(LOG_LEVEL_ERROR, ...)`

Also, is the largest positive integer a good limit? It could be a problem too if streams go out-of-sync. Let's have some reasonable limit. Maybe 8192.

jsorg71: We should go though(at some point) and move all g_writeln calls to log calls.

xrdp/xrdp_mm.c
jsorg71: @proski I think @speidy is right, that code is not needed and can actually cause issue.

proski:
OK, I just wanted to make sure it's not needed.

libxrdp/xrdp_caps.c
proski: OK, no objection for this patch. We can fix it separately.

xrdp/xrdp_mm.c
proski: OK, I just wanted to make sure.

PR #603, opened 2016-12-30
Some good things extracted from the X11rdp MacOS port

proski:
The most important fix is restoring the download cache functionality that was broken by 6b589569ee2aeb20f157ef73b2ca240166fef6a4


proski:
I've added a patch that is even more important. X11rdp was using standard path to fonts. It would not find fonts on a system without existing fonts. With the fix, X11rdp uses the fonts that are built together with X11rdp.

metalefty:
Looks good.

PR #605, opened 2016-12-31
Allow PAM file selection in configure, improve autodetection, add SUSE

proski:
Use easy to understand names for config files (debian, redhat, suse,
unix). Move all autodetection to a separate script.


metalefty:
Looks good to me for FreeBSD. `xrdp-sesman.unix` formerly named as `xrdp-sesman.other`  is valid pamfile for FreeBSD and your patch correctly choose it.

speidy:
Lets resolve the conflict and merge

PR #607, opened 2017-01-01
Exit with error message if pkg-config is not found

proski:
PKG_PROG_PKG_CONFIG does not exit if it fails to find pkg-config.

Continuing without pkg-config leads to misleading error messages when
checks for packages fail.


speidy:
+1

jsorg71:
bootstrap checks for pkg-config but I think the released code will already have ./bootstrap run.
+1

PR #611, opened 2017-01-04
fix build with --enable-xrdpdebug=yes

volth:
support build with ```./configure --enable-xrdpdebug=yes```


common/Makefile.am
```diff
@@ -18,6 +18,12 @@ AM_CPPFLAGS = \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
   -DXRDP_LOG_PATH=\"${localstatedir}/log\"
 
+if XRDP_DEBUG
+AM_CPPFLAGS += -DXRDP_DEBUG
+else
+AM_CPPFLAGS += -DXRDP_NODEBUG
```
proski: XRDP_NODEBUG is not used anywhere in the code. Instead of adding it, let's remove it from the makefiles where it's defined.

volth: it is used in ```defines.h``` as
```
#ifdef XRDP_DEBUG
#define DEBUG(args) g_writeln args;
#define LIB_DEBUG(_mod, _text) _mod->server_msg(_mod, _text, 1);
#else
#define DEBUG(args)
#define LIB_DEBUG(_mod, _text)
#endif
```
then ```DEBUG``` and ```LIB_DEBUG``` used in the code compiled by those makefiles

proski: Either you misread my comment, or you quoted wrong piece of code. How would this code break if XRDP_NODEBUG is not defined? I only see XRDP_NODEBUG in Makefile.am, not in the actual code.

volth: There are two ```XRDP_DEBUG```: the environment value set by ```./configure --enable-xrdpdebug=yes``` which then in some of Makefiles results in the second ```XRDP_DEBUG``` of C preprocessor.
If a Makefile in a particular directory does not have this code, ```./configure --enable-xrdpdebug=yes``` does not result in passing ```-DXRDP_DEBUG``` in the command line of C preprocessor/compiler and those files will be compiled without ```XRDP_DEBUG``` defined.

Almost every ```.c```-files depend on ```XRDP_DEBUG```.
Even if a particular ```.c```-file does not have ```XRDP_DEBUG``` string in the source code, it directly or indirectly includes [defines.h](https://github.com/neutrinolabs/xrdp/blob/devel/common/defines.h#L25) which has macroses dependend on ```XRDP_DEBUG```.

Removing the snippet from Makefile does not "break" the code if we call "break" the imposibility to produce the output. I meant that without this 
```
if XRDP_DEBUG
AM_CPPFLAGS += -DXRDP_DEBUG
else
AM_CPPFLAGS += -DXRDP_NODEBUG
```
debug logging is always off irregardless of ```--enable-xrdpdebug=yes``` because the ```XRDP_DEBUG``` is not passed to the C preprocessor, and ```LIB_DEBUG``` in the files in that directory compiles to no code.

proski: Please read my comments. I'm talking of XRDP_NODEBUG only, not of XRDP_DEBUG.

proski:
+1

Issue #551, opened 2016-12-15
Move keymap "conffiles" out of the way


Natureshadow:
We have been experiencing quite some issues with the `km-*.ini` files in Debian, because of how conffiles are handled by dpkg (and probably other packaging systems).

I do not think that these are config files at all. They might be in a strictly technical sense, but then, do you really expect users to chagne these files in order to modify the behaviour of xrdp?

From my point of view, these files should go to `/usr/share` as simple data files.

I could prepare this as a pull request for 0.9.1 if you agree.

proski:
There may be situations where `/usr/share` is read-only but `/etc` is not. Some users may actually want to add or adjust a keymap for their language, which could hinder them.

I agree that the distributed keymaps should go to `/usr/share`, but the users should also have a way to override or augment them, and I don't think such feature existing now. If it's possible, it should be documented.

I suggest that we concentrate of true bugs for 0.9.1.

PR #614, opened 2017-01-05
added cmdline param --cfg-path to xrdp and xrdp-sesman

### LINKED ISSUE #609: path to xrdp.ini and sesman.ini as command line parameter
Please add possibility to specify path to ```xrdp.ini``` and ```sesman.ini``` in command-line of ```xrdp``` and ```xrdp-sesman``` (e.g. ```xrdp --conf=/somepath/xrdp.ini```).

It would allow to port ```xrdp``` to [nixos](http://nixos.org).

volth:
added cmdline param ```--cfg-path``` to xrdp and xrdp-sesman.

it allows to change the config directory (which is ```/etc/xrdp``` by default) without recompiling the project.

fixes #609

### DIFF
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -635,6 +635,7 @@ main(int argc, char **argv)
     }
 
     g_free(startup_params);
+    log_end();
     g_deinit();
     return 0;
 }


metalefty:
I roughly agree your fix but I like `-f` and `--config` for the cli option. It's a personal preference though.
What everyone else do you think about that?

metalefty:
Ah, `-f` is already in use for fork option.

proski:
`--config` would be good for a single configuration file (e.g. `xrdp.ini`). But the option specifies path for all configuration files. Calling it `--config` would be misleading.

That said, I would prefer to have the `--config` option. Multiple instances of xrdp could use the same settings apart from the port and log files. Why duplicate the whole configuration directory?

volth:
Because there are other files in ```XRDP_CFG_PATH``` besides ```xrdp.ini``` and ```xrdp-sesmain.ini```.
There are ```rsakeys.ini```, ```xrdp_keyboard.ini``` and a lot of ```km-*.ini```.
1. should each of the files have its own command-line key?
2. or own entry in ```xrdp.ini```?
3. or should they reside in the directory specified in compile-time? 

Each of these solutions brings extra complexity and could be inconvenient to somebody (e.g. to those people who make .deb packages).
It requires community feedback.
What I added is simple overriding of the compile-time setting ```XRDP_CFG_PATH``` with command line key ```--cfg-path```

volth:
if you want to run few ```xrdp``` instance simultaneously, perhaps you should also override ```/tmp/.xrdp```

proski:
`rsakeys.ini` should be specified in `xrdp.ini`, it's an omission that it's not configurable. All keyboard related configs should be in `/usr/share/xrdp` as already suggested in #551. `startwm.sh` is in `sesman.ini`, it should be possible to specify if with full path. We should have configurable paths in the master configuration files, i.e. `xrdp.ini` and `sesman.ini`.

The commands line argument for utilities like `xrdp-chansrv` is not useful, as they are not supposed to be run by the end users. They should get their configuration from other utilities, as discussed in #582.

You are correct that there are issues besides `/etc/xrdp`, pidfiles would be an issue too.

volth:
paths to ```startwm.sh```, ```cert.pem``` and ```key.pem``` are already can be set up in ```xrdp.ini```.

directory with keyboard configs should be configurable as well (there is no ```/usr/share/xrdp``` on nixos).

```xrdp-chansrv``` gets the path from other utility [sesman launches it with cfg-path as the command line parameter](https://github.com/neutrinolabs/xrdp/pull/614/files#diff-0a29fa4d88ec74be7f54cbe04c8d6db6R118). How else ```could``` sesman pass this info to ```xrdp-chansrv``` ? I see no answer in #582

**Comment by volth** [comment]
Also, ```xrdp``` reads both config files ```xrdp.ini``` and ```sesman.ini```.

**Comment by proski** [comment]
The path to `rsakeys.ini` should be in `xrdp.ini`. The contents should be separate, `rsakeys.ini` is readable by root only.

Keyboard configs should be in the directory where bitmaps, cursors and fonts are installed. That's `$(datadir)/xrdp`. It can be made configurable if needed.

I believe `xrdp-chansrv` should not use any configuration files. I don't see a good reason to have configuration files for every utility.

`xrdp` reading `sesman.ini` is bad, we should fix it.

File: libxrdp/xrdp_rdp.c [hunk 1]
```diff
@@ -244,6 +234,48 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
...
+                            "rsakeys.ini file should use absolute path, using "
+                            "default instead: %s", client_info->rsakeys_ini_file);
+            }
+            else
+            {
+                /* use user defined rsakeys.ini */
+                g_strncpy(client_info->rsakeys_ini_file, value, 1023);
+            }
+        }
+        else if (g_strcasecmp(item, "xrdp_keyboard_ini") == 0)
```
proski: Is this feature really important? Why would different instances of xrdp need to use different keyboard configuration?

sesman/config.c
```diff
@@ -100,11 +106,12 @@ config_read(struct config_sesman *cfg)
 
 /******************************************************************************/
 int DEFAULT_CC
-config_read_globals(int file, struct config_sesman *cf, struct list *param_n,
+config_read_globals(char *base_dir, int file, struct config_sesman *cf, struct list *param_n,
```
proski: Please always use `const char *` for string arguments unless the string is used for output.

xup/xup.c
```diff
@@ -1358,7 +1358,7 @@ lib_send_client_info(struct mod *mod)
 
     g_writeln("lib_send_client_info:");
     make_stream(s);
-    init_stream(s, 8192);
+    init_stream(s, (sizeof(mod->client_info))+0xFFF & ~0xFFF);
```
proski: What is this? I believe such calculations need to have comments. Or they should be done in macros or functions with descriptive names.

common/os_calls.c
```diff
@@ -2267,6 +2267,19 @@ g_strchr(const char* text, int c)
 }
 
 /*****************************************************************************/
+/* locates char in text */
```
volth: copy-paste from few lines above changing ```strchr``` to ```strrchr```

libxrdp/xrdp_rdp.c
```diff
@@ -36,25 +36,37 @@
...
     values->auto_free = 1;
-    g_snprintf(cfg_file, 255, "%s/xrdp.ini", XRDP_CFG_PATH);
-    DEBUG(("cfg_file %s", cfg_file));
-    file_by_name_read_section(cfg_file, "globals", items, values);
+    DEBUG(("xrdp_ini_file %s", xrdp_ini_file));
+
+    g_strncpy(xrdp_ini_dir, xrdp_ini_file, 255);
+    p = (char*)g_strrchr(xrdp_ini_dir, '/');
+    if (p == NULL) {
+        log_message(LOG_LEVEL_ERROR, "cannot calculate dirnane of: %s", xrdp_ini_file);
```
volth: here actually should be ```assert``` and validation that ```--config``` contains '/' where command line is parsed

sesman/config.c
volth: it looks like a proposal for global refactoring.

xup/xup.c
volth: the structure became bigger that the magic value of 8192 resulted in xrdp crashes.

common/xrdp_client_info.h
```diff
@@ -128,6 +128,9 @@ struct xrdp_client_info
 
   char certificate[1024];
   char key_file[1024];
+  char rsakeys_ini_file[1024];
+  char xrdp_keyboard_ini_file[1024];
+  char keymaps_path[1024];
```
volth: is it a problem? touching any file in common/ resulted in xordxrdp recompilation.

common/os_calls.c
```diff
@@ -2267,6 +2267,19 @@ g_strchr(const char* text, int c)
...
+const char *APP_CC
+g_strrchr(const char* text, int c)
```
volth: it is copy-paste from few lines above changing ```strchr``` to ```strrchr```

libxrdp/xrdp_rdp.c
```diff
@@ -222,18 +222,8 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
         }
         else if (g_strcasecmp(item, "key_file") == 0)
         {
-            g_memset(client_info->key_file, 0, sizeof(char) * 1024);
-            if (g_strlen(value) == 0)
```
volth: > This change makes the default configuration invalid, so it's not good.
I do not understand.
The default values are set above. before this change they was not set if ```key_file``` was absent, ```key_file=``` was required to set the default value.

libxrdp/xrdp_rdp.c
volth: If the keyboard configuration is the same for all the the instances, why do not compile it into the code?

common/os_calls.c
volth: I notices that the purpose of these wrappers is to avoid using ```#include <string.h>``` and other standard headers, only this file with wrappers includes them and the rest of the xrdp's (but not sesman) sources use only wrappers.
I do not know why it is so, just followed the style.
Whole project could be refactored removing those wrappers.

volth:
yes, I think it makes sense to push the first version (with ```--cfg-path```) because it does one straightforward thing and solves the problem.
All the following (```---config```, ```sesmanport```, ....),  is more about the style than about solving the problem.

proski:
I disagree. Adding `--cfg-path` to all utilities creates a user interface that is not necessary and that will be removed by later commits. We should respect our users and avoid tricking them.

Please use smaller PRs for the issues. They need to be discussed separately. It's great that you make a patch to stop xrdp from reading sesman.ini. That patch should be a made a separate PR so that others can review the code, the design, possible side effects. For example, I don't think `sesmanport` should be added to every entry in `xrdp.ini`.

volth:
```sesmanport``` is replacement for magic ```port=-1``` which had meaning "sesman should be used and its port to be read from ```sesman.ini```".

Issue #608, opened 2017-01-02
Parallels Client (MacOS) drops the connection with 'Internal Error <131088>'


speidy:
When using Parallels Client with Audio Redirection turned on, the client is dropping the connection right after user logon to backend (Xorg for example).

I found that the reason for them to drop the connection is our propriety microphone implementation inside rdpsnd channel, which is working with rdesktop, NeutrinoRDP clients.

microphone redirection is not defined inside rdpsnd channel in MS doc ([[MS-RDPEA]](https://msdn.microsoft.com/en-us/library/cc240933.aspx)). they use another dynamic channel for microphone (as defined in [MS-RDPEAI] doc)

xrdp sends some extended microphone messages inside rdpsnd channel, which are in fact ignored by most of the clients (mstsc.exe, Android, etc.). Paralells client just drops the connection if they get an unknown message,  instead of ignoring it. 

I think we should talk to Parallels guys about that.

A quick 'patch' for this issue can be commenting out those lines: https://github.com/neutrinolabs/xrdp/blob/devel/sesman/chansrv/sound.c#L948-L949

jsorg71:
I would not say it's propriety messages.  They are just a non MS extensions to RDPSND.  xrdp has this in a few other places and we should document them.
The client should simply ignore capabilities it does not understand.

matt335672:
Addressed by #615

PR #615, opened 2017-01-06
Channels improvements

### LINKED ISSUE #594: Jump Desktop no connection: protocol error 3
Hi guys,

I'm currently messing around with several RDP clients. I want at least one decent RDP client on iOS getting to work but as already reported, Parallels client doesn't work (at least not for me) and my favorite one, Jump Desktop, cannot connect at all.
I get an error message in the client stating "protocol error, error code = 3" for iOS and the same is for Jump Desktop on macOS.
The syslog just states:
Dec 28 02:11:56 eos xrdp[11893]: (11893)(140493878143104)[INFO ] A connection received from: 10.234.4.135 port 51007
Dec 28 02:11:56 eos xrdp[11893]: (11893)(140493878143104)[DEBUG] Closed socket 12 (AF_INET 10.234.4.131:3389)
Dec 28 02:11:56 eos xrdp[12319]: (12319)(140493878143104)[DEBUG] Closed socket 11 (AF_INET 0.0.0.0:3389)
Dec 28 02:11:56 eos xrdp[12319]: (12319)(140493878143104)[INFO ] Using default X.509 certificate: /etc/xrdp/cert.pem
Dec 28 02:11:56 eos xrdp[12319]: (12319)(140493878143104)[INFO ] Using default X.509 key file: /etc/xrdp/key.pem
Dec 28 02:11:56 eos xrdp[12319]: (12319)(140493878143104)[DEBUG] Security layer: requested 3, selected 1
and after pressing OK in the client:
Dec 28 02:12:36 eos xrdp[12319]: (12319)(140493878143104)[DEBUG] Closed socket 12 (AF_INET 10.234.4.131:3389)
Dec 28 02:12:36 eos xrdp[12319]: (12319)(140493878143104)[ERROR] Listening socket is in wrong state, terminating listener

Using iFreeRDP works but this client is really bad in performance and usablity...

Anything I can help with? I can test alot of thing if you want but I can't code (at least not more than a little perl scripting :-) )

Cheers 
Markus

speidy:
* Check for empty channel names, they are unusable. Jump Desktop client send empty channel names for disabled channels. That made xrdp think we got more channels then really negotiated via  Client MCS Channel Join Request PDUs. That fix makes Jump Desktop client work with xrdp (should resolve #594)

* Respect xrdp.ini channel blocking settings. The restriction was only applied on client to server channel data.  block PDUs coming from chansrv if channel is not allowed. This change can help with Parallels Client, which doesn't support our rdpsnd channel (#608). One can now block rdpsnd channel in xrdp.ini so Parallels Client will not drop the connection.

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -1862,6 +1862,12 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
             return 1;
         }
         in_uint8a(s, channel_item->name, 8);
+        if (g_strlen(channel_item->name) == 0)
+        {
+            g_writeln("xrdp_sec_process_mcs_data_channels: got an empty channel name, ignoring it");
+            g_free(channel_item);
+            continue;
+        }
         in_uint32_le(s, channel_item->flags);
         channel_item->chanid = MCS_GLOBAL_CHANNEL + (index + 1);
         list_add_item(self->mcs_layer->channel_list, (tintptr) channel_item);
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -1829,12 +1829,12 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
     client_info = &(self->rdp_layer->client_info);
 
 
-    DEBUG(("processing channels, channel_code is %d", client_info->channel_code));
+    DEBUG(("processing channels, channels_allowed is %d", client_info->channels_allowed));
 
     /* this is an option set in xrdp.ini */
-    if (client_info->channel_code != 1) /* are channels on? */
+    if (client_info->channels_allowed != 1) /* are channels on? */
     {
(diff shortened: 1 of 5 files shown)


libxrdp/xrdp_sec.c
```diff
@@ -1862,6 +1862,12 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
             return 1;
         }
         in_uint8a(s, channel_item->name, 8);
+        if (g_strlen(channel_item->name) == 0)
+        {
+            g_writeln("xrdp_sec_process_mcs_data_channels: got an empty channel name, ignoring it");
```
proski: Please use proper log for all new messages.

jsorg71:
+1 awesome, I like the second fix more than the first.
@proski only allow channel one way won't work for clipboard issue as acks are used in protocol.

speidy:
I  also think the first change is not the long term solution. I think a. better solution will require to improve the xrdp <-> chansrv protocol. We should initialize channels by handling channel join pdus.

PR #618, opened 2017-01-10
sesman: print max_disc_time as DisconnectedTimeLimit

metalefty:
not max_idle_time. 

It looks like just a copy & paste mistake.


proski:
+1

Issue #623, opened 2017-01-14
Windows RemoteApp can not be opened more than one


dudustone:
I can open multiple Windows RemoteApp application without going through the xrdp agent. When accessing an application published by Window RemoteApp via an xrdp proxy, the first application can be opened, but the client does not respond when a new application is reopened.

speidy:
Can you test it with #624 ?
I think it should solve that problem.

speidy:
Sorry, #624  should not fix that. 

But #625  should be a fix.
Can you try?

speidy:
Can this issue be closed?

dudustone:
This issue has been solved perfectly.Thank you very much.

jumpwubo:
@dudustone how do you accessing remoteapp via xrdp?can detail, thanks

PR #624, opened 2017-01-15
Rail text length fix

speidy:
This is an old fix we didn't merge.


libxrdp/xrdp_orders_rail.c
```diff
@@ -191,15 +191,18 @@ xrdp_orders_send_as_unicode(struct stream *s, const char *text)
     int str_chars;
     int index;
     int i32;
-    twchar wdst[256];
-
-    str_chars = g_mbstowcs(wdst, text, 255);
+    twchar *wdst;
 
+    wdst = (twchar *) g_malloc(sizeof(twchar) * 64 * 1024, 1);
```
proski: Please use `g_new0`, it's more readable.

Why is 64*1024 used? I believe `strlen(text) + 1` should be safe. It's suggested by the glibc documentation for `mbstowcs`: https://www.gnu.org/software/libc/manual/html_node/Non_002dreentrant-String-Conversion.html

libxrdp/xrdp_orders_rail.c
```diff
@@ -191,15 +191,18 @@ xrdp_orders_send_as_unicode(struct stream *s, const char *text)
...
+    if (wdst == 0)
+    {
+        return 1;
+    }
+    str_chars = g_mbstowcs(wdst, text, 2 * 1024);
```
proski: The size argument should be the same as for the memory allocation.

libxrdp/xrdp_orders_rail.c
```diff
@@ -256,6 +259,8 @@ xrdp_orders_send_window_new_update(struct xrdp_orders *self, int window_id,
     {
         /* titleInfo */
         num_chars = g_mbstowcs(0, window_state->title_info, 0);
+        num_chars = MIN(num_chars, 2 * 1024);
```
proski: If `2 * 1024` is used in so many places, it should be a named constant.

speidy:
Sorry, I think i was wrong and it doesn't fix #623.

metalefty:
This doesn't fix #623 but still useful fix?

speidy:
yes.

speidy:
ok, fixed the sizes issue.

libxrdp/xrdp_orders_rail.c
```diff
@@ -256,6 +262,8 @@ xrdp_orders_send_window_new_update(struct xrdp_orders *self, int window_id,
     {
         /* titleInfo */
         num_chars = g_mbstowcs(0, window_state->title_info, 0);
+        num_chars = MIN(num_chars, sizeof(twchar) * (g_strlen(window_state->title_info) + 1));
+        num_chars = MAX(num_chars, 0);
```
proski: That `MAX` is not giving me confidence. How can `num_chars` be negative? Would it happen if `g_mbstowcs` returns -1 or when `sizeof(twchar) * (g_strlen(window_state->title_info) + 1)` overflows? Let's remove that line. If error handling is needed, it should be explicit. The same applies to every other `MAX`. At least please add a comment.

jsorg71: Yes, it happens from g_mbstowcs returning -1.  We can change it to make it more clear what we are checking.  I'm still not sure why "window_state->title_info" get corrupt but this check should be here anyway.

libxrdp/xrdp_orders_rail.c
```diff
@@ -195,15 +195,21 @@ xrdp_orders_send_as_unicode(struct stream *s, const char *text)
...
 
-    str_chars = g_mbstowcs(wdst, text, 255);
+    len = g_strlen(text) + 1;
 
+    wdst = (twchar *) g_malloc(sizeof(twchar) * len, 1);
+    if (wdst == 0)
+    {
+        return 1;
+    }
+    str_chars = g_mbstowcs(wdst, text, sizeof(twchar) * len);
```
proski: The last argument to mbstowcs and g_mbstowcs is the number of wide characters, not bytes. So it should be `len`, not `sizeof(twchar) * len`. We are promising more space than what we have.

libxrdp/xrdp_orders_rail.c
```diff
@@ -260,6 +266,12 @@ xrdp_orders_send_window_new_update(struct xrdp_orders *self, int window_id,
     {
         /* titleInfo */
         num_chars = g_mbstowcs(0, window_state->title_info, 0);
+        num_chars = MIN(num_chars, sizeof(twchar) * (g_strlen(window_state->title_info) + 1));
```
proski: What is happening here? I think g_mbstowcs() should be fully trusted, as it returns the number of the characters. But we should add 1 for the terminating zero and multiply the result by `sizeof(twchar)` to get the number of bytes.

g_strlen() can overestimate the number of the wide characters, as it would count multi-byte characters as multiple characters.

libxrdp/xrdp_orders_rail.c
```diff
@@ -536,6 +548,12 @@ xrdp_orders_send_notify_new_update(struct xrdp_orders *self,
     {
         /* ToolTip (variable) UNICODE_STRING */
         num_chars = g_mbstowcs(0, notify_state->tool_tip, 0);
+        num_chars = MIN(num_chars, sizeof(twchar) * (g_strlen(notify_state->tool_tip) + 1));
```
proski: Same thing here and below. If that code is correct, let's have a function for it and describe the purpose of that calculation.

libxrdp/xrdp_orders_rail.c
speidy: len is the multi-byte string length, the last argument should be max wide-chars to write to dest.

speidy: Good catch, thanks

proski:
Much better! Looks good to me.

speidy:
I think its ready to merge now.

metalefty:
Jay approved merging this in gitter chat.

PR #625, opened 2017-01-15
Pass session info in proxy mode

speidy:
Support Save Session Info PDU in proxy scenario
https://msdn.microsoft.com/en-us/library/cc240636.aspx

That one should solve #623

### DIFF
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -1282,8 +1282,7 @@ xrdp_rdp_send_session_info(struct xrdp_rdp *self, const char *data,
 
     s_mark_end(s);
 
-    /* SAVE_SESSION_INFO 0x26 */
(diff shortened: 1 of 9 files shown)


libxrdp/xrdp_rdp.c
```diff
@@ -1252,3 +1252,44 @@ xrdp_rdp_send_deactivate(struct xrdp_rdp *self)
...
+    else
+    {
+        free_stream(s);
+        return 1;
+    }
+
+    s_mark_end(s);
+
+    /* SAVE_SESSION_INFO 0x26 */
+    if (xrdp_rdp_send_data(self, s, 0x26) != 0)
```
proski: Can we stop doing that at least in the new code? If 0x26 is SAVE_SESSION_INFO, let's define SAVE_SESSION_INFO to 0x26 and use it.

dudustone:
Thank you very much, this issue has been solved perfectly.

Issue #592, opened 2016-12-25
librdp.so module broken


proski:
The librdp.so module (RDP proxy) is not working with the current devel branch. That's the only module that reports its version as 2; other modules report version 3.

There are some differences in `struct mod` for librdp.so and the genetic `struct xrdp_mod`. The last common field is `si` in `struct xrdp_mod` and `sck` in `struct mod`, and it's not the same thing.

It is worth the trouble to fix librdp.so? What are the use cases? Isn't it obsoleted by libxrdpneutrinordp.so?

speidy:
This module is obsolete. we should remove it.
for RDP proxy use neutrinordp module.

PR #626, opened 2017-01-15
 Remove librdp.so sources, it's obsoleted by neutrinordp

proski:
Rename "neutrinordp-and" section in xrdp.ini to "rdp-any", as it's the only available RDP client now.

The `rdp` directory had no code changes since year 2010 (apart from warning fixes, cleanup and changes affecting the whole source tree).

This fixes issue #592


speidy:
+1

xrdp/xrdp.ini
```diff
@@ -208,12 +208,6 @@ password=ask
 
 [rdp-any]
 name=rdp-any
-lib=librdp.so
-ip=ask
-port=ask3389
-
-[neutrinordp-any]
-name=neutrinordp-any
```
speidy: that change can make some people confused.
it should be documented.

proski: Actually, I think all "*-any" sections should not be in the default install. They allow using xrdp as a proxy to another network, which it not an expected behavior. Instead, those sections should be in a sample file with extensive documentation.

I guess I'll just keep using `neutrinordp-any` for now, as this PR is about a different issue.

proski:
The change for the section name has been removed from this PR.

jsorg71:
merged as 2b614ed4ab7b2c022325edb1dd32c3d9ca0d088b

PR #627, opened 2017-01-15
Check for openssl binary in configure, don't fail if not found

proski:
Prefer openssl binary from the same location as the openssl library.

Make it possible to suppress certificate generation during install by
passing OPENSSL=: to make.


speidy:
+1

metalefty:
+1

PR #628, opened 2017-01-16
x11rdp: respect disconnect timeout set by XRDP_SESMAN_MAX_DISC_TIME

metalefty:
The old code overwrites disconnect timeout to 60 sec when
XRDP_SESMAN_KILL_DISCONNECT is set to truthy value.


speidy:
+1

proski:
Good catch! Ship it!

PR #630, opened 2017-01-17
Fix calling XChangeProperty in clipboard provide

selu:
[XChangeProperty](https://tronche.com/gui/x/xlib/window-information/XChangeProperty.html)'s last parameter is number of elements, not number of bytes.
Because of this bug, Paste did not work in any java applications.
Java checks all formats first and because bytes are always bigger than length, it reads further and once it finds a 0 value, check fails with java.lang.NullPointerException
As I checked other calls of XChangeProperty in clipboard.c are fine.


proski:
+1, good catch, and thanks for checking other calls!

speidy:
+1

PR #631, opened 2017-01-17
Update m4/ax_type_socklen_t.m4, it works with -Wall now

proski:
Please see https://github.com/peti/autoconf-archive/pull/103 for the original PR


metalefty:
+1

PR #632, opened 2017-01-18
Don't use ACCESS macro

proski:
ACCESS is defined unconditionally in the files that use it. No comments
exist about that macro. The code guarded by the macro is PAM related.

### DIFF
--- a/xrdp/xrdp_login_wnd.c
+++ b/xrdp/xrdp_login_wnd.c
@@ -19,7 +19,6 @@
  */
 
 #include "xrdp.h"
-#define ACCESS
 #include "log.h"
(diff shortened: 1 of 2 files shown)


proski:
I checked the git history, the ACCESS definition comes from commit 10fdc5c17f7aec2df8885d2cf7354f6681af76f6

Even in that commit, ACCESS is defined unconditionally in both files where it's used. Apparently, it was used by the commit author to disable or enable the new functionality.

Since the gateway functionality is controlled by `xrdp.ini`, the ACCESS definition is not needed.

speidy:
👍

PR #633, opened 2017-01-18
Only check for systemd on Linux, no need to warn users of other OSes

proski:
That's the text that pkg-config and configure output to stderr:
```
Package systemd was not found in the pkg-config search path.
Perhaps you should add the directory containing `systemd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'systemd' found
```
That's not useful on FreeBSD or Mac OS. It is still possible to specify
`--with-systemdsystemunitdir=DIR` manually on any OS.


metalefty:
+1 Thank you for taking care of operating systems other than Linux.

PR #645, opened 2017-01-25
Add GNU/kFreeBSD support

mirabilos:
This depends on https://github.com/neutrinolabs/xorgxrdp/pull/65 and https://github.com/neutrinolabs/librfxcodec/pull/14 however.

I’ve tested this with Debian (compile test only).


proski:
+1

metalefty:
+1

metalefty:
My change I'm trying to make will conflict with this. Let's merge this first.

PR #649, opened 2017-01-26
arch.h changes to include stdint.h

jsorg71:


jsorg71:
I didn't notice we added -Werror to Travis

sesman/chansrv/devredir.c
```diff
@@ -1265,8 +1265,8 @@ dev_redir_file_write(void *fusep, tui32 DeviceId, tui32 FileId,
     IRP           *new_irp;
     int            bytes;
 
-    log_debug("DeviceId=%d FileId=%d Length=%d Offset=%lld",
-              DeviceId, FileId, Length, Offset);
+    log_debug("DeviceId=%d FileId=%d Length=%d Offset=%d",
+              DeviceId, FileId, Length, (int)Offset);
```
proski: Let's not truncate any data. The purpose of log_debug() is to present the data as is, even (an especially) if it's unreasonable. Please cast Offset to `unsigned long long`, as the original type (tui64) is unsigned. I know, is a lot of typing, but I'm not aware of a better approach.

proski:
+1

PR #650, opened 2017-01-27
sesman: remove unnecessary fork for FreeBSD

metalefty:
it is not only unnecessary but causes another bug.

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -513,32 +513,23 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
          *  $OpenBSD: session.c,v 1.252 2010/03/07 11:57:13 dtucker Exp $
          *  with some ideas about BSD process grouping to xrdp
          */
-        pid_t bsdsespid = g_fork();
 
-        if (bsdsespid == -1)
+        /**
+         * Create a new session and process group since the 4.4BSD
+         * setlogin() affects the entire process group
+         */
+        if (g_setsid() < 0)
         {
-        }
-        else if (bsdsespid == 0) /* BSD session leader */
-        {
-            /**
-             * Create a new session and process group since the 4.4BSD
-             * setlogin() affects the entire process group
-             */
-            if (g_setsid() < 0)
-            {
-                log_message(LOG_LEVEL_ERROR,
-                            "setsid failed - pid %d", g_getpid());
-            }
-
-            if (g_setlogin(s->username) < 0)
-            {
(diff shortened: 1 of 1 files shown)


sesman/session.c
```diff
@@ -513,33 +508,30 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
...
+        {
+            log_message(LOG_LEVEL_ERROR,
+                        "setlogin failed for user %s - pid %d", s->username,
+                        g_getpid());
         }
-
-        g_waitpid(bsdsespid);
 #endif
+        g_delete_wait_obj(g_term_event);
+        g_tcp_close(g_sck);
```
proski: Why is this code moved after the FreeBSD specific code? It's better not to hold resources longer than they are needed.

metalefty:
The old code runs session twice due to unnecessary deeper forking. After first session exits, new session starts again.  Besides, the second session doesn't passes FreeBSD specific code.

It's difficult for me to explain what happens by the old code.  Please see the below figure.

```
            this sesman becomes sessvc and forks Xorg, chansrv, session when child sessvc exits
                   |
xrdp-sesman -+- xrdp-sesman --- xrdp-sessvc -+- Xorg        <- first session: setlogined
             +                               +- xrdp-chansrv    this session passes thorough FreeBSD specific patch
             +                               +- xrdp-sesman --- (session manager, ex. Xfce4)
             +
             +
             +- xrdp-sessvc --- Xorg                         <- second session: NOT setlogined
                    |        +- xrdp-chansrv                     this session doesn't passes FreeBSD specific patch, runs after first session exits
                    |        +- xrdp-sesman --- (session manager, ex, Xfce4)   
                    |
              this sessvc was above sesman
```

sesman/session.c
metalefty: Oh, it shouldn't be moved.

metalefty:
The bug is born by myself, found by myself, fixed by myself :pensive: so it is not reported to elsewhere.

proski:
OK then. Just please fix the unnecessarily moved code and let's merge.

By the way, we should split huge functions that do multiple things, fork, wait, loop etc. Bad code structure contributes to such bugs.

Also, it seems strange that xrdp-sesman serves both as a root-owned authenticating daemon and as a user-owned session manager. The later functionality could be split into a separate binary, say, xrdp-session.

metalefty:
I agree with the last sentence. Let's merge later.

PR #651, opened 2017-01-27
Update README.md about installing / adjusting firewall on Fedora,Centos and RHEL

itamarjp:


README.md
```diff
@@ -47,7 +47,14 @@ apt-get install xrdp
 
 ### RedHat / CentOS / Fedora
 ```bash
-yum install xrdp
+yum -y install xrdp
+dnf -y install xrdp
```
proski: Please add a note that yum is for older distros.

README.md
```diff
@@ -47,7 +47,14 @@ apt-get install xrdp
...
+
+After installing Xrdp in Centos, RHEL or Fedora  you need to adjust the firewall
```
proski: This should not be in the code block.

README.md
```diff
@@ -47,7 +47,14 @@ apt-get install xrdp
...
 ### RedHat / CentOS / Fedora
 ```bash
-yum install xrdp
+yum -y install xrdp
+dnf -y install xrdp
+
+After installing Xrdp in Centos, RHEL or Fedora  you need to adjust the firewall 
+
+# firewall-cmd --permanent --zone=public --add-port=3389/tcp
+# firewall-cmd --reload
```
proski: Why are those commands commented out?

README.md
metalefty: Why you're adding `-y` ?

itamarjp: to install without asking to confirm

itamarjp:
I will merge please update on top of it if needed.

proski:
I believe it is the responsibility of the PR author to address all objections. In some cases it may be OK to push changes and promise fixes later. It's much worse to merge the PR as is and ask *others* to submit changes. That shifts the responsibility to other developers, who may have completely different priorities and perspectives. I, for one, would prefer to have that information in the Wiki.

proski:
The requested fixes are part of #654

README.md
metalefty: I know what `-y` does but why it's added?
I think users should confirm it on their own responsibility.

`apt` also ask users to confirm installation. Adding it only to `yum/dnf` is inconsistent.

PR #652, opened 2017-01-30
docs: remove IdleTimeLimit until it is actually implemented

metalefty:
it gives users false hope and finally disappoints them.

### DIFF
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -126,12 +126,6 @@ If set to \fB1\fR, \fBtrue\fR or \fByes\fR, every session will be killed
 within 60 seconds after the user disconnects.
 
 .TP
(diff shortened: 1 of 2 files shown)


proski:
+1

The configuration files and the manuals are not the right place for TODO lists.

Please note that there are still references to `IdleTimeLimit`, `max_idle_time`, and `XRDP_SESMAN_MAX_IDLE_TIME` in the code. If nobody plans to implement idle timeout, they should be removed too.

metalefty:
I think everything to support idle timeout is done in xrdp (as frontend). The rest part should be done in xorgxrdp/x11rdp.  We can remove such code and whenever resurrect it from git history if implementation is done in backend.

proski:
OK, let's merge

metalefty:
No logic change, merging immediately.

PR #654, opened 2017-01-30
Documentation improvements

proski:


README.md
```diff
@@ -47,15 +47,16 @@ apt-get install xrdp
 
 ### RedHat / CentOS / Fedora
 ```bash
-yum -y install xrdp
-dnf -y install xrdp
-
-After installing Xrdp in Centos, RHEL or Fedora  you need to adjust the firewall 
+dnf install xrdp
```
metalefty: +1 Users should confirm installation on their own responsibility.

README.md
```diff
@@ -47,15 +47,16 @@ apt-get install xrdp
...
+```
 
-# firewall-cmd --permanent --zone=public --add-port=3389/tcp
-# firewall-cmd --reload
+Older distribution should use `yum` instead of `dnf`. To allow outside
```
metalefty: I don't think the words "Older distribution" is not always correct.   `dnf` replaced `yum` since Fedora 22.  RHEL 7.2 is released later than Fedora 22.   It's a trivial issue though.

proski:
Rephrased text about `dnf` and `yum`.

itamarjp:
I noted something, xrdp is not packaged for RedHat / CentOS 

in centos is required to enable epel repo doing a 

yum -y install epel-release

and only after that doing a yum -y install xrdp

I don't know about what should be the procedure in a  RedHat system.

metalefty:
Indeed.  EPEL needs to be installed first.

Once again, users are responsible to confirm what changes will be made to their system.
According to yum help, `-y` option answers yes to any question not only installation confirmation.
We shouldn't add it. It should be added by users' own decision.

> -y, --asumeyes
>         Assume yes; assume that the answer to any question which would be asked is yes.

metalefty:
@proski What do you think about adding instruction how to enable EPEL?

proski:
Done.

metalefty:
Now looks good to me.

metalefty:
Doc changes only, merging.

PR #656, opened 2017-02-01
chansrv: fix warnings in pulse module

metalefty:
It was undiscovered due to pulse module is currently not tested by travis.

### DIFF
--- a/sesman/chansrv/pulse/module-xrdp-sink.c
+++ b/sesman/chansrv/pulse/module-xrdp-sink.c
@@ -171,12 +171,12 @@ static void sink_update_requested_latency_cb(pa_sink *s) {
 
     u->block_usec = BLOCK_USEC;
     //u->block_usec = pa_sink_get_requested_latency_within_thread(s);
-    pa_log("1 block_usec %d", u->block_usec);
+    pa_log("1 block_usec %llu", (unsigned long long) u->block_usec);
 
     u->got_max_latency = 0;
     if (u->block_usec == (pa_usec_t) -1) {
         u->block_usec = s->thread_info.max_latency;
-        pa_log_debug("2 block_usec %d", u->block_usec);
+        pa_log_debug("2 block_usec %llu", (unsigned long long) u->block_usec);
         u->got_max_latency = 1;
     }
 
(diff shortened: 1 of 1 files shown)


proski:
We should incorporate audio into the build.

The `pa_usec_t` type is defined as `uint64_t`. We cannot assume that it's `long`. The standard approach is to cast the argument to `unsigned long long` and use `%llu` format.

metalefty:
I already have a Makefile to build pulse module. I'll brush it up and integrate into travis build.

proski:
It's not just Travis. It should be built and installed as part of the build. Then Travis would pick it up automatically.

metalefty:
Sure. But pulseaudio's configured source is required to build. The build process will be different depending on distro.

jsorg71:
I've tried to push the pulseaudio guys to release a development package for building sinks and sources out of tree.  Sometime like pulseaudio-dev for Debian.
They did not seem very interested.
It's too bad, they have a really nice module system.

metalefty:
Anyway, let's fix warnings first. If the change looks good, please merge.

jsorg71:
+1
Looks good to me, remember, in this file we are trying to match pulseaudio's coding style.

sesman/chansrv/pulse/module-xrdp-sink.c
```diff
@@ -171,12 +171,12 @@ static void sink_update_requested_latency_cb(pa_sink *s) {
 
     u->block_usec = BLOCK_USEC;
     //u->block_usec = pa_sink_get_requested_latency_within_thread(s);
-    pa_log("1 block_usec %d", u->block_usec);
+    pa_log("1 block_usec %llu", (unsigned long long)u->block_usec);
```
proski: Please use space after cast here and elsewhere. I checked pulseaudio 10.0 sources, they use space too.

metalefty: ok. I always forget put a space after cast :disappointed:

metalefty: done.

PR #658, opened 2017-02-02
fix warning new since stdint.h change

jsorg71:


common/log.c
```diff
@@ -592,7 +592,7 @@ log_message(const enum logLevels lvl, const char *msg, ...)
         /* log to syslog*/
         /* %s fix compiler warning 'not a string literal' */
         syslog(internal_log_xrdp2syslog(lvl), "(%d)(%ld)%s", g_getpid(),
-               tc_get_threadid(), buff + 20);
+               (long) tc_get_threadid(), buff + 20);
```
proski: Thread ID should fit `long` on any supported OS, even on Win64. I think the fix belongs to the tc_get_threadid() return type.

xup/xup.c
```diff
@@ -212,7 +212,7 @@ lib_mod_connect(struct mod *mod)
         if (trans_connect(mod->trans, mod->ip, con_port, 3000) == 0)
         {
             LLOGLN(0, ("lib_mod_connect: connected to Xserver "
-                   "(Xorg or X11rdp) sck %ld", mod->trans->sck));
+                   "(Xorg or X11rdp) sck %ld", (long) (mod->trans->sck)));
```
proski: I'm surprised why `sck` is not defined as `int` in `common/trans.h`. If it's an opaque pointer, it should be printed as a pointer. If it needs to be a 64-bit integer, let's print it as `unsigned long long`.

proski:
If we want to support Win64, we should not cast pointers or pointer-sized scalar types to `long`. They would not fit. Even if the value is expected to be small (e.g. a file descriptor), cutting the upper bytes can turn an invalid value into a valid one, hiding an issue.

Two other approaches are available, dependent on the nature of the data. One is to cast to `void *` and use `%p` format (good for actual pointers), another is to cast to `long long` (possibly `unsigned`) and use `%lld` or `%llx` format (for data other than pointers).

Whether we should support Windows and Win64 in particular is a question too big for this PR. My feeling is that we should not, but let's not make such important decisions based on the need to add a few casts.

common/list.c
```diff
@@ -221,6 +221,6 @@ list_dump_items(struct list *self)
 
     for (index = 0; index < self->count; index++)
     {
-        g_writeln("%d: 0x%lx", index, list_get_item(self, index));
+        g_writeln("%d: 0x%lx", index, (long) list_get_item(self, index));
```
jsorg71: That is a good point

xup/xup.c
proski: I checked the sources, it's just a socket. It should be `int`.

jsorg71: I think it's that way because SOCKET on Windows is defined as a HANDLE which is void*
I think it's really just an int and this is a strange Microsoft thing.  I think openssl on Windows just cast it to an int anyway.  Lets not change this now as it's always big enough and this PR is just for warnings.

common/log.c
jsorg71: same here,
Lets not change this now as it's always big enough and this PR is just for warnings.

jsorg71:
@proski I don't think we have to "officially support" Windows but it would be nice if it worked.  I can see a use for VNC or RDP proxies on Windows.  Of course sesman, chansrv, Xorg does not make sense.

proski:
Let's discuss Windows support on gitter. Sure, it would be nice. The problem is that it take a lot of effort to keep the code compatible so that two or three people run xrdp on Windows as a proxy. It's hard to keep a high quality Windows port.

The issue with casts is that it's hard to identify where they are unnecessary, especially programmatically. Once a cast is added to the code, it's here to stay. The problem with casts in C is that they allow too many kinds of data conversion and cannot be annotated to limit their effect. A cast added to widen an scalar type can later have effect of shorting an scalar type or converting a pointer to an integer.

common/log.c
proski: Alright, this is OK for now. Anybody changing to the thread ID type would change tc_get_threadid() and would need to recheck the callers, including this one.

xup/xup.c
proski: Then let's use `long long` here to be sure were are not cutting away the top half of the pointer on Windows. `%p` won't be good, as sockets are not generally printed in hexadecimals on Linux.

proski:
@jsorg71 Could you please get this PR in a good shape and merge? I understand your position regarding Windows support, we can keep it for now. I'm trying to add i386 support to Travis to catch such warnings in the future.

PR #659, opened 2017-02-04
Remove --break-blocks option, it doesn't match the existing code style

proski:
Strip trailing newlines.


metalefty:
+1 Simple fix with no logic change. Let's merge.

Issue #524, opened 2016-12-04
rdpDeferredUpdateCallback with large screens and RFX


moobyfr:
with large screen (1900x1100) and using RFX, I get always a black screen , using current devel branch for xrdp and neutrinoRDP or xfreerdp for client.
The xrdp-sesman being filled by 
```rdpDeferredUpdateCallback: reschedule rect_id 1 rect_id_ack 0```

in ``xrdp/xrdp_encoder.c:368``
``` 
   if ((enc->num_crects > 512) || (enc->num_drects > 512))
    {
        return 0;
    }
 ```

According to xrdp.log:
```xrdp:xrdp_encoder [3403921343]: process_enc_rfx: num_crects 540 num_drects 1```

That's explain why RFX doesn't work with large resolution

tfischer77:
I'm facing the exact same issue here. Compiling xrdp 0.9.1 without rfx "solves" it...
I'm using mstsc from Windows 7. When using a resolution of 1900x1200 and highest quality (LAN), the screen is "black" and the sesman log is filled with the above mentioned message..
Selecting a lower quality (WAN) is working, but if I try to show a picture with a lot of details in it, the RDP client (mstsc) disconnects immediaetely with a message "protocol error".

macdanger:
Hi guys,
any idea how to fix this or work around this? Because working with large resolutions without remoteFX is no fun :-) And higher resolutions than full HD with remoteFX result in a black or empty sceen. 
what I noticed is that the modlines in the xorg.conf only go up to 1920x1080. Maybe it helps when we add more modlines for higher resolutions?

proski:
Please try compiling xrdp and xorgxrdp with `--without-simd`

moobyfr:
I've tried, the error is the same. The problem comes when the resolution of the screen needs more than 512 rect (each rect is 64x64) to be covered.

sylvain121:
Hi, 
I confirm problem. I have 2 large screen (1920x1080) and when I use RFX, I get black screen when connection crash.

proski:
I looked at `process_enc_rfx`, it's full of magic numbers and it doesn't even return an error or logs anything when enc->num_crects or enc->num_drects exceeds 512. It's not clear where 512 comes from. Is that a protocol limitation? I don't see any references to 512 in librfxcodec sources.

jsorg71:
That is just a sanity check, it can be increased or removed.

moobyfr:
[...]

PR #664, opened 2017-02-12
rfx fixes for large tile sets, performance change, Xorg will start ne…

jsorg71:
…xt frame earlier

### DIFF
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -328,8 +328,7 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
     out_data = NULL;
     out_data_bytes = 0;
 
-    if ((enc->num_crects > 0) && (enc->num_crects <= UINT16_MAX) &&
-        (enc->num_drects > 0) && (enc->num_drects <= UINT16_MAX))
+    if ((enc->num_crects > 0) && (enc->num_drects > 0))
     {
         alloc_bytes = XRDP_SURCMD_PREFIX_BYTES;
         alloc_bytes += self->max_compressed_bytes;
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -320,21 +320,21 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
     mutex = self->mutex;
     event_processed = self->xrdp_encoder_event_processed;
 
-    if ((enc->num_crects > 512) || (enc->num_drects > 512))
+    if ((enc->num_crects > 4096) || (enc->num_drects > 4096))
     {
         return 0;
     }
 
     out_data_bytes = 16 * 1024 * 1024;
-    index = 256 + sizeof(struct rfx_tile) * 512 +
-                  sizeof(struct rfx_rect) * 512;
+    index = 256 + sizeof(struct rfx_tile) * 4096 +
+                  sizeof(struct rfx_rect) * 4096;
     out_data = (char *) g_malloc(out_data_bytes + index, 0);
     if (out_data == 0)
     {
         return 0;
     }
     tiles = (struct rfx_tile *) (out_data + out_data_bytes + 256);
-    rfxrects = (struct rfx_rect *) (tiles + 512);
+    rfxrects = (struct rfx_rect *) (tiles + 4096);
 
     count = enc->num_crects;
     for (index = 0; index < count; index++)
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -39,6 +39,8 @@
   } \
   while (0)
 
+#define XRDP_SURCMD_PREFIX_BYTES 256
+
 /*****************************************************************************/
 static int
 process_enc_jpg(struct xrdp_encoder *self, XRDP_ENC_DATA *enc);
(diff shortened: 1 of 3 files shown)


proski:
This patch is on the par with the existing code, but it's behind the modern standards of software engineering. Numeric constants should have names. It should be clear from the name whether 4096 is a protocol limitation, a "large enough number" or something else. If a constant number is derived from other numbers, it should be represented as an expression.

Likewise, when two variables are compared, the difference between those numbers would normally have a meaning that can be expressed with a concise term (e.g. unacknowledged packets). A local variable could be created to hold that difference and describe its meaning. That would not make the program slower; it's trivial for a compiler to optimize out local variables.

Especially when the original code didn't get it right, it's important that the code that replaces it goes beyond fixing the immediate issue.

If the code fixes issues for the users, let's merge this PR. But I cannot review this PR in a meaningful way. I can say it has no typos and gets formatting right. I could look deeper if the assumptions were expressed in the code.

tfischer77:
As I told in #524, the PR does not solve it completely, as im some cases, at least "drects" can be > 4096. I did some tests with the code in `xrdp/xrdp_encoder.c` and ended up with this code in `process_enc_rfx`, which now runs for several hours without crashes and frozen screens:
```
    mutex = self->mutex;
    event_processed = self->xrdp_encoder_event_processed;

/*
    if ((enc->num_crects > 8192) || (enc->num_drects > 8192))
    {
        return 0;
    }
*/
    out_data_bytes = 16 * 1024 * 1024;
    index = 256 + sizeof(struct rfx_tile) * enc->num_crects +
                  sizeof(struct rfx_rect) * enc->num_drects;
    out_data = (char *) g_malloc(out_data_bytes + index, 0);
    if (out_data == 0)
    {
        return 0;
    }
    tiles = (struct rfx_tile *) (out_data + out_data_bytes + 256);
    rfxrects = (struct rfx_rect *) (tiles + enc->num_crects);

    count = enc->num_crects;

```
I don't know if I will produce any "ugly" side effects like memory leaks, but I think I will leave the code running in my test environment for some days and look what happens.

speidy:
+1

Natureshadow:
Btw, why are num_crects and num_drects signed integers, and the code does not check for overruns?

Natureshadow:
What about this:

```
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -22,6 +22,7 @@
 #include "xrdp.h"
 #include "thread_calls.h"
 #include "fifo.h"
+#include "limits.h"
 
 #ifdef XRDP_RFXCODEC
 #include "rfxcodec_encode.h"
@@ -320,21 +321,25 @@ process_enc_rfx(struct xrdp_encoder *sel
     mutex = self->mutex;
     event_processed = self->xrdp_encoder_event_processed;
 
-    if ((enc->num_crects > 512) || (enc->num_drects > 512))
-    {
+    out_data_bytes = 16 * 1024 * 1024;
+
+    if ((enc->num_crects > (INT_MAX / sizeof(struct rfx_tile))) ||
+        (enc->num_drects > (INT_MAX / sizeof(struct rfx_rect) -
+                            sizeof(struct rfx_tile) * enc->num_crects -
+                            256 - out_data_bytes)) ||
+        (enc->num_crects < 0) || (enc->num_drects < 0)) {
         return 0;
     }
 
-    out_data_bytes = 16 * 1024 * 1024;
-    index = 256 + sizeof(struct rfx_tile) * 512 +
-                  sizeof(struct rfx_rect) * 512;
+    index = 256 + sizeof(struct rfx_tile) * enc->num_crects +
+                  sizeof(struct rfx_rect) * enc->num_drects;
     out_data = (char *) g_malloc(out_data_bytes + index, 0);
     if (out_data == 0)
     {
         return 0;
     }
     tiles = (struct rfx_tile *) (out_data + out_data_bytes + 256);
-    rfxrects = (struct rfx_rect *) (tiles + 512);
+    rfxrects = (struct rfx_rect *) (tiles + enc->num_crects);
 
     count = enc->num_crects;
     for (index = 0; index < count; index++)
```

It does a proper check on the structure sizes the number of rects will later be used in and imposes no further artifical limits.

proski:
Much better. There is still 256 is a couple of places, it needs to be clarified.

Please fix the reuse of the `index` variable. A variable should serve a single purpose. `index` is a very generic name. Loop variables can be `i`, `j` etc. The large number in the malloc() call could probably be `max_index`.

jsorg71:
The 256 is an empty header that the lower layers fill out when sending.  Other parts call it pad_bytes.  Pad usually means put it at end, maybe we can thing of a better name.
num_crects and num_drects comes in as uint16 and put into int.  So it can be 0 to 64K.
https://github.com/neutrinolabs/xrdp/blob/2424a9c020d3644323caca8f29ae5b3b2c355de1/xup/xup.c#L1162-L1186

Natureshadow:
So count yourself lucky that xrdp compiles on no system that runs on 16 bit architectures :D!

jsorg71:
xrdp wont startup if size of int is not 4.  So far no one has asked for 16 bit int systems.  The only one I can think of is Window 3.1.
https://github.com/neutrinolabs/xrdp/blob/2424a9c020d3644323caca8f29ae5b3b2c355de1/xrdp/xrdp.c#L324-L328

Natureshadow:
>https://github.com/neutrinolabs/xrdp/blob/2424a9c020d3644323caca8f29ae5b3b2c355de1/xrdp/xrdp.c#L324-L328


Aaah... oh. Uhu...

PR #665, opened 2017-02-12
aarch64 doesn't require pointers to be aligned

proski:
Without this patch, a warning is printed for every file: https://kojipkgs.fedoraproject.org/packages/xrdp/0.9.1/2.fc26/data/logs/aarch64/build.log

### DIFF
--- a/common/arch.h
+++ b/common/arch.h
@@ -84,7 +84,7 @@ typedef int bool_t;
 #define NEED_ALIGN
 #elif defined(__x86__) || defined(__x86_64__) || \
       defined(__AMD64__) || defined(_M_IX86) || defined (_M_AMD64) || \
-      defined(__i386__)
+      defined(__i386__) || defined(__aarch64__)
 #define NO_NEED_ALIGN
 #else
 #warning unknown arch


jsorg71:
It's the integer reading and writing that sometimes should be aligned to avoid bus error.  Some kernels like armel can read unaligned but the kernel catches the bus error and corrects the read.  If this happens, it's slower than if you avoid the bus error.  Maybe that is happening on aarch64 too and we are better off putting it in the NEED_ALIGN group.  I don't know, I don't have access to one yet.
+1 for this PR, it's better than what we have now.

proski:
I don't have access to those machines either. I made some research online. There would be no crash, the operation would work correctly, but it may be slow and non-atomic.
http://stackoverflow.com/questions/38535738/does-aarch64-support-unaligned-access

In most cases the the data would be aligned. But if NEED_ALIGN is defined, all data is processed one byte at a time. That's highly ineffective. Avoiding it would offset the slowdown from a few unaligned.

Eventually we should add the functions where the caller assures that the data is aligned. That data would be processed using the original data size, not bytes, as long as no byte swapping is needed.

jsorg71:
i just got a Pine64 board running Debian 8 arm64 so I can test.

jsorg71:
It looks like this architecture behaves a lot like x86 so NO_NEED_ALIGN and L_ENDIAN is correct.

proski:
Thank you for checking that!

Issue #663, opened 2017-02-09
fails to detect some Japanese keyboard


yyyjajp:
Since commit 9c31bd5cc428d912194fc11d5706520473922853 my laptop's keyboard layout in xrdp is "us" not "jp". Reverting the commit gets back the "jp" layout.
xrdp.log looks like the same, but `setxkbmap -print` output is different.
current:
[xrdp.log.txt](https://github.com/neutrinolabs/xrdp/files/764598/xrdp.log.txt)
[setxkbmap.txt](https://github.com/neutrinolabs/xrdp/files/764614/setxkbmap.txt)
after the revert:
[xrdp.log.txt](https://github.com/neutrinolabs/xrdp/files/764603/xrdp.log.txt)
[setxkbmap.txt](https://github.com/neutrinolabs/xrdp/files/764620/setxkbmap.txt)

metalefty:
I can reproduce the issue.

metalefty:
The issue is related #418 not only #516.  

proski:
> Could you please open an issue for it? What's in your xrdp.log? The code is supposed to fall back to the basic key layout (0x00000411 in this case). The actual key maps for jp and jp2 were the same.

The xrdp's key layout properly fallbacks to `0x00000411` layout but even in fallback case, xrdp still lookups xorg's keylayout by `0xe0010411`. So if no definition for `0xe0010411` is found in xrdp_keyboard.ini, xorg keylayout fallbacks to us.

If fallback occurred, `client_info->keylayout` also should be set to  `client_info->keylayout & 0x0000ffff`.

PR #667, opened 2017-02-16
Workaround for a regression of JP keyboard detection #663

metalefty:
caused by 9c31bd5.  This is not a complete fix but just a workwround
because keylayout value of Japanese keyboard can be other than these
4 values. This workdaround still doesn't cover all JP keyboards.

  - 0x00000411
  - 0xe0010411
  - 0xe0200411
  - 0xe0210411

Looking for `0411` in lower 16 bits is enough to detect JP keyboards.


metalefty:
Fixed typo in commit message, force pushed.

proski:
+1

metalefty:
Already tested by yyyjajp in https://github.com/neutrinolabs/xrdp/commit/9c31bd5cc428d912194fc11d5706520473922853#commitcomment-20834035,  I think this can be merged immediately.

PR #669, opened 2017-02-22
Add i686 check in Travis CI

proski:
This depends on #658

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,41 +4,104 @@ language: c
 branches:
   except: /^(gh-pages|v[0-9]\..*)/
 
-compiler:
-  - clang
-  - gcc
-  - g++
-
 addons:
   apt:
-    packages:
-      # xrdp
-      - libpam0g-dev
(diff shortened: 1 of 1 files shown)


proski:
Some of the i386 packages conflict with the x86_64 packages, which necessitated using `matrix` with explicit `include`. YAML anchors are used to reduce duplication.

Using `matrix` allowed me to limit "make distcheck" to just three configurations (one per compiler) to speed up the checks.

g++ on i386 and `-Werror` have been removed temporarily until the breakage is fixed (#658).

I was unable to get libfuse-dev:i386 to install, so FUSE is not checked on i386. I'll keep trying on a separate branch.

Setting `LDFLAGS=-m32` will be unneeded once nasm.m4 in librfxcode is updated (see https://github.com/neutrinolabs/librfxcodec/pull/26)

proski:
Rebased on top of devel, restored g++ and -Werror, all the tests pass. Should be OK to merge now.

Issue #666, opened 2017-02-14
xrdp spins 100% CPU


atanamir:
I'm not sure if this is FreeRDP doing this or xrdp, but I'll file it here first.  Right now my xrdp process spins 100% of CPU and is extremely laggy to the point of being unusuable (for example, typing a character takes >1sec to show up on the screen).  I did an strace on the process and it only consists of this line ad infinitum:

`read(12, 0x2321853, 16709)              = -1 EAGAIN (Resource temporarily unavailable) `

Is it a problem where the async read is not being handled properly (i.e. infinite loop of polling)?  

My xrdp configure flags were:

`./configure --enable-pixman --enable-fuse --disable-rfxcodec --disable-painter --disable-ipv6 --prefix=/usr --sysconfdir=/etc --localstatedir=/var `

And FreeRDP was:

`cmake . -DWITH_SSE2=ON -DCMAKE_INSTALL_PREFIX=/usr -DWITH_FFMPEG=OFF -DWITH_JPEG=ON -DWITH_X264=ON -DWITH_XCURSOR=ON -DWITH_XI=ON -DWITH_XV=ON -DWITH_XRENDER=ON -DWITH_XFIXES=ON`

atanamir:
Did some more debugging into the issue.  Execution is stuck inside `ssl_tls_read` of `ssl_calls.c`.  It goes between the `SSL_read`, which is returning `SSL_ERROR_WANT_READ` (or `SSL_ERROR_WANT_WRITE`).  Then that causes the `continue` and it then calls `SSL_read` again, over and over.  I think there needs to be a call to `select` somewhere in that loop?

metalefty:
Probably your pointing out is correct but let me clarify the issue.
If you don't use TLS, still get 100% cpu usage?  Try `security_layer=rdp`.

atanamir:
Adding that line to my xrdp.ini seems to have solved the CPU usage issue.  It was set to `negotiate` previously.

metalefty:
OK, the issue will occur only if `security_layer` is set to `negotiate` or `tls`.
The issue is caused by SSL/SSL related code as you say.

jsorg71:
This should be fixed with #670

metalefty:
@atanamir can you confirm? close if fixed for you.

atanamir:
It works!  Thanks guys!

PR #670, opened 2017-02-23
common: use select for SSL_ERROR_WANT_READ, SSL_ERROR_WANT_WRITE tls …

jsorg71:
…errors


jsorg71:
@speidy How does this look for fixing #666 ?

metalefty:
I tested quickly. Looks good to me.
@atanamir can you check if this fixes #666?

PR #673, opened 2017-02-27
Update submodules to devel branches

proski:
The librfxcodec branch was unset (defaulting to master). The libpainter branch was set to devel. Neither is a good choice for both devel and master branches of xrdp. Use `branch = .` - it's a brand new feature of git: https://github.com/git/git/commit/c8386962d6590ec6f71f353bc5b1b573b1e71b16

At this point it's more documentation than configuration. Submodule updates are made rarely and manually. Still, it's good to document the parent branch and treat both submodules the same way.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,10 +1,11 @@
 [submodule "librfxcodec"]
 	path = librfxcodec
 	url = https://github.com/neutrinolabs/librfxcodec.git
+	branch = .
 	ignore = untracked
 
 [submodule "libpainter"]
 	path = libpainter
 	url = https://github.com/neutrinolabs/libpainter.git
-	branch = devel
+	branch = .
 	ignore = untracked


metalefty:
+1 the new feature is nice.

speidy:
+1

jsorg71:
One nice thing about submodules it that the parent can point to a older sha1 in the submodule and allow an additional staging before updating the parent.  We can still do that I hope?
Will older git systems be ok?

proski:
In practice, nothing changes. The branch setting is only used when updating the submodules from the remote repositories using
```
git submodule update --remote
```
That would allow users to ignore our SHA1 and use the current branch for the submodule. That's not something we encourage users to do.

But they do it anyway, they would get submodule branches with the same name as the top tree branch.

Without this PR, git 2.12.0 or git 2.7.3
```
$ git submodule update --remote
Submodule path 'libpainter': checked out '898058cfcdc29bd745e6479327a2ed52e32cdc24'
Submodule path 'librfxcodec': checked out 'b9300b2e03a397f77bd8d60c8db8909083d630f0'
```

That's origin/devel for libpainter and origin/master for librfxcodec.

With this PR, git 2.7.3
```
$ git submodule update --remote
fatal: Needed a single revision
Unable to find current origin/. revision in submodule path 'libpainter'
```

git 2.12.0
```
$ git submodule update --remote
Submodule path 'libpainter': checked out '898058cfcdc29bd745e6479327a2ed52e32cdc24'
Submodule path 'librfxcodec': checked out '72efbf96ba1f8eb3e301edddf164d245135edb44'
```

That's origin/devel for both submodules.

Another approach would be to have `branch = devel` on both submodules. That way, users on the master branch would get devel branches of the submodules. It may be OK. After all, if users care to update, they probably want the latest code. But on the other hand, I don't want someone to package devel branches of submodules with a release made from master. So I prefer something that either breaks or does the right thing.

As for the updates submitted as PR (like this one), they are made on a separate branch (not devel), so I would not rely on `git submodule update --remote` at all. I'd rather do it manually.

jsorg71:
Alright, thanks for explanation.

PR #675, opened 2017-02-28
Fix compilation on MacOS

proski:
xrdp_rdp.c includes openssl/ssl.h now, make sure the OpenSSL include path
is used.

---

That's a compile issue, let's merge it before 0.9.2.

### DIFF
--- a/libxrdp/Makefile.am
+++ b/libxrdp/Makefile.am
@@ -8,6 +8,8 @@ AM_CPPFLAGS = \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
   -I$(top_srcdir)/common
 
+AM_CFLAGS = $(OPENSSL_CFLAGS)
+
 AM_LDFLAGS =
 
 LIBXRDP_EXTRA_LIBS =


metalefty:
+1

easyw:
Hi, then xrdp is compilable/usable on osx Sierra?
Thx
Maurice

metalefty:
We don't officially support OS X. I know some users use xrdp on OS X but we don't provide any support. Users do it in their responsibility. If you want to use xrdp on OS X, we welcome your contributions.

easyw:
Hi, I successfully built xrdp 0.92 on osx sierra and xrdp is working with vnc...
I would like to build also xorgxrdp on sierra and I followed the wiki, but I have an issue that is probably easy to fix but I don't know in which way...
I miss xserver-xorg-dev or xorg-x11-server-sdk or xorg-x11-server-devel... in which way can I install it on sierra? I cannot find any tip around ...
thx
Maurice

easyw:
ok, I managed to compile xorgxrdp mixing the instruction found here
https://github.com/neutrinolabs/xrdp/wiki/Building-on-OSX
and here
https://github.com/neutrinolabs/xrdp/wiki/Compiling-and-using-xorgxrdp
> ./bootstrap
./configure PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig
make
sudo make install

probably <b>./bootstrap</b> would need to be added at the Building-on-OSX page
Maurice

PR #677, opened 2017-03-02
common: increase tls_ciphers size to 1024 bytes

### LINKED ISSUE #544: Inclease maximum length of tls_ciphers
The length of `tls_ciphers` parameter in xrdp.ini is limited up to 63 characters.
This is due to xrdp_client_info struct has a member tls_ciphers and its length is hardcoded as 64.

```C
struct xrdp_client_info
{
  (snip)
  int disableSSLv3; /* 0 = no, 1 = yes */
  char tls_ciphers[64];
  (snip)
};
```

To fully configure TLS cipher suites not using aliases such as HIGH, MEDIUM or LOW, the current maximum length is not enough.

Here's an example of cipher suites (for web servers):
https://mozilla.github.io/server-side-tls/ssl-config-generator/

metalefty:
Fixes #544

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -148,7 +148,7 @@ struct xrdp_client_info
   int max_unacknowledged_frame_count;
 
   long ssl_protocols;
-  char tls_ciphers[64];
+  char tls_ciphers[1024];
 
   int client_os_major;
   int client_os_minor;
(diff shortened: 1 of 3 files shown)


proski:
I considered that approach, but I don't like such radical expansion a structure that is sent to backends. I would rather allocate memory dynamically.

metalefty:
Yes, I don't think this is the best approach. I welcome alternative PRs.

metalefty:
#680 looks much better.

PR #678, opened 2017-03-03
Include config_ac.h from *.c files, not from headers

### LINKED ISSUE #647: Include config_ac.h ① in all files, and ② first, before anything else
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=852617#42 for a lecture from GNU autoconf upstream:

> If the real-world package is not including config.h first, then that is                                     
> a bug in the real-world package that should be fixed there.  Autoconf                                       
> clearly documents that if you create config.h, it MUST be included                                          
> first, before any system headers.                                                                           

The [workaround](https://anonscm.debian.org/cgit/pkg-remote/xrdp.git/plain/debian/patches/lfs.diff?id=e17430063641d44f5596b5bfc1b32ac4ba39f9f1) I had to add to the xrdp Debian package to enable and fix Large File Support was… not nice. So, apparently, the correct fix involves having __every__ __single__ source file (`*.c`, but also C++ et al.) `#include "config_ac.h"` first, *before* __any__ other `#include` or nōn-preprocessor statement (the inclusion from various header files, such as `arch.h`, can then be removed).

Child issues:
- https://github.com/neutrinolabs/xorgxrdp/issues/66
- https://github.com/neutrinolabs/librfxcodec/issues/15
- https://github.com/neutrinolabs/libpainter/issues/8

proski:
Directories without Makefile.am are exempt from the change, which includes X11rdp. They have their own build systems.

Resolves #647.

### DIFF
--- a/common/arch.h
+++ b/common/arch.h
@@ -19,10 +19,6 @@
 #if !defined(ARCH_H)
 #define ARCH_H
 
-#if defined(HAVE_CONFIG_H)
-#include "config_ac.h"
-#endif
-
 #include <stdlib.h>
 
 #if defined(HAVE_STDINT_H)
--- a/common/fifo.c
+++ b/common/fifo.c
@@ -18,6 +18,10 @@
  * FIFO implementation to store pointer to data struct
  */
 
+#if defined(HAVE_CONFIG_H)
+#include <config_ac.h>
+#endif
+
 #include "fifo.h"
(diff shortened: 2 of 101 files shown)


jsorg71:
Thanks, I'm sure that was an extremely boring task.
+1

proski:
No, that was a ruby script :-P

PR #680, opened 2017-03-04
Allocate space for tls_ciphers dynamically

proski:


proski:
The patch has not been tested at all. Consider it work in progress until further notice. The patch is published to show an alternative approach to #677

There is already `struct xrdp_config`, which holds parts of `xrdp.ini` including some security bits (e.g. `crypt_level`). However, that structure is only available through `struct xrdp_wm` (window manager), which makes it a poor choice for security related configuration.

There is also `struct xrdp_sec` for security related configuration. However, that structure is not populated directly from the config file. For example, `crypt_level` is read to `struct xrdp_client_info` (in addition to `struct xrdp_config`) and then converted to a value acceptable for `struct xrdp_sec`.

One possible approach would be to have a single config loader, but it would be too intrusive for 0.9.2. On the other hand, stretching fields is not a good idea even in the short term. It increases data exchanged with the backends and doesn't protect against buffer overflow - it just makes the buffer bigger.

proski:
I was able to test this patch. It appears to work correctly. So it's not work in progress anymore.

metalefty:
I also tested. Works fine.

speidy:
LGTM

metalefty:
@jsorg71 please merge if looks good to you.

metalefty:
Let's merge before v0.9.2.

jsorg71:
I'll merge it.
This struct gets send between processes so the pointer will be invalid in Xorg and X11rdp but we'll plan to change that later.  I'm sure Xorg and X11rdp do not use this member.

PR #681, opened 2017-03-05
Fix for slow encoder with fast client and network.  Handle misbehaving clients that send frame acks out of order.  Other miscellaneous code cleanup.

jsorg71:


xrdp/xrdp_encoder.c
```diff
@@ -132,6 +132,9 @@ xrdp_encoder_create(struct xrdp_mm *mm)
     g_snprintf(buf, 1024, "xrdp_%8.8x_encoder_term", pid);
     self->xrdp_encoder_term = g_create_wait_obj(buf);
     self->max_compressed_bytes = client_info->max_fastpath_frag_bytes & ~15;
+    self->frames_in_flight = client_info->max_unacknowledged_frame_count;
```
proski: Would it be possible to eliminate client_info->max_unacknowledged_frame_count? It looks like self->frames_in_flight is taking its place for all purposes.

xrdp/xrdp_mm.c
```diff
@@ -2165,17 +2165,98 @@ xrdp_mm_check_chan(struct xrdp_mm *self)
...
-int APP_CC
-xrdp_mm_check_wait_objs(struct xrdp_mm *self)
+static int APP_CC
+xrdp_mm_update_module_frame_ack(struct xrdp_mm *self)
+{
+    int fif;
+    struct xrdp_encoder *encoder;
+
+    encoder = self->encoder;
+    fif = encoder->frames_in_flight;
```
proski: Please combine declaration and initialization. It makes the code shorter and more readable.
```
int fif;
// what is that "fif" thing?
// more declarations
// more code
fif = encoder->frames_in_flight;
// Oh, I know what "fif" means.
// By the way, is it an integer? Where's the declaration?
```

xrdp/xrdp_mm.c
```diff
@@ -2165,17 +2165,98 @@ xrdp_mm_check_chan(struct xrdp_mm *self)
...
     int y;
     int cx;
     int cy;
-    int use_frame_acks;
-    int ex;
+
+    while (1)
+    {
+        tc_mutex_lock(self->encoder->mutex);
+        enc_done = (XRDP_ENC_DATA_DONE*)
```
proski: Missing space before `*`

xrdp/xrdp_mm.c
```diff
@@ -2314,26 +2326,24 @@ xrdp_mm_check_wait_objs(struct xrdp_mm *self)
...
+    if (frame_id < 0)
     {
-        if (self->encoder->frame_id_server > self->encoder->frame_id_server_sent)
-        {
-            LLOGLN(10, ("xrdp_mm_frame_ack: frame_id_server %d", self->encoder->frame_id_server));
-            self->encoder->frame_id_server_sent = self->encoder->frame_id_server;
-            self->mod->mod_frame_ack(self->mod, 0, self->encoder->frame_id_server);
-        }
+        /* if frame_id == -1 ack all sent frames */
+        encoder->frame_id_client = encoder->frame_id_server;
```
proski: If -1 has that special property, let's make a define for it. I'm not sure if other negative numbers should have that properly. The code and the comment differ here.

speidy:
Jay, I like that change.
I would like to test Parallels Client as well, it requires zero frames in flights.
Let me do some tests and update.

xrdp/xrdp_encoder.c
jsorg71: no, max_unacknowledged_frame_count is the value received from the client.  frames_in_flight is our value, possibly modified for usage.

xrdp/xrdp_mm.c
jsorg71: Thanks for narrating your thought process but this is your opinion.  I find separate declaration easier to read.

xrdp/xrdp_mm.c
jsorg71: Yes, I think I'll change the comment to state the all negative numbers are treated the same.
I do not like the constant "make a define for it"  I think clear comments is what's really needed.  A define can be misleading and lead to more problems.

speidy:
Parallels client seems to work OK, tested from Mac.

PR #682, opened 2017-03-07
chansrv: dvc, check channel exists on get/remove api struct

speidy:
This issue fixes a case when xrdpapi (WTS calls) is used in an unexpected way, there is a case which can lead chansrv to get an api struct which is already removed, then chansrv can crash with segfault.


proski:
+1

Please describe how this issue manifests itself. It's good to know in case if someone else reports it.

speidy:
PR description updated.

PR #686, opened 2017-03-08
Merge in latest changes; edited error window to say 'Osirium'

MUzzell:


proski:
Any comment on what it was? I believe some changes could be useful for xrdp.

jsorg71:
I think this was an accidental PR.  @speidy and I have been working with them.  We need to split these up into smaller PRs.  Actually, we have already been doing this.  Some already passed review, some in the hopper.  There is more good PRs to come from Osirium but some that we just don't want because they are too specific to Osirium use case.

proski:
Thank you for reply. I tried rebasing the PR on top of the devel branch, and I see that most of the remaining patches are for neutrinordp. But still, there are some logic changes that might be useful for xrdp users.

MUzzell:
Sorry for the confusion guys. I selected the wrong target branch!

PR #687, opened 2017-03-08
Log user-friendly message when certificate/privkey is inaccessible

metalefty:
We shouldn't assume that xrdp daemon is running under root privilege.
In many cases, root privilege is not really needed for xrdp daemon.
xrdp may fail to load certificate/privkey due to lack of permissions
when running under user privilege. Checking existence of files is not
enough and xrdp should output user-friendly log in such case.

Reported by Debian user in bug 856436 [1].

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=856436

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2234,6 +2234,18 @@ g_file_exist(const char *filename)
 }
 
 /*****************************************************************************/
+/* returns boolean, non zero if the file is readable */
+int
+g_file_readable(const char *filename)
+{
+#if defined(_WIN32)
+    return 0; /* TODO: what should be done here? */
+#else
+    return access(filename, R_OK) == 0;
+#endif
+}
+
+/*****************************************************************************/
 /* returns boolean, non zero if the directory exists */
 int
 g_directory_exist(const char *dirname)
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2239,7 +2239,7 @@ int
 g_file_readable(const char *filename)
 {
 #if defined(_WIN32)
-    return 0; /* TODO: what should be done here? */
+    return _waccess(filename, 04) == 0;
 #else
     return access(filename, R_OK) == 0;
 #endif
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -269,6 +269,12 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
                 /* use user defined certificate */
                 g_strncpy(client_info->certificate, value, 1023);
             }
+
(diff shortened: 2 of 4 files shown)


common/os_calls.c
```diff
@@ -2117,6 +2117,18 @@ g_file_exist(const char *filename)
 }
 
 /*****************************************************************************/
+/* returns boolean, non zero if the file is readable */
+int APP_CC
+g_file_readable(const char *filename)
+{
+#if defined(_WIN32)
+    return 0; /* TODO: what should be done here? */
```
metalefty: Please let me know what should be done for WIN32.

cosmo0920: How about use `_access` or `_waccess`?
ref: https://msdn.microsoft.com/en-us/library/aa246685(v=vs.60).aspx

Natureshadow:
Actually, the *only* valid way to find out whether you can read a file is trying to read it. On all platforms.

Checking existence and access rights is not enough. Just the... open it and see what happens.

-nik

metalefty:
Indeed.

metalefty:
OK, these calls should error when failed to load valid certificate or privkey.
Use `log_message` instead of `g_writeln` should help users.

```c
    if (SSL_CTX_use_RSAPrivateKey_file(self->ctx, self->key, SSL_FILETYPE_PEM)
            <= 0)
    {   
        g_writeln("ssl_tls_accept: SSL_CTX_use_RSAPrivateKey_file failed");
        return 1;
    }   

    if (SSL_CTX_use_certificate_chain_file(self->ctx, self->cert) <= 0)
    {   
        g_writeln("ssl_tls_accept: SSL_CTX_use_certificate_chain_file failed");
        return 1;
    }  
```

metalefty:
Switching to `log_message` from `g_writeln` doesn't fail on security negotiation.

libxrdp/xrdp_rdp.c
```diff
@@ -268,6 +268,12 @@ xrdp_rdp_read_config(struct xrdp_client_info *client_info)
                 /* use user defined certificate */
                 g_strncpy(client_info->certificate, value, 1023);
             }
+
+	    if (!g_file_readable(client_info->certificate))
+            {
+                log_message(LOG_LEVEL_ERROR, "Cannot open certificate file %s: %s",
```
proski: "Cannot open" should only be logged is opening the file fails.

metalefty:
As @Natureshadow says, checking readability is not enough.  Fundamentally, certificate should be validated. Access rights check and user-friendly log help some users.  Better than nothing. Let's ship this.

PR #691, opened 2017-03-14
Propel xorgxrdp as default backend, give xorgxrdp the first place

metalefty:
If we recommend xorgxrdp as preferred backend, we should set xorgxrdp as default.

This is a part of "Merge Debian local patches project".


proski:
+1

I was thinking of submitting a patch like that. Let's have it in 0.9.2.

speidy:
+1

Issue #692, opened 2017-03-14
Make socket directory configurable


metalefty:
I looked at patches in Debian package, one of them named [sockpath.diff](http://sources.debian.net/src/xrdp/0.9.1-7/debian/patches/sockpath.diff/) changes the directory where xrdp's socket files are located. We have hard-coded path points to `/tmp/.xrdp` everywhere.

Can this be configurable? Like this for example. If the directory was configurable, the patch can be reduced.

`./configure --sockdirprefix=/foo/bar`

proski:
I like the idea of using `/var/run/xrdp`. One big advantage is that directories under `/var` can be made part of the package and have permissions preconfigured. They will persist over reboots. I also like the idea of using `./configure`.

I'm not sure about `/var/run/xrdp/sockdir` - that's one more directory to create and configure. FHS doesn't mention the second level directories under `/var/run`:
http://www.pathname.com/fhs/pub/fhs-2.3.html#VARRUNRUNTIMEVARIABLEDATA

I'll try to implement it for `/var/run/xrdp`. If Debian really wants `/var/run/xrdp/sockdir`, I want to hear their arguments.

proski:
I just noticed that the patch includes X11rdp and xorgxrdp. They can be installed separately from xrdp. I think they should get the socket directory as an environment variable. But it would be unsafe to change the default directory now. Somebody can end up with the new xrdp and an old X11rdp that is not aware of the environment variables and the changed default.

Still, we can provide tools for a distribution to change the socket directory across the board and set up the dependencies accordingly.

I think in the long term, xrdp should talk to xorgxrdp using an X11 protocol extension rather than use any additional sockets. But that would be a big change.

proski:
Done. Please see https://github.com/neutrinolabs/xrdp/pull/695 and https://github.com/neutrinolabs/xorgxrdp/pull/85

metalefty:
Both #695 and neutrinolabs/xorgxrdp#85 has been merged, closing. Thank you for implementing my idea.

proski:
It's a great idea. It will simplify packaging.

I think we need two follow-ups. chmod should not be called on a pre-existing socket directory - that would preserve the permissions set by the distribution. file_loc.h should be renamed to xrdp_sockets.h or something and installed, as it's a part of the xrdp API.

metalefty:
Indeed. This PR covers only socket path. Let's raise another PRs for premission.
And another one for header file name.

[...]

PR #695, opened 2017-03-15
Make socket directory configurable

proski:
The default remains `/tmp/.xrdp` for compatibility with older versions of X1rdp and xorgxrdp.

The socket directory is configured by passing `--with-socketdir=DIR` to `configure`.

xrdp-sesman passes the socket directory to the backends as XRDP_SOCKET_PATH environment variable. X11rdp uses it.

`file_loc.h` has been striped of the defines that should not have a default value.

Remove cleanup code in xrdp-sessvc, it is obsolete. xrdp-chansrv should be able to clean up after itself.


metalefty:
Great work! Much cleaner than I tried to do that.

metalefty:
@jsorg71 @speidy What do you think? See also #692.

PR #696, opened 2017-03-17
remove sessvc, one less process, move auth_start_session, auth_stop_session, auth_end to same process

jsorg71:


sesman/session.c
```diff
@@ -420,15 +349,51 @@ wait_for_xserver(int display)
...
+    int cspid;
+
+    cspid = g_fork();
+    if (cspid == 0)
+    {
+        chansrv_params = list_create(); 
+        chansrv_params->auto_free = 1;
+
+        /* building parameters */
+        g_snprintf(exe_path, 261, "%s/xrdp-chansrv", XRDP_SBIN_PATH);
```
proski: Please use sizeof(exe_path) instead of 261. g_snprintf() and underlying vsnprintf() count the final zero towards the maximum length.

sesman/session.c
```diff
@@ -420,15 +349,51 @@ wait_for_xserver(int display)
...
-session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
+session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
+                   struct SCP_SESSION *s)
 {
     int display = 0;
     int pid = 0;
     int wmpid = 0;
-    int pampid = 0;
     int xpid = 0;
+    int cspid = 0;
```
proski: I would appreciate if we use longer names for new variables. It's really helpful. "cs" can be anything: chanserv, client session, critical section, whatever. Also, underscores increase readability.

sesman/session.c
```diff
@@ -828,12 +779,17 @@ session_start_fork(tbus data, tui8 type, struct SCP_SESSION *s)
...
-                g_snprintf(text, 255, ":%d.0", display);
-                g_setenv("DISPLAY", text, 1);
-                /* new style waiting for clients */
-                session_start_sessvc(xpid, wmpid, data, s->username, display);
+                cspid = session_start_chansrv(s->username, display);
+                log_message(LOG_LEVEL_ALWAYS, "waiting for window manager "
+                            "pid(%d) to exit", wmpid);
+                g_waitpid(wmpid);
+                log_message(LOG_LEVEL_ALWAYS, "window manager pid(%d) did "
+                            "exit, cleaning up session", wmpid);
```
proski: It would be better to write `window manager (pid %d)` than `window manager pid(%d)` here and two lines above.

metalefty: +1 I agree.

jsorg71:
I'll create 2 new PR that will replace this one taking replies here into consideration.

PR #708, opened 2017-03-23
pulse fixes after socket directory change

metalefty:
Pulse module doesn't build after recent socket directory change.  This is not a complete fix. One more thing needs to be done. `XRDP_SOCKET_PATH`  needs to be passed pulse module.


proski:
+1

Sorry for that.

metalefty:
No problem :smile:

PR #717, opened 2017-03-30
Fix to handle OS disabled IPv6, issue #714.

MichaelSweden:
- Changes made only in the os_calls.c file (keep the libcommon interface intact)
- Exported functions changed: g_tcp_bind g_tcp_bind_address g_tcp_connect
- Support three network configurations:
  1) Normal network, with IPv6
  2) Partly disabled IPv6 via sysctl.conf
  3) Total disabled IPv6 via grub

Please see #714

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -419,8 +419,6 @@ g_tcp_socket(void)
     rv = (int)socket(AF_INET6, SOCK_STREAM, 0);
     if (rv < 0)
     {
-        log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
-
         switch (errno)
         {
             case EAFNOSUPPORT: /* if IPv6 not supported, retry IPv4 */
@@ -429,6 +427,7 @@ g_tcp_socket(void)
                 break;
 
             default:
+                log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
                 return -1;
         }
     }
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -727,8 +727,69 @@ g_sck_close(int sck)
 #endif
 }
 
+#if defined(XRDP_ENABLE_IPV6)
 /*****************************************************************************/
-/* returns error, zero is good */
+/* Helper function for g_tcp_connect.                                        */
+static int
+connect_loopback(int sck, const char *port)
+{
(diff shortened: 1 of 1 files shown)


metalefty:
I'm testing this.

metalefty:
Works fine for me.

metalefty:
I'll merge this in 24 hours if no objections.

MichaelSweden:
Okay, great.

I have now tested on Xubuntu 17.04 Beta2 and it works OK.

PS. One thing I notice was that if IPv6 isn't available and the g_tcp_socket function will falling back to IPv4. It will be an ERROR message in log files "g_tcp_socket: Address family not supported by protocol". This isn't so nice! After this it will be an INFO message "IPv6 not supported, falling back to IPv4". I haven't changed this function, so it is as it was before.

MichaelSweden:
Sorry, I'm a bit annoyed that I missed that old error message. This error message will always come at start if the system have IPv6 disabled. I think it should be removed. I have looked at it and it is easy fixed by moving the log_message line at row 422 little bit down to the default case. Like this:
```
index ca8ab42..524d819 100644
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -419,8 +419,6 @@ g_tcp_socket(void)
     rv = (int)socket(AF_INET6, SOCK_STREAM, 0);
     if (rv < 0)
     {
-        log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
-
         switch (errno)
         {
             case EAFNOSUPPORT: /* if IPv6 not supported, retry IPv4 */
@@ -429,6 +427,7 @@ g_tcp_socket(void)
                 break;
 
             default:
+                log_message(LOG_LEVEL_ERROR, "g_tcp_socket: %s", g_get_strerror());
                 return -1;
         }
     }
```
@metalefty Is it too late to commit a fix to this pull request? You wrote that you will merge this and I don't want to mess things up now.

metalefty:
@MichaelSweden No problem. I'm waiting for your fix.

moobyfr:
@MichaelSweden : just make another commit in your branch, and gitlab will add it into this pull request.

MichaelSweden:
Since the diff/system split the change in two. I show you this picture that more clearly show the change I did.
commitfixerrmsg

metalefty:
LGTM.

Issue #723, opened 2017-04-03
Pulse audio modules does not compile


gkyildirim:
I've cloned the latest code (7a7445fefd9a61535881ceedf5ed82febb52d2dc) and followed to build guide of [xrdp audio support](https://github.com/neutrinolabs/xrdp/wiki/Audio-Output-Virtual-Channel-support-in-xrdp) . ```make``` outputs error such as below.
```
gky@kamu_26102016:~/Downloads/xrdp/sesman/chansrv/pulse$ make
cc -Wall -O2 -I/home/gky/Downloads/pulseaudio-5.0 -I/home/gky/Downloads/pulseaudio-5.0/src -DHAVE_CONFIG_H -fPIC   -c -o module-xrdp-sink.o module-xrdp-sink.c
In file included from module-xrdp-sink.c:71:0:
module-xrdp-sink.c: In function ‘data_send’:
../../../common/xrdp_sockets.h:25:31: error: ‘XRDP_SOCKET_PATH’ undeclared (first use in this function)
 #define CHANSRV_PORT_OUT_STR  XRDP_SOCKET_PATH "/xrdp_chansrv_audio_out_socket_%d"
                               ^
module-xrdp-sink.c:311:37: note: in expansion of macro ‘CHANSRV_PORT_OUT_STR’
         snprintf(s.sun_path, bytes, CHANSRV_PORT_OUT_STR, u->display_num);
                                     ^
../../../common/xrdp_sockets.h:25:31: note: each undeclared identifier is reported only once for each function it appears in
 #define CHANSRV_PORT_OUT_STR  XRDP_SOCKET_PATH "/xrdp_chansrv_audio_out_socket_%d"
                               ^
module-xrdp-sink.c:311:37: note: in expansion of macro ‘CHANSRV_PORT_OUT_STR’
         snprintf(s.sun_path, bytes, CHANSRV_PORT_OUT_STR, u->display_num);
                                     ^
../../../common/xrdp_sockets.h:25:48: error: expected ‘)’ before string constant
 #define CHANSRV_PORT_OUT_STR  XRDP_SOCKET_PATH "/xrdp_chansrv_audio_out_socket_%d"
                                                ^
module-xrdp-sink.c:311:37: note: in expansion of macro ‘CHANSRV_PORT_OUT_STR’
         snprintf(s.sun_path, bytes, CHANSRV_PORT_OUT_STR, u->display_num);
                                     ^
[...]

PR #724, opened 2017-04-04
pulse: define default socket directory in Makefile

metalefty:
The build should be integrated to xrdp and socket directory should be inherited from xrdp but let's apply workaround for the present.

This helps users such as #723.

### DIFF
--- a/sesman/chansrv/pulse/Makefile
+++ b/sesman/chansrv/pulse/Makefile
@@ -2,8 +2,11 @@
 # build xrdp pulseaudio modules
 #
 
+# change this to your pulseaudio source directory
 PULSE_DIR = /home/lk/pulseaudio-1.1
-CFLAGS    = -Wall -O2 -I$(PULSE_DIR) -I$(PULSE_DIR)/src -DHAVE_CONFIG_H -fPIC
+# change this if you're using non-default socket directory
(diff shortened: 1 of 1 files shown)


sesman/chansrv/pulse/Makefile
```diff
@@ -2,8 +2,11 @@
 # build xrdp pulseaudio modules
 #
 
+# change this to your pulseaudio source directory
 PULSE_DIR = /home/lk/pulseaudio-1.1
-CFLAGS    = -Wall -O2 -I$(PULSE_DIR) -I$(PULSE_DIR)/src -DHAVE_CONFIG_H -fPIC
+# change this if you're using non-default socket directory
+SOCK_DIR  = /tmp/.xrdp
+CFLAGS    = -Wall -O2 -I$(PULSE_DIR) -I$(PULSE_DIR)/src -DHAVE_CONFIG_H -fPIC -DXRDP_SOCKET_DIR=\"$(SOCK_DIR)\"
```
gkyildirim: I guess XRDP_SOCKET_DIR should be XRDP_SOCKET_PATH. Otherwise it does not compile.

metalefty: You're correct. Thanks.

PR #726, opened 2017-04-04
sesman: LOGNAME should be set to username

### LINKED ISSUE #725: LOGNAME env should be set
In [sesman/env.c](https://github.com/neutrinolabs/xrdp/blob/6ed4c969f4d646a7751fe2da29ba94eddd3d6477/sesman/env.c#L134), the environment variable `LOGNAME` should be set to the same value as `USER`. I checked the source code of other X login managers, and it is being set there. There was even a bug in Firefox that makes [Firefox crash](https://bugzilla.mozilla.org/show_bug.cgi?id=1345438) when this isn't set.

metalefty:
which is used to store login name. The login name shall be the string
that would be returned by the getlogin() function [1]. Some applications
only refers LOGNAME environment variable and don't fallback to other
variables such as USER.

Fixes #725.  Reported by @seidler2547.

[1] The Single UNIX Specification, Issue 7

### DIFF
--- a/sesman/env.c
+++ b/sesman/env.c
@@ -132,6 +132,7 @@ env_set_user(const char *username, char **passwd_file, int display,
             g_setenv("SHELL", pw_shell, 1);
             g_setenv("PATH", "/sbin:/bin:/usr/bin:/usr/local/bin", 1);
             g_setenv("USER", username, 1);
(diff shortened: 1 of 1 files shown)


jsorg71:
+1

PR #732, opened 2017-04-18
pulse: define default socket directory in Makefile

metalefty:
Is it OK to ship this to v0.9 branch? It includes a build fix.

### DIFF
--- a/sesman/chansrv/pulse/Makefile
+++ b/sesman/chansrv/pulse/Makefile
@@ -2,8 +2,11 @@
 # build xrdp pulseaudio modules
 #
 
+# change this to your pulseaudio source directory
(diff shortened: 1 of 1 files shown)


proski:
Would it be possible to generate `sesman/chansrv/pulse/Makefile` from `sesman/chansrv/pulse/Makefile.in` if we cannot use Automake in that directory?

Also, would it be possible to read the socket directory from an environment variable and fall back to the configure default? Mismatching socket directory can lead to hard to diagnose issues.

I agree that build fixes should be merged to the stable branch. But let's make sure we have a solid fix.

metalefty:
OK, let's make a solid fix.

metalefty:
Can be replaced by #743.

PR #741, opened 2017-05-06
startwm.sh: do not call pseudo shell code for /etc/X11/xdm/Xsession

ftake:
Since /etc/X11/xdm/Xsession calls user's login shell, startwm.sh
should not execute the psuedo code; It causes to execute
~/.profile twice.


metalefty:
If no objections, I'll merge this within 24 hours.

PR #742, opened 2017-05-06
Created PID folder before starting the daemon

vincentkersten:
This is needed for Debian at least.


Natureshadow:
>This is needed for Debian at least.

We already have this fix in Debian...

If you have an issue with the Debian package, why don't you talk to the Debian maintainer?

metalefty:
This doesn't make sense to me. There's no reason to create `/var/run/xrdp` directory.  PID directory is not `/var/run/xrdp` but `/var/run` because `PIDFile` is set to `/var/run/xrdp-sesman.pid`. See 3 lines above the line you changed.

Probably you should talk to @Natureshadow or other Debian maintainers.

PR #743, opened 2017-05-09
make audio sockets path configurable

metalefty:
* pass xrdp socket path to user session (chansrv) via environment variable
* respect XRDP_SOCKET_PATH environment variable 
* other minor changes

I confirmed this properly works with non-default socket directory. Compiled `--with-socket-dir=/tmp/.xrdp2` option. If this is OK to merge, let's ship this to v0.9 branch too.


metalefty:
Any opinions? If no objections, I'll merge this in 24 hours.

PR #747, opened 2017-05-09
common: ssl_tls_write / read return 0 on socket close

jsorg71:
The non TLS send / recv return 0 on socket close, tls_read / write should too.


metalefty:
+1 Looks good to me.

jsorg71:
I think the problem was this while loop
https://github.com/neutrinolabs/xrdp/blob/devel/common/ssl_calls.c#L681-L696
SSL_shutdown is always failing.
Looking at the documentation, we should only need to call it twice, at most.

PR #768, opened 2017-06-07
Fix UK/GB keyboard layout

ben-cohen:
The UK/GB keyboard doesn't work properly connecting to xrdp with xorgxrdp.
It does work when connecting to xrdp with x11vnc, however.

This is because the layout is not declared in xrdp_keyboard.ini.  It needs
to be called "gb" not "uk".  (There are other layouts that have
km-nnnn.ini files that aren't declared in xrdp_keyboard.ini, so they might
have the same bug, but I haven't tested that.  This is analagous to the
commits for the "ch" and "pl" layouts.)

Test case:

1. Use a PC with the UK/GB keyboard layout.

2. Create /etc/xrdp/startwm.sh as follows:

     #!/bin/sh
     export LANG=en_GB.UTF-8
     export MDM_LANG=en_GB.UTF-8
     export XTERM_LOCALE=en_GB.UTF-8
     xterm
     exit 0

3. Connect using rdesktop to localhost (from a session where the same
   environment variables are defined).

4. Test layout-specific keys such as:
      \ (backslash)
      | (bar)
      `#` (numbersign)
      ~ (asciitilde)
      £ (sterling: Shift-3)
      € (EuroSign: AltGr-4)

These keys should produce the appropriate symbol but they don't.  With
the change in this commit the keys produce the correct symbol.


metalefty:
+1

metalefty:
Easy and no logic change, let's merge.

PR #789, opened 2017-06-30
vnc: do not send any data if cliprdr channel not available

### LINKED ISSUE #755: Xvnc backend disconnects when some data copied to clipboard
I'm trying to use xrdp with Xvnc on CentOS 7.3, I've managed to get to the point where the VNC server is started and I briefly see the desktop before the RDP client is abruptly closed.  Checking on the server I can see that Xvnc is still running, and I can connect to it successfully with vncviewer.

It seems like xrdp gets clipboard data from Xvnc that it can't decode for some reason and then bombs out:

```
xrdp: [20170520-18:53:30] [DEBUG] VNC got clip data
xrdp: lib_data_in: lib_mod_process_message failed
```

I'm using xrdp 0.9.2-5.el7 from EPEL, with TigerVNC 1.8.0 (which was released a few days ago), but I get the same problem with tigervnc-server-1.3.1-9.el7.

metalefty:
This caused disconnection when client connecting without cliprdr
channel because VNC module tries to send data to a non-existent
channel (channel_id = -1).

Should fix #755.

### DIFF
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -802,7 +802,15 @@ lib_clip_data(struct vnc *v)
         out_uint8s(out_s, 4);
         s_mark_end(out_s);
         size = (int)(out_s->end - out_s->data);
-        error = v->server_send_to_channel(v, v->clip_chanid, out_s->data, size, size, 3);
+        if (v->clip_chanid < 0)
+        {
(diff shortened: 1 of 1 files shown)


metalefty:
This approach is not correct.

PR #792, opened 2017-07-01
Unify chanserv logging

geiseri:
I removed all instances of ```LOG(...)``` and replaced them with ```LOGM(...)```.  The following mappings where used:

* ```0``` -> ```LOG_LEVEL_INFO``` (or ```LOG_LEVEL_ERROR``` if it was an error)
* ```1``` -> ```LOG_LEVEL_INFO```
* ```10``` -> ```LOG_LEVEL_DEBUG```
* ```20``` -> ```LOG_LEVEL_DEBUG```


metalefty:
+1

metalefty:
I think macros such as `log_error`, `log_always`, `log_info`, `log_debug` in `chansrv/clipboard.c` should use LOGM. These macros use `g_write` or `g_writeln`. They writes log to stdout but it is not helpful.

geiseri:
@metalefty okay, I will get that one too.

geiseri:
looks like ```devredir.c``` needs it too

geiseri:
After looking at ```mlog.h``` it looks like there are some big differences in functionality.  it might be better for the output logic in mlog.h to use the ```log_message(...)``` plumbing and then have all of the modules use the ```log_error```, ```log_always```, ```log_info``` and ```log_debug```.  Does that sound like a better plan?

metalefty:
Don't forget to remove removed files in Makefile.am.

I think `log_*` functions are debug logs. So converting `log_error` -> `LOGM((LOG_LEVEL_ERROR` might be too noisy.

metalefty:
I'm working on this based on your work.

aquesnel:
This pull request can be closed now since PR #1633 includes these changes and has been merged to devel.

PR #797, opened 2017-07-05
xrdp-sesadmin: fix error when there are no sessions

ben-cohen:
Test case:  On a system running xrdp with no sessions running run:

  xrdp-sesadmin -u=<user> -p=<password> -c=list

Expected result: "No sessions." (ignoring debug output)
Observed result: "Error getting session list."

In the SCP_SERVER_STATE_MNG_LISTREQ case in scp_v1_mng_process() if
there are no sessions it ends the scp session, which causes an error in
the client.  In commit 0017081d the client was changed to report errors,
giving the result above.

Fix by calling scp_v1s_mng_list_sessions() from scp_v1_mng_process()
even when there are no sessions, and if so sending a packet with a count
of zero so that the client gets what it expects.


metalefty:
I quickly did test run, it looks fine. Good catch!

Issue #801, opened 2017-07-07
If socket directory not present xrdp fails to connect


geiseri:
In cases where the unix sockets are kept in a tmpfs directory the structure may not be present when the unix sockets are requested.  This will cause the connection to fail after the socket create timeout.  If we ensure the directory structure is present right before it is requested it will continue to run.

geiseri:
@metalefty #809 effectively breaks all of the sockets since it cannot actually create the directory.

geiseri:
I also noticed that the code chmod the socket directory so that others can create sockets in the created directory.  I think it needs to be 0x1777.

geiseri:
Found it, I used ```g_create_path``` not ```g_create_dir```.  This ensured the entire structure was present in case it was not already.

metalefty:
Fix has been merged into devel, closing.

PR #802, opened 2017-07-07
Ensure socket directory present

geiseri:
In cases where the socket directory is in tmpfs it cannot always be certain to exist.  Debian has a hack around this by running a shell script to generate the needed directory structure when the socket dir is not ```/tmp/.xrdp```.  This does the equivalent at startup of sessman.  This should fix issue #801.

### DIFF
--- a/sesman/sesman.c
+++ b/sesman/sesman.c
@@ -417,6 +417,17 @@ main(int argc, char **argv)
     log_message(LOG_LEVEL_INFO,
                 "starting xrdp-sesman with pid %d", g_pid);
 
+    /* make sure socket (XRDP_SOCKET_PATH) directory exists */
+    if (!g_directory_exist(XRDP_SOCKET_PATH))
+    {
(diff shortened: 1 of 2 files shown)


metalefty:
+1

sesman/sesman.c
```diff
@@ -417,6 +417,17 @@ main(int argc, char **argv)
     log_message(LOG_LEVEL_INFO,
                 "starting xrdp-sesman with pid %d", g_pid);
 
+    /* make sure socket (XRDP_SOCKET_PATH) directory exists */
```
metalefty: I noticed we can use `g_mk_temp_dir()` instead.

geiseri: Where does that go?  I think debian likes to put it in the socket path.   It needs to be deterministic because sound needs it.

metalefty: The function name is not very good but it creates sockets directory. See what `g_mk_temp_dir()` function does.

metalefty:
Replaced by #809.

Issue #803, opened 2017-07-09
Possibly, a glitch with IPv4 struct initialization


andrecbarros:
Look on "s" initialization (zero fill) on line 755 of "common/os_calls.c", it should be, it seems, "g_memset(&s, 0, sizeof(s));" ("s" instead of "sa"). Copy and past is salvation, copy and past is a sin!

[xrdp-wrong_struct_zeroed.diff.txt](https://github.com/neutrinolabs/xrdp/files/1133466/xrdp-wrong_struct_zeroed.diff.txt)

metalefty:
Indeed.

metalefty:
Merged, thanks!

PR #805, opened 2017-07-10
common: fix a glitch with IPv4 struct initialization

metalefty:
Pointed out by: andrecbarros
Closes: #803


metalefty:
This is simple and clear. Let me merge this immediately.

jsorg71:
+1

PR #806, opened 2017-07-10
sesman: move auth_start_session, auth_stop_session, and auth_end to c…

jsorg71:
…hild process

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -457,7 +457,6 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
         return 0;
     }
 
-    auth_start_session(data, display);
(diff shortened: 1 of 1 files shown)


jsorg71:
This PR should replace #779

metalefty:
Tested on FreeBSD, I confirmed that this PR breaks nothing for FreeBSD.
If this solves systemd sessions issue, I'm OK with merging this.

moobyfr:
Tested on linux jessie + ldap/krb5, the users have the pam settings applied, so not disturbing the  CVE-2017-6967 fix. And it's fix the original issue #778  about systemd

metalefty:
Let's merge and prepare release.

PR #813, opened 2017-07-13
Adjustments to XRDP_SOCKET_PATH make process

geiseri:
This does 2 things.  The first is use ```g_create_path``` instead of ```g_create_dir``` to ensure the entire path is present when creating the socket path.  The second is to rename ```g_mk_temp_dir``` to something more meaningful.  It is now called ```g_mk_socket_path```.


metalefty:
+1 Thanks!

PR #819, opened 2017-07-16
Add shared memory file to rediscover sessions #800

ben-cohen:
Allow xrdp-sesman to discover sessions still running that were created
by a previous xrdp-sesman process.

Implement this using a file-backed shared mmap region, visible to both
the daemon and session instances of xrdp-sesman.  Add a heartbeat
timestamp updated by the session instance so that the daemon instance
can infer whether the sessions in the mmap region are current or stale.

The shared memory can also be used to pass other data between daemon and
session instances, for example the session's idle time.

Add locking around access to g_sessions and the mmap region.  This is a
PTHREAD_PROCESS_SHARED lock which protects shared memory used by
separate processes.  Defining DEBUG_SESSION_LOCK enables logging to help
debug locking bugs.

Notes:

1. The number of sessions is limited by the size of the array in shared
memory, SESMAN_SHAREDMEM_MAX_SESSIONS.  This could be made dynamic
instead by growing the file and the mmap region.

2. If sesshm_try_open_existing_shm() finds a shm file but it looks
wrong, it creates a new one.  Perhaps it should instead exit with an
error?

3. In sesshm_thread() if the session xrdp-sesman notices that it has
been removed from the daemon's list it exits.  It should kill the X
server and/or window manager first.

4. This doesn't yet update idle times.  I need to work out how to get
this information from the X server.

5. This uses an array of sessions in the mmap region so the linked list
g_sessions might be redundant now.

6. Defining DONT_USE_SHM will let xrdp-sesman run without creating or
trying to use the mmap file.  This could be removed in the future.

7. I think the locking fixes an theorised bug where a session can be
removed from the g_sessions linked list in session_kill() (called from
the signal handler) at the same time as a session is added in
session_start_fork().


sesman/sesman.c
```diff
@@ -413,6 +413,12 @@ main(int argc, char **argv)
         g_file_close(fd);
     }
 
+    if (session_init_shared())
```
speidy: Please always add brackets for if body statements. Check out xrdp coding style in wiki.

sesman/session.c
```diff
@@ -37,10 +40,22 @@
...
 #endif
 
+#include <sys/mman.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <pthread.h>
```
speidy: Are all those includes really needed? we better handle pthread stuff inside thread_calls.c

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+ * @return 0 on success; non-zero otherwise
+ *
+ */
+static int
+sesshm_lock()
+{
+#ifdef DONT_USE_SHM
+    return 0;
+#else
+    int rc = pthread_mutex_lock(&g_shm_mapping->mutex);
```
speidy: Please use thread_calls.c calls, or add a wrapper there.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+#ifndef DONT_USE_SHM
+static int
+sesshm_try_lock()
+{
+    return 0;
+    struct timespec timeout;
+    timeout.tv_sec = 1;
+    timeout.tv_nsec = 0;
+
+    int rc = pthread_mutex_timedlock(&g_shm_mapping->mutex, &timeout);
```
speidy: Same as above.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+                    "pthread_mutex_unlock() failed (%d)", errno);
+    }
+    return rc;
+#endif
+}
+
+
+#ifdef DEBUG_SESSION_LOCK
+#define sesshm_try_lock() \
+       ({  \
```
speidy: Consider using do { ... } while(0); pattern.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+ *
+ * @brief map the file fd into memory using mmap
+ * @return pointer to the mapping
+ *
+ */
+#ifndef DONT_USE_SHM
+static struct session_shared_data *
+sesshm_map(int fd)
+{
+    g_shm_mapping = (struct session_shared_data *) mmap(NULL,
```
speidy: We should also add wrapper for this one, maybe in os_calls.c.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+/******************************************************************************/
+/**
+ *
+ * @brief thread to poll the shared memory for various changes
+ * @return NULL
+ *
+ */
+#ifndef DONT_USE_SHM
+static void *
+sesshm_thread(void *arg)
```
speidy: We might handle this stuff in a separate file. 
session.c is already too big.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+#ifdef DONT_USE_SHM
+    return 0;
+#else
+    int rc;
+
+    /* Look for an existing shared file */
+    rc = sesshm_try_open_existing_shm();
+
+    /* If it's not okay then create it */
+    if (rc != 0)
```
speidy: Missing brackets.

sesman/session.c
```diff
@@ -90,6 +112,577 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+ * @brief validate whether a session_item pointer is valid
+ * @param item a pointer to a session_item
+ * @return 1 if valid or 0 if not
+ *
+ */
+#ifndef DONT_USE_SHM
+static int
+validate_session_item_ptr(struct session_item *item)
+{
+    char *start = (char *) &g_shm_mapping->sessions[0];
```
speidy: Please separate deceleration from initialization.

ben-cohen:
@speidy - I have implemented your suggestions in the two commits above.

mrichar1:
Just hit #800 and was looking to see if it has been fixed and found this PR - any chance this could get picked up and worked on again, as this would be really useful to have!

luisdalves:
Hey guys. Do you have any plan to include this?

Prototyped:
This is of significant interest to people who have things like Debian/Ubuntu unattended-upgrades set up with needrestart. The latter willl restart systemd services impacted by an update, and this includes xrdp-sesman. As a result, after an upgrade of xrdp, I lose the ability to reattach to already-running sessions.

It seems merging this change will allow xrdp to "reclaim" the previously running sessions (modulo maintaining backward compatibility with the contents of the file across versions, or at least detecting incompatibility upon load, which is a whole other issue). This would resolve the problem.

(Instead of using a memory-mapped file which carries the risk that the structures change across xrdp upgrades, I would suggest instead using a SIGINT/SIGTERM/SIGQUIT handler that serializes out the structures into a forgiving format, e.g. JSON, where a newer version can load what is present and default the rest, and also the object can carry a file format version that a newer version can check against to determine whether it still has the ability to load it.)

matt335672:
Hi all,

I'm a little late to this conversation I'm afraid. I absolutely agree this functionality is necessary from a restarting services perspective. 

At the moment we're working on a rather important fix for a systemd change (#1684), which is looking like it's going to break this PR I'm afraid.

It's not all bad news however. Some of the functions of sesman need to be moved into the session leader process which runs the user session. This is going to involve active communication between sesman and the session leader.

Once this is done, we can arrange for the session leader to detect when sesman restarts and re-send the state data sesman needs to be able to re-establish connections.

I think this is a better systems design. The PR here, whether or not we're using a separate file, can't track session state changes while sesman is down. The session leaders however will be able to do this while sesman is stopped.

Please feel free to come back to me with any questions or comments about this.

metalefty:
This should be reconsidered after #1961.

PR #823, opened 2017-07-17
Make the create path a bit more robust

geiseri:
append a / to ensure the full path is created even when the config variable lacks a trailing /


metalefty:
+1

Issue #830, opened 2017-07-20
zombie sessions?


alex-p-uk:
I'm running xrdp 0.9.2 on Ubuntu 14.04 with `MaxSessions=50` and persistent sessions. 

After a while clients can't connect any longer because: `[20170720-14:50:35] [ERROR] X server -- no display in range is available` . `ps` only shows around 20 Xorg sessions, but I can see some of them have display IDs above 50, which means xrdp for some reason didn't reuse lower numbers when clients disconnected and eventually reached the number that matches `MaxSessions` + `X11DisplayOffset=10`. I think this may be related to https://github.com/neutrinolabs/xrdp/issues/812

Below is a log from freshly restarted xrdp. 2 things to note:
* Xorg :11 and Xorg :12 were followed by Xorg :15 while :13 and :14 displays were free
* it repeats that it closed more and more sockets, which may suggest those weren't real sockets or weren't actually closed

This requires to restart and drop sessions once in a while, which did not happen with xrdp-0.8. I'm happy to try any suggestions. Thank you!

```
[20170720-14:51:49] [DEBUG] libscp initialized
[20170720-14:51:49] [INFO ] starting xrdp-sesman with pid 8829
[20170720-14:51:49] [INFO ] listening to port 3350 on 127.0.0.1
[20170720-14:53:29] [INFO ] A connection received from 127.0.0.1 port 58417
[20170720-14:53:30] [INFO ] ++ created session (access granted): username USER1, ip 123.456.789.10:53635 - socket: 12
[20170720-14:53:30] [INFO ] starting Xorg session...
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:5910)
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:6010)
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:6210)
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:5911)
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:6011)
[20170720-14:53:30] [DEBUG] Closed socket 10 (AF_INET 0.0.0.0:6211)
[20170720-14:53:31] [DEBUG] Closed socket 8 (AF_INET 127.0.0.1:3350)
[...]

Issue #812, opened 2017-07-13
some socket files are remained after session exits


metalefty:
I noticed socket files are remained after session exits.

When session running
```
% LANG=C ls -l /tmp/.xrdp
total 2                       
srw-rw----  1 meta  wheel  0 Jul 13 13:29 xrdp_chansrv_audio_in_socket_10                                               
srw-rw----  1 meta  wheel  0 Jul 13 16:41 xrdp_chansrv_audio_out_socket_10                                              
srwxr-xr-x  1 root  wheel  0 Jul  6 16:37 xrdp_display_10   
srw-rw----  1 meta  wheel  0 Jul  6 16:37 xrdpapi_10        
```

After session exits
```
% LANG=C ls -l /tmp/.xrdp
total 2                       
srw-rw----  1 meta  wheel  0 Jul 13 13:29 xrdp_chansrv_audio_in_socket_10                                               
srw-rw----  1 meta  wheel  0 Jul 13 16:41 xrdp_chansrv_audio_out_socket_10                                              
srw-rw----  1 meta  wheel  0 Jul  6 16:37 xrdpapi_10   
```

Only `xrdp_display_10` is removed. It is created by xorgxrdp and xorgxrdp removes the socket file on exit.
https://github.com/neutrinolabs/xorgxrdp/blob/f83423e980f4e1cd53af90c093955cbff44f7452/module/rdpClientCon.c#L1262

Other 3 files are created by sesman however sesman doesn't cleanup these files.  These remained socket files prevent future session from reusing display number because any of socket files exist, sesman regards the display number as in use.
https://github.com/neutrinolabs/xrdp/blob/devel/sesman/session.c#L213

Maybe these audio in/out socket and xrdp api socket file should be removed around here.
https://github.com/neutrinolabs/xrdp/blob/b6bd8acef19763b1116bb4c881b46b0359c84f90/sesman/session.c#L799

metalefty:
Fixed by #831.

DS455:
Again, there is a problem with the connection. Installed version of hrdp v0.9.4.

xrdp-sesman.log

```
[20171025-14:26:56][INFO ] A connection received from 127.0.0.1 port 35290
[20171025-14:26:57] [INFO ] ++ created session (access granted): username tes, ip 192.168.0.187:42654 - socket: 12
[20171025-14:26:57] [INFO ] starting Xorg session...
[20171025-14:26:57] [ERROR] X server -- no display in range is available
[...]

PR #831, opened 2017-07-21
sesman: delete socket files when session exits

metalefty:
Solves #812 and #830.

I'm not sure if pulse module can delete sockets by itself so deleting sockets on session exit.


metalefty:
@jsorg71 Do you think there's better way to delete orphans socket files?

metalefty:
I want to ship this to the next release.

moobyfr:
+1, working fine.

metalefty:
Let's merge.

PR #833, opened 2017-07-24
Small refactoring before adding new audio codec

metalefty:
I'm planning to introduce additional audio codecs such as ADPCM or AAC.
Let me refactor a little bit. Supporting more codecs make if-else-if-else deeper. Using switch-case instead.

### DIFF
--- a/common/xrdp_constants.h
+++ b/common/xrdp_constants.h
@@ -394,11 +394,13 @@
 #define CF_GDIOBJFIRST                 768
 #define CF_GDIOBJLAST                  1023
 
-/* Sound format constants */
-#define WAVE_FORMAT_PCM                1
-#define WAVE_FORMAT_ADPCM              2
-#define WAVE_FORMAT_ALAW               6
-#define WAVE_FORMAT_MULAW              7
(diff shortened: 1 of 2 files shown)


jsorg71:
+1
BTW, I've got a working branch with AAC here
https://github.com/jsorg71/xrdp/tree/fdk-aac
I tested with MSTSC from windows 10.

metalefty:
Nice! I haven't actually started AAC yet.

What about ADPCM? I think I can use [libsox](http://sox.sourceforge.net/libsox.html) for MS-ADPCM and researching how to use.  In my understanding, libsox supports not only MS-ADPCM but also A-law, mu-law, u-law without external libraries so we can add more than one codec by one library. Another advance of libsox is in RHEL repository. RHEL has very fewer number of multimedia libraries than Debian like distro. 

If you know good alternatives, please let me know.

jsorg71:
@metalefty libsox sounds good, it's a matter of if the output is compatible with MS client.
I'll make a new PR for the my AAC branch.  I wanted to separate out the commits anyway.

PR #838, opened 2017-07-27
xfuse_create_share() called before xfuse_init()

matt335672:
Running xrdp 0.9.2 on CentOS/RHEL 7.3, we occasionally see coredumps of chansrv, always when logging in.

Investigation of these coredumps shows a SEGV in `xfuse_create_share()`, where the new xinode is inserted into the inode_table. At this point, `g_xrdp_fs` is uninitialized so it seems that `xfuse_init()` has not yet been called.

```
(gdb) print g_xrdp_fs
$1 = {inode_table = 0x0, max_entries = 0, num_entries = 1, next_node = 1}
```

This patch caters for this situation by creating the inode table if required during `xfuse_create_share()`

My local testing shows chansrv to be more robust with this patch. Is there maybe a preferable way to handle this situation, like calling `xfuse_init()` from elsewhere?

### DIFF
--- a/sesman/chansrv/chansrv_fuse.c
+++ b/sesman/chansrv/chansrv_fuse.c
@@ -611,6 +611,12 @@ int xfuse_create_share(tui32 device_id, const char *dirname)
     if (dirname == NULL || strlen(dirname) == 0)
         return -1;
 
+    /* Do we have an inode table yet? */
+    if (xfuse_init_xrdp_fs())
+    {
+        return -1;
+    }
+
     xinode = g_new0(struct xrdp_inode, 1);
     if (xinode == NULL)
     {
@@ -857,6 +863,12 @@ static int xfuse_init_lib(struct fuse_args *args)
 static int xfuse_init_xrdp_fs(void)
 {
     struct xrdp_inode *xino;
+
+    /* Already called? */
+    if (g_xrdp_fs.inode_table != NULL)
+    {
+        return 0;
+    }
 
     g_xrdp_fs.inode_table = g_new0(struct xrdp_inode *, 4096);
     if (g_xrdp_fs.inode_table == NULL)


sesman/chansrv/chansrv_fuse.c
```diff
@@ -443,8 +443,11 @@ xfuse_init(void)
     }
 
     /* setup xrdp file system */
-    if (xfuse_init_xrdp_fs())
-        return -1;
```
metalefty: Always use braces for newly touched codes. The exiting code is not always good example. The original code violates our coding standard.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -611,6 +614,13 @@ int xfuse_create_share(tui32 device_id, const char *dirname)
     if (dirname == NULL || strlen(dirname) == 0)
         return -1;
 
+    /* Do we have an inode table yet? */
+    if (g_xrdp_fs.inode_table == NULL)
+    {
+        if (xfuse_init_xrdp_fs())
+            return -1;
```
metalefty: Same thing here. Braces.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -443,8 +443,11 @@ xfuse_init(void)
     }
 
     /* setup xrdp file system */
-    if (xfuse_init_xrdp_fs())
-        return -1;
+    if (g_xrdp_fs.inode_table == NULL)
+    {
+        if (xfuse_init_xrdp_fs())
+            return -1;
```
metalefty: Always use braces for newly touched codes. The exiting code is not always good example. The original code violates our coding standard.

metalefty:
I understand the cause of the issue but I'm not sure why your patch solves the issue.
Your change makes `xfuse_init_xrdp_fs()` inside `xfuse_init()` will be not allways called, right?

matt335672:
Hi metalefty

Apologies for the braces. I've found and read the coding standard now.

As regards your other comment, the change inside `xfuse_init()` was made to cater for the situation where `xfuse_init_xrdp_fs()` is called before `xfuse_init()`.  In this situation, `xfuse_init()` will eventually be called, and we don't want to initialise the inode table again.

I agree my intent could be clearer however, so I've cleaned up the patch a bit. I've place the guard against being called twice inside `xfuse_init_xrdp_fs()` now, and simply called the function unconditionally from both places where it's needed. The original code in `xfuse_init()` is now untouched.

Is that clearer now, or have I misunderstood your question?

jsorg71:
+1

metalefty:
Now the fix looks pretty clearer.

Let's make the commit message clearer next. This PR has two point, right?

 * make sure inode table be initialized
 * add a guard not to initialize twice

You can squash the commits into one commit and edit commit message before the PR merged.

matt335672:
How's that?

metalefty:
This will help you. https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/

matt335672:
Apologies metalefty - bad terminology on my part in my comment. I've done the squash already, and all I really wanted was a quick sanity check. Nice link though.

metalefty:
I also apologize you. I didn't aware you've done squash. It seems pretty good. Ready to merge.

PR #840, opened 2017-07-28
add audio codec fdk-aac

jsorg71:
fix for skipped acks, add fdk-aac decode option


common/xrdp_constants.h
```diff
@@ -401,6 +401,7 @@
 #define WAVE_FORMAT_MULAW              0x0007
 #define WAVE_FORMAT_MPEGLAYER3         0x0055
 #define WAVE_FORMAT_OPUS               0x0069
+#define WAVE_FORMAT_AAC_MS             0xA106
```
metalefty: Let's use the same name as RFC or MS document.

https://msdn.microsoft.com/en-us/library/dd341376.aspx
https://tools.ietf.org/html/rfc2361

jsorg71: good point

configure.ac
```diff
@@ -230,6 +234,13 @@ then
     [AC_MSG_ERROR([please install libfuse-dev or fuse-devel])])
 fi
 
+# checking for fdk aac
+if test "x$enable_fdkaac" = "xyes"
+then
+  AC_CHECK_HEADER([fdk-aac/aacenc_lib.h], [],
+    [AC_MSG_ERROR([please install libfdk-aac-dev or dontknow-devel])])
```
speidy: `dontknow-devel` --> `fdk-aac-devel`
* (Ref: https://www.rpmfind.net/linux/rpm2html/search.php?query=fdk-aac-devel)

sesman/chansrv/sound.c
```diff
@@ -436,6 +469,206 @@ sound_process_output_formats(struct stream *s, int size)
...
+    int mode;
+    int bitrate;
+    int afterburner;
+    int channel_order;
+    AACENC_InfoStruct info = { 0 };
+    AACENC_BufDesc in_buf = { 0 };
+    AACENC_BufDesc out_buf = { 0 };
+    AACENC_InArgs in_args = { 0 };
+    AACENC_OutArgs out_args = { 0 };
+    void *in_buffer[1];
```
speidy: Whats is the point to use an array of single element for the following variables?
I would switch `void **`, `int`

sesman/chansrv/sound.c
```diff
@@ -436,6 +469,206 @@ sound_process_output_formats(struct stream *s, int size)
...
+
+        error = aacEncOpen(&g_fdk_aac_encoder, 0, 2);
+        if (error != AACENC_OK)
+        {
+            LOG(0, ("sound_wave_compress_fdk_aac: aacEncOpen() failed"));
+            return rv;
+        }
+
+        aot = 2; /* MPEG-4 AAC Low Complexity. */
+        //aot = 129; /* MPEG-2 AAC Low Complexity. */
```
speidy: Do we need this commented out code?

configure.ac
jsorg71: great, thanks

sesman/chansrv/sound.c
jsorg71: This is just to be consistent with there API.  You can encode many buffers.  We just have one but just trying to follow their style.

sesman/chansrv/sound.c
jsorg71: This was left over from the effort to figure out why MSTSC does not decode.  I'll remove it.

configure.ac
```diff
@@ -230,6 +234,13 @@ then
     [AC_MSG_ERROR([please install libfuse-dev or fuse-devel])])
 fi
 
+# checking for fdk aac
+if test "x$enable_fdkaac" = "xyes"
+then
+  AC_CHECK_HEADER([fdk-aac/aacenc_lib.h], [],
+    [AC_MSG_ERROR([please install libfdk-aac-dev or fdk-aac-devel])])
+fi
+
```
metalefty: Not bad but using pkg-config would be better.  fdk-aac install .pc files.

sesman/chansrv/Makefile.am
```diff
@@ -23,6 +23,11 @@ AM_CPPFLAGS += -DXRDP_FUSE $(FUSE_CFLAGS)
 CHANSRV_EXTRA_LIBS += $(FUSE_LIBS)
 endif
 
+if XRDP_FDK_AAC
+AM_CPPFLAGS += -DXRDP_FDK_AAC
+CHANSRV_EXTRA_LIBS += -lfdk-aac
```
metalefty: can be written like this with pkg-config
`CHANSRV_EXTRA_LIBS += $(FDKAAC_LIBS)`

PR #859, opened 2017-09-01
xrdpapi: change WTSVirtualChannelWrite return codes to be consistent with WTSVirtualChannelRead

speidy:

PR #863, opened 2017-09-06
Fix WTSVirtualChannelWrite() returncode

ben-cohen:
The function WTSVirtualChannelWrite() in xrdpapi.c returns 0 on success
and -1 on error.  It should return 1 on success and 0 on error, because:

1. It is the behaviour described in Microsoft's [API documentation](https://msdn.microsoft.com/en-us/library/aa383855(v=vs.85).aspx).
2. It is the behaviour expected by xrdpapi/[simple.c](https://github.com/neutrinolabs/xrdp/blob/devel/xrdpapi/simple.c#L124).
3. It is consistent with xrdpapi's behaviour for other functions
   including [WTSVirtualChannelRead()](https://github.com/neutrinolabs/xrdp/blob/devel/xrdpapi/xrdpapi.c#L280).

### DIFF
--- a/tcutils/utils.cpp
+++ b/tcutils/utils.cpp
@@ -37,7 +37,7 @@ int Utils::getMountList(void *wtsChannel, QList<QListWidgetItem *> *itemList)
 
     /* send command */
     rv = WTSVirtualChannelWrite(wtsChannel, s.data, bytesToSend, &bytesWritten);
-    if (rv)
+    if (!rv)
     {
         QMessageBox::information(NULL, "Get device list", "\nError sending "
                                  "command to client");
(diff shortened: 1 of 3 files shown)


ben-cohen:
Apparently the Travis CI breakage was caused by commit [e6b7d12](https://github.com/neutrinolabs/xrdp/commits/e6b7d12191200fec68fed07f61c54e44bc69d4f5) on devel and not this change!

ben-cohen:
NB The existing returncode for WTSVirtualChannelWrite() are expected in xrdpvr/xrdpvr.c and tcutils/utils.cpp.

ben-cohen:
Closing as a duplicate of #859.

Issue #860, opened 2017-09-03
Display issues in NeutrinoRDP 


Zhao150101:
I RDP to a windows server(OS: windows server 2008) by NeutrinoRDP and then ssh to a Linux server. I find an issue that some lines are missing when I try to scan a txt file. Like the picture below.
![image](https://user-images.githubusercontent.com/15672593/30002970-82333bf6-90e9-11e7-93af-7442c60165fc.png)
The real file content is "Aaaaaaa\nBbbbbbb" but the second line is missing in my picture.
What I had done is  

- ssh to a Linux server using SecureCRT

- input `clear` to clear screen

- input `cat /tmp/1.t` to show the txt file

By my testing, the second line is missing if it has the same length with the first line.

Besides , I also find that if I type <Ctrl+L> in windows `cmd.exe` and then type `Enter`, the screen will show some black areas in wrong place like below picture.
![image](https://user-images.githubusercontent.com/15672593/30003090-2559d83e-90eb-11e7-888f-9563a685fa73.png)

All versions of xrdp and NeutrinoRDP are latest (branch devel).
I'm not sure whether these 2 issues caused by the same reason, so would you please help to have a look?
Thanks!

cocoon:
you can try my branch:

`git clone https://github.com/cocoon/xrdp -b 2017-9-04-devel-fixed`

and if you want to avoid a crash and running on Ubuntu1604:

`git clone https://github.com/cocoon/NeutrinoRDP -b Ubuntu1604-CrashFix-UserHomeEnv`

Zhao150101:
It works, thanks @cocoon

metalefty:
@cocoon Let's merge it. It is worth enough. Can you separate your fix and create a PR?

PR #869, opened 2017-09-08
workaround for corrupted display with W2K8

cocoon:
workaround for corrupted display like black bars left of cmd with W2K8
as requested here: https://github.com/neutrinolabs/xrdp/issues/860

But maybe someone can do proper fixes.

BTW:
I reported this looong timea go ;)

here:
http://xrdp-devel.766250.n3.nabble.com/Xrdp-devel-NeutrinoRDP-black-lines-with-ping-in-a-remote-console-screenshot-td4025274.html

or here:
https://github.com/neutrinolabs/NeutrinoRDP/issues/1

BTW2: still open bugs:

- ~~misuse of hidelogwindow instead of correct own hideloginwindow~~

- ~~xrdp systemd setting User=root and fix neutrinordp https://github.com/neutrinolabs/NeutrinoRDP/blob/devel/libfreerdp-utils/file.c#L81
(question is if it can cause any harm if home is set to / if it can't get from env)~~

- new cursor + Windows 10

### DIFF
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -1468,7 +1468,9 @@ lfreerdp_pre_connect(freerdp *instance)
 
     instance->settings->glyph_cache = true;
     /* GLYPH_SUPPORT_FULL and GLYPH_SUPPORT_PARTIAL seem to be the same */
-    instance->settings->glyphSupportLevel = GLYPH_SUPPORT_FULL;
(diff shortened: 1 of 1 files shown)


metalefty:
@jsorg71 what do you think of this workaround? If you're ok with this, let's ship to v0.9.5. Or can you fix the issue fundamentally?

jsorg71:
I can fix this later.  Let go ahead and turn glyphs off for now.

metalefty:
Okay, let's fix this later and merge this workaround now.

Issue #414, opened 2016-08-25
black screen if hidelogwindow=yes


cocoon:
Hi!
I had the problem, that I got a black screen since the following commit if I set hidelogwindow=yes in xrdp.ini:

https://github.com/neutrinolabs/xrdp/commit/f49e2058dfd857b441daa4b18bcde6e341d8b9da

xrdp: change when using xrdp.ini hidelogwindow
jsorg71 authored on 2 Dec 2015

Problem seems to be the change from:

if (self->session->client_info->rdp_autologin)

to:
if (self->session->client_info->rdp_autologin || self->hide_log_window)

I only want to hide the "Log-Window" that appears if something wents wrong when clicking the OK button to connect and I had to change it back to have it working.

Maybe you can confirm it.

Best regards!

jsorg71:
Yes, that is my fault, I'll fix it.
I need to add another option hideloginwindow for when user does not want the login window

metalefty:
@jsorg71 Have this been fixed?

cocoon:
No it's still in there:
https://github.com/neutrinolabs/xrdp/blob/devel/xrdp/xrdp_wm.c#L569

metalefty:
OK, thanks. This is still marked as TODO.

PR #876, opened 2017-09-14
fix misuse of hidelogwindow

cocoon:
fix black screen if hidelogwindow=yes
https://github.com/neutrinolabs/xrdp/issues/414

### DIFF
--- a/xrdp/xrdp_wm.c
+++ b/xrdp/xrdp_wm.c
@@ -567,7 +567,7 @@ xrdp_wm_init(struct xrdp_wm *self)
     xrdp_wm_load_static_pointers(self);
     self->screen->bg_color = self->xrdp_config->cfg_globals.ls_top_window_bg_color;
 
-    if (self->session->client_info->rdp_autologin || self->hide_log_window)
+    if (self->session->client_info->rdp_autologin)
     {
         /*
          * NOTE: this should eventually be accessed from self->xrdp_config


xrdp/xrdp_wm.c
```diff
@@ -567,7 +567,10 @@ xrdp_wm_init(struct xrdp_wm *self)
     xrdp_wm_load_static_pointers(self);
     self->screen->bg_color = self->xrdp_config->cfg_globals.ls_top_window_bg_color;
 
-    if (self->session->client_info->rdp_autologin || self->hide_log_window)
+    /* fix black screen if hidelogwindow=yes in xrdp.ini, only the log window should be hidden. not everything */
+    /* https://github.com/neutrinolabs/xrdp/issues/414 */
+    /*if (self->session->client_info->rdp_autologin || self->hide_log_window)*/
```
metalefty: The fix seems good but don't leave the old code in comment. You can just remove it.

cocoon: should I remove all other comments or leave them in place?

metalefty: I think no comments needed there.

metalefty:
Can you squash commits into one commit and include "Fixes #414." like 06ce0d7? Then #414 will be automatically closed when this PR merged.

Issue #719, opened 2017-03-31
[crash] [Systemd] neutrinordp cannot get homedir from env


cocoon:
I had a crash on Ubuntu 16.04.02 when running xrdp compiled with neutrinordp with systemd and then tried to connect using the nutrinordp modul to any ip.

I compiled with debugging enabled
```./configure --enable-neutrinordp --sbindir=/usr/sbin --enable-xrdpdebug```

modified the crash dump in /var/crash/_usr_sbin_xrdp.0.crash added a line
```Package:xrdp```

so that apport-retrace could use the crashdump:
```apport-retrace -s /var/crash/_usr_sbin_xrdp.0.crash```

and that lead me to:
https://github.com/neutrinolabs/NeutrinoRDP/blob/devel/libfreerdp-utils/file.c#L81

Then I found that others had this problem in the past, too:
https://glyptodon.org/jira/si/jira.issueviews:issue-html/GUAC-154/GUAC-154.html

```
Comment by Michael Jumper [ 2012-06-19 ]

This actually appears fixed upstream. Checking the latest from FreeRDP's git repositories, they now default to / if $HOME is unavailable.
```

I made a simple test and added an additional NULL check and set it to "/"
```c
        if(settings->home_path == NULL)
        {
                settings->home_path = "/";
        }
```

And after that it was not crashing anymore.

An additional fix is to set a username in the service file, then it is able to get the HOME var:

```
vi /lib/systemd/system/xrdp.service

[Service]
Type=forking
PIDFile=/var/run/xrdp.pid
EnvironmentFile=-/etc/sysconfig/xrdp
EnvironmentFile=-/etc/default/xrdp
User=root
...
```

then reload:
systemctl daemon-reload

just for documentation a helpful command:
```systemctl show xrdp.service```

metalefty:
I think the issue should be fixed neutrinordp side.

metalefty:
Solved by the fix on NeutrinoRDP.

PR #877, opened 2017-09-14
fix crash neutrinordp missing home env

cocoon:
https://github.com/neutrinolabs/xrdp/issues/719

One possible way to fix the crash is to add User=xxx in xrdp.service-file

Other fix would be in neutrinordp to set a fallsback path if no home env exists.

### DIFF
--- a/instfiles/xrdp.service.in
+++ b/instfiles/xrdp.service.in
@@ -6,6 +6,7 @@ After=network.target xrdp-sesman.service
 
 [Service]
 Type=forking
+User=root
 PIDFile=@localstatedir@/run/xrdp.pid
 EnvironmentFile=-@sysconfdir@/sysconfig/xrdp
 EnvironmentFile=-@sysconfdir@/default/xrdp


metalefty:
Actually, this will avoid the crash but I don't think this is a fix. IMO, this should be fixed at NeutrinoRDP.

metalefty:
neutrinolabs/NeutrinoRDP#12 is a proposed fix on the NeutrinoRDP side.

cocoon:
"IMO, this should be fixed at NeutrinoRDP".
I totally agree. NeutrinoRDP fix is a must.

What advantage/disadvantage do we get if the User=root line is added or not to the service file?
Is it planned to be running as a different user in the future?

speidy:
seems like it fixed and merged in neutrinolabs/NeutrinoRDP#12
Can you close this PR ?

metalefty:
>  Is it planned to be running as a different user in the future?

Yes. root privilege is not necessarily for xrdp. Debian runs it user privilege.

metalefty:
Closing.

PR #895, opened 2017-09-28
xrdp: get port from configfile in access_control()

### LINKED ISSUE #894: xrdp hard coded port for sesman, config settings are not fully read
After changing Globals/ListenPort in sesman.ini to a number other than 3350, clients show an error:
`"Reply from access control: Error connecting to PAM"`
and the next line is logged to xrdp.log:
`[20170926-10:34:23] [ERROR] Failure connecting to socket sesman`

It seems the issue is caused due to the sesman port hard coded to 3350 in xrdp/xrdp_mm.c, access_control(). My colleague  @osalvadorvilardaga has a patch for it which works fine in test environments.

osalvadorvilardaga:
This fixes #894

access_control() does not read the port from the configuration file and it uses a hardcoded port instead.
So,  in case we use another port which is not the default one(3350), we are going to fail while connecting.

### DIFF
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -1471,11 +1471,13 @@ access_control(char *username, char *password, char *srv)
     unsigned long size;
     int index;
     int socket = g_tcp_socket();
+    char port[8];
 
     if (socket != -1)
     {
+        xrdp_mm_get_sesman_port(port, sizeof(port));
         /* we use a blocking socket here */
-        reply = g_tcp_connect(socket, srv, "3350");
+        reply = g_tcp_connect(socket, srv, port);
 
         if (reply == 0)
         {


xrdp/xrdp_mm.c
```diff
@@ -1471,11 +1471,13 @@ access_control(char *username, char *password, char *srv)
     unsigned long size;
     int index;
     int socket = g_tcp_socket();
+    char port[8];
 
     if (socket != -1)
     {
+	xrdp_mm_get_sesman_port(port, sizeof(port));
```
metalefty: Indent.

metalefty:
Good catch! Looks good to me.

osalvadorvilardaga:
Is there anything else I can do here?

jsorg71:
+1

There is more to do, sesman should use common/trans and we should use a unix domain socket for xrdp <-> sesman since they are run on the same machine.

osalvadorvilardaga:
@jsorg71 About the common/tranns, I guess you mean changing sesman_listen_test() and sesman_main_loop() functions to use trans_* functions.
But maybe we want to merge first this.

jsorg71:
Yes, for sure.

metalefty:
Thanks, I think there's no problem on this PR now.

metalefty:
Let's merge. Then think about using unix domain socket as Jay say.

PR #928, opened 2017-10-30
Properly support horizontal scroll wheel

### LINKED ISSUE #730: Support horizontal scroll wheel
I don't actually have a mouse with a horizontal scroll wheel, but I use X11 Wheel Emulation with a mouse button. It would be nice if the horizontal scroll events would be handled by xrdp. These are mouse buttons 6 and 7 I believe.

metalefty:
To support horizontal scroll wheel properly, 

* advertise capability for horizontal scroll wheel (TS_INPUT_MOUSE_HWHEEL)
* handle horizontal scroll wheel events as mouse events not only **extended** mouse events
* make mouse button 6 and 7 should act as well as button 4 and 5 (vertical scrolling)

Additionally, I've made some constifications and cleanup.

Tested and working with:
* FreeRDP 2.0.0-rc0
* Windows 10 1709 17025.1000 mstsc.exe
* MSRDP Beta for Mac Version: 8.2.37 (Build: 808)

Not working with:
* MSRDP for Mac Version 8.0.43 (Build 27325)

because horizontal scroll events are sent as vertical scroll. xrdp never receive horizontal scroll.

Closes #730.


common/xrdp_constants.h
```diff
@@ -190,13 +190,21 @@
...
-#define MOUSE_FLAG_BUTTON2             0x2000
-#define MOUSE_FLAG_BUTTON3             0x4000
-#define MOUSE_FLAG_BUTTON4             0x0280
-#define MOUSE_FLAG_BUTTON5             0x0380
-#define MOUSE_FLAG_DOWN                0x8000
+/* Mouse Event (MS-RDPBCGR 2.2.8.1.1.3.1.1.3) */
+#define PTRFLAGS_HWHEEL                0x0400
+#define PTRFLAGS_WHEEL                 0x0200
+#define PTRFLAGS_WHEEL_NEGATIVE        0x0100
+#define WheelRorationMask              0x01FF
```
speidy: Please follow the other defines styling (capitalizing and naming convention)

common/xrdp_constants.h
```diff
@@ -595,11 +603,16 @@
...
-#define  INPUT_FLAG_FASTPATH_INPUT2 0x0020
+/* Input Capability Set (MS-RDPBCGR 2.2.7.1.6) */
+#define  INPUT_FLAG_SCANCODES          0x0001
+#define  INPUT_FLAG_MOUSEX             0x0004
+#define  INPUT_FLAG_FASTPATH_INPUT     0x0008
+#define  INPUT_FLAG_UNICODE            0x0010
+#define  INPUT_FLAG_FASTPATH_INPUT2    0x0020
+#define  INPUT_FLAG_UNUSED1            0x0040
+#define  INPUT_FLAG_UNUSED2            0x0080
+#define  TS_INPUT_FLAG_MOUSE_HWHEEL    0x0100
```
speidy: TS_INPUT --> INPUT ?

xrdp/xrdp_wm.c
```diff
@@ -1279,21 +1281,23 @@ xrdp_wm_mouse_click(struct xrdp_wm *self, int x, int y, int but, int down)
...
-                }
-                else if (but == 6 && !down)
-                {
-                    self->mm->mod->mod_event(self->mm->mod, WM_BUTTON6UP, x, y, 0, 0);
-                }
-                if (but == 7 && down)
+
+                /* horizontal scroll */
+
+                if (but == 6)
```
speidy: Can you add some constants to describe the values 6 , 7 ?

common/xrdp_constants.h
metalefty: I brought these constants as-is from MS-RDPBCGR. I think this is odd but using the same name as MS will be clear. What would be the best?
https://msdn.microsoft.com/en-us/library/cc240586.aspx

common/xrdp_constants.h
metalefty: TS is correct. https://msdn.microsoft.com/en-us/library/cc240563.aspx

xrdp/xrdp_wm.c
metalefty: Simply the values indicates mouse button. The constants will be MOUSE_BUTTON1 ... MOUSE_BUTTON7. Is it fine?

common/xrdp_constants.h
speidy: I think the coding style should be consistent with the exisiting code, also there is a typo Roration -> Rotation

common/xrdp_constants.h
speidy: As we already avoid the TS_ prefix we should be consistent with the that.

xrdp/xrdp_wm.c
speidy: I don't familiar with this, if they have a known position you can call them by its position .

common/xrdp_constants.h
jsorg71: I think these are from the old T128 doc.  We can add update them and add TS_ prefix at a later time.

xrdp/xrdp_wm.c
jsorg71: Defines here don't help much, 6 means button 6.  BUTTON6 not much better then 6.

common/xrdp_constants.h
metalefty: Okay, I'll remove TS_ prefix at least from the constants I added. BTW, lots of TS_ prefixed constants are introduced around painter.

jsorg71: No, then leave it.  I'll change them all later.

metalefty: I get it.

common/xrdp_constants.h
metalefty: I've only fixed the typo at this time. Actually it is not used anywhere. There's only definition so far.

common/xrdp_constants.h
speidy: I'm fine with TS_ prefix, just lets change them all :)

jsorg71: I added an issue for this and assigned to me.
#931

xrdp/xrdp_wm.c
metalefty: Right.

metalefty:
Anything else?

metalefty:
Let's merge.

PR #930, opened 2017-10-31
Implements XRDP over vsock

jterry75:
1. Implements the ability to use AV_VSOCK for the transport rather than TCP.
2. Updates the ini file to be able to conditionally turn this feature on.

### DIFF
--- a/common/trans.c
+++ b/common/trans.c
@@ -792,7 +792,9 @@ trans_listen_address(struct trans *self, char *port, const char *address)
     {
         self->sck = g_tcp_socket();
         if (self->sck < 0)
+        {
             return 1;
+        }
 
         g_tcp_set_non_blocking(self->sck);
 
@@ -814,7 +816,9 @@ trans_listen_address(struct trans *self, char *port, const char *address)
 
         self->sck = g_tcp_local_socket();
         if (self->sck < 0)
+        {
             return 1;
+        }
 
         g_tcp_set_non_blocking(self->sck);
 
@@ -831,6 +835,28 @@ trans_listen_address(struct trans *self, char *port, const char *address)
             }
         }
     }
+#if defined(XRDP_ENABLE_VSOCK)
+    else if (self->mode == TRANS_MODE_VSOCK) /* vsock socket */
+    {
+        self->sck = g_vsock_socket();
+        if (self->sck < 0)
+        {
+            return 1;
+        }
+
+        g_tcp_set_non_blocking(self->sck);
+
+        if (g_vsock_bind(self->sck, port) == 0)
+        {
+            if (g_tcp_listen(self->sck) == 0)
+            {
+                self->status = TRANS_STATUS_UP; /* ok */
+                self->type1 = TRANS_TYPE_LISTENER; /* listener */
+                return 0;
+            }
+        }
+    }
+#endif
 
     return 1;
 }
--- a/xrdp/xrdp_listen.c
+++ b/xrdp/xrdp_listen.c
@@ -153,6 +153,7 @@ static int
 xrdp_listen_get_port_address(char *port, int port_bytes,
                              char *address, int address_bytes,
                              int *tcp_nodelay, int *tcp_keepalive,
+                             int *mode,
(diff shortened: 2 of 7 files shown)


jsorg71:
We are going to need an autotools/configure check for this.

jterry75:
@jsorg71 - Do you mean that you want this to be conditionally compiled only when someone runs `./configure --enable-vsock` or something?

jsorg71:
@jterry75 yes, as you can see, travis-ci fails because it does not have that header

metalefty:
xrdp is not used only on GNU/Linux. Not many but certain users use xrdp on FreeBSD or macOS or else.  Conditional switches are fine for such OSes.

BTW, disabling vsock build by not giving `--enable-vsock` means vsock is not tested by Travis CI.  For the time being, that is OK but we need to make some changes on travis config someday.

xrdp/xrdp_listen.c
```diff
@@ -204,7 +206,16 @@ xrdp_listen_get_port_address(char *port, int port_bytes,
...
                         }
                     }
-
+#if defined(XRDP_ENABLE_VSOCK)
+                    if (g_strcasecmp(val, "use_vsock") == 0)
+                    {
+                        val = (char *)list_get_item(values, index);
+                        if (g_text2bool(val) == 1)
+                        {
+                            *mode = TRANS_MODE_VSOCK;
```
metalefty: Let's be friendly to users. The message is just for an example.

```C
if (g_text2bool(val) == 1)
{
    log_message(LOG_LEVEL_WARNING, "use_vsock is enabled but not built with vsock option.");
#if defined(XRDP_ENABLE_VSOCK)
    *mode = TRANS_MODE_VSOCK;
#endif
}
```

jterry75: Good feedback. Fixed.

jterry75:
@metalefty - I think you are saying to add `--enable-vsock` to the `max_amd64_conf` section? I'm not sure how to make it distro specific to only use this option on linux?

jterry75:
@metalefty - Is that what you are asking me to do? If so I can make the change and update but just wanted to make sure that's what you mean.

jsorg71:
@jterry75 I think he is saying we need to change that(travis) some day.
I always prefer to have less #ifdefs, I'll do a PR for my idea to simplify.

metalefty:
@jterry75 Sorry for late reply. I'm not asking you to do that. I meant `--enable-vsock` also should be pass the Travis build test. Currently, build environment doesn't have `linux/vm_sockets.h`. I have some thoughts to improve travis build test so I'll change the build test after merging this.

common/trans.c
```diff
@@ -831,6 +831,26 @@ trans_listen_address(struct trans *self, char *port, const char *address)
             }
         }
     }
+#if defined(XRDP_ENABLE_VSOCK)
+    else if (self->mode == TRANS_MODE_VSOCK) /* vsock socket */
+    {
+        self->sck = g_vsock_socket();
+        if (self->sck < 0)
+            return 1;
```
metalefty: We always use braces for if statements. See `coding_style.md`.

jterry75: Good catch. Updated the other locations that I based this off of.

jterry75:
@metalefty - Updated based on your suggestions leaving out the Travis changes.

@jsorg71 - I am happy to leave all the functions in and simply `#ifdef` in the implementations in `os_calls.c` to avoid the `#ifdef` in other files. Would you like me to do that? I figured that it was better to remove the code associated with the `--enable-vsock` so it was a compile time error rather than runtime but I'm fine either way.

jsorg71:
@jterry75 , lets use #935 for this

jsorg71:
I think we can delete this.  Sorry, I didn't want to create a new PR, I wanted to do a PR to your PR.  I guess I can't do that ??

metalefty:
You can't do a PR to another PR but you can do a PR to other people's branch. It is almost as same as doing a PR to another PR.

metalefty:
Merged as #935.

jterry75:
Thanks!

PR #935, opened 2017-11-07
vsock

jsorg71:
Always have the os_calls vsock functions defined.   More consistent with other function.
Print an error if vsock is requested but not compiled.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1011,7 +1011,7 @@ g_sck_vsock_bind(int sck, const char *port)
 #if defined(XRDP_ENABLE_VSOCK)
     struct sockaddr_vm s;
 
-    memset(&s, 0, sizeof(struct sockaddr_vm));
+    g_memset(&s, 0, sizeof(struct sockaddr_vm));
     s.svm_family = AF_VSOCK;
     s.svm_port = atoi(port);
     s.svm_cid = VMADDR_CID_ANY;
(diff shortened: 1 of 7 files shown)


common/os_calls.c
```diff
@@ -990,6 +1004,24 @@ g_sck_local_bind(int sck, const char *port)
...
 }
 
+/*****************************************************************************/
+int
+g_sck_vsock_bind(int sck, const char *port)
+{
+#if defined(XRDP_ENABLE_VSOCK)
+    struct sockaddr_vm s;
+
+    memset(&s, 0, sizeof(struct sockaddr_vm));
```
metalefty: g_memset?

metalefty:
Maybe I should comment on Jay's change here. Please forget.

common/os_calls.c
```diff
@@ -583,6 +586,17 @@ g_sck_local_socket(void)
...
 
 /*****************************************************************************/
+int
+g_sck_vsock_socket(void)
+{
+#if defined(XRDP_ENABLE_VSOCK)
+    return socket(PF_VSOCK, SOCK_STREAM, 0);
+#else
+    return 0;
+#endif
```
hardening: should be -1

jsorg71: Yea, xrdp currently treats sockets < 1 as invalid.  This should be changed in many places.  I'll add an issue for this.

jsorg71: I add #938 for this

PR #936, opened 2017-11-07
sound: g_bbuf_size should be determined after codec is determined

### LINKED ISSUE #910: AAC audio redirection is not smooth with Windows client
Other codecs such as MP3 or PCM are working fine with Windows client.

AAC is tested with NeutrinoRDP but not with Windows (mstsc).

speidy said: 
> might be related to the format settings

metalefty:
Fixes #910, if both client and server support more than one codecs,
g_bbuf_size might be set to wrong value.

There might be a better fix but this certainly fixes the issue.


jsorg71:
+1

metalefty:
The diff is simple enough, let's merge.

Issue #931, opened 2017-11-01
Add TS_ to INPUT_* defines in xrdp_constants.h


jsorg71:
As #928 suggests

metalefty:
Closing as #946 has been closed.

PR #946, opened 2017-11-13
add TS_ to INPUT constants in xrdp_constants.h

jsorg71:


jsorg71:
fix for #931

metalefty:
I thought you were going to respect the original document. If the original constant has TS_ prefix and the prefix removed in xrdp, resurrect the prefix. Am I misunderstood?

jsorg71:
Yes, so we don't need this PR.  If we're all good,  I'll delete this after day or so and close #931

metalefty:
BTW, I have some thoughts on constants.  I'll raise a new issue and let's discuss.

PR #948, opened 2017-11-14
Implements the accept/close logic for vsock

jterry75:
Looks like I forgot to add this in the original commit. This now will avoid logging messages about unknown connections on unknown families and give better output.

Ex: "AF_VSOCK connection received from cid: -1 port: 3389"

> -1 for clarity:
> ```C
> #define VMADDR_CID_ANY -1U
> ```

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -670,6 +670,9 @@ g_sck_close(int sck)
 #if defined(XRDP_ENABLE_IPV6)
         struct sockaddr_in6 sock_addr_in6;
 #endif
+#if defined(XRDP_ENABLE_VSOCK)
+        struct sockaddr_vm sock_addr_vm;
+#endif
     } sock_info;
     socklen_t sock_len = sizeof(sock_info);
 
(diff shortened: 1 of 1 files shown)


speidy:
Hi @jterry75 
How can we test xrdp over vsock ? 

Thx.

jterry75:
@speidy - [Simple VSOCK setup for QEMU](https://gist.github.com/mcastelino/9a57d00ccf245b98de2129f0efe39857)

I found this link that might be helpful. The scenario here would be to directly remote to a linux VM hosted in QEMU that might not have networking. IE: The XRDP channel to the guest can be done over VSOCK between the host/guest. The easiest way to do this is to create a VM with QEMU and enable VSOCK as your RDP client destination (using socat or others) and then set `use_vsock=true` in the `xrdp.ini` file in the guest. (To change the default `port=3389` you can also update the `xrdp.ini`). Now you can start connecting to XRDP over the host/guest channel and send/receive connections.

jsorg71:
+1

PR #956, opened 2017-11-21
common: avoid 100% cpu on ssl accept, can be fake client

jsorg71:


alexbers:
I can confirm that 100% load is fixed after applying this patch.

metalefty:
+1

PR #961, opened 2017-11-24
Constify color depth and capability set

metalefty:
Define capset constants in hex as MS-RDPBCGR does. Easier to compare with MS documents.

### DIFF
--- a/libxrdp/xrdp_caps.c
+++ b/libxrdp/xrdp_caps.c
@@ -610,89 +610,89 @@ xrdp_caps_process_confirm_active(struct xrdp_rdp *self, struct stream *s)
         len -= 4;
         switch (type)
         {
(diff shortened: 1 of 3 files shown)


libxrdp/xrdp_caps.c
```diff
@@ -610,90 +610,90 @@ xrdp_caps_process_confirm_active(struct xrdp_rdp *self, struct stream *s)
         len -= 4;
         switch (type)
         {
-            case RDP_CAPSET_GENERAL: /* 1 */
+            case RDP_CAPSET_GENERAL: /* 0x01 */
```
speidy: We don't need to have the comments here now

metalefty:
Please check decimal to hex conversion. I've already checked but to make it doubly sure.

metalefty:
My colleague has checked decimal to hex conversion.

PR #972, opened 2017-12-13
Add Finnish rdp_layout_ definition

rkantos:
Add the missing? Finnish rdp_layout_ definitions. Makes my install work immediately with Finnish keyboard layout. (same as Swedish)


metalefty:
👍 Good catch!

Issue #923, opened 2017-10-24
[BUG/MITIGATION] XRDP/Xorg not starting correctly.


Hacksawfred3232:
This i don't think classifies as a pull request. But this is a mitigation to solve the problem concerning XRDP not starting correctly.

When connecting to RDP on a computer running with XRDP, it does not start the Xorg desktop program for the connecting RDP session.
This is present in the log:

```
[20171024-12:35:10] [INFO ] A connection received from ::1 port 51480
[20171024-12:35:11] [INFO ] ++ created session (access granted): username pi, ip 0.0.0.0:35408 - socket: 12
[20171024-12:35:11] [INFO ] starting Xorg session...
[20171024-12:35:11] [DEBUG] Closed socket 9 (AF_INET6 :: port 5912)
[20171024-12:35:11] [DEBUG] Closed socket 9 (AF_INET6 :: port 6012)
[20171024-12:35:11] [DEBUG] Closed socket 9 (AF_INET6 :: port 6212)
[20171024-12:35:11] [DEBUG] Closed socket 7 (AF_INET6 ::1 port 3350)
[20171024-12:35:11] [DEBUG] Closed socket 8 (AF_INET6 ::1 port 3350)
[20171024-12:35:11] [INFO ] Xorg :12 -auth .Xauthority -config xrdp/xorg.conf -noreset -nolisten tcp  
[20171024-12:35:21] [ERROR] X server for display 12 startup timeout
[20171024-12:35:21] [INFO ] starting xrdp-sessvc - xpid=18021 - wmpid=18020
[20171024-12:35:21] [ERROR] X server for display 12 startup timeout
[20171024-12:35:21] [ERROR] another Xserver might already be active on display 12 - see log
[20171024-12:35:21] [DEBUG] aborting connection...
[20171024-12:35:21] [INFO ] ++ terminated session:  username pi, display :12.0, session_pid 18019, ip 0.0.0.0:35408 - socket: 12
```
Trying to run Xorg as the user xrdp (After modifying the passwd file to allow a shell) prompts this error
`/usr/lib/xorg/Xorg.wrap: Only console users are allowed to run the X server`
Normally this would not cause a problem. However on this version of XRDP, this causes the timeout error you see in the log.

MITIGATION: Due to the fact that i am still able to run this as root, i considered this to be a problem concerning who is allowed to run the Xserver on my system.
[...]

PR #974, opened 2017-12-14
sesman: comment on non-suid Xorg in sesman.ini

metalefty:
Pointed out by: #923


jwmullally:
LGTM

(As for actually picking a sensible default: as these are different for each distro its really up to them to patch the right value into sesman.ini at packaging time)

metalefty:
Yes, I expect each distro package maintainer to replace the parameter with the right value for each distro. The help text is mainly for the people installs xrdp from source.

metalefty:
Comment changes only, no logic change. Merging.

PR #979, opened 2017-12-20
Fix memory corruption introduced by CVE-2017-16927 fix.

Natureshadow:


speidy:
wow, good catch.
Thanks.

sesman/libscp/libscp_v0.c
```diff
@@ -226,7 +226,7 @@ scp_v0s_accept(struct SCP_CONNECTION *c, struct SCP_SESSION **s, int skipVchk)
 
         /* reading username */
         in_uint16_be(c->in_s, sz);
-        buf = g_new0(char, sz);
+        buf = g_new0(char, sz + 1);
         in_uint8a(c->in_s, buf, sz);
         buf[sz] = '\0';
```
speidy: I think this is redundant now. we can leave it for the sake of readability

metalefty:
Good catch. Let's ship it to v0.9.5.

Issue #919, opened 2017-10-18
Issues reported by static analysis tool


keithrob91:
Hi,

I recently reported a few issues raised by a static analysis tool on xorgxrdp, so I thought I would do the same for xrdp. As before, I hope this is OK? Apologies if any of them are false positives (I haven’t reported any that were obviously incorrect), and apologies for not providing any patches, unfortunately I’m snowed under doing other things. Finally, thanks for the good work on xrdp!

## Double free:
Calls to xrdp_wm_key() can free “self->popup_wnd” and within xrdp_wm_key_unicode() there are multiple calls to xrdp_wm_key() (depending on code path) which leads to the situation where the first call to xrdp_wm_key() frees the memory, and then then subsequent call does the same.

The actual free itself takes place in xrdp_bitmap_delete()

## Several resource leaks:
~~scp_v1c_get_session_list() – Leaks “ds” in the do/while loop if any of the error conditions occur after ds has been allocated and the function immediately returns~~

~~scp_v1c_mng_get_session_list() – Same as above~~

sesman_listen_test() – Leaks “sck” if either of the “error == 0” tests returns false

~~dev_redir_proc_query_dir_response() – Leaks “xinode”~~

~~xrdp_load_keyboard_layout() – Leaks “fd” if g_file_open() happens to return an fd equal to zero. The tool questioned whether the if test should actually be if (fd >= 0)~~

## Out of bands access:
inputSession() – Test function only? The line,

g_strncpy(s->locale, "it_IT  0123456789", 18);

causes an out of bands access as s->locale is size 18 and a ‘0’ is inserted in position 18

main() in genkeymap.c – The line,

e.keycode = xfree86_to_evdev[i-8];

causes an out of bounds read when i == 137

jsorg71:
Thanks for correcting this.  I'm not sure were the original came from.  It may have been that way because it stopped something else from crashing.  When originally written there was not many examples on RandR.

metalefty:
No feed back from the reporter, closing.

keithrob91:
Apologies, I've had zero chance to do anything about this, but definitely appreciate the work put in. As and when I do get a chance I will happily provide feedback!

PR #999, opened 2018-01-17
fix potential fd leak

metalefty:
In most cases, checking fd > 0 is not valid. open(2) returns -1 on
error, 0 on stdin, 1 on stdout, 2 on stderr, >2 . The border should be
between -1 and 0. Additionally, between 2 and 3.

Pointed out by: #919


jsorg71:
+1
Since we never leave fd 0 closed, this has never been a problem but good to fix.

PR #1000, opened 2018-01-17
fix for zombie processes sometimes appearing

jsorg71:


smithAchang:
what's the triggering condition? I have encounter this preoblem in my ubuntu18.04 LTS. Using TOP shell，sometime there will be a zombie process od xrdp

PR #1007, opened 2018-01-22
sesman: add hostname to vnc password file

jsorg71:


sesman/env.c
```diff
@@ -184,9 +194,8 @@ env_set_user(const char *username, char **passwd_file, int display,
                                         "password file %s", *passwd_file);
                             g_file_delete(*passwd_file);
                         }
-
-                        g_sprintf(*passwd_file, "%s/.vnc/sesman_%s_passwd:%d",
-                                  pw_dir, username, display);
+                        g_sprintf(*passwd_file, "%s/.vnc/sesman_%s_passwd:%d:%s",
+                                  pw_dir, username, display, hostname);
```
metalefty: bikeshed discussion 😜

It doesn't affect the fix at all but HOSTNAME DISPLAY order looks more natural for me. 
For example, `DISPLAY` environment variable sometimes includes hostname.

metalefty: @jsorg71  what about like this? just a cosmetic point.

```
g_snprintf("%s/.vnc/sesman_passwd-%s@%s:%d", pw_dir, username, hostname, display);
```

I'd like to ship this to v0.9.6 release.

jsorg71: sure, looks safe

metalefty:
Merge and I'll change the filename later.

PR #1008, opened 2018-01-23
pulse: bring include path fix from Debian

metalefty:
The original patch made by Dominik George:
https://anonscm.debian.org/cgit/pkg-remote/xrdp.git/commit/debian/patches/pulse-debian.patch?id=4158025abba8fc84fb8375c9cb3b9b2e206758e7

### DIFF
--- a/sesman/chansrv/pulse/module-xrdp-sink.c
+++ b/sesman/chansrv/pulse/module-xrdp-sink.c
@@ -68,7 +68,7 @@ typedef bool pa_bool_t;
 #endif
 
 #include "module-xrdp-sink-symdef.h"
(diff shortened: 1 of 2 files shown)


Natureshadow:
I don't think the patch is complete.

xrdp_socket.h is not on any include path referenced in the Makefile.

Did you make sure you understand the patch in Debian? xrdp_socket.h is copied to the package containing the pulse module code there.

metalefty:
Ah, I've certainly tested the build but there's a mistake. It was false negative.
xrdp is already installed to my build environment so `xrdp_sockets.h` is included from  `/usr/include`.

PR #1014, opened 2018-01-29
fix bug in common/log.c and libxrdp/libxrdp.c

daixj-shterm:


metalefty:
Please don't change permission. Leave it as-is.

daixj-shterm:
Sorry about that. I used beyond Compare to merge the code, maybe it change the permission.

metalefty:
Good catch. Now looks good to me. Thank you for the nice first contribution!

PR #1017, opened 2018-01-31
Added /etc/xrdp.command script for Mac OS

Vraiment:
This commit depends on my previous pull request.

It consists of only three changes:
* *instfiles/xrdp.command.in*: The template for the command file, based in *instfiles/xrdp.sh*
* *instfiles/Makefile.am*: To add the template for the command file
* *.gitignore*: To ignore the generated file

Everything else is from my previous pull request, as this commit depends on that one


Vraiment:
Need to verify a bug with the script, will reopen after fix

PR #1021, opened 2018-02-01
Added PAM support for Mac OS

Vraiment:
Mac OS supports PAM, so it makes sense to add support for it, at least I can login after this commit (there are more issues down the line, but this commit fixes that).

Some notes:
* I'm not sure at all if the file */etc/pam.d/authorization* is unique to Mac OS, but it seems to be standard for it (is available in both machines I use)
* The PAM configuration file I uploaded (*xrdp-sesman.macos*) is basically the same as Mac OS' default SSH's PAM configuration, I don't know how to fine tune it but it makes sense you need the same permissions as SSH

### DIFF
--- a/instfiles/pam.d/xrdp-sesman.macos
+++ b/instfiles/pam.d/xrdp-sesman.macos
@@ -1,0 +1,12 @@
+# xrdp-sesman: auth account password session
+# based on Apple's sshd PAM configuration
+auth       optional       pam_krb5.so use_kcminit
+auth       optional       pam_ntlm.so try_first_pass
+auth       optional       pam_mount.so try_first_pass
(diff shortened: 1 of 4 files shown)


Vraiment:
ping?

instfiles/pam.d/xrdp-sesman.macos
```diff
@@ -0,0 +1,11 @@
+# sshd: auth account password session
```
metalefty: xrdp-sesman rather than sshd

Vraiment: good catch

Vraiment:
Updated comments

PR #1036, opened 2018-02-08
Make xrdp.ini to be dynamic for each operating system

Vraiment:
Mac OS needs libraries to be `.dylib` instead of `.so`. By using the same logic available in _docs/man/Makefile.am_ this commit makes the _xrdp/xrdp.ini_ file to depend on the operating system and replace the libraries extensions where necessary

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ sesman/tools/xrdp-sesrun
 sesman/tools/xrdp-sestest
 sesman/tools/xrdp-xcon
 sesman/xrdp-sesman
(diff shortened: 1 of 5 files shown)


metalefty:
Good work! Related to this, pulse script path in `sesman.ini` should refer `@sysconfdir@`.
```
[SessionVariables]
PULSE_SCRIPT=/etc/xrdp/pulse/default.pa
```

Vraiment:
ping

metalefty:
I'm testing this.

PR #1037, opened 2018-02-09
Log certificate/privkey readability problem

c0r0n3r:


libxrdp/xrdp_iso.c
```diff
@@ -101,16 +101,21 @@ xrdp_iso_negotiate_security(struct xrdp_iso *self)
...
+                if (g_file_readable(client_info->certificate) &&
+                    g_file_readable(client_info->key_file))
+                {
+                    /* that's a patch since we don't support CredSSP for now */
+                    self->selectedProtocol = PROTOCOL_SSL;
+                }
+                else
+                {
+                    /* certificate or privkey is not readable */
+                    log_message(LOG_LEVEL_DEBUG, "No readable certificates or "
```
speidy: It should be `LOG_LEVEL_WARN`.
Can you adapt the error message to something like: can't read certificate file ...

speidy: btw, we should move this `g_file_readable` calls to the config parsing part i think.
it's definetly not `libxrdp` responsibility.

metalefty:
No feedback for a long time closing.

Issue #1033, opened 2018-02-06
tls_ciphers cannot contain "!" because it is a comment out symbol


metalefty:
Many people want to configure `tls_ciphers` for examle like this.

```
tls_ciphers=HIGH:!aNULL:!eNULL:!EXPORT:!RC4
```

However, `!` is treated as a comment out symbol. That line will be interpreted as folows:
```
tls_ciphers=HIGH:
```

Neither
```
tls_ciphers="HIGH:!aNULL"
```
nor
```
tls_ciphers=HIGH:\!aNULL"
```
won't help as config file parser is not very smart. This may cause `tls_ciphers` unintended state. 

Removing `!` from comment out symbols is the easiest way to fix. Do you mind if I remove it?
`!` is not used anywhere in default config. Should we make the parser smarter?

PR #1040, opened 2018-02-13
Do not treat `!` as a comment out symbol

metalefty:
The simplest fix for #1033.


metalefty:
@jsorg71 @speidy If no objections, I'll apply this fix for the time being. Ideally, the config parser should be smarter.

jsorg71:
+1

PR #1052, opened 2018-03-03
common: temporarily disable DHE

metalefty:
until make it possible to use generated DH parameters per installation.

speidy concern about using static DH parameters:
https://github.com/neutrinolabs/xrdp/commit/1690950cc89220366a4dea0495a6948b2f248c33#r27897446

### DIFF
--- a/common/ssl_calls.c
+++ b/common/ssl_calls.c
@@ -492,6 +492,7 @@ see also
  * https://wiki.openssl.org/index.php/Diffie-Hellman_parameters
  * https://wiki.openssl.org/index.php/Manual:SSL_CTX_set_tmp_dh_callback(3)
 */
+#if 0 /* temprarily disable DHE until make DH parameters not static */
 DH *get_dh2236()
 {
     static unsigned char dh2236_p[] = {
@@ -538,6 +539,7 @@ DH *get_dh2236()
     }
     return dh;
 }
+#endif
 
 /*****************************************************************************/
 struct ssl_tls *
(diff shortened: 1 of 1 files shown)


speidy:
Lets merge this for now, it also breaks compatibility becuase of OpenSSL 1.1.0 API changes

speidy:
oh and we should `#if 0` the function `get_dh2236()` as well so xrdp builds ok with newer gcc.

metalefty:
Okay, will do it  later.

enricotagliavini:
SSL_CTX_set_dh_auto is a good idea for openssl 1.1.0 and newer, very simple, no user interaction needed. But I don't get what do you mean with "add a backport". Do you mean use it with older openssl versions?

Also a small note: is SSL_CTX_set_dh_auto just going to use precomputed keys anyway? You cannot generate DH keys on the fly, it takes too long. To my best understanding even Apache httpd mod_ssl uses the builtin openssl per-generated keys, when a dhparam is not provided (note: it is provided inside the cert file, which basically means almost nobody even knows it's there). See

https://github.com/apache/httpd/blob/2b9e9b4c4226c22d9f5c489661507e7547de051c/modules/ssl/ssl_engine_init.c#L108

metalefty:
Regarding the next release, I'm going to disable non-EC DHE because speidy concerns about static DH parameters. I think static DH params are OK as I mentioned before [1] but not 100% sure. 

Also, I agree with enricotagliavini overall. To make per-install DH params will resolve concern about static DH params but I don't think spending time on that is worth the effort. I don't object doing that but I don't because ECDHE is enough at least in my case. 

[1] https://github.com/neutrinolabs/xrdp/commit/1690950cc89220366a4dea0495a6948b2f248c33#r27899026

PR #1053, opened 2018-03-05
Update xrdp_keyboard.ini to handle Swiss French

fbourqui:
Add the missing Swiss French rdp_layout_ definitions, ch(fr) in X11


metalefty:
Merged. Good work!

PR #1078, opened 2018-03-23
instfiles: Add pam.d config for arch linux.

boppbo:
system-auth didn't work for gnome under arch linux. 

According to https://wiki.archlinux.org/index.php/PAM#PAM_base-stack system-remote-login seemed to be the right choice. Works for gnome at least.


metalefty:
Also modify `configure.ac`  and `instfiles/pam.d/mkpamrules` to fully care Arch.

metalefty:
Ah, `mkpamrules` already modified.

boppbo:
I only found it in the description for the possible values. Was this the correct place? Furthermore the macos option was missing in the description. So I added it too

metalefty:
Yes, that's the place.

boppbo:
Anything else missing for merge?

PR #1082, opened 2018-03-26
Corrected endianness detection on ppc64el.

ferseiti:
In common/arch.h, the endianness detection considers all powerpc
architectures as big endian. Since that is not true for ppc64el, I
added a verification that checks other preprocessor macros, only for
ppc cases.

Signed-off-by: Fernando Seiti Furusato <ferseiti@gmail.com>

### DIFF
--- a/common/arch.h
+++ b/common/arch.h
@@ -63,8 +63,10 @@ typedef int bool_t;
 #endif
 
 #if !(defined(L_ENDIAN) || defined(B_ENDIAN))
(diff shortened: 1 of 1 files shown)


ferseiti:
This patch corrects an error "endian wrong, edit arch.h" when running xrdp on a ppc64el machine.

metalefty:
Good catch. I remember PPC is bi-endian. Thanks!

PR #1097, opened 2018-04-12
xrdp_encoder leaks

jsorg71:
when running xrdp in thread mode, these leaks can add up


jsorg71:
opps, ok I'll delete my branch

metalefty:
And the other commit is also merged as acd05f365f885e8a4aab23d23ba4ce48b6f31f71.

PR #1098, opened 2018-04-12
fix for zombie processes sometimes appearing

jsorg71:


metalefty:
Already merged as f763cb37885b02c310aa906147669a29b99405dc.

Related to #987. That's the disadvantage of "Rebase and merge". The merged commit and the comit in your local branch has different commit hash. We cannot know if the commit merged or not by commit hash. This is the one of the reasons why I'd like to turn back to merge commits.

jsorg71:
That is a good point although my house keeping is to blame here.
I'm ok with either way.

metalefty:
Maybe you're warned like this when you delete your branch?

```
error: The branch 'foobar' is not fully merged.
If you are sure you want to delete it, run 'git branch -D foobar'.
```

This is the side effect of rebase merging. git also cannot detect if the branch is merged as well as human because of different commit hashes. 

@jsorg71 could you re-enable creating merge commit in repository config?

jsorg71:
Ok done.  I'll miss the clean easy to follow history.

PR #1103, opened 2018-04-17
docs: '!' is no longer a comment out symbol

metalefty:
Removed in #1040.


metalefty:
Simple enough, merging.

Issue #1106, opened 2018-04-18
Blue screen after login


nopyhe:
xrdp log:
```
[20180418-21:14:24] [DEBUG] Testing if xrdp can listen on 0.0.0.0 port 3389.
[20180418-21:14:24] [DEBUG] Closed socket 7 (AF_INET 0.0.0.0:3389)
[20180418-21:14:26] [INFO ] starting xrdp with pid 24795
[20180418-21:14:26] [INFO ] listening to port 3389 on 0.0.0.0
[20180418-21:15:12] [INFO ] Socket 12: AF_INET connection received from 33.33.33.33 port 55792
[20180418-21:15:12] [DEBUG] Closed socket 12 (AF_INET 192.168.1.2:3389)
[20180418-21:15:12] [DEBUG] Closed socket 11 (AF_INET 0.0.0.0:3389)
[20180418-21:15:12] [INFO ] Using default X.509 certificate: /etc/xrdp/cert.pem
[20180418-21:15:13] [INFO ] Using default X.509 key file: /etc/xrdp/key.pem
[20180418-21:15:13] [DEBUG] TLSv1.2 enabled
[20180418-21:15:13] [DEBUG] TLSv1.1 enabled
[20180418-21:15:13] [DEBUG] TLSv1 enabled
[20180418-21:15:13] [DEBUG] Security layer: requested 11, selected 1
[20180418-21:15:13] [INFO ] connected client computer name: PC-1.
[20180418-21:15:13] [INFO ] TLS connection established from 33.33.33.33 port 55792: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384
[20180418-21:15:13] [DEBUG] xrdp_0000614d_wm_login_mode_event_00000001
[20180418-21:15:13] [INFO ] Loading keymap file /etc/xrdp/km-00000409.ini
[20180418-21:15:13] [WARN ] local keymap file for 0x00000409 found and doesn't match built in keymap, using local keymap file
[20180418-21:15:25] [DEBUG] xrdp_wm_log_msg: connecting to sesman ip 127.0.0.1 port 3350
[20180418-21:15:26] [INFO ] xrdp_wm_log_msg: sesman connect ok
[20180418-21:15:26] [DEBUG] xrdp_wm_log_msg: sending login info to session manager, please wait...
[20180418-21:15:26] [DEBUG] return value from xrdp_mm_connect 0
[...]

PR #1107, opened 2018-04-18
sesman: Update Xorg help comment for Arch Linux

metalefty:
As reported in #1106, Arch Linux looks to me like using Xorg wrapper now.


metalefty:
@nopyhe has confirmed `param=/usr/lib/xorg-server/Xorg` works.

PR #1108, opened 2018-04-19
sesrun support start other session based on session_code argument

aiden0z:
Support sesrun to start different session based on session_code argument, like `sesrun <server> <username> <password> <width> <height> <bpp> <session_cod>`.

Session code 0 for Xvnc, 10 for X11RDP, 20 for Xorg.

### DIFF
--- a/sesman/tools/sesrun.c
+++ b/sesman/tools/sesrun.c
@@ -46,6 +46,7 @@ main(int argc, char **argv)
     int height;
     int bpp;
     int display;
+    int session_code;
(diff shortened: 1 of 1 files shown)


metalefty:
Good work but can you remove the merge commit?

aiden0z:
removed

metalefty:
Thanks for your contribution. BTW, I'm planning to improve sesrun like we can do:
```
xrdp-sesrun --user foobar --type xvnc|xorgxrdp
```

PR #1120, opened 2018-05-03
Copy the PAM session environment for the reconnect script

matt335672:
In some environments, the reconnect script can benefit from access to environment variables set during the PAM session creation. For example, in a Kerberos-based environment, the reconnect script will need KRB5CCNAME to be set to make use of network services.


metalefty:
Thanks, I'll look at it. BTW, please always respect our coding standard.

matt335672:
squashed and rebased.

PR #1134, opened 2018-05-24
add ipv6only to configure echo and add configure parameter to xrdp -h…

jsorg71:
… output

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,7 +58,7 @@ AC_ARG_WITH([socketdir],
 AC_SUBST([socketdir], [$with_socketdir])
 
 AC_ARG_WITH([systemdsystemunitdir],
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -396,3 +397,4 @@ echo "  exec_prefix             $exec_prefix"
 echo "  libdir                  $libdir"
 echo "  bindir                  $bindir"
 echo "  sysconfdir              $sysconfdir"
+echo "#define CONFIGURE_STRING \"$@\"" > configure_params.h
```
metalefty: this doesn't work out-of-tree building. I'm fixing this.

metalefty: It was my misunderstanding but I'd like to make it more pretty.

Issue #1181, opened 2018-07-31
building xrdp with pulseaudio support for ubuntu 18.04 fails


bmullan:
I have an ubuntu 18.04 VM.

I followed the instructions found here:
https://github.com/neutrinolabs/xrdp/wiki/How-to-set-up-audio-redirection

Everything seems to install correctly until 

> Finally, let's make. You'll have two .so files module-xrdp-sink.so and module-xrdp-source.so.
> 
> cd ~/xrdp/sesman/chansrv/pulse
> make PULSE_DIR="~/pulseaudio-11.1"

At that point I get the following error...

> 
> :~/xrdp/sesman/chansrv/pulse$ make PULSE_DIR="~/pulseaudio-11.1"
> cc -Wall -O2 -I~/pulseaudio-11.1 -I~/pulseaudio-11.1/src -DHAVE_CONFIG_H -fPIC   -c -o module-xrdp-sink.o module-xrdp-sink.c
> module-xrdp-sink.c:25:10: fatal error: config.h: No such file or directory
>  #include <config.h>
>           ^~~~~~~~~~
> compilation terminated.
> <builtin>: recipe for target 'module-xrdp-sink.o' failed
> make: *** [module-xrdp-sink.o] Error 1

NOTE: the contents of the following directory... there is no config.h file ??

> ~/xrdp/sesman/chansrv/pulse$ ls
> Makefile            module-xrdp-sink-symdef.h  module-xrdp-source-symdef.h
> module-xrdp-sink.c  module-xrdp-source.c       README.md

I have tried re-doing this several times just to make sure I didn't skip a step but always the same failure.

Are there files missing from the github ?

Any ideas would be welcome.

Brian

metalefty:
I confirmed it builds. Did you changed enviroment-dependent values to fit yours?

metalefty:
Where's your PULSE_DIR?

metalefty:
BTW, I  didn't say envirnment variables, environment-dependent values.

bmullan:
Sorry.. it's really late here.

I'll recheck any makefiles etc tomorrow.

metalefty:
The `config.h` being searched during the build is `config.h` at the top of the pulseaudio source directory. Set the correct PULSE_DIR then everything will work well.

shadoxx:
I can confirm this bug.
[...]

PR #1185, opened 2018-08-02
Corrected spacing between arguments to `cc` for Pulse chansrv Makefile

shadoxx:
This PR fixes https://github.com/neutrinolabs/xrdp/issues/1181 where the Pulseaudio Chanserv module for xrdp fails to compile due to "missing config.h" on Ubuntu 18.04.

This is caused by incorrect spacing between the `-I` parameter and argument to `cc` in the Makefile.

### DIFF
--- a/sesman/chansrv/pulse/Makefile
+++ b/sesman/chansrv/pulse/Makefile
@@ -4,7 +4,7 @@
 
 # change this to your pulseaudio source directory
 PULSE_DIR = /tmp/pulseaudio-10.0
-CFLAGS    = -Wall -O2 -I$(PULSE_DIR) -I$(PULSE_DIR)/src -DHAVE_CONFIG_H -fPIC
+CFLAGS    = -Wall -O2 -I $(PULSE_DIR) -I $(PULSE_DIR)/src -DHAVE_CONFIG_H -fPIC
 
 all: module-xrdp-sink.so module-xrdp-source.so
 


metalefty:
Can you provide a pointer of the issue whitespace is required after `-I`? Not placing whitespace after `-I` is not strange at all. Many people don't put a space after `-I`. If that is true, many software will be affected. Many software needs to change its Makefile. I don't think it makes sense.

Even if this commit fixes the issue, we should find out the true issue.

shadoxx:
> Can you provide a pointer of the issue whitespace is required after -I?
The syntax listed in the manpage for `cc` is `-I <dir>`. I understand that many people don't put a space after `-I`, but this is what fixed it.
Currently compiling this module under Ubuntu 18.04 is broken. This the what fixes it. :man_shrugging: I don't know what to tell you.

metalefty:
If Ubuntu is broken, Ubuntu should be fixed. This is a workaround to avoid the issue but not our issue.

shadoxx:
I've rechecked this again and was able to get this module to compile by specifying an absolute file system path in PULSE_DIR. If PULSE_DIR is set to a relative path (`~/pulseaudio-11.1`), `cc` will complain about missing `config.h`. However, if PULSE_DIR is set to something like `/home/user/pulseaudio-11.1` the compilation will succeed.
Still not sure why this is broken on Ubuntu 18.04, but considering the documentation suggests using a relative path to the pulse source tree, it's definitely a confusing issue.

metalefty:
Okay, I understand. If it's not a real path, for example `~` as , shell doesn't expand `-I~/somewhere` -> `-I/home/username/somewhere`. To expand `~` to full path, space after `-I` is needed.

In this context,  your PR makes sense 😄 .

PR #1187, opened 2018-08-05
Fix comparison of IPv6 addresses in session_get_bydata()

s-k2:
I encountered a bug in sesman when UBI-policy is used and xrdp was compiled with IPv6 enabled (Debian does this for the xrdp-package). In this case sesman does not compare the IP-addresses correctly and reuses sessions from different addresses.

The problem can be located in sesman/session.c, line 144 in master. The client's address and the session-item's address are only compared up to the first colon using g_strncmp_d() to ignore the port-number. This approach works well for IPv4 but not for IPv6.

I attached a quick-and-dirty-hack to solve that problem which compares the addresses up to the last colon. It works with either IPv4 or IPv6 but assumes that a colon followed by a port-number is part of the client_ip-string.

I am not quite sure if session.c is the right place for the two new functions I added, maybe you have a better idea where to put them (or know of a better approach)

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -89,6 +89,37 @@ dumpItemsToString(struct list *self, char *outstr, int len)
     return outstr ;
 }
 
+/* find the last occurrence of ch in str limited by n */
+static int
+last_index_of(const char *str, char ch, int n)
+{
+    int last_colon = -1;
+    int i;
+
+    for (i = 0; i < n && str[i] != 0; i++)
+    {
+        if (str[i] == ch)
+            last_colon = i;
+    }
+
+    return last_colon;
+}
+
+/* compare two strings containing ip-addresses and ports but ignore the latter */
+static int
+is_equal_ip_ignoring_port(const char *ip1, const char *ip2, size_t n)
+{
+    int last_colon_ip1 = last_index_of(ip1, ':', n);
+    int last_colon_ip2 = last_index_of(ip2, ':', n);
+
+    if (last_colon_ip1 != last_colon_ip2)
+        return 0;
+
+    if (last_colon_ip1 == -1)
+        last_colon_ip1 = n;
+
+    return strncmp(ip1, ip2, last_colon_ip1) == 0;
+}
 
 /******************************************************************************/
 struct session_item *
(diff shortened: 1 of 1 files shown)


metalefty:
Good catch.

sesman/session.c
```diff
@@ -89,6 +89,37 @@ dumpItemsToString(struct list *self, char *outstr, int len)
     return outstr ;
 }
 
+/* find the last occurrence of ch in str limited by n */
+static int
+last_index_of(const char *str, char ch, int n)
```
metalefty: This function should be placed `common/` because it has nothing sesman-specific.
It looks general helper function handling string.

s-k2:
Would os_calls.c be the right place for it?

sesman/session.c
```diff
@@ -89,6 +89,21 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
 }
 
+/* compare two strings containing ip-addresses and ports but ignore the latter */
+static int
+is_equal_ip_ignoring_port(const char *ip1, const char *ip2, size_t n)
+{
+    int last_colon_ip1 = g_last_ch_pos(ip1, ':', n);
+    int last_colon_ip2 = g_last_ch_pos(ip2, ':', n);
+
+    if (last_colon_ip1 != last_colon_ip2)
```
speidy: please always use brackets.

sesman/session.c
```diff
@@ -89,6 +89,21 @@ dumpItemsToString(struct list *self, char *outstr, int len)
...
+        return 0;
+
+    if (last_colon_ip1 == -1)
```
speidy: same here

sesman/session.c
```diff
@@ -89,6 +89,21 @@ dumpItemsToString(struct list *self, char *outstr, int len)
     return outstr ;
 }
 
+/* compare two strings containing ip-addresses and ports but ignore the latter */
+static int
+is_equal_ip_ignoring_port(const char *ip1, const char *ip2, size_t n)
```
speidy: `const size_t n` ?

common/os_calls.c
```diff
@@ -2822,6 +2822,22 @@ g_pos(const char *str, const char *to_find)
     return (pp - str);
 }
 
+/*****************************************************************************/
+int
+g_last_ch_pos(const char *str, char ch_to_find, int n)
```
speidy: you want `strrchr` from `string.h`

then do pointer arithmetic:
```c
e = strchr(string, 'e');
index = (int)(e - string);
```

sesman/session.c
s-k2: Ouch, I accidentally used size_t as I always do. But in xrdp's codebase is hardly used. Should I change it to an signed const int here, too?

common/os_calls.c
s-k2: No I can't do this. I would not like to trust on the input beeing correctly terminated with \0. Just like the the code I'd like to patch.
Unfortunately the C standard does not know of strnrchr() so I had to write my own...

metalefty:
`os_calls.c` is a possible candidate to put it on. but please don't hesitate to create a new file if needed.

PR #1198, opened 2018-08-27
sesman: pass pulse socket name via environment variable

metalefty:
See also: neutrinolabs/pulseaudio-module-xrdp#15


metalefty:
@speidy @jsorg71 any thoughts?

Issue #1202, opened 2018-09-02
xrdp-chansrv crashes on RHEL 7.5


savisko:
I experienced several time that copy&paste stops working after several days of open RemoteDesktop session. Eventually I've found that xrdp-chansrv process crashes. Last time I ran gdb on the  xrdp-chansrv process and I got it caught SIGSEGV signal:

```
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f30003ea700 (LWP 10957)]
0x000055691a872c26 in xfuse_create_file_in_xrdp_fs ()

(gdb) info threads 
  Id   Target Id         Frame 
* 2    Thread 0x7f30003ea700 (LWP 10957) "xrdp-chansrv" 0x000055691a872c26 in xfuse_create_file_in_xrdp_fs ()
  1    Thread 0x7f300a3dd840 (LWP 10938) "xrdp-chansrv" 0x00007f3008888c03 in select () from /lib64/libc.so.6

(gdb) bt
#0  0x000055691a872c26 in xfuse_create_file_in_xrdp_fs ()
#1  0x000055691a8735ff in xfuse_add_clip_dir_item ()
#2  0x000055691a878253 in clipboard_c2s_in_files ()
#3  0x000055691a87640a in clipboard_data_in ()
#4  0x000055691a86fa63 in my_trans_data_in ()
#5  0x00007f3009fed5f1 in trans_check_wait_objs () from /usr/lib64/xrdp/libcommon.so.0
#6  0x000055691a86ef22 in channel_thread_loop ()
#7  0x00007f3009dcadd5 in start_thread () from /lib64/libpthread.so.0
#8  0x00007f3008891b3d in clone () from /lib64/libc.so.6
(gdb) 


---Type <return> to continue, or q <return> to quit---
   0x000055691a872bf1 <+177>:   callq  0x55691a86d850 <__strcpy_chk@plt>
   0x000055691a872bf6 <+182>:   cmp    $0x4000,%r14d
   0x000055691a872bfd <+189>:   je     0x55691a872c50 <xfuse_create_file_in_xrdp_fs+272>
   0x000055691a872bff <+191>:   or     $0x1a4,%r14d
   0x000055691a872c06 <+198>:   movq   $0x0,0x20(%rbx)
   0x000055691a872c0e <+206>:   mov    %r14d,0x8(%rbx)
[...]

PR #1225, opened 2018-10-11
Fix crash of xrdp-chansrv process, issue #1202.

savisko:
In xfuse_create_file_in_xrdp_fs insure that xfuse_fs is properly initialized.


metalefty:
I'm looking at this. The changes are simple, looks good so far.

PR #1253, opened 2018-11-23
neutrinordp: don't enable remote_app if the INFO_RAIL flag is not set

daixj-shterm:
I'm trying to connect xrdp with Microsoft Remote Desktop for Mac 8(8.0.30030),
Nothing displayed after logged in with neutrinordp-any.This is the output:
```
lfreerdp_context_new: 0x1fae370
loaded module 'libxrdpneutrinordp.so' ok, interface size 408984, version 3
lfreerdp_pre_connect:
The colormap is all NULL
Railsupport !!!!!!!!!!!!!!!!!!
host 10.10.xx.xx port 3389
connected to 10.10.xx.xx:3389
null server certificate
lxrdp_connect: freerdp_connect returned 1
...
```
I think remote_app should not been enable if INFO_RAIL is not set.
MS-RDPERP 3.3.5.1.3 and 3.3.5.1.4

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -155,6 +155,7 @@ struct xrdp_client_info
 
   int no_orders_supported;
   int use_cache_glyph_v2;
+  int rail_enable;
 };
 
 #endif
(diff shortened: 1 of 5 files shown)


metalefty:
Good catch.

PR #1257, opened 2018-11-27
Fix build with FDK AAC v2

metalefty:
Original fix provided by takefu. See also:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=233556


metalefty:
Tested with FreeBSD 12 / fdk-aac 2.0.0. It properly sounds.

PR #1270, opened 2018-12-13
Solve problem with run bash script with parameters on xrdp. 

d-yacenko:
Problem with start bash script with parameters on xrdp as follows - **rdesktop -s "/my/bash/script.sh parameters" ip_address** . If you try this - xrdp run startwm.sh instead you script.
This patch fixed this issue.

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -556,9 +556,9 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                     if (s->program[0] != 0)
                     {
                         //g_execlp3(s->program, s->program, 0); // old start call
-                        if(strchr(s->program, ' '))
+                        if(g_strchr(s->program, ' '))
                         {
-                           log_message(LOG_LEVEL_DEBUG, "starting program with parameres: %s ", s->program);
+                           log_message(LOG_LEVEL_DEBUG, "starting program with parameters: %s ", s->program);
 			   char *params[] = { "sh", "-c", s->program, NULL};
                            g_execvp("/bin/sh", params);
                         }
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -45,7 +45,7 @@
 #ifndef PR_SET_NO_NEW_PRIVS
 #define PR_SET_NO_NEW_PRIVS 38
 #endif
-
+#include <string.h>
 
 extern unsigned char g_fixedkey[8];
 extern struct config_sesman *g_cfg; /* in sesman.c */
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                 {
                     if (s->program[0] != 0)
                     {
-                        g_execlp3(s->program, s->program, 0);
+			char * params=strchr(s->program+1,' ');
+			if(params)
+				params=strchr(s->program+1,' ');
+			if(params) {
+				*params='\0';
+				params++;
+			}
+			if(params && strlen(params)>0) {
+//			   log_message(LOG_LEVEL_DEBUG, "			program = %s params = %s",s->program,params);
+                           g_execlp3(s->program,"-c", params);
+			}
+			else{ 
+//			   log_message(LOG_LEVEL_DEBUG, "			program = %s ",s->program);
+                           g_execlp3(s->program, s->program, 0);
+			}
                         log_message(LOG_LEVEL_ALWAYS,
                                     "error starting program %s for user %s - pid %d",
                                     s->program, s->username, g_getpid());
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -45,7 +45,6 @@
 #ifndef PR_SET_NO_NEW_PRIVS
 #define PR_SET_NO_NEW_PRIVS 38
 #endif
-#include <string.h>
 
 extern unsigned char g_fixedkey[8];
 extern struct config_sesman *g_cfg; /* in sesman.c */
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -37,6 +37,10 @@
 #include <sys/prctl.h>
 #endif
 
+#ifdef HAVE_STRING_H
+#include <string.h>
+#endif
+
 #include "sesman.h"
 #include "libscp_types.h"
 #include "xauth.h"
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -547,21 +547,24 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                 {
                     if (s->program[0] != 0)
                     {
-			char * params=strchr(s->program+1,' ');
-			if(params)
-				params=strchr(s->program+1,' ');
-			if(params) {
-				*params='\0';
-				params++;
-			}
-			if(params && strlen(params)>0) {
-//			   log_message(LOG_LEVEL_DEBUG, "			program = %s params = %s",s->program,params);
-                           g_execlp3(s->program,"-c", params);
-			}
-			else{ 
-//			   log_message(LOG_LEVEL_DEBUG, "			program = %s ",s->program);
+                        char *params = strchr(s->program+1, ' ');
+                        if(params)
+                        {
+                            params = strchr(s->program+1, ' ');
+                        }
+                        if(params)
+                        {
+                            *params = '\0';
+                            params++;
+                        }
+                        if (params && strlen(params) > 0)
+                        { 
+                           g_execlp3(s->program, "-c", params);
+                        }
+                        else
+                        { 
                            g_execlp3(s->program, s->program, 0);
-			}
+                        }
                         log_message(LOG_LEVEL_ALWAYS,
                                     "error starting program %s for user %s - pid %d",
                                     s->program, s->username, g_getpid());
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -547,44 +547,20 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                         g_set_current_dir(s->directory);
                     }
                 }
+                for(int i=strlen(s->program)-1; i >= 0 && s->program[i] == ' '; i--)
+                {
+                    s->program[i]= '\0';
+                }
                 if (s->program != 0)
                 {
                     if (s->program[0] != 0)
                     {
                         //g_execlp3(s->program, s->program, 0); // old start call
-                        int len=256, counter=0;
-                        char *tmp=0, *buf=malloc(len), *last_delim=0, **arg_array=0;
-                        strncpy(buf,s->program,len-1);
-                        buf[len-1]='\0';
-                        //count space char in execution string
-                        tmp=buf;
-                        for(counter=0;*tmp;tmp++)
+                        if(strchr(s->program, ' '))
                         {
-                                if(' '==*tmp)
-                                {
-                                        counter++;
-                                        last_delim=tmp;
-                                }
-                        }
-                        counter += last_delim < (buf + len - 1);
-                        counter++; // add on to trailing zero element
-                        // create and fill array of string for execvp call
-                        arg_array = malloc(sizeof(char*) * counter);
-                        if (arg_array)
-                        {
-                                size_t idx  = 0;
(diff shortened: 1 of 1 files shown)


sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
                     {
-                        g_execlp3(s->program, s->program, 0);
+			char * params=strchr(s->program+1,' ');
+			if(params)
+				params=strchr(s->program+1,' ');
+			if(params) {
+				*params='\0';
+				params++;
+			}
+			if(params && strlen(params)>0) {
```
metalefty: braces

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                 {
                     if (s->program[0] != 0)
                     {
-                        g_execlp3(s->program, s->program, 0);
+			char * params=strchr(s->program+1,' ');
+			if(params)
+				params=strchr(s->program+1,' ');
+			if(params) {
```
metalefty: braces

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                 {
                     if (s->program[0] != 0)
                     {
-                        g_execlp3(s->program, s->program, 0);
+			char * params=strchr(s->program+1,' ');
+			if(params)
+				params=strchr(s->program+1,' ');
```
metalefty: space

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+			if(params) {
+				*params='\0';
+				params++;
+			}
+			if(params && strlen(params)>0) {
+//			   log_message(LOG_LEVEL_DEBUG, "			program = %s params = %s",s->program,params);
+                           g_execlp3(s->program,"-c", params);
+			}
+			else{
```
metalefty: braces

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+			char * params=strchr(s->program+1,' ');
+			if(params)
+				params=strchr(s->program+1,' ');
+			if(params) {
+				*params='\0';
+				params++;
+			}
+			if(params && strlen(params)>0) {
+//			   log_message(LOG_LEVEL_DEBUG, "			program = %s params = %s",s->program,params);
+                           g_execlp3(s->program,"-c", params);
```
metalefty: spaces

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+			}
+			else{ 
+//			   log_message(LOG_LEVEL_DEBUG, "			program = %s ",s->program);
```
metalefty: indent

sesman/session.c
```diff
@@ -547,7 +547,21 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+                           g_execlp3(s->program, s->program, 0);
```
metalefty: indent

sesman/session.c
metalefty: and space

sesman/session.c
metalefty: and space

metalefty:
You must read coding_style.md first of all.

d-yacenko:
style fixed

d-yacenko:
finally fixed

d-yacenko:
Hi.
Why was removed my code (to pass a parameter when running a script in a session)? It code works fine for me in production.

metalefty:
What do you mean? I didn't remove anything.

d-yacenko:
I am sory, I apparently misunderstood. I thought you removed my patch from the next build.
image

matt335672:
Hi @d-yacenko 

I've had a look at your code above, and it doesn't quite seem to be doing what I'd hoped it would be doing.

In particular more than one parameter does not seem to be being passed through to the script properly. All the parameters get clumped into a single parameter, which probably isn't what the user is expecting.

I ran a few tests on an Ubuntu 18.04 system with bash 4.4.18

Here's params.sh:-
```bash
#!/bin/bash

echo Param count : $#
declare -i n=0
echo "$0"
for i in "$@"; do
    ((++n))
    echo "[$n] : $i"
done
```

Looking at your code, if `program` contains `params.sh aaa bbb ccc` when your code is entered, your code results in the following call:-

```
g_execlp3("params.sh", "-c", "aaa bbb ccc");
```

This translates into 

```
execlp("script.sh", "-c", "aaa bbb ccc",NULL);
```

and the output of the script is:-

```
Param count : 1
./params.sh
[1] : aaa bbb ccc
```

So bash (my copy at least) is ignoring the `-c`.

I replaced `params.sh` with the equivalent C code and got very similar results.

I can suggest a couple of ways to fix this, at least one of which isn't too different from where you are now. However, I'm a bit concerned I may not have fully grasped the reasoning behind your `-c` parameter, so I might be going off on a bit of a tangent. Before we go any further I though I should share my findings with you and see what your comments are.

d-yacenko:
Hi!
You're right. I solved myself problem of transferring only one parameter.
It will be my great pleasure to modify the suggested code to accept any number of parameters. Unfortunately I'm busy right now and can do it in a few days - probably over the weekend.

matt335672:
That's great - I'll have a look at it after the weekend.

d-yacenko:
Hi.
I am passing new code to access many parameters. 

Note that in this code, the execution string is limited to 255 characters. May by is this unnecessary overhead?

matt335672:
Thanks for that @d-yacenko 

I can see what the code is doing (it's clear enough to me). However, there are quite a few stylistic issues here, and some functional issues.

Functional issues:-
- Whitespace can also be tabs.
- If the user accidentally uses multiple spaces, the use of `strtok()` picks up on this, but your `counter` value used for logging is incorrect.
- There's no way to escape a space in a directory path. So if I have a script in '/home/Program Files', I have no way to access it easily.

Stylistic issues:-
- The XRDP coding standard is [here](/neutrinolabs/xrdp/wiki/Coding-Style). In particular, you've got an inconsistent use of braces and indentation. Using `astyle` is recommended to pick up on most things.
- Please make use of functions in `os_calls.h` where available. In particular, use `g_malloc()` in preference to `malloc()`, `g_strncpy()` rather than `strncpy()` and `g_strdup()` rather than `strdup()`. 
- I couldn't understand why you'd use `malloc()/strncpy()` for the intial copy of `s->command`, and then use `strdup()` for the individual parameters. Why not use `strdup()` (or rather `g_strdup()`) for both functions?
- In any case the calls you have to `strdup()` are not needed. The buffer you're parsing with `strtok()` is already a copy of `s->command`, so all you need is `*(arg_array + idx++) = token;`

Other stylistic observations. These are more personal preference than anything else.
- For accessing the pointer array in `arg_array`, using array syntax is possibly easier to parse for humans than using pointer syntax. So for example, use `arg_array[idx++]` rather than `*(arg_array + idx++)`.
- This code is crying out to be wrapped in a function to limit the scope of the temporary variables you've added, and (possibly) to make use of it in other situations.

These are all pretty easily solved apart from escaping spaces I think.

I do wonder if we're over-thinking this. After all, what we're really wanting here is a non-blocking analogue of the C `system(3)` call for multiple parameters. The Linux manpage for `system(3)` is worth a read.

Something like the following pseudo-code would do the job, and be easier to test:-

```
IF << whitespace in s->program >>
THEN
    /* Emulate a system() call to parse whitespace */
    char *params[] = { "sh", "-c", s->program, NULL};
    g_execvp("/bin/sh", params);
ELSE
    g_execlp3(s->program, s->program, 0);
 END IF
```

There's the small overhead of a 'sh' invocation to parse the whitespace, but the big win is that `s->program` behaves exactly how the user expects.

There's a fair bit in here, so have a think and come back to me, particularly if you disagree with anything I've said.

Thanks again for taking the time to put this together.

d-yacenko:
Hey.
Thanks for the answer. I agree with the indicated direction. I need to think (and work) a couple more days and I will send a new commit.

mds99:
May i ask to check if s->program has no trailing spaces as well? sometimes we could not start the script, as the program value xrdp tries to execute is "script "

d-yacenko:
Hi.
The new version looks better and easier. Additionally I removed trailing zeros as @mds99 said.
I hope /bin/ sh will be available in all linux distributives :)

WBR Dmitry

matt335672:
Couple of things:-
- You don't need <string.h> any more if you're using `g_strchr()`. You seem to have included it twice.
- Lines 550-553 don't seem to be necessary - `sh` should do this for you.
- Line 559: Call to `g_strchr()` needs to look for spaces and tabs.
- Line 559: You've now got exactly the same logging message in both if statement branches here, so it could be moved above the if statement.
- Lines 561-563 : Indentation is not regular.
- Line 565 : Coding standard needs opening brace on separate line from `else`.

Hope that's useful - let me know if I've not been clear.

d-yacenko:
Hi.
All done.
It seems...

matt335672:
Thanks.

Nearly there!

The CI build has failed, you'll have noticed. This is because your `char *params[] =` line needs to be `const char *params[] =` because otherwise you're asking for a `char *` to be pointing to a constant string.

Once you fix this you'll find the call to `g_execvp()` also fails.This is more complex to explain. The actual system call `execvp` takes a `char *const argv []` for its second parameter. This is for historical reasons explained on the OpenGroup [exec manpage](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html) under 'RATIONALE'. The original XRDP developers took the decision to use a `char *argv[]` for this. `execvp` will never modify its arguments however, and so it is safe in the `g_execvp()` call to cast `params` to a `char **`, as is done elsewhere in this module.

That will fix the compilation issues. Then:-
- Could you please remove the commented out 'old start call' line? We have git to track this sort of thing for us, so we don't need to keep old code hanging about in the source.
- I appreciate this module in general is rather badly indented which certainly isn't your fault. However, our coding style [line wrapping section](/neutrinolabs/xrdp/wiki/Coding-Style#line-wrapping) calls for 80 character lines where possible. Could you move the `"starting program with parameters: %s "` string onto a new line and align it under `LOG_LEVEL_DEBUG`? I wouldn't worry about the rest of it.

When that's done, if you squash to a single commit in git we should be able to accept this.

Let me know if you have any comments or questions on the above.

d-yacenko:
Hey.
Excuse me for the weak programming. I was just careless due to lack of time.

WBR Dmitry

matt335672:
Hi @d-yacenko 

No worries - thanks for sticking with me.

I must apologise too - I missed the extra `#include <string.h>` at line 48. In my defence I was testing your patch so that I'm certain it does what we think it does.

Could you take the unnecessary include out and squash all the commits? I'll be happy to accept the PR then.

If you've not come across squashing before, I can do it from here, but it's a useful technique to be able to use when preparing a PR for any project. Let me know what you want to do.

matt335672:
Great - ta.

Are you going to do the squash, or shall I do it from here?

d-yacenko:
please do you

matt335672:
Done!

Thanks very much for your contribution to XRDP and for sticking with it.

d-yacenko:
Thank you!

mds99:
Thanks for the fix.
Which code tree state i should use to make a test build? 'devel' or a specific commit? Could you advise?

matt335672:
Use the devel branch:-

```
git clone https://github.com/neutrinolabs/xrdp.git
cd xrdp
```

- `git log` will sohw you the most recent commits, which include the one above.
- See the [wiki](/neutrinolabs/xrdp/wiki) for build instructions.

There's a possibility this may not fix your issue - see my comment in #1695. If it doesn't I'll be happy to supply you with a patch and instructions on how to apply it so you can try that.

PR #1298, opened 2019-02-26
Restrict outbound clipboard

jaroslaw-osmanski:
This patch adds configuration option to xrdp-chansrv that turns copying from the session to clipboard.


metalefty:
Awesome! Good work!

metalefty:
Existing codes use environment variable to pass config parameter. For example, CHANSRV_LOG_PATH.
Of course environment variable is not the best way but consider using environment variable or let's discuss.

sesman/chansrv/clipboard.c
```diff
@@ -2494,7 +2521,14 @@ clipboard_xevent(void *xevent)
...
     {
         case SelectionNotify:
-            clipboard_event_selection_notify(lxevent);
+            if (g_outbound_clipboard_restricted == 0)
+            {
+                clipboard_event_selection_notify(lxevent);
+            }
+            else
+            {
+                return 1;
```
metalefty: I want some log here. Something like "Outbound clipboard is restricted by config.".

jaroslaw-osmanski:
Should env be the only way of setting restriction on the clipboard or additional method? In my opinion config is less surprising.

metalefty:
One unacceptable point is chansrv is opening and reading sesman's config file `sesman.ini`. Please don't do that. Then you'll see how clipboard restriction flag should be passed sesman to chansrv.

metalefty:
Adding `RestrictOutboundClipboard` parameter to sesman.ini is OK. It's not surprising at all.  The parameter must be fetched by sesman because it is **sesman**.ini.

jaroslaw-osmanski:
OK. Sesman gathers info from sesman.ini and then passes information about restriction to chansrv, right?

jaroslaw-osmanski:
`chnsrv` is already reading `sesman.ini` here: https://github.com/neutrinolabs/xrdp/blob/devel/sesman/chansrv/chansrv.c#L1639
So it is config.c that shouldn't be used?

sesman/chansrv/clipboard.c
jaroslaw-osmanski: Is it ok now?

metalefty:
Indeed. That should be fixed but not at this time. I'll change it later.

I'm thinking of this. Similar thing is done as `CHANSRV_LOG_LEVEL` to set chansrv log level. The difference is setting environment variable directly or via config parameter.

### sesman
1. sesman read if RestrictOutboundClipboard is on/off
2. sesman sets env var something like `RESTRICT_OUTBOUND_CLIPBOARD` to something like `1`, `yes`, `true`
3. start chansrv with that env var

### chansrv
3. read the env var at startup
4. if set to `1`, restrict outbound clipboard

sesman/chansrv/clipboard.c
jaroslaw-osmanski: Is this comment enough? And is the logging level ok?

jaroslaw-osmanski:
@metalefty Can I do something more to improve the code?

sesman/chansrv/clipboard.c
metalefty: Good enoug, thanks.

metalefty:
@jaroslaw-osmanski Give me some time to test.

sesman/chansrv/clipboard.c
```diff
@@ -185,38 +185,38 @@ x-special/gnome-copied-files
...
-{                                                       \
-    g_write("[%10.10u]: CLIPBOARD  %s: %d : ERROR: ",   \
-            g_time3(), __func__, __LINE__);             \
-    g_writeln (_params);                                \
-}
+    {                                                       \
+        g_write("[%10.10u]: CLIPBOARD  %s: %d : ERROR: ",   \
+                g_time3(), __func__, __LINE__);             \
+        g_writeln (_params);                                \
+    }
```
metalefty: Thank you for reformatting in a separate commit. It helps me to see the true diff.

sesman/chansrv/chansrv.c
```diff
@@ -1791,6 +1792,13 @@ main(int argc, char **argv)
         return 1;
     }
 
+    restrict_outbound_clipboard_env = g_getenv("CHANSRV_RESTRICT_OUTBOUND_CLIPBOARD");
+    if (restrict_outbound_clipboard_env != 0) {
```
metalefty: See `coding_style.md`.  I know some existing codes don't obey our coding style but newly added codes should obey the coding style.

sesman/config.h
```diff
@@ -60,6 +60,7 @@
 #define SESMAN_CFG_SEC_USR_GROUP     "TerminalServerUsers"
 #define SESMAN_CFG_SEC_ADM_GROUP     "TerminalServerAdmins"
 #define SESMAN_CFG_SEC_ALWAYSGROUPCHECK "AlwaysGroupCheck"
+#define SESMAN_CFG_RESTRICT_CLIPBOARD "RestrictOutboundClipboard"
```
metalefty: I think `SESMAN_CFG_RESTRICT_CLIPBOARD` should include `OUTBOUND` because this might mislead clipboard is restricted in both directions. It is a little bit too long but fair enough.

sesman/session.c
```diff
@@ -374,6 +374,10 @@ session_start_chansrv(char *username, int display)
                      g_cfg->env_names,
                      g_cfg->env_values);
 
+        if (g_cfg->sec.restrict_outbound_clipboard == 1) {
```
metalefty: coding style.

sesman/config.c
```diff
@@ -273,6 +274,13 @@ config_read_security(int file, struct config_security *sc,
         {
             sc->ts_always_group_check = g_text2bool((char *)list_get_item(param_v, i));
         }
+
+        if (0 == g_strcasecmp(buf, SESMAN_CFG_RESTRICT_CLIPBOARD))
+        {
+            sc->restrict_outbound_clipboard = g_text2bool((char *)list_get_item(param_v, i));
+        }
+
+
```
metalefty: Redundant continuous blank line.

metalefty:
I commented on some style & format issues but logic looks OK. After some test, I'll merge. I'd appreciate if you fix style & format before I merge.

sesman/chansrv/clipboard.c
jaroslaw-osmanski: Tried to make review as easy as possible.

jaroslaw-osmanski:
@metalefty Fixed style and format

metalefty:
I'm testing this today and will be merged shortly. Sorry for the delay and thank you very much for your work!

metalefty:
Tested. Clipboard text/image/file LGTM.

black1220:
Can I only pass text copy&paste and limit file transfer ?

PR #1318, opened 2019-03-28
sesman: Fix default_wm and reconnect_sh refer wrong path after SIGHUP

### LINKED ISSUE #1315: error starting window manager after KILL -HUP
Version: xrdp 0.9.8-2~bpo9+1 (debian backports)

We used the following configuration
```
# /etc/xrdp/sesman.ini
[Globals]
...
UserWindowManager=startwm.sh
DefaultWindowManager=startwm.sh
```

After kill -HUP on the main xrdp-sesman process (after a config change for max sessions) we get the following error message.
```
[CORE ] error starting default wm for user .... pid 23400
[CORE ] waiting for window manager (pid 23400) to exit
[DEBUG] errno: 2, description: No such file or directory
```

Maybe the process is running in the wrong path (cwd) after forking or so and doesn't find the script. After changing it to an absolute path everything is running fine. Not sure if it's a packaging bug or code.

metalefty:
Changes in #1147 had a bug. Fixes #1315.

Reported by:    Daniel Hoffend

### DIFF
--- a/sesman/config.c
+++ b/sesman/config.c
@@ -185,11 +185,11 @@ config_read_globals(int file, struct config_sesman *cf, struct list *param_n,
     if (cf->default_wm[0] != '/')
     {
         /* sizeof operator returns string length including null terminator  */
-        length = sizeof(XRDP_CFG_PATH) + g_strlen(g_cfg->default_wm) + 1; /* '/' */
+        length = sizeof(XRDP_CFG_PATH) + g_strlen(cf->default_wm) + 1; /* '/' */
         buf = (char *)g_malloc(length, 0);
-        g_sprintf(buf, "%s/%s", XRDP_CFG_PATH, g_cfg->default_wm);
-        g_free(g_cfg->default_wm);
-        g_cfg->default_wm = g_strdup(buf);
+        g_sprintf(buf, "%s/%s", XRDP_CFG_PATH, cf->default_wm);
+        g_free(cf->default_wm);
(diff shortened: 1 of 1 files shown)


metalefty:
@mirabilos I'd appreciate if you apply this to Debian.

mirabilos:
Hi @metalefty,

>@mirabilos I'd appreciate if you apply this to Debian.

we’re in deep freeze, but I can see the kind of bug this fixes
would be annoying to have in a stable release.

AFAICT the change is just s/g_cfg/cf/g in two if blocks
(and some logging we don’t need to take), so small and
easily enough reviewable to qualify.

@Natureshadow do you wish to take this up with the release
managers or should I?

metalefty:
I think you'll need some more information. v0.9.7, v0.9.8, v0.9.9 have this issue. The upcoming v0.9.10 includes this fix. I know Debian hardly apply version update except unstable so I want you to consider applying this fix to the branches such as stable-bpo. The original issue #1315 is reported by a Debian user. At least a Debian user is annoyed because of our bug.

This issue is a kind of availability.  New sessions cannot be started after xrdp-sesman caught SIGUP unless `DefaultWindowManager` in sesman.ini is specified in full path. Same with `ReconnectScript`. It won't be executed unless it is specified in full path.

I'm not sure Debian rotates xrdp log, the major scenario that xrdp-sesman catch SIGUP is log rotation.

For the record, I post this to the Debian bug, too.

PR #1319, opened 2019-03-28
xrdp: do not exit when caught SIGHUP

### LINKED ISSUE #1316: xrdp stops when using kill -HUP for logrotation
Version: xrdp 0.9.8-2~bpo9+1 (debian backports)

xrdp-sesman reopens the logfile for writing. xdrp itself doesn't and just thinks it gets terminated and stops. Would be nice for consistency

metalefty:
Fixes #1316.

### DIFF
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -164,6 +164,13 @@ xrdp_child(int sig)
 }
 
 /*****************************************************************************/
+void
+xrdp_hang_up(int sig)
+{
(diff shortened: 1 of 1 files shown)


jsorg71:
This looks good, maybe rebase

metalefty:
There's no problem with this. The true goal is to reload the config and reopen log files as described in #1316. Required for log rotation.  However, not to exit with SIGHUP is required sperately.

Let's implement reloading the config and reopening logs later, merge this first.

Issue #996, opened 2018-01-12
Drive redirection not getting file information correctly


vscfreire:
When using drive redirection with Microsoft Remote Desktop client (`mstsc.exe`), file and directory information are not being correctly fetched/updated.

Apparently, `xinode` data for files on redirected drives is only created when directory contents are listed, and is never updated from that point on. This causes two separate (although related) issues.

### Issue 1: directory contents only update with `ls`

For example, if I logon through xrdp (using `mstsc.exe`), launch a terminal window and go to the redirected drive folder, until I run a `ls` command, it's like the drive contents don't exist:

    $ cd thinclient_drives
    $ ls
    C:
    $ cd C\:
    $ cd Windows
    bash: cd: Windows: No such file or directory
    $ cd Users
    bash: cd: Users: No such file or directory
    $ 

So, at this point it seems like the redirected `C:` drive is empty. However, if I run an `ls` command, it seems to fetch the information and it works correctly from that point on:

    $ ls
     'Documents and Settings'   langpacks      'Program Files'         swapfile.sys
     Drivers                    pagefile.sys   'Program Files (x86)'   'System Volume Information'
     hiberfil.sys               PerfLogs       Recovery                Users
     Intel                      ProgramData    '$Recycle.Bin'          Windows
    $ cd Windows
    $ pwd
    /home/user/thinclient_drives/C:/Windows

However, it works just for that directory (I have to use `ls` inside the directory for it to be updated as well):

    $ cd System32
    bash: cd: System32: No such file or directory
    $ ls System32
    ls: cannot access 'System32': No such file or directory
    $ ls
     addins                     INF            schemas
     appcompat                  InfusedApps    security
     ...
     IME                        RtCRU64.exe    WMSysPr9.prx
     ImmersiveControlPanel      SchCache       write.exe
    $ cd System32
    $ pwd
    /home/user/thinclient_drives/C:/Windows/System32
    $

This problem is not visible when using a graphical file manager, because it will list the directory contents automatically when we enter a directory.

I believe this is due to the `ls` command (without arguments) triggering the `xfuse_cb_opendir` callback, which uses `dev_redir_get_dir_listing`/`dev_redir_proc_query_dir_response` to create the `xinode` data.

A simple `cd` or `ls [filename]` starts by triggering the `xfuse_cb_lookup` callback, which fails immediately because there's no `xinode` corresponding to the name.

[...]

PR #1327, opened 2019-04-02
XFuse file overwrite and lookup improvements

matt335672:
This pull request fixes a couple of annoyances I've had with the xfuse filing system:-

- When connecting for the first time, file can't be viewed without browsing down to them (e.g.):-
```
$ ls -l ~/thinclient_drives/U\:/Downloads/temp.txt
ls: cannot access /home/disktest/thinclient_drives/U:/Downloads/temp.txt: No such file or directory
$ ls -l ~/thinclient_drives/U\:/Downloads
ls: cannot access /home/disktest/thinclient_drives/U:/Downloads: No such file or directory
$ ls -l ~/thinclient_drives/U\:/
<snipped>
$ ls -l ~/thinclient_drives/U\:/Downloads/
<snipped>
$ ls -l ~/thinclient_drives/U\:/Downloads/temp.txt
-rw-r--r--. 0 disktest gdat_hdv_disktest 43 Feb 28 16:08 /home/disktest/thinclient_drives/U:/Downloads/temp.txt
```
- Existing files can't be overwritten or appended (e.g.):-
```
$ echo "More text" >~/thinclient_drives/U\:/Downloads/temp.txt
bash: echo: write error: Input/output error
$ echo "More text" >>~/thinclient_drives/U\:/Downloads/temp.txt
bash: echo: write error: Input/output error
```

The first is fixed by changing the lookup functionality for remote files. The original lookup functionality only looks for file attributes in the local cache, even for remote files. Not only does this fail if the directory hasn't been read before, but long-term caching problems are introduced. This implementation always goes to the remote system to look for file attributes, relying on the fuse implementation to do any local caching.

The second of these is relatively simple; the open file functionality needs to have support added for O_WRONLY mode.

For the purpose of review and discussion, I've left this pull request as a series of four commits for now. The first two implement quite a few tidy-ups to the code, and the second two implement the features above.

I appreciate this is a fairly significant change, so I'm happy to split this up into separate pull requests if this helps.


metalefty:
Thank you for your great work! Make sure to pass CI test.

matt335672:
As an extra bit of info, I'm pretty sure, this PR addresses the issues raised in #996

metalefty:
Let's ship this to the upcoming v0.9.10 release.

sesman/chansrv/devredir.c
```diff
@@ -91,6 +91,13 @@
     }                                                   \
 }
 
+/*
+ * Size of structs supported by IRP_MJ_QUERY_INFORMATION without
+ * trailing RESERVED fields (MS-RDPEFS 2.2.3.3.8)
```
metalefty: Thank you for recording the source (MS-RDP*)!

PR #1363, opened 2019-06-23
chansrv: don't let audio fall behind

jsorg71:
When playing long audio, audio can fall behind.  This uses the acks to keep it close.


acharintsev:
xrdp-0.9.11.
Problem with long delay in being sent sound from server still exists. When starting audio playback on the server, a pause occurs. About 5-7 seconds. Such a feeling that for a long time the channel for redirecting sound to the RDP is established. If you start a constant sound stream (for example, a song in VLC), then system sounds and sound from YouTube begin to be transmitted from the server without pauses.

PR #1366, opened 2019-07-02
Listen list

jsorg71:
Allow xrdp to listen on more than one port.  Controlled by xrdp.ini.  Support old file format and -p command line option.


metalefty:
Great work!

PR #1427, opened 2019-10-23
default rdpsnd audin off, can enable with --enable-rdpsndaudin

jsorg71:
Since MS-RDPEAI works disable rdpsnd audio


metalefty:
LGTM

PR #1441, opened 2019-11-12
make vsock config accept -1 for cid and port

Belinsky-L-V:
[linux/vm_sockets.h](https://github.com/torvalds/linux/blob/v5.3/include/uapi/linux/vm_sockets.h) defines `VMADDR_CID_ANY` and `VMADDR_PORT_ANY` (both
equal to `-1U`) for vsock cid and port respectively. This change aims to
add the capability to parse negative cid and port numbers for vsock from
the config and pass them to the kernel.

Allows for valid configurations such as `port=vsock://-1:3389` to be
correctly processed, but will parse port and cid inputs like `---1` to
`-`, which should currently get silently turned into 0 by `atoi` inside
`g_sck_vsock_bind_address`. Inputs that do not contain `-` get parsed as
by `xrdp_listen_parse_integer`.

Addresses #1403

### DIFF
--- a/xrdp/xrdp_listen.c
+++ b/xrdp/xrdp_listen.c
@@ -358,6 +358,53 @@ xrdp_listen_parse_integer(char *strout, int strout_max,
 
 /*****************************************************************************/
 static int
+xrdp_listen_parse_vsock(char *strout, int strout_max,
+                        const char *strin, int strin_max)
+{
+    int count;
+    int in;
+    int strin_index;
+    int strout_index;
+
+    strin_index = 0;
+    strout_index = 0;
+    in = 0;
+    count = 0;
+    while ((strin_index < strin_max) && (strout_index < strout_max))
+    {
+        if (in)
+        {
+            if ((strin[strin_index] >= '0') && (strin[strin_index] <= '9'))
+            {
+                strout[strout_index++] = strin[strin_index++];
+                count++;
+                continue;
+            }
+            else
+            {
+                break;
+            }
+        }
+        else
+        {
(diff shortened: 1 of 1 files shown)


Belinsky-L-V:
May be worthwhile to add more validation to various parsers in xrdp/xrdp_listen.c including this vsock one in a different RP. For the sake of simplicity here I just implemented the fix suggested [here](https://github.com/neutrinolabs/xrdp/issues/1403#issuecomment-533837144) by @prydom and restricted it only to vsock, as opposed to port parsing in general.

metalefty:
Overall, LGTM. Nice work!

jsorg71:
Looks good to me.  I think it would be fine to change xrdp_listen_parse_integer too.

Belinsky-L-V:
> 
> 
> Looks good to me. I think it would be fine to change xrdp_listen_parse_integer too.

Not sure that'd be a good idea. `xrdp_listen_parse_integer` is used to parse TCP ports. Unlike vsock's `unsigned int` port and cids, those are supposed to be 16-bit unsigned so `atoi("-1")` inside `g_sck_vsock_bind_address` wouldn't map to the range of valid ports in that case. I assume it's preferable to fail on the parsing stage, as opposed to passing an invalid port on.

prydom:
@Belinsky-L-V, @jsorg71: I likewise agree that `xrdp_listen_parse_integer` should not change it's behavior and your patch LGTM. Thanks for taking the reins on this!

jsorg71:
@Belinsky-L-V Thanks for explaining, looks good.

PR #1442, opened 2019-11-13
mkpamrules: Support openSUSE's usage of /usr/etc/pam.d to contain the pam configuration files.

yifanjiang:
Please see the details in https://lists.opensuse.org/opensuse-factory/2019-08/msg00113.html. The old directory for probing the pam configuration file is also kept for backward compatibility reason.


metalefty:
+1 Simple enough and no side effects with other distro.

PR #1449, opened 2019-12-05
Significant remote file system improvements

matt335672:
- Reimplemented inode store in separate module `chansrv_xfs.[hc]`
- Allowed atimes and mtimes to be written to Windows side
- Mapped directory/file user write bit to (inverted) Windows FILE_ATTRIBUTE_READONLY bit
- Mapped file user execute bit to Windows FILE_ATTRIBUTE_SYSTEM bit
- Implemented improved security for remotely mounted drives (see man sesman.ini)
- Implemented USB device removal, allowing hot-plug/remove of memory sticks
- Fixed `pagefile.sys` (etc) breaking Ubuntu file browser
- Fixed write offset bug
- Allowed renaming of open files
- Improved reported error codes
- Fixed various memory leaks
- Removed hardcoded buffer sizes where appropriate
- Addressed valgrind/memcheck errors related to struct fuse_file_info pointers.

This improves the remote filesystem support to a point where it can even be used to support an XRDP build, although it is necessary to disable PAM support for this, as `instfiles/pam.d/mkpamrules` creates a soft-link which isn't supported. I used the following commands on an Ubuntu 18.04 LTS system with the XRDP 0.9.11 release tarball:-

```
cd ~/thinclient_drives/C\:/Users/mjb/Documents/
tar xf ~/xrdp-0.9.11.tar.gz
cd xrdp-0.9.11
./configure --enable-pixman --enable-fuse --disable-pam
make
```

There are a lot of code changes here. I've tried to simplify the code where possible to make it easier to follow. A lot of private stuff has been moved out of `devredir.h` into other places. primarily the new includes `ms-erref.h`, `ms-fscc.h`, `ms-rdpefs.h` and `ms-smb2.h`. The separate inode store in `chansrv_xfs.c` is significantly faster than the previous inode store, and removes a lot of complications from the code paths in `chansrv_fuse.c`.

### DIFF
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -225,6 +225,14 @@ Following parameters can be used in the \fB[Chansrv]\fR section.
 Directory for drive redirection, relative to the user home directory.
 Created if it doesn't exist. If not specified, defaults to \fIxrdp_client\fR.
 
+.TP
+\fBFileUmask\fR=\fImode\fR
+Additional umask to apply to files in the \fBFuseMountName\fR directory.
+The default value of 077 prevents other users on the system from reading
+files on your redirected drives. This may not be approprate for all
+environents, and so you can change this value to allow other users to
+access your remote files if required.
+
 .SH "SESSIONS VARIABLES"
 All entries in the \fB[SessionVariables]\fR section are set as
 environment variables in the user's session.
--- a/sesman/chansrv/Makefile.am
+++ b/sesman/chansrv/Makefile.am
@@ -53,6 +53,8 @@ xrdp_chansrv_SOURCES = \
   chansrv_common.h \
   chansrv_fuse.c \
   chansrv_fuse.h \
+  chansrv_xfs.c \
+  chansrv_xfs.h \
   clipboard.c \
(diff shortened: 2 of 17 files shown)


metalefty:
Brilliant!  You really make lots of contributions on drive redirection. I trust your QA. Let's ship it to the next v0.9.12 release.

matt335672:
Thanks for vote of confidence!

I've got a question for you. I've created four new includes `ms-erref.h`, `ms-fscc.h`, `ms-rdpefs.h` and `ms-smb2.h` in the chansrv directory containing definitions from the relevant MS docs. I did this to remove unnecessary definitions from `devredir.h`.

These includes could be merged with `common/xrdp_constants.h`, or alternatively some of the constants in `common/xrdp_constants.h` could be moved out to new files `common/ms-rdpbcgr.h` and `common/ms-rdpegdi.h` to slim it down a bit. It's just a bit of tidying up.

Let me know which you prefer and I'll put a separate PR together after Xmas.

metalefty:
The latter plan sounds tidy and clear which constants come from which MS document for me.
+1 to the latter.

PR #1469, opened 2020-01-06
sesman/chansrv/chansrv_fuse.c: resolve double free found by cppcheck

chipitsine:
sesman/chansrv/chansrv_fuse.c:1719:9: error: Memory pointed to by 'full_path' is freed twice. [doubleFree]


matt335672:
Good catch.

metalefty:
Oops.

PR #1470, opened 2020-01-06
Move MS-defined constants into separate includes

matt335672:
This change was discussed very briefly in PR #1449 

The constants in `xrdp_constants.h` which can be attributed to specific MS documents have been moved out of `xrdp_constants.h` into include files named after the documents.  So for example, the NTSTATUS code STATUS_SUCCESS from the MS document MS-ERREF is now defined in `ms-erref.h`.

To avoid unnecessary coupling, I've not added nested includes for these files to `xrdp_constants.h`. I've simply included these file in the places where they are needed.

Some of these files were created locally to the chansrv directory for #1449 These files are now moved to the common directory. 

Other notes/questions:-
- The constants NT_STATUS_xxx introduced in #1449 are renamed to STATUS_xxx to correspond with the definitions in `ms-erref.h`
- Some of the constants in `xrdp_constants.h` were commented `TODO: to be renamed`, where MS names were not actually used. I've not changed these yet, but would be happy to do so if required. Now might be a good time, straight after v0.9.12 is released.
- The file `X11R7.6/rdp/rdp.h` in the x11rdp repository includes `xrdp_constants.h`, and so will not longer compile against xrdp. I'm assuming this is OK as x11rdp is now unmaintained.
- The `xrdp_constants.h` file is part of the RHEL/CentOS [xrdp-devel RPM](https://koji.fedoraproject.org/koji/rpminfo?rpmID=19931645). Presumably the new files should also be added to this RPM at some stage, although I'm not aware of anything other than x11rdp which includes this RPM.


metalefty:
Looks much tidied-up. Great job.

metalefty:
@jsorg71 is this fine for you?

jsorg71:
Yes, looks good.

PR #1474, opened 2020-01-08
FreeBSD CI: simplify CI test for a while

metalefty:
some tests are failing due to FreeBSD OpenSSL version change.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,14 +1,10 @@
 FreeBSD_task:
   matrix:
     env:
-      SSL: openssl
-    env:
       SSL: libressl
   matrix:
     freebsd_instance:
       image_family: freebsd-12-1
-    freebsd_instance:
-      image_family: freebsd-11-3-snap
   prepare_script:
     - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm
     - git submodule update --init --recursive


chipitsine:
btw, I came to some (ugly) solution

https://www.mail-archive.com/haproxy@formilux.org/msg35826.html

I'd like to find something really better

metalefty:
Alternatively, I believe this should also solve the package name mismatch (not tested). See `/etc/pkg/FreeBSD.conf`.
```diff
--- FreeBSD.conf.orig   2020-01-09 14:19:10.405306000 +0900
+++ FreeBSD.conf        2020-01-09 14:19:36.760731000 +0900
@@ -8,7 +8,7 @@
 #

 FreeBSD: {
-  url: "pkg+http://pkg.FreeBSD.org/${ABI}/quarterly",
+  url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest",
   mirror_type: "srv",
   signature_type: "fingerprints",
   fingerprints: "/usr/share/keys/pkg"
```

FreeBSD pkgs (binary packages) are built quarterly. There was [openssl package name changes](https://github.com/freebsd/freebsd-ports/blob/096b70bb35ada3166546eaa7209bd805f65d6b33/UPDATING#L31-L44) in 2019-2020 new year holidays. This changes are already applied to HEAD but not completely applied yet to quarterly (2020Q1).

OS image `freebsd-12-1` refers quarterly 2020Q1 packages. OTOH, I think `freebsd-11-3-snap` refers HEAD (latest) because it is snapshot. This mismatch and incompleteness are causing the failure.

I believe the changes will be applied also to quarterly build shortly.  The workaround will be really short life. openssl111 will be removed from package repo shortly. So better solution I think is to use the same package repo `latest` or `quarterly` both on 12 and 11 OS image.

metalefty:
Performing CI test on multiple versions of the same OS will be robust but honestly I think a single version is enough. If performing CI test on multiple versions, I should have use `-snap` image both with 11 and 12 (and 13).

`-snap` works fine: https://github.com/metalefty/xrdp/runs/380799278

chipitsine:
I'm not that skilled in freebsd world. I used to run it years ago.

let me know what would be the most appropriate

metalefty:
Short answer: not to mix `-snap` images and other images

If we're using `freebsd-11-x-snap` image, FreeBSD 12 image also should be `freebsd-12-x-snap`. Because snapshot images are ahead of releases, there's some difference in package name such as `openssl` vs `openssl111`.

Issue #1473, opened 2020-01-07
various cppcheck findings


chipitsine:
```
sesman/chansrv/pcsc/xrdp_pcsc.c:1123:9: warning: Either the condition 'pioRecvPci==0' is redundant or there is possible null pointer dereference: pioRecvPci. [nullPointerRedundantCheck]
sesman/chansrv/pcsc/xrdp_pcsc.c:1125:9: warning: Either the condition 'pioRecvPci==0' is redundant or there is possible null pointer dereference: pioRecvPci. [nullPointerRedundantCheck]
sesman/chansrv/pcsc/xrdp_pcsc.c:1198:26: warning: Either the condition 'mszGroups==NULL' is redundant or there is possible null pointer dereference: mszGroups. [nullPointerRedundantCheck]
xrdp/xrdp_cache.c:366:13: warning: Either the condition 'bitmap!=0' is redundant or there is possible null pointer dereference: bitmap. [nullPointerRedundantCheck]
```

matt335672:
Hi @chipitsine 

I'm having problems reproducing exactly what you've got above.

In particular, if I check out v0.9.12 on Ubuntu 18.04.3 I get the following set of errors with `cppcheck -q --force  --enable=warning .`:-

```
[sesman/chansrv/chansrv_fuse.c:1719]: (error) Memory pointed to by 'full_path' is freed twice.
[sesman/chansrv/chansrv_xfs.c:390] -> [sesman/chansrv/chansrv_xfs.c:398]: (warning) Either the condition 'xfs!=NULL' is redundant or there is possible null pointer dereference: xfs.
[sesman/chansrv/chansrv_xfs.c:390] -> [sesman/chansrv/chansrv_xfs.c:399]: (warning) Either the condition 'xfs!=NULL' is redundant or there is possible null pointer dereference: xfs.
[sesman/chansrv/devredir.c:207]: (error) Uninitialized variable: s
[sesman/chansrv/pcsc/xrdp_pcsc.c:1183] -> [sesman/chansrv/pcsc/xrdp_pcsc.c:1198]: (warning) Either the condition 'mszGroups==NULL' is redundant or there is possible null pointer dereference: mszGroups.
[tests/tcp_proxy/main.c:40]: (error) Failed to parse #define
[vrplayer/decoderthread.cpp:33]: (warning) Member variable 'DecoderThread::pktTimer' is not initialized in the constructor.
[...]

PR #1481, opened 2020-01-16
chansrv-fuse Fix cppcheck 1.89+1.90 warnings

matt335672:
This PR addresses the cppcheck warning generated by cppcheck for `xrdp/xrdp_{xfs,fuse}.c` and `devredir.c` mentioned in #1473

The warning in `chansrv_fuse.c` generated by cppcheck 1.90 (but not 1.89 or 1.82) appears to be a regression in cppcheck which has already been reported - see cppcheck TRAC [9292](https://trac.cppcheck.net/ticket/9292) and [9437](https://trac.cppcheck.net/ticket/9437). Following a `realloc()` failure, a `free()` of the original region results in a false positive.

cppcheck allows for false positives to be flagged with specially formed comments. The `--inline-suppr` command-line flag is needed to enable these. This approach has been taken for this file.

The warning in `devredir.c` appears to be valid code used to generate a random number if `/dev/urandom` is unavailable. I've simply replaced the code with a call to `g_random()` in this case.

So despite the warnings, no actual bugs have been found here.

**Update:** After #1487 was merged, the warning which had previously been generated for `chansrv_xfs.c` has now gone. This file has been removed from the original PR.

### DIFF
--- a/sesman/chansrv/devredir.c
+++ b/sesman/chansrv/devredir.c
@@ -185,28 +185,10 @@ devredir_init(void)
 {
     struct  stream *s;
     int     bytes;
-    int     fd;
-
-    union _u
-    {
-        tui32 clientID;
-        char buf[4];
-    } u;
+    tui32 clientID;
 
     /* get a random number that will act as a unique clientID */
-    if ((fd = open("/dev/urandom", O_RDONLY)) != -1)
-    {
-        if (read(fd, u.buf, 4) != 4)
-        {
-        }
-        close(fd);
-    }
-    else
-    {
-        /* /dev/urandom did not work - use address of struct s */
-        tui64 u64 = (tui64) (tintptr) &s;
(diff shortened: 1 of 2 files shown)


sesman/chansrv/devredir.c
```diff
@@ -185,28 +185,10 @@ devredir_init(void)
...
-        }
-        close(fd);
-    }
-    else
-    {
-        /* /dev/urandom did not work - use address of struct s */
-        tui64 u64 = (tui64) (tintptr) &s;
-        u.clientID = (tui32) u64;
-    }
+    g_random((char *) &clientID, sizeof(clientID));
```
metalefty: Good catch. It is good to use existing wrapper function.

matt335672:
Updated following commit of #1487 to devel branch. Header comment also updated.

PR #1485, opened 2020-01-19
misc: Fix cppcheck 1.82 + 1.90 warnings

matt335672:
This PR addresses the warnings generated by cppcheck 1.82 and 1.90 for `tests/tcp_proxy/main.c`, `xrdpapi/xrdp-ssh-agent.c` and `vrplayer/decoderthread.cpp` mentioned in #1473.

The first two are trivial and are not described further.

As far as `vrplayer/decoderthread.cpp` goes, this file is not longer used.  It was removed from the QT project file ` vrplayer/vrplayer.pro` by cae2adb75e2b339017746436c63f7d2b8c35fb63 in December 2012 and is not referenced anywhere else.

So despite the cppcheck warnings, no actual bugs have been found here.


matt335672:
Second commit added for `vrplayer/ourinterface.cpp` which I missed first time round.

PR #1487, opened 2020-01-25
Use g_free for xfs/g_xfs in xfs_delete_xfs_fs

derekschrock:
It's possible for xfs (g_xfs) to be null due to fuse mounts failing.
If g_xfs is null chansrv will crash with sig. fault.

### DIFF
--- a/sesman/chansrv/chansrv_xfs.c
+++ b/sesman/chansrv/chansrv_xfs.c
@@ -387,7 +387,12 @@ xfs_create_xfs_fs(mode_t umask, uid_t uid, gid_t gid)
 void
 xfs_delete_xfs_fs(struct xfs_fs *xfs)
 {
-    if (xfs != NULL && xfs->inode_table != NULL)
+    if (xfs == NULL)
+    {
+        return;
+    }
+
+    if (xfs->inode_table != NULL)
     {
         size_t i;
         for (i = 0 ; i < xfs->inode_count; ++i)


derekschrock:
By @metalefty request @matt335672 for review.

matt335672:
Hi @derekschrock 

Thanks for the PR. I've had a look at the code, and you're quite right that entering `xfs_delete_xfs_fs()` will cause a segfault if called with a NULL parameter. It's down to bad coding on my part which I didn't catch in my unit testing.

That said, I think your PR needs a bit of work for two reasons:-
- It's valid to call `free(p)` with p being NULL. This has been the case for a very long time. It's even in the rather dog-eared copy of the 2nd edition K&R I have at home on page 252 (printed 1988) that _[free] does nothing if P is NULL_. So for XRDP, `free(p)` behaves identically to `g_free(p)`, and whether to use one or the other is a question of style rather than behaviour.
- To get to your patched line 400, the path of execution has to pass through the two preceding lines. These both de-reference the NULL pointer before calling `free()`, and so either would cause a segfault.

Assuming the above make sense, and I'm not missing something obvious could you update and re-submit the PR?

@metalefty - this code is referenced in my cppcheck PR #1481 (which co-incidentally also addresses this). Since Derek's putting some work in now, I'll happily rework that PR when we're done with this one. The other cppcheck PRs are not affected.

Thanks both

metalefty:
Thank you both.

@matt335672 can you review @derekschrock 's new patch?

matt335672:
LGTM - simplest solution to known problem. Thanks @derekschrock.

metalefty:
We need to make a release  for this fix.

matt335672:
Agreed, as it seems to be affecting other users. I can only apologise for not picking it up in my unit testing.

metalefty:
ref. #1449

PR #1494, opened 2020-02-06
Fix empty rhost in /var/log/auth.log

dimapnz:
I found that the service xrdp-sesman writes an empty rhost into log file /var/log/auth.log. Because of this, the rule "pam-generic" (/etc/fail2ban/filter.d/pam-generic.conf) in the package "fail2ban" does not work. This PR fixing this issue.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1453,7 +1453,7 @@ g_write_ip_address(int rcv_sck, char *ip_address, int bytes)
 
         if (ok)
         {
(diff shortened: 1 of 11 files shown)


metalefty:
Thanks, I will test it.

dimapnz:
With option --enable-ipv6 also works properly

moobyfr:
This pull request can probably be closed: the current release code has valid rhost values (thanks to #2223)

matt335672:
Good spot - thanks.

Issue #1502, opened 2020-02-21
xrdp-chansrv coredump upon second connection attempt


Smit-tay:
Consistant core dump when initializing remote connection via rdp - a session already exists for this user.

Using remmina 1.3.10 RDP client - default RDP protocol settings.

Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[INFO ] A connection received from 127.0.0.1 port 57402
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: pam_sss(xrdp-sesman:auth): authentication success; logname= uid=0 euid=0 tty=xrdp-sesman ruser= rhost= user=MYCOMPANY\ME
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[INFO ] ++ created session (access granted): username MYCOMPANY\ME, ip XXX.XXX.XXX.XXX:42558 - socket: 12
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[INFO ] starting Xvnc session...
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[DEBUG] Closed socket 13 (AF_INET 0.0.0.0:5911)
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[DEBUG] Closed socket 13 (AF_INET 0.0.0.0:6011)
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[DEBUG] Closed socket 13 (AF_INET 0.0.0.0:6211)
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[4963]: (4963)(140636279781184)[DEBUG] Closed socket 10 (AF_INET 127.0.0.1:3350)
Feb 21 10:59:19 MY_MACHINE.MYCOMPANY.com xrdp-sesman[7079]: (7079)(140636279781184)[INFO ] calling auth_start_session from pid 7079
[...]

PR #1504, opened 2020-02-23
Check before free

Smit-tay:
-- inode table array element can be null (ex: element zero intentionally null)

### DIFF
--- a/sesman/chansrv/chansrv_xfs.c
+++ b/sesman/chansrv/chansrv_xfs.c
@@ -397,7 +397,8 @@ xfs_delete_xfs_fs(struct xfs_fs *xfs)
         size_t i;
         for (i = 0 ; i < xfs->inode_count; ++i)
         {
-            free(xfs->inode_table[i]);
+            if(xfs->inode_table[i] != NULL)
+                free(xfs->inode_table[i]);
         }
     }
     free(xfs->inode_table);


matt335672:
I can't see the point of this check.

`free(NULL)` is perfectly valid C, and has been since at least K&R days. The reference from ISO/IEC 9899:1999 (which is the only one I have by my desk) is 7.20.3.2 which quotes for `free(ptr)`:-

_If ptr is a null pointer, no action occurs._

See also my comment in #1487

What am I missing? Are you getting a coredump which is pointing you in this area? Are you hitting the issue with 0.9.12 which is addressed by #1487?

Smit-tay:
Indeed.

Although because I am looking at a production server it's not so easy to debug.

The call stack led me to this quite simple function and my (naive) review caught what look(s/ed) like a potential problem.

I was unaware that free(0) is safe.

So, that implies that something that the array holds is no longer available.

For me debugging beyond this level isn't an option, not due to personal ability, but due to environmental problems.  I really don't have the resources to investige in a meaningful way

matt335672:
I might be able to do something with the stack trace you've provided in #1502 to work out whether this is indeed #1487 or something else.

Could you maybe close this PR, re-open #1502, and add some information about the server operating system and version of the XRDP package you've installed? I couldn't see that information.

Is falling back to v0.9.11 something you could do quite easily? I'm not familiar with the remmina client, but if it allows you to map a network drive, this might be a workaround you could use.

PR #1545, opened 2020-04-06
Remotefx painter

J-W-C:
A proposed addition that make the Painter send screen image data via RDP surface commands using the RemoteFX codec instead of using RDP bitmap updates.

Makes the login screen come up a little faster, and make xrdp compatible with clients that require RemoteFX.

Work in progress, includes some to-do notes and developmental asserts.


metalefty:
BTW, can you create commits as your name, not EC2 Default User?
You can redo commit and force-push.

J-W-C:
Yes, I noticed somewhat late I hadn't set my git email address on that EC2 instance.  

Off to learn how to redo commits and force-push now...

xrdp/xrdp_painter.c
```diff
@@ -76,6 +77,10 @@ xrdp_painter_add_dirty_rect(struct xrdp_painter *self, int x, int y,
     return 0;
 }
 
+#define RFX_TILE_SQUARE 64
+
+#define XRDP_SURCMD_PREFIX_BYTES 256
```
metalefty, line 82: Where do these come from?

J-W-C: Thank you for looking at my stuff!

I just looked again and I think the first one might want to be replaced by CT_TILE_64x64 from src/rfxconstants.h, assuming that's visible outside the RFX encoder implementation.  (I wonder what "CT" stands for?)

The second, XRDP_SURCMD_PREFIX_BYTES, is a copy of an existing #define in xrdp_encoder.c, where it serves the same purpose (leaving room for an RDP surface command so that the image encoding doesn't have to be copied).  Maybe the #define belongs in libxrdp/libxrdpinc.h?

xrdp/xrdp_painter.c
```diff
@@ -94,47 +99,145 @@ xrdp_painter_send_dirty(struct xrdp_painter *self)
...
+            num_rfxrects++;
+
+            int top_ty = rect.top / RFX_TILE_SQUARE;
+            int bot_ty = (rect.bottom - 1) / RFX_TILE_SQUARE;
+            for (int ty = top_ty; ty <= bot_ty; ty++)
+            {
+                int left_tx = rect.left / RFX_TILE_SQUARE;
+                int right_tx = (rect.right - 1) / RFX_TILE_SQUARE;
+                for (int tx = left_tx; tx <= right_tx; tx++)
+                {
```
J-W-C, line 147: For loop initial declarations are only allowed in C99 -- not otherwise required by xrdp.

xrdp/xrdp_painter.c
```diff
@@ -94,47 +99,145 @@ xrdp_painter_send_dirty(struct xrdp_painter *self)
...
 
-        jndex++;
-        error = xrdp_region_get_rect(self->dirty_region, jndex, &rect);
-    }
+        int last_wide = 1 + (width - 1) % RFX_TILE_SQUARE;
+        int last_high = 1 + (height - 1) % RFX_TILE_SQUARE;
+        for (int ty = 0; ty < tiles_high; ty++)
+        {
+            for (int tx = 0; tx < tiles_wide; tx++)
+            {
```
J-W-C, line 161: Ditto.

metalefty:
Has this been finished?

J-W-C:
Trying to get back to it now.  I just noticed I should have some #ifdefs so it'll build w/ --disable-rfxcodec.

PR #1551, opened 2020-04-18
Check for overflow before g_malloc() calls

### LINKED ISSUE #1533: memory out of bounds write in scard_process_list_readers
**version**

```
https://github.com/neutrinolabs/xrdp/blob/1c4e14415d923666ac60cb77c7e753ca24e0268d/sesman/chansrv/smartcard_pcsc.c#L600
```

**vuln**

`scard_process_list_readers` first read 4 bytes to `bytes_groups`, then use `g_malloc(bytes_groups + 1, 1)` to alloc memory.

if  `bytes_groups=0xffffffff` , it could trigger Integer overflow (`0xffffffff+1=0`), it could alloc 0 size memory to `groups`.

Then it could use `in_uint8a`  to copy data to  `groups`. resulting in **out-of-bounds write**

```
int
scard_process_list_readers(struct trans *con, struct stream *in_s)
{
    ..............................
    ..............................
    in_uint32_le(in_s, hContext);
    in_uint32_le(in_s, bytes_groups);
   
    groups = (char *) g_malloc(bytes_groups + 1, 1); // 0xffffffff + 1 = 0
    in_uint8a(in_s, groups, bytes_groups);  // copy 0xffffffff data to groups
```

### LINKED ISSUE #1534: memory out of bounds write in my_api_trans_data_in
**version**

```
https://github.com/neutrinolabs/xrdp/blob/1c4e14415d923666ac60cb77c7e753ca24e0268d/sesman/chansrv/chansrv.c#L1070
```

**vuln**

`my_api_trans_data_in` first read 4 bytes to `channel_name_bytes`, then use `g_new0(char, channel_name_bytes + 1)` to alloc memory.

if  `channel_name_bytes=0xffffffff` , it could trigger Integer overflow (`0xffffffff+1=0`), it could alloc 0 size memory to `chan_name`.

Then it could use `in_uint8a`  to copy data to  `chan_name`. resulting in **out-of-bounds write**

```
int
my_api_trans_data_in(struct trans *trans)
{
    ..........................
    ..........................
    else if (trans->extra_flags == 1)
    {
        rv = 1;
        in_uint32_le(s, channel_name_bytes);
        chan_name = g_new0(char, channel_name_bytes + 1);  // Integer overflow
        if (chan_name == NULL)
        {
            return 1;
        }
        in_uint8a(s, chan_name, channel_name_bytes); // out of bounds write
```

### LINKED ISSUE #1544: Integer overflow in xrdp_bitmap_create
**version**

```
https://github.com/neutrinolabs/xrdp/blob/1c4e14415d923666ac60cb77c7e753ca24e0268d/xrdp/xrdp_bitmap.c#L126
```

**vuln**

`process_server_create_os_surface`  first read some data from stream， then it could pass argument to `server_create_os_surface`.

```
static int
process_server_create_os_surface(struct mod *mod, struct stream *s)
{
    int rv;
    int rdpid;
    int width;
    int height;

    in_uint32_le(s, rdpid);
    in_uint16_le(s, width);
    in_uint16_le(s, height);
    rv = mod->server_create_os_surface(mod, rdpid, width, height);
    return rv;
}
```

then it could finally call to xrdp_bitmap_create

```
struct xrdp_bitmap *
xrdp_bitmap_create(int width, int height, int bpp,
                   int type, struct xrdp_wm *wm)
{
   	...........................
   	
    switch (bpp)
    {
        case 8:
            Bpp = 1;
            break;
        case 15:
            Bpp = 2;
            break;
        case 16:
            Bpp = 2;
            break;
    }
    
    if (self->type == WND_TYPE_BITMAP || self->type == WND_TYPE_IMAGE)
    {
    	// width, height come from user input data.
        self->data = (char *)g_malloc(width * height * Bpp, 0);
    }
```

`width` and `height` is come from data, the **max value** of `width * height * Bpp` is `0x1fffc0002`.

```
>>> hex(0xffff*0xffff*2)
'0x1fffc0002'
```

the first argument  type of `g_malloc` is `int` , which **max value** is `0xffffffff`, So the size value will be truncated.

then It could lead **out of bounds write** to `self->data` later.

matt335672:
Fixes #1533, fixes #1534, fixes #1544

This is the simplest fix I could come up with. Arguably `g_malloc()` should take an unsigned type of some sort (`size_t` even) but I think implementing this would involve a lot of minor changes to many files, given the general absence of this type in the codebase. Maybe now is a good time to do that(?) Anyone got any thoughts on that?

### DIFF
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -21,6 +21,8 @@
 #include <config_ac.h>
 #endif
 
+#include <limits.h>
+
 #include "arch.h"
 #include "os_calls.h"
 #include "thread_calls.h"
@@ -1067,6 +1069,11 @@ my_api_trans_data_in(struct trans *trans)
         rv = 1;
         in_uint32_le(s, channel_name_bytes);
         //g_writeln("my_api_trans_data_in: channel_name_bytes %d", channel_name_bytes);
+        if (channel_name_bytes == INT_MAX)
+        {
+            // Adding one to this will overflow
+            return 1;
+        }
         chan_name = g_new0(char, channel_name_bytes + 1);
         if (chan_name == NULL)
         {
--- a/sesman/chansrv/smartcard_pcsc.c
+++ b/sesman/chansrv/smartcard_pcsc.c
@@ -27,6 +27,8 @@
 #include <config_ac.h>
 #endif
 
+#include <limits.h>
+
 #define JAY_TODO_CONTEXT    0
 #define JAY_TODO_WIDE       1
 
@@ -597,6 +599,11 @@ scard_process_list_readers(struct trans *con, struct stream *in_s)
     uds_client = (struct pcsc_uds_client *) (con->callback_data);
     in_uint32_le(in_s, hContext);
     in_uint32_le(in_s, bytes_groups);
+    if (bytes_groups == INT_MAX)
+    {
+        // Adding one to this will overflow
+        return 1;
+    }
     groups = (char *) g_malloc(bytes_groups + 1, 1);
     in_uint8a(in_s, groups, bytes_groups);
     in_uint32_le(in_s, cchReaders);
--- a/xrdp/xrdp_bitmap.c
+++ b/xrdp/xrdp_bitmap.c
@@ -25,6 +25,8 @@
 #include <config_ac.h>
 #endif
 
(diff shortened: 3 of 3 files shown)


matt335672:
Just been thinking some of this through - it needs more work. I'll resubmit it when it's in a better state.

PR #1557, opened 2020-04-23
Check for overflow before g_malloc() calls

### LINKED ISSUE #1533: memory out of bounds write in scard_process_list_readers

### LINKED ISSUE #1534: memory out of bounds write in my_api_trans_data_in

### LINKED ISSUE #1544: Integer overflow in xrdp_bitmap_create

matt335672:
Here's a better attempt to replace #1551 

Fixes  #1533, fixes #1534, fixes #1544

#1533 - Checked the supported strings for the group parameter (see here). They're all relatively small, so code restructured to just allow a short string here, and check for it.
#1534 - Similarly the allowed string for a virtual channel name is quite small here - see here.
#1544 - Checked for integer overflow explicitly when requesting buffer.

I've not attempted to change the argument to g_malloc() to an unsigned integer here - I think that needs more thought and a separate PR

### DIFF
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -39,6 +39,9 @@
 #include "xrdp_sockets.h"
 #include "audin.h"
 
+#define CHANNEL_NAME_BYTES 7
+#define MAX_PATH 260
+
 static struct trans *g_lis_trans = 0;
 static struct trans *g_con_trans = 0;
 static struct trans *g_api_lis_trans = 0;
@@ -1042,7 +1045,7 @@ my_api_trans_data_in(struct trans *trans)
     int rv;
     int bytes;
     int ver;
-    int channel_name_bytes;
+    unsigned int channel_name_bytes;
     struct chansrv_drdynvc_procs procs;
     char *chan_name;
 
@@ -1067,6 +1070,13 @@ my_api_trans_data_in(struct trans *trans)
         rv = 1;
         in_uint32_le(s, channel_name_bytes);
         //g_writeln("my_api_trans_data_in: channel_name_bytes %d", channel_name_bytes);
+        /*
+         * Name is limited to CHANNEL_NAME_BYTES for an SVC, or MAX_PATH
+         * bytes for a DVC */
+        if (channel_name_bytes > MAX(CHANNEL_NAME_BYTES, MAX_PATH))
+        {
+            return 1;
+        }
         chan_name = g_new0(char, channel_name_bytes + 1);
         if (chan_name == NULL)
         {
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -39,7 +39,8 @@
 #include "xrdp_sockets.h"
 #include "audin.h"
 
-#define CHANNEL_NAME_BYTES 7
+#include "ms-rdpbcgr.h"
+
 #define MAX_PATH 260
 
 static struct trans *g_lis_trans = 0;
@@ -1047,10 +1048,10 @@ my_api_trans_data_in(struct trans *trans)
     int ver;
     struct chansrv_drdynvc_procs procs;
     /*
-     * Name is limited to CHANNEL_NAME_BYTES for an SVC, or MAX_PATH
+     * Name is limited to CHANNEL_NAME_LEN for an SVC, or MAX_PATH
      * bytes for a DVC
      */
-    char chan_name[MAX(CHANNEL_NAME_BYTES, MAX_PATH) + 1];
+    char chan_name[MAX(CHANNEL_NAME_LEN, MAX_PATH) + 1];
     unsigned int channel_name_bytes;
 
     //g_writeln("my_api_trans_data_in: extra_flags %d", trans->extra_flags);
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -1045,9 +1045,13 @@ my_api_trans_data_in(struct trans *trans)
     int rv;
     int bytes;
     int ver;
-    unsigned int channel_name_bytes;
     struct chansrv_drdynvc_procs procs;
-    char *chan_name;
+    /*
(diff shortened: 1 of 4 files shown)


matt335672:
Just realised that the change to `chansrv.c` can be further improved. The call to `g_malloc()`/`g_free()` for the channel name has now been replaced with a simple fixed-size stack allocation.

sesman/chansrv/chansrv.c
```diff
@@ -39,6 +39,9 @@
 #include "xrdp_sockets.h"
 #include "audin.h"
 
+#define CHANNEL_NAME_BYTES 7
```
metalefty, line 42: Isn't this defined in MS specification?

matt335672: You're quite right. There's no name for it, but it's in section 2.2.1.3.4.1 of MS-RDPBCGR. I've added the definition to the include file and referenced that. Let me know if/when you want me to squash.

sesman/chansrv/chansrv.c
```diff
@@ -1042,9 +1046,13 @@ my_api_trans_data_in(struct trans *trans)
     int rv;
     int bytes;
     int ver;
-    int channel_name_bytes;
     struct chansrv_drdynvc_procs procs;
-    char *chan_name;
+    /*
+     * Name is limited to CHANNEL_NAME_LEN for an SVC, or MAX_PATH
+     * bytes for a DVC
+     */
```
metalefty, line 1053: Good comment, thanks.

sesman/chansrv/chansrv.c
metalefty: Thanks!

metalefty:
@matt335672 
BTW, I and Jay are willing to invite you to xrdp members. We really appreciate your work not only by code but also on issue triages and answering users' questions. What about that?

After you joined xrdp members, you'll be able to close/reopen/edit issues, editing labels (issue triage), merge PRs (but usually don't) by yourself. Thank you very much for your work on xrdp!

matt335672:
Well, I'm honoured, truly.

In short, 'yes'. At the moment I've got time to work on the project, and if you're prepared to allow me to help in a more hands-on way that would be great.

Thank you.

metalefty:
Anyway, you should join the gitter chatrooms.
* https://gitter.im/neutrinolabs/xrdp
* https://gitter.im/neutrinolabs/xrdp-questions

metalefty:
@matt335672 You've already added to xrdp team. You should now able to close or label issues. Feel free to do that.

PR #1595, opened 2020-05-29
Prevent chansrv input channels being scanned during a server reset

matt335672:
A server reset is used to implement  a client resize.

Currently this is done by sending a TS_DEACTIVATE_ALL_PDU and then resending capabilities with TS_DEMAND_ACTIVE_PDU by `libxrdp_reset()`

It's possible at the moment for channel data to be read from chansrv and sent to the RDP client between these two messages. This is because at the end of `xrdp_mcs_send()` for the 'deactivate all' we call `xrdp_mcs_call_callback()`. If there's pending chansrv data it's sent to the client.

This appears to be at odds with [MS-RDPBCGR 3.2.5.5.1](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/e3dddf9e-ddc7-4622-a0d4-cc1d68f1f1ac) which calls for the client to disable all its input handlers.

This is also what FreeRDP 2.0.0 appears to do. If a message is received between TS_DEACTIVATE_ALL_PDU and TS_DEMAND_ACTIVE_PDU for a channel other than the MCS global channel an error is reported.

This PR disables the chansrv input checks while a `libxrdp_reset()` is in progress.

### DIFF
--- a/libxrdp/xrdp_fastpath.c
+++ b/libxrdp/xrdp_fastpath.c
@@ -146,7 +146,10 @@ xrdp_fastpath_send(struct xrdp_fastpath *self, struct stream *s)
     {
         return 1;
     }
-    xrdp_fastpath_session_callback(self, 0x5556, 0, 0, 0, 0);
+    if (self->session->check_for_app_input)
+    {
+        xrdp_fastpath_session_callback(self, 0x5556, 0, 0, 0, 0);
+    }
     return 0;
 }
 
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -47,6 +47,7 @@ libxrdp_init(tbus id, struct trans *trans)
     session->rdp = xrdp_rdp_create(session, trans);
     session->orders = xrdp_orders_create(session, (struct xrdp_rdp *)session->rdp);
     session->client_info = &(((struct xrdp_rdp *)session->rdp)->client_info);
+    session->check_for_app_input = 1;
     return session;
 }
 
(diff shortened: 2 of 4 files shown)


metalefty:
As my quick reading of the specification, the logic looks good to me but I concern some trivial points.

xrdp/xrdp_types.h
```diff
@@ -382,6 +382,9 @@ struct xrdp_wm
   int hints;
   char pamerrortxt[256];
 
+  /* Disable checks for channel input when channels are unavailable */
+  int disable_channel_input_check;
```
metalefty, line 386: First of all, I'm not strong position on this. However I concern about this may cause double-negative expression and spoil code-readability. If you have a fair reason to choose `disabled_`, it's OK.

matt335672:
I had a think over the weekend, and there may be better ways to do this.

I could push the logic down into the MCS layer which would move it into libxrdp rather than the xrdp process itself. That might make it a bit cleaner.

xrdp/xrdp_wm.c
```diff
@@ -1930,7 +1930,10 @@ callback(intptr_t id, int msg, intptr_t param1, intptr_t param2,
             rv = xrdp_wm_process_channel_data(wm, param1, param2, param3, param4);
             break;
         case 0x5556:
-            rv = xrdp_mm_check_chan(wm->mm);
+            if (!wm->disable_channel_input_check)
```
metalefty, line 1933: Just a thought, this double-negative expression is a little little bit confusing to me non-native English speaker. 
The current code is acceptable because I always see such double-negative conditions everywhere, every project.

metalefty:
Yup, handling in libxrdp layer would be nice.

matt335672:
OK - I'll refactor it down there, retest and update the PR.

Thanks.

matt335672:
Changes now implemented in the session layer, which I'm happier with.

The variable names should be clearer now. Happy to change if necessary.

libxrdp/xrdp_fastpath.c
```diff
@@ -146,7 +146,10 @@ xrdp_fastpath_send(struct xrdp_fastpath *self, struct stream *s)
     {
         return 1;
     }
-    xrdp_fastpath_session_callback(self, 0x5556, 0, 0, 0, 0);
+    if (self->session->check_for_app_input)
+    {
+        xrdp_fastpath_session_callback(self, 0x5556, 0, 0, 0, 0);
```
metalefty, line 151: It's not your fault but us, `0x5556` should be constified someday.

metalefty:
LGTM now but let me some more time to test.

PR #1602, opened 2020-06-10
sesman: work on moving sesman to trans, v0 scp working

jsorg71:


matt335672:
I see you're returning `SCP_SERVER_STATE_INTERNAL_ERR` for missing data (i.e. `s_check_rem()` fails). Would `SCP_CLIENT_STATE_SIZE_ERR` / `SCP_SERVER_STATE_SIZE_ERR` (as appropriate) be a better choice?

I ask, as it's relevant to another change I'm looking at at the moment.

Thanks,

jsorg71:
@matt335672 yes, I noticed that too, I think SCP_SERVER_STATE_SIZE_ERR would be the right one.  I'll change that when I push the v1 changes.  Thanks.

matt335672:
@jsorg71 - I'm quite keen to get the move to the UNIX domain socket completed, as I think it's a good idea to get this in before a fix to #1684 which is a big chunk of work.

When we've got the domain socket move done we can use something like `getpeereid()` for authentication for the management functions used by  `xrdp-sesadmin` which will really help usability.

I've got some effort available at the moment. Would you be happy for me to take your `sesman_trans` branch and work with it?

Thanks.

matt335672:
@jsorg71 - I think we've done this one now(?)

jsorg71:
absolutely, thanks, I'll close this.

PR #1604, opened 2020-06-11
xrdp: add newline at eol of PID file

metalefty:
sesman puts newline at eol of PID file but xrdp don't. I found that it
is reported by Ubuntu user 11 years ago.

Reported by:    https://bugs.launchpad.net/ubuntu/+source/xrdp/+bug/484116

### DIFF
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -623,7 +623,7 @@ main(int argc, char **argv)
         }
         else
         {
-            g_sprintf(text, "%d", pid);
+            g_snprintf(text, "%d", pid);
             g_file_write(fd, text, g_strlen(text));
             g_file_close(fd);
         }


metalefty:
Oops, I was insane. This code don't work at all. I need to take a break. Sorry for the noise.

metalefty:
As far as I read the current code, both xrdp and xrdp-sesman put no newline at eol.  However, I'm not sure which is the standard format.

matt335672:
I can't find any documentation on the standard, but I expect a lot of them did have newlines as they were written (in the past) by System-V scripts using `echo`. I think either should be fine.

metalefty:
Thanks, I misrecognized the issue. The Ubuntu issue is not exist in the current version of xrdp. Closing.

PR #1606, opened 2020-06-14
Read only configuration for virtual X server

RolKau:
When the X server is launched, only the configuration _main file_
xrdp/xorg.conf is specified, but the configuration _directory_ is
retained as default, as is indicated by the output:

    (==) Using config directory: "/etc/X11/xorg.conf.d"

in ~/.xorgxrdp.10.log. If the host has an existing X server running for
the console, and this is not entirely auto-configured, configuration
files from this directory will be picked up and cause the virtual X
server that XRDP launches to attempt to load modules for it, which may
result in errors such as:

    (EE) parse_vt_settings: Cannot open /dev/tty0 (Permission denied)

or

    (EE) xf86OpenConsole: Cannot open virtual console 2 (Permission denied)

Searching the Internet for these errors will typically lead the user to
a myriad of non-solutions, involving added the user to the tty group, or
editing /etc/X11/Xwrapper.config.

This changeset changes the configuration of the virtual X server so that
it considers only /etc/X11/xrdp as its proper configuration directory,
and pick up the xorg.conf file by default from there.

### DIFF
--- a/sesman/sesman.ini.in
+++ b/sesman/sesman.ini.in
@@ -85,8 +85,8 @@ SyslogLevel=DEBUG
 ;
 param=Xorg
 ; Leave the rest paramaters as-is unless you understand what will happen.
-param=-config
-param=xrdp/xorg.conf
+param=-configdir
+param=xrdp
 param=-noreset
 param=-nolisten
 param=tcp


moobyfr:
I had a look some years ago with the autoloading of modules. I came to the conclusion that the HAL part of Xorg loads modules directly without having to look for configuration files. For example, the load of `nv`, `nouveau` or `nvidia` can be made without having to specify it in the configuration file. Does the usage of `configdir` stop these mecanisms ?

RolKau:
> Does the usage of configdir stop ... [automatic loading of nv through HAL]

I am not really sure, but I _think_ that the loading code works by first reading the file specified by -config, then appending all the files found in -configdir, and in the end auto-configuring what is missing. I know that with an X server connected to real hardware, you can for instance specify mouse settings in /etc/X11/xorg.conf.d, and have the monitor hardware probed automatically.

RolKau:
I found that there is another potential bug lurking: If -config is not specified, it will read the default /etc/X11/xorg.conf, which of course can contain whatever irrelevant to xrdp. There doesn't seem to be some way of saying that you _don't_ want a -config or -configdir (other than setting it to an invalid directory, which leaves an error message); over-specifying by pointing both to the same destination seems to work just as fine.

RolKau:
> load of `nv`, `nouveau` or `nvidia` can be made without having to specify it in the configuration file. Does the usage of `configdir` stop these mecanisms

FWIW, if I add myself to the render group (I am on a Debian derivative):

    sudo adduser $(whoami) render

then I get these lines in ~/.xorgxrdp.10.log

    [598498.241] rdpProbe: found DRMDevice xorg.conf value [/dev/dri/renderD128]
    [598498.241] rdpProbe: found DRI3 xorg.conf value [1]
    [598498.241] (II) XRDPDEV(0): using default device
    ...
    [598498.241] rdpPreInit: /dev/dri/renderD128 open ok, fd 10
    ...
    [598498.241] rdpPreInit: glamor module load ok
    [598498.291] (II) XRDPDEV(0): glamor X acceleration enabled on Radeon RX 580 Series (POLARIS10, DRM 3.35.0, 5.4.0-42-generic, LLVM 10.0.0)
    [598498.291] rdpPreInit: glamor init ok

even if both -config and -configdir are now redirected to /etc/X11/xrdp

PR #1623, opened 2020-06-29
Fixed CVE-2020-4044 CI errors

matt335672:
Fix errors identified by Travis CI in PR for above CVE.

In all cases where a tui32 has been replaced by an int, the read value is followed by a bounds check for both less-than or greater-than, so no unsigned overflows have been introduced by this change


metalefty:
Thanks!

PR #1638, opened 2020-07-22
Fixing the build with --enable-xrdpdebug and CFLAGS=-Werror -Wall

aquesnel:
This pull request fixes the build errors that occur when xrdp is built with --enable-xrdpdebug and CFLAGS=-Werror -Wall. This pull request also configures the build to use -Werror -Wall as the default CFLAGS for make when run on a veleopper's machine or on the build servers.

This issue was debugged by @matt335672 in https://github.com/neutrinolabs/xrdp/pull/1633#issuecomment-660892330 and this change should prevent this issue from affecting other developers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,17 @@ max_amd64_conf: &max_amd64_conf
       packages:
         - *max_amd64_deps
 
+max_debug_amd64_conf: &max_debug_amd64_conf
+  env:
+    - CONF_FLAGS="--enable-ipv6 --enable-jpeg --enable-fuse --enable-mp3lame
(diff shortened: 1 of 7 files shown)


metalefty:
Why closed?

aquesnel:
I had made this pull request too eagerly before I saw that the changes broke the i386 build. I didn't think this pull request was worth looking at yet, so I closed the pull requst. I'll make a new pull request once I figure out how I broke the bulid and I fix it.

aquesnel:
this pull request is a broken duplicate of pull request https://github.com/neutrinolabs/xrdp/pull/1640

Issue #1588, opened 2020-05-18
Feature request - separate config file for xrdp and sesman


jkeltz:
I install xrdp on 100's of hosts.  I have a config file directory that is automatically duplicated to all the systems in my department.  My /etc/xrdp is replaced by [config dir]/xrdp, but since the directory is the same everywhere, I then lose the ability to customize my installation because [config dir] is the same on every system.  On certain systems, I need changes in just the sesman.ini file.  I would like xrdp-sesman to be able to accept on the command line a "-c" option for config file.  In this way, I could have [config dir]/xrdp/sesman.ini.[host].  My systemd start script for xrdp-sesman could start with the host-specific config if one was available, or the default otherwise.  This is what I do with most other software packages.  However, xrdp doesn't offer the "-c" option on daemon startup.  It would be nice to put this into xrdp and xrdp-sesman.  
Thanks for your consideration.

derekschrock:
Couldn't you just symlink the host-based conf. to xrdp.ini or sesman.ini?

jkeltz:
The configuration system (which I inherited)  duplicates the configuration  tree on every host, so I can't.  If I change it on one system then it would be reverted.  That's why being able to specify the configuration file on startup would be perfect.
On May 21, 2020, at 11:12 p.m., Derek Schrock <notifications@github.com<mailto:notifications@github.com>> wrote:

Couldn't you just symlink the host-based conf. to xrdp.ini or sesman.ini?

matt335672:
I think this could be quite a handy feature, particularly for testing - I could have made use of it very recently.

I have a look when I get a mo', if no-one else already has.

matt335672:
I've had a look at this, and the obvious thing is to add a `-c` option to xrdp and sesman so that the main config file for these two programs can be overridden.

Two less obvious implications are:-
- The `sesrun` tool needs the same argument. The other tools should probably need it too, but at the moment they have other ways to find `sesman`.
- The xrdp daemon also reads `sesman.ini` to get the port to connect to sesman.  This override will not be available for xrdp, so if you're using this facility sesman will need to be listening on the standard port.

If anyone's not happy with that, please say something.

PR #1650, opened 2020-08-10
Add -c/--config option to sesman

matt335672:
This PR is the second of three required to implement #1588 (separate config file for xrdp and sesman).

The first PR is #1635, and this PR depends on that one for correct chansrv operation. The last PR (TBD) is independent of the first two and concerns xrdp only.

This PR makes the following changes:-
### sesman
- command-line argument parser added which adds common functionality with XRDP (`--help` and `--version` arguments)
- Simplification of `config.c` interface. The `struct config_sesman` is allocated as part of the read, and unnecessary global functions are removed.
- ~~Passes --config on to chansrv when it is started (hence dependency on #1635)~~ (8/9/2020 - no longer needed)
- Manpage updated

### sesrun
- --config argument added here too for completeness.
- Manpage updated (and corrected)

### DIFF
--- a/sesman/sesman.c
+++ b/sesman/sesman.c
@@ -28,7 +28,19 @@
 #include <config_ac.h>
 #endif
 
+#include <stdarg.h>
+
 #include "sesman.h"
+#include "xrdp_configure_options.h"
+
+struct sesman_startup_params
+{
+    const char *sesman_ini;
+    int kill;
+    int no_daemon;
+    int help;
+    int version;
+};
 
 int g_sck;
 int g_pid;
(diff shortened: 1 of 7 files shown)


sesman/sesman.c
```diff
@@ -177,110 +282,142 @@ sesman_main_loop(void)
     return rv;
 }
 
+/*****************************************************************************/
+static void
+print_version(void)
+{
+    g_writeln("xrdp-sesman %s", PACKAGE_VERSION);
+    g_writeln("  The XRDP session manager");
```
metalefty, line 290: We see both the uppercase XRDP and the lowercase xrdp. There are no explicit rules but there more lowercase forms than uppercase forms. This is really my personal opinion, the lowercase xrdp looks cooler.

No problem, this is not important at all.

metalefty (approved):

sesman/sesman.c
matt335672: Let's go with that then.

I've added a [page](/neutrinolabs/xrdp/wiki/Referring-to-the-project) to the wiki  specifying this as the project preference. It's linked from the main page.

PR #1651, opened 2020-08-11
Added -c /--config option to xrdp

matt335672:
This is the third and last PR in a series to implement #1588  (separate config file for xrdp and sesman).

The other two are for sesman (#1650) and chansrv (#1635), and are largely independent of this PR. This PR affects xrdp and libxrdp.

This PR makes the following changes:-

### common/log.c
Opened the .ini file read-only with `g_file_open_ex` rather than `g_file_open`. Without this change, the logging code creates an empty ini file if one wasn't present, which is probably not what is required!

### xrdp.c
- A few methods were made static where not required to be global
- File run through astyle which found a few issues.
- A few hard-coded string value sizes have been removed.
- local variable `no_daemon` replaced with `daemon` to avoid double-negatives (i.e. `!no_daemon`)

### xrdp (other files)
- Manpage updated
- Other changes made to remove hard-coded references to the location of xrdp.ini

### libxrdp
- Extra parameter added to `libxrdp_init()` which allows the location of `xrdp.ini` to be specified for the session. NULL can be specified for this parameter to implement the existing behaviour with the default file.

### DIFF
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -22,6 +22,8 @@
 #include <config_ac.h>
 #endif
 
+#include <stdarg.h>
+
 #include "xrdp.h"
 #include "log.h"
 #include "xrdp_configure_options.h"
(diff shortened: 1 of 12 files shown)


xrdp/xrdp.c
```diff
@@ -417,14 +443,14 @@ main(int argc, char **argv)
...
     int exit_status = 0;
     int test;
-    char cfg_file[256];
     enum logReturns error;
-    struct xrdp_startup_params *startup_params;
+    struct xrdp_startup_params startup_params = {0};
     int pid;
     int fd;
-    int no_daemon;
+    int daemon;
```
metalefty, line 450: Good point.

metalefty:
@matt335672  Hi, it's been a long time.  Overall, it looks good to me. Let's ship it if you're sure it works.

matt335672:
Hi @metalefty - good to hear from you.

Yes, I'm happy with it - I just wanted you to check it before a merge. I'll do that then.

The PR I'd really like to see merged in is #1633. I'm happy with it, but until it's merged it's acting as a bit of a blocker. I've got other changes to chansrv which need to be submitted but I don't want to do so until the logging is stable. I'd be grateful if you could take a look at it when you're able.

Thanks,

matt335672:
Just find a minor issue around libxrdp initialisation. I'll fix it and re-test thoroughly before I merge.

Issue #1648, opened 2020-08-07
Internal username / password (256) buffer is smaller than max username / password in rdp protocol (511)


bolkedebruin:
While experimenting with SSO with JWT tokens supplied in the username field I encountered that usernames above 256 characters are not accepted (xrdp drops the connection). However on the protocol level it is specified te be a maximum of 511 characters. 

The internal buffers in xrdp use indeed 256 although it checks for 511 when accepting the connection

I suggest (and will provide a PR) to have a larger buffer.

metalefty:
Good catch.

PR #1653, opened 2020-08-16
Fix buffer len for client info

bolkedebruin:
The MS docs state that the maximum length of a username/password/domain etc is 511. xrdp checks for this but it's internal buffers are 256. This makes it into a define and sets the maximum size to 512 (null terminator needs to be counted)

Fixes: #1648

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -18,6 +18,8 @@
  * xrdp / xserver info / caps
  */
 
+#include "xrdp_constants.h"
+
 #if !defined(XRDP_CLIENT_INFO_H)
 #define XRDP_CLIENT_INFO_H
 
@@ -57,11 +59,11 @@ struct xrdp_client_info
   char hostname[32];
   int build;
   int keylayout;
-  char username[256];
-  char password[256];
-  char domain[256];
-  char program[256];
-  char directory[256];
+  char username[INFO_CLIENT_MAX_CB_LEN];
+  char password[INFO_CLIENT_MAX_CB_LEN];
+  char domain[INFO_CLIENT_MAX_CB_LEN];
+  char program[INFO_CLIENT_MAX_CB_LEN];
+  char directory[INFO_CLIENT_MAX_CB_LEN];
   int rdp_compression;
   int rdp_autologin;
   int crypt_level; /* 1, 2, 3 = low, medium, high */
--- a/common/xrdp_constants.h
+++ b/common/xrdp_constants.h
@@ -38,6 +38,10 @@
  ******************************************************************************/
 
 #define INFO_CLIENT_NAME_BYTES  32
+/**
+ * Maximum length of a string (two bytes + len), excluding the terminator
+ * [MS-RDPBCGR] TS_INFO_PACKET(2.2.1.11.1.1)
+ */
 #define INFO_CLIENT_MAX_CB_LEN  512
 
 #define XRDP_MAX_BITMAP_CACHE_ID  3
--- a/common/xrdp_constants.h
+++ b/common/xrdp_constants.h
@@ -39,7 +39,7 @@
 
 #define INFO_CLIENT_NAME_BYTES  32
 /**
- * Maximum length of a string (two bytes + len), excluding the terminator
+ * Maximum length of a string including the mandatory null terminator
  * [MS-RDPBCGR] TS_INFO_PACKET(2.2.1.11.1.1)
  */
 #define INFO_CLIENT_MAX_CB_LEN  512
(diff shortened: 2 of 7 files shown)


vnc/vnc.h
```diff
@@ -133,8 +133,8 @@ struct vnc
     int mod_mouse_state;
     int palette[256];
     int vnc_desktop;
-    char username[256];
-    char password[256];
+    char username[8192];
```
metalefty, line 136: Does this come from VNC protocol specification?

bolkedebruin: RealVNC (authoritative) is currently 255 characters. Older legacy implementations may still be 8. RFB protocol itself is agnostic[0]. Since the user password is used as the DES key, and DES key size is limited to 56 bits (plus 8 parity bits), your key can only be up to 7 8-byte characters long. If the password is shorter than 8 characters, it's just padded with zeroes.

But let me rebase, as you are looking at an older commit

bolkedebruin:
Ping @metalefty

metalefty:
@matt335672 can you also review this?

matt335672:
Will do.

common/xrdp_constants.h
```diff
@@ -38,6 +38,7 @@
  ******************************************************************************/
 
 #define INFO_CLIENT_NAME_BYTES  32
+#define INFO_CLIENT_MAX_CB_LEN  512
```
matt335672, line 41: This is an interesting one. The value is directly derived from [MS-RDPBCGR] TS_INFO_PACKET(2.2.1.11.1.1), but there's no actual name for this length (as far as I can tell). It might be better placed in common/ms-rdpbcgr.h than here, but given the lack of an actual definition on the docs I don't really have string feelings either way. If it stays here, a reference to where it's from would be useful - either [MS-RDBBCGR] of elsewhere.

bolkedebruin: Can do that, it is also in libscp_v0 btw (definition seem to be all over the place or hardcoded :/).

bolkedebruin:
Done @matt335672

metalefty:
LGTM. I think this should be shipped to August release.

metalefty (approved):

common/xrdp_client_info.h
```diff
@@ -57,11 +59,11 @@ struct xrdp_client_info
...
-  char username[256];
-  char password[256];
-  char domain[256];
-  char program[256];
-  char directory[256];
+  char username[INFO_CLIENT_MAX_CB_LEN];
+  char password[INFO_CLIENT_MAX_CB_LEN];
+  char domain[INFO_CLIENT_MAX_CB_LEN];
+  char program[INFO_CLIENT_MAX_CB_LEN];
+  char directory[INFO_CLIENT_MAX_CB_LEN];
```
metalefty, line 66: This fix is a good catch. However, this will break compatibility with xorgxrdp. 

See about 80 lines below. No worries but I need to take care about xrdp and xorgxrdp version in sync.
xorgxrdp v0.2.14 (to be released in August) must depend on v0.9.14.

metalefty: ref. https://github.com/neutrinolabs/xrdp/pull/542#discussion_r92231986

bolkedebruin: Got it, it would be nice if that "comment" inside the source was a bit clearer on the why or have the file detail on how it is used or what the assumptions are.

common/xrdp_constants.h
```diff
@@ -38,6 +38,10 @@
  ******************************************************************************/
 
 #define INFO_CLIENT_NAME_BYTES  32
+/**
+ * Maximum length of a string (two bytes + len), excluding the terminator
+ * [MS-RDPBCGR] TS_INFO_PACKET(2.2.1.11.1.1)
+ */
 #define INFO_CLIENT_MAX_CB_LEN  512
```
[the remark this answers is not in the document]
bolkedebruin: Sure, it reads a bit ambigous. I'll fix that

xup/xup.h
```diff
@@ -154,8 +155,8 @@ struct mod
   int height;
   int bpp;
   int sck_closed;
-  char username[256];
-  char password[256];
+  char username[INFO_CLIENT_MAX_CB_LEN];
+  char password[INFO_CLIENT_MAX_CB_LEN];
```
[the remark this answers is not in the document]
bolkedebruin: @matt335672 in `xup.h` I made the change. Or is there anything else you are referring to?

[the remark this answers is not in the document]
matt335672: In `xup/xup.c` at (your) line 1549 and 1553, the code to initialise these two fields is still capping the lengths at 255 rather than `INFO_CLIENT_MAX_CB_LEN-1`. 

I don't think the VNC client needs updating at all - sorry about the noise. I see from the comments above that you were looking in this area with earlier commits which have now been removed. The VNC username/password interface could certainly do with tidying up, but that can be left alone for now.

PR #1662, opened 2020-08-24
Adding checks to prevent buffer overruns during data chunk re-assembly

aquesnel:
Problem:
The data chunk buffer is allocated based on the total length field in a DYNVC_DATA_FIRST message. Upon receiving subsequent DYNVC_DATA messages, the data payload is appended to the chunk buffer without first checking that there is enough space remaining in the chunk buffer. This could lead to a buffer overrun in the server.

Solution:
This pull request adds a check to verify that the chunk buffer has enough remaining space to accept the received payload, and returns an error if there is not enough space in the buffer.


metalefty:
I think this can be merged.

matt335672:
Agreed.

PR #1663, opened 2020-08-25
Detect and setup Programmer Dvorak keyboard layout

RolKau:
This patch will add the keyboard layout identifier for the Programmer
Dvorak keyboard layout, so that if a Windows client has that layout
active, it will be mapped to the corresponding xkeyboard-config layout
in the X server.

An XFreeRDP client knows about this layout too, and will correspondingly
map it to the identifier given here, making the layout propagate
correctly through X-to-X connections as well.

To replicate the full Windows keyboard layout several options have to
be set as well as the main layout. To avoid having these options spill
over to other layouts that are dumped, the old settings are stored
before the dump and then restored afterwards.


instfiles/Makefile.am
```diff
@@ -50,7 +50,9 @@ dist_startscript_DATA = \
   km-00000813.ini \
   km-00000816.ini \
   km-0000100c.ini \
-  km-00010409.ini
+  km-00010409.ini \
+  km-19360409.ini \
```
metalefty, line 54: Backslash must not be put at the end.

metalefty (changes requested):
Nice first contribution, thank you!
Can you fix a tiny breakage?

metalefty:
```
[20200817-18:05:31] [INFO ] TLS connection established from ::ffff:192.168.10.20 port 40455: TLSv1.3 with cipher TLS_AES_256_GCM_SHA384
[20200817-18:05:31] [DEBUG] xrdp_0000dc44_wm_login_mode_event_00000001
[20200817-18:05:31] [INFO ] Loading keymap file /usr/local/etc/xrdp/km-00000409.ini
[20200817-18:05:31] [WARN ] local keymap file for 0x00000409 found and doesn't match built in keymap, using local keymap file
```

Also, can you attach the log `km-19360409.ini` is properly loaded when your patch applied? Like above.

instfiles/Makefile.am
RolKau: It was actually intentional to have a blank line at the end, to avoid having every change overwrite the blame of the previous line to add the backslash, but I'm not invested in it, so it's changed now.

RolKau:
> Also, can you attach the log `km-19360409.ini` is properly loaded when your patch applied? Like above.

I have these four lines in my /var/log/xrdp.log when logging on with this layout specified:

    [20200826-13:03:35] [DEBUG] xrdp_000cc3f5_wm_login_mode_event_00000001
    [20200826-13:03:35] [INFO ] Loading keymap file /etc/xrdp/km-19360409.ini
    [20200826-13:03:35] [WARN ] local keymap file for 0x19360409 found and doesn't match built in keymap, using local keymap file
    [20200826-13:03:41] [DEBUG] xrdp_wm_log_msg: connecting to sesman ip 127.0.0.1 port 3350

(I read the source code but I couldn't figure out why the message was classified as "WARN")

instfiles/Makefile.am
metalefty: Thanks, it actually caused a build failure.

metalefty:
> I read the source code but I couldn't figure out why the message was classified as "WARN"

Indeed, that should be reconsidered.

PR #1666, opened 2020-08-28
Adding a Travis CI build using Ubuntu Focal

aquesnel:
Follow up for #1659 to add a CI build using the Ubuntu Focal platform version of the gcc compiler and dependencies.

This pull request depends on compiler error fixes in #1659 being merged.

### Open questions

* should all CI builds be updated to use the latest platform/compiler versions? Or should one/most builds use the older platform version to verify that xrdp continutes to build on the older platform?

#### TODO:

* add/upgrade a clang build to the latest version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false # use new container
 language: c
+dist: focal
 
 branches:
   except: /^(gh-pages|v0\.[0-8].*)/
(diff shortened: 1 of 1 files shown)


matt335672:
Thanks for raising this.

At the moment the benefits of this PR can be seen here, as #1659 has not been merged:-

[https://travis-ci.org/github/aquesnel/xrdp/builds/721903430](https://travis-ci.org/github/aquesnel/xrdp/builds/721903430)

My own preference would be to turn this round and make all the builds `focal` apart from a single `xenial` build. However, when I tried this myself I discovered libfdk-aac-dev [seems to be missing for i386](https://travis-ci.org/github/matt335672/xrdp/builds/721954540).

So another approach might be to keep i386 at xenial (for 'legacy') and move everything else forward to focal.

I hope this helps - sorry if it raises more questions than it answers.

aquesnel:
I've updated this pull request with @matt335672 's suggestion of using the latest Ubuntu for all 64-bit builds, and only using ubuntu xenial for the 32-bit builds.

matt335672:
I'm certainly happy with this. I just pulled the commits into a temp branch in my own repo and ran it through Travis. Results are [here](https://travis-ci.org/github/matt335672/xrdp/builds/725519201). Using the `name:` tags helps the readability I think.

What do you think @metalefty?

metalefty:
Totally agree.

aquesnel:
### note

This pull request depends on the pull request #1659 being merged in order to not break the build. Since #1659 has been merged, this pull request is now unblocked and can be safely merged once it has been approved.

metalefty (approved):

Issue #1667, opened 2020-08-28
Proposal: Allow token sign in without autologon for SSO


bolkedebruin:
**Challenge**

RDP files do not allow storing for passwords that are set by the server. This limits single sign on possibilities without NLA. So for example if you use a remote desktop gateway (like, shameless plug, https://github.com/bolkedebruin/rdpgw) which creates a downloadable RDP file you cannot make this single sign on at the moment.

**Proposed solution**
If a "enable_token_login" is specified in the config file and a ASCII record separator character is detected everything before the separator is considered to be the username and everything after the separator is considered to be the authentication token (like JWT). So something like

bolke|JWT_TOKEN

At logon time the username is scanned and if the ASCII record separator is found the token is copied over to the password and the username is set to everything before the separator.

Copying the token to the password very early in the provess ensures that the password does not get accidentally logged. Detection is also important to ensure a login attempt does not get triggered without cause and users do not get locked out. 

Of course an alternative separator (character or string) can be used, but not that the username string is limited to 511 characters.

I have a working solution.

what do you thing @matt335672 @metalefty @jsorg71 ?

matt335672:
As a workaround for your particular use case it sounds OK to me.

I think it's important it's disabled by default otherwise we'll be encouraging users to store passwords in .ini files which won't go down well with enterprise security departments.

bolkedebruin:
@matt335672 that makes sense. The linked PR is doing exactly that. Do you mind reviewing it?

matt335672:
Since we've merged the PR, I'll close this.

PR #1668, opened 2020-08-29
Add support for token authentication

bolkedebruin:
This feature allows to embed a token in the username field. Tokens
are separated from the username by the ASCII field separator character
0x1F (unicode 0x001F).

See https://github.com/neutrinolabs/xrdp/issues/1667

cc @metalefty @matt335672

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,3 +12,4 @@ as possible.
 Our email eddress for security report is:
 
 * [xrdp-core@googlegroups.com](mailto:xrdp-core@googlegroups.com)
+
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -9,7 +9,8 @@ If you find a security vulnerability please kindly inform us about the problem i
 so that we can fix the security problem to protect a lot of users around the world as soon
 as possible.
 
-Our email eddress for security report is:
+Our email address for security report is below. This is a private mailing list and not open 
+for public viewing.
 
 * [xrdp-core@googlegroups.com](mailto:xrdp-core@googlegroups.com)
 
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2591,6 +2591,19 @@ g_strcat(char *dest, const char *src)
 }
 
 /*****************************************************************************/
+/* returns dest */
+char *
+g_strncat(char *dest, const char *src, int len)
+{
+    if (dest == 0 || src == 0)
+    {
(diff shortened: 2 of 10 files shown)


SECURITY.md
```diff
@@ -12,3 +12,4 @@ as possible.
 Our email eddress for security report is:
 
 * [xrdp-core@googlegroups.com](mailto:xrdp-core@googlegroups.com)
+
```
metalefty, line 15: Is this an intended change?

bolkedebruin: Ah no, but I think I should make it so: see the typo aboven and does this email list actually exist? I can only find xrdp-devel.

metalefty: Yes, the mailing list certainly exists. It is not a public list because the purpose of the list is to receive security (vulnerability) reports privately.

bolkedebruin: Ah gotcha. It might be nice to make that evident as the name of the maling list doesnt make that very clear (=assumption)

matt335672:
I've had a look at this too, and I've got no comments - looks very readable.

SECURITY.md
metalefty: Will do, thanks for the point.

metalefty:
I haven't understood what's the benefit of this PR (just haven't read #1667 cover to cover) but the code looks very readable.

You don't need to wait until I understand the benefit. If there's no disbenefit, feel free to merge this.

bolkedebruin:
@metalefty sure I understand that if you read it out of context that might be the case. The PR allows you to store, for example, a JWT that is short lived in an RDP file. So we deploy this together with a remote desktop gateway (https://github.com/bolkedebruin/rdpgw) that allows to embed custome tokens. We then enable "pam_jwt" (https://github.com/bolkedebruin/pam-jwt) for xrdp-sesman to verify those token. The end result for the user is that it is single sign on without the need for kerberos / network level authentication.

matt335672:
Hi @bolkedebruin 

One thing I missed - could you add a brief description of the extra parameters to `docs/man/xrdp.ini.5.in` then we can merge this.

Thanks

bolkedebruin:
Absolutely. Next week most likely AFK mostly

bolkedebruin:
@matt335672 ready for review/merge

matt335672:
That's great - thanks.

PR #1673, opened 2020-09-01
Musl fixes

maxice8:


matt335672:
Sorry this is inline but I'm having problems with the review function right now.

Can you replace the 'Needed for putpwent' comment with 'Needed for putpwent on MUSL'?

Most of the user are expecting libc, and this won't make any sense on those platforms as the test macros are different. I can't see any platform-specific way to detect MUSL.

Also move the whole define up above all the includes. From https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html:-

> You should define these macros by using ‘#define’ preprocessor directives at the top of your source code files. These directives must come before any #include of a system header file. It is best to make them the very first thing in the file, preceded only by comments. You could also use the ‘-D’ option to GCC, but it’s better if you make the source files indicate their own meaning in a self-contained way.

PR #1675, opened 2020-09-02
Update xrdp_keyboard.ini for Norwegian keyboards

per-ok:
Connecting via RDP using Norwegian keyboard always results in a default (US English?) keyboard mapping.


metalefty:
The keymap file `km-00000414.ini` is also needed.

metalefty:
See also other `new keyboard layout` labeled PRs.
https://github.com/neutrinolabs/xrdp/pulls?q=is%3Apr+label%3A%22new+keyboard+layout%22+is%3Aclosed

per-ok:
I believe it is already there:
https://github.com/neutrinolabs/xrdp/blob/devel/instfiles/km-00000414.ini

Issue #1678, opened 2020-09-04
Previous warnings in chansrv_fuse.c are now errors due to -Werror


gdsotirov:
There's a difference in latest [version 0.9.14](https://github.com/neutrinolabs/xrdp/releases/tag/v0.9.14) compared to [0.9.13.1](https://github.com/neutrinolabs/xrdp/releases/tag/v0.9.13.1) coming from commit 7e58209 by @aquesnel, which now turns all warnings into errors on Linux (i.e. with `-Werror` compiler option). However, thus previous warnings in chansrv_fuse.c are now turned into errors and thus are blocking the build of the software.

The relevant part of my build log is:

```
Making all in chansrv
make[3]: Entering directory '/usr/src/tmp/xrdp-0.9.14/sesman/chansrv'
i586-slackware-linux-gcc -DHAVE_CONFIG_H -I. -I../..  -DXRDP_CFG_PATH=\"/etc/xrdp\" -DXRDP_SBIN_PATH=\"/usr/sbin\" -DXRDP_SHARE_PATH=\"/usr/share/xrdp\" -DXRDP_PID_PATH=\"/var/run\" -DXRDP_SOCKET_PATH=\"/tmp/.xrdp\" -I../../common  -DXRDP_FUSE -D_FILE_OFFSET_BITS=64 -I/usr/include/fuse  -DXRDP_OPUS -DXRDP_MP3LAME    -O3 -march=i586 -mtune=i686 -Wall -Wwrite-strings -Werror -O2 -c -o chansrv_fuse.o chansrv_fuse.c
chansrv_fuse.c: In function 'xfuse_cb_readdir':
chansrv_fuse.c:1610:20: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     else if ((dh = (struct xfs_dir_handle *) fi->fh) == NULL)
                    ^
chansrv_fuse.c: In function 'xfuse_cb_read':
chansrv_fuse.c:2098:15: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     if ((fh = (XFUSE_HANDLE *)fi->fh) == NULL)
               ^
chansrv_fuse.c: In function 'xfuse_cb_write':
chansrv_fuse.c:2171:15: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     if ((fh = (XFUSE_HANDLE *)fi->fh) == NULL)
               ^
[...]

PR #1686, opened 2020-09-12
Use a single type for the fuse file handle callback context

aquesnel:
Follow up from the discussion in #1678 about the multiple types assigned to the fuse file handle callback context.

#### problem

fuse_file_info.fh field is used to store a pointer two different types: an xfuse_handle and a xfs_dir_handle. There is no type identifier in the handle, so without further investigation it is not clear if there type error in the code.

#### solution

use a single type in `fuse_file_info.fh` by adding a `xfs_dir_handle` pointer as a field of `xfuse_handle` and then only set the `xfuse_handle` pointer as the `fuse_file_info.fh` field.

#### risks

* No testing performed.
  * I haven't setup xrdp with a client that uses the fuse file system for manual testing, nor do I know which rdp functionality to execute to exercise this code.
  * No unit test coverage of this functionality 
* I'm not an expert on the fuse callbacks, therefore there might be edge cases of memory leaks that I'm not aware of. I tried to follow the existing logic for allocation and freeing of the structs to reduce the risk of introducing a bug.

### DIFF
--- a/sesman/chansrv/chansrv_fuse.c
+++ b/sesman/chansrv/chansrv_fuse.c
@@ -298,6 +298,14 @@ struct xfuse_handle
     tui32 DeviceId;
     tui32 FileId;
     int   is_loc_resource; /* this is not a redirected resource */
+    
+    /* a directory handle, if this xfuse_handle represents a directory.
+     * NULL, if this xfuse_handle represents a file.
+     *
+     * Note: when this xfuse_handle represents a directory, then the other 
+     *       fields of this structure contain invalid values.
+     */
+    struct xfs_dir_handle *dir_handle;
 };
 typedef struct xfuse_handle XFUSE_HANDLE;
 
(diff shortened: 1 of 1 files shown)


matt335672:
Looks good, but I haven't tested it yet.

Couple of comments at this stage:-
- Probably worth avoiding mixing `intptr_t` and `tintptr` in the same module. Could you stick with `tintptr` for now? I believe `intptr_t` was introduced with C99, and we don't currently require that. Personally I think we should be moving in that direction but that could be debated with a separate PR.
- Please add a comment for the `dir_handle` member variable, as it's only required for directory operations.

Thanks again

matt335672:
Thanks for the prompt response.

I've run regression checks on this by pulling and building xrdp on a redirected filesystem and it all works fine.

Can you fix the two 'dierctory' mis-spellings and squish this into a single commit? Then I think we can merge it.

matt335672:
Thanks for the contribution!

PR #1688, opened 2020-09-15
Update README.md

zero77:
This quickly shows all package versions across different distros at a glance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
 
 *Current Version:* 0.9.14
 
+[![Packages](https://repology.org/badge/latest-versions/xrdp.svg)](https://repology.org/metapackage/xrdp/versions)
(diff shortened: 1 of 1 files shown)


matt335672:
It's a nice idea.

I'd like to see some sort of disclaimer however that the link is provided by a third party which is nothing to do with project directly.

@metalefty - what do you think?

zero77:
It may be a bit big but, this is more self explanatory.
[Packaging status](https://repology.org/project/xrdp/versions)

matt335672:
Thanks,

I've got a slight preference for a link rather than an image myself but I'm not that fussed personally about the presentation.

However it's done, I think it's important that we make it clear that the information is being provided by a third party which is nothing to do with us - we have no control over what is actually being presented by the website on the other end of the link.

I think it would be wrong to give an impression to the casual visitor that we're responsible for any packaging. Otherwise we'll get packaging issues raised against us which we can't do anything about directly.

PR #1691, opened 2020-09-18
genkeymap: Fix an array declaration conflict

FelixZhang:
The conflict on the size of xfree86_to_evdev between genkeymap.c and
evdev-map.c is causing build failures in openSUSE builds

### DIFF
--- a/genkeymap/genkeymap.c
+++ b/genkeymap/genkeymap.c
@@ -44,7 +44,7 @@
 #include <X11/XKBlib.h>
 #include <locale.h>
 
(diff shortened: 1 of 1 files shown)


matt335672:
Nice find.

I'm sure this could do with tidying up (there's a lot of magic numbers in here), but I can see why the static analysis is picking it up.

metalefty:
Hi @FelixZhang , it's been a while. Thank you for your xrdp housekeeping for openSUSE. Nice find from me, too.

Issue #1612, opened 2020-06-19
Support for HiDPI


okhowang:
I use an Thinkpad X1 with 200% DPI.
when I connect to xrdp, remote X server's dpi is 96.
but when I connect to windows rdp, system will process dpi to 200% auto.

metalefty:
I think it depends on the WM.

okhowang:
but when I use Thinkpad X1 to connect xrdp
```shell
$ xdpyinfo | grep resolution
  resolution: 96x96 dots per inchs
```
It seem like incorrect dpi was reported by virtual screen

metalefty:
What's your backend?

okhowang:
xorgxrdp

metalefty:
okay cool

okhowang:
@metalefty 
I have a branch dpi for support monitor info in [xrdp](https://github.com/okhowang/xrdp/tree/dpi) and [xorgxrdp](https://github.com/okhowang/xorgxrdp/tree/dpi).
but there is some trouble when processing scaling for xorg server.
do you have any idea in this?
```
static RRTransformPtr
rdpGetTransform(uint32_t scale, int width, int height) {
    double d_scale;
    XFixed f_scale;
    if (scale == 0 || scale == 100) return NULL;
    d_scale = (double)scale / 100;
    f_scale = XDoubleToFixed(d_scale);
    RRTransformPtr transform = malloc(sizeof(RRTransformRec));
    RRTransformInit(transform);
    pixman_transform_init_scale(&transform->transform, f_scale, f_scale);
//    PictFilterPtr filter = malloc(sizeof(PictFilterRec));
//    filter->name = PictureGetFilterName(PictFilterBest);
//    filter->ValidateParams = NULL;
//    filter->id = PictFilterBest;
//    filter->height = height;
//    filter->width = width;
//    RRTransformSetFilter(transform, filter, &f_scale, 0, width, height);
    return transform;
}
```
is filter necessary? and how make transform work properly?

blackcrack:
by the way, KDE/plasma have also a bit problem with HiDPI with TV's(WebOS) with/without kscreen2,
have make a bugreport under bugs.kde.org
https://bugs.kde.org/show_bug.cgi?id=425111
just as info and hint ...

Nexarian:
[...]

PR #1692, opened 2020-09-22
parse monitor/display info

okhowang:
add parsing monitorex pdu and display info in core data pdu,
and store it into client_info
relate #1612


common/xrdp_client_info.h
```diff
@@ -38,6 +46,11 @@ struct xrdp_client_info
   int bpp;
   int width;
   int height;
+  int physical_width;
+  int physical_height;
+  int orientation;
+  int desktop_scale_factor;
+  int device_scale_factor;
```
metalefty, line 53: These new variables must be appended to the end.

okhowang: fixed

PR #1702, opened 2020-10-19
startwm.sh: Additionally support /usr/etc/X11/xdm/Xsession

yifanjiang:
SUSE distributions are switched to use /usr/etc for system configuration, Xsession startup files is involved.

https://lists.opensuse.org/opensuse-factory/2019-08/msg00113.html


metalefty:
Merged, thanks! Reference is always important.

Issue #1634, opened 2020-07-14
xrdp memory leak


yuyiliang1993:
Playing video in an xrdp session will cause memory growth for a long time, and the memory leak in the vpn environment is more serious. How to solve it?

matt335672:
Hi @yuyiliang1993 

Can you please give us a bit more information?

The following would be useful:-
- Server operating system
- Version of XRDP
- Does this happen with Xorg backend or Xvnc backend or both?
- If Xorg backend, what is the version of xorgxrdp?
- If Xvnc backend, what is the colour depth you are using?
- Which processes are using more memory

Thanks,

yuyiliang1993:
Hello there
I connected the Windows target machine through the neutrinordp module, and then watched the video, and found that the memory of the xrdp process increased relatively fast, and the CPU occupancy rate was 100%, especially in the vpn environment. The version I use is xrdp-0.9.11.
Thank you

yuyiliang1993:
> Hi @yuyiliang1993
> 
> Can you please give us a bit more information?
> 
> The following would be useful:-
> 
> * Server operating system
> * Version of XRDP
> * Does this happen with Xorg backend or Xvnc backend or both?
> * If Xorg backend, what is the version of xorgxrdp?
> * If Xvnc backend, what is the colour depth you are using?
> * Which processes are using more memory
> 
> Thanks,

yuyiliang1993:
Hello there
I connected the Windows target machine through the neutrinordp module, and then watched the video, and found that the memory of the xrdp process increased relatively fast, and the CPU occupancy rate was 100%, especially in the vpn environment. The version I use is xrdp-0.9.11.
Thank you

matt335672:
What operating system are you running xrdp on?

Thanks.

yuyiliang1993:
Debian7.5 and Centos7

yuyiliang1993:
On 7/14/2020 21:00，matt335672<notifications@github.com> wrote：

What operating system are you running xrdp on?

Thanks.

matt335672:
I've spent a bit of time looking at this, and while I'm seeing memory usage by XRDP increasing to what seems a lot, it all seems to be caching of various sorts. Eventually XRDP reaches a certain level of memory usage and stays there. I can't find any actual leaks as such.

So it's possible I'm not seeing exactly what you're seeing.

Can you try the following for me?
[...]

Issue #1608, opened 2020-06-17
xrdp memory growth is particularly fast when using VPN proxies


limenghui614:
Hi,
Memory has been growing rapidly with VPNS, especially for batch operations or video playback.

metalefty:
I don't think VPNs matter but can you confirm it reproduces also without VPN?

limenghui614:
I can confirm that the memory is stable without VPN

matt335672:
Possibly a duplicate of #1634

PR #1706, opened 2020-10-20
Fix neutrinordp buffering

matt335672:
This PR primarily addresses #1634 and #1608 where displaying a video over a slow link causes XRDP to buffer unsent packets without a limit.

The basic issue is that the neutrinordp module is not making use of the source_info mechanism which is designed to throttle inputs when they are producing too much output.

I've also made some code improvements while implementing this as follows:-

1) A comment has been added to the `struct source_info`definition in `trans.h` explaining what this structure is used for.
2) An `enum` is now used for the `cur_source` member variable of this struct rather than an `int` to help when using a symbolic debugger.
3) The various XRDP modules for the backends had `tintptr si` (i.e. void *) member definitions in their header files for the module source_info pointer. I've replaced the `void *` pointers with pointers to incomplete type declarations as follows:-
   ```
   struct source_info;  /* Incomplete type */
   . . .
   struct mod
   {
       . . .
       struct_source_info *si;
       . . .
   ```
   This prevents the need for pointer casting and hence improves compile-time checking.


metalefty:
I haven't tested this yet but the code looks fine. Feel free to merge.

PR #1711, opened 2020-10-30
Add log_config_init_for_console() for utilities

matt335672:
This is a bit of tidy-up work for the standalone utilities in `sesman/tools`.

These utilities call library functions which use the logging facility.  Since they're utilities, they don't really need to log anywhere other than the current console. This change adds a function `log_config_init_for_console()` to the logging sub-system which allows this to be done in a straightforward manner.

The utility `sesadmin` has been modified to use the new facility.

Other changes to log.c:-
- A log file is no longer mandatory.
- A way to disable the logging config dump on startup has been added.
- The logging dump to stdout generates slightly less output for the console and syslog

@aquesnel - if you have the time, I'd be interested in any comments you have as you were the last person to make significant changes to the logging.

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -150,13 +150,6 @@ internal_log_start(struct log_config *l_cfg)
         return ret;
     }
 
-    /* if logfile is NULL, we return error */
-    if (0 == l_cfg->log_file)
-    {
-        g_writeln("log_file not properly assigned");
-        return ret;
-    }
-
     /* if progname is NULL, we return error */
     if (0 == l_cfg->program_name)
     {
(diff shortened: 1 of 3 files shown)


common/log.c
```diff
@@ -408,16 +408,35 @@ internal_log_config_dump(struct log_config *config)
...
+    {
+        internal_log_lvl2str(config->syslog_level, str_level);
+    }
+    else
+    {
+        g_strcpy(str_level, "<disabled>");
+    }
     g_printf("\tSyslogLevel:   %s\r\n", str_level);
     
     g_printf("per logger configuration:\r\n");
```
aquesnel, line 442: Tangential to your work:

Reading this diff, I'm realizing that the per logger config may be confusing to xrdp admins since it is only relevant for XRDP_DEBUG builds. If you have time can you clean it up?

aquesnel:
I like the idea for this change and the code looks good to me.

common/log.c
aquesnel: Never mind about my comment here. Since this is tangential to your work, and I am also updating the logging code, I went ahead and implement removing the per logger config dump for non-XRDP_DEBUG builds in #1708

aquesnel (approved):

matt335672:
Hi @aquesnel 

Following the merge of #1708, I've made a minor change to allow the logging level to be specified for `log_config_init_for_console()`

Output of `xrdp-sesadmin` is now something like:-

```
[20201130-11:08:19] [INFO ] [_scp_v1c_mng_check_response(libscp_v1c_mng.c:411)] [v1c_mng:411] connection ok
Session ID: 27668
	Session type: 2
	Screen size: 1364x768, color depth 32
	Idle time: 0 day(s) 0 hour(s) 0 minute(s)
	Connected: 2020/11/30 10:53
```

There's still a lot of work to do on this utility, but the logging is at least sane now.

Just thought I'd run it past you before I merge. Still OK with it?

Thanks.

aquesnel:
yes it still looks good to me.

PR #1713, opened 2020-11-04
Add delay between exit of window manager and termination of Xvnc

### LINKED ISSUE #1712: kwin_x11 runs indefinitely because Xvnc exits too early
I have a server running Fedora 33 with KDE session through Xrdp and I connect to it with RDP from Mac (Microsoft’s RDP) and another Fedora 33 (Remmina). For the past month, I noticed a symptom: if I click “logout” in KDE and RDP session ends. But kwin_x11 keeps running in getTimestamp(), occupying 100% of a core. 

TL;DR: kwin_x11 still waits for message from X server but Xvnc is terminated already. kwin_x11 cannot exit event loop to process quit events.

1. How Xrdp works

Xrdp relies on Xrdp-sesman to launch Xvnc and start a X server. A basic process graph is as following:

```
xrdp-sesman(68415)───xrdp-sesman(70909)─┬─Xvnc(70915)─┬─{Xvnc}(70924)
                                        │             ├─{Xvnc}(70925)
                                        │             ├─{Xvnc}(70926)
                                        │             ├─...
                                        ├─sh(70914)─┬─ssh-agent(71221)
                                        │           └─startkde(71222)───{startkde}(71234)
                                        └─xrdp-chansrv(70954)───{xrdp-chansrv}(70959)
```

xrdp-sesman waits for end of life of at least two processes: a display (Xvnc) and a window manager (startkde).

2. How startkde works

In my case, startkde is startplasma-x11.The call graph is:

```
startplasma-x11 -> plasma_session -> kwin_x11
```

startplasma-x11 calls plasma_session with QProcess and it ends execution based on unregistration of two dubs services: org.kde.ksmserver and org.kde.Shutdown.

plasma_session is more complicated, it runs the following eight jobs in sequence (refer to _second debug), kwin_x11 is one of them:

* kcminit_startup
* kded5
* kwin_x11
* ksmserver
* startup phase 0
* startup phase 1
* restore session jobs
* startup phase 2

3. What’s the exit condition of kwin_x11

It should be KSelectionOwner::lostOwnership. My understanding is that it is triggered by exiting of X server in this case.

4. What’s the problem

When I click logout in KDE, X server ends and display is closed. This _should_ trigger  KSelectionOwner::lostOwnership to end kwin_x11. However, kwin_x11 calls getTimestamp from Qt and it is a *blocking* call. The getTimestamp() works by sending dummy message to X server to get a timestamp. But now the X server is _already dead_. This blocks the *main thread* and thus  KSelectionOwner::lostOwnership will never be processed.

5. How to solve that?

I add a delay between exit of window manager and termination of Xvnc.

ivzhh:
kwin_x11's termination is not waited by startkde. Xvnc is closed
by sesman after sesman detects exit of startkde.
kwin_x11 runs indefinitely on getTimestamp due to the display is
closed. Thus kwin_x11 cannot process exit event from event loop.
Therefore, kwin_x11 will run a loop and consume a whole core of
CPU.

The fix adds a g_sleep(1000) to allow kwin_x11 to exit gracefully.

Fix #1712 kwin_x11 runs indefinitely because Xvnc exits too early

### DIFF
--- a/sesman/session.c
+++ b/sesman/session.c
@@ -818,6 +818,7 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                             "auth_end from pid %d", g_getpid());
                 auth_stop_session(data);
                 auth_end(data);
+                g_sleep(2000);
                 g_sigterm(display_pid);
                 g_sigterm(chansrv_pid);
                 cleanup_sockets(display);


metalefty:
I'm still not sure why this is required to Xvnc only.

ivzhh:
This is my post on [kde-devel](https://mail.kde.org/pipermail/kde-devel/2020-November/000168.html). In the post, I gave three solutions:

* KDE: use a preemptive method to handle exit condition; or find a way to delay ending X server
* Qt: provide a non-blocking getTimestamp() (may lead to many changes)
* Xrdp-sesman: delay ending Xvnc; or use systemd slice to find active processes

In short, I am still working on patches for KDE now in my spare time.

The reason I create pull-request for Xrdp _first_ is: 

* This is the minimum change that makes this problem gone.
* I did not read about that startx will terminate all processes, which are created by it. So I guess maybe some other WM may have this issue some how. This is a precaution.
* Xrdp is closer to downstream distros. Users can have fixes soon. Distros are slow on updating Qt and KDE.

Thank you!

PR #1727, opened 2020-11-17
Update verify_user_pam.c

RisingWater:
when a system give a tip message in function verify_pam_conv, authenticate will fail.
so it need skip this message to make sure authenticate success.

### DIFF
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -77,6 +77,9 @@ verify_pam_conv(int num_msg, const struct pam_message **msg,
                 reply[i].resp = g_strdup(user_pass->pass);
                 reply[i].resp_retcode = PAM_SUCCESS;
                 break;
+            case PAM_TEXT_INFO:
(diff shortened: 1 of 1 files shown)


sesman/verify_user_pam.c
```diff
@@ -77,6 +77,9 @@ verify_pam_conv(int num_msg, const struct pam_message **msg,
                 reply[i].resp = g_strdup(user_pass->pass);
                 reply[i].resp_retcode = PAM_SUCCESS;
                 break;
+            case PAM_TEXT_INFO:
+                memset(&reply[i], 0, sizeof(struct pam_response));
```
metalefty, line 81: use `g_memset` instead.

metalefty (changes requested):

metalefty (approved):
I'm fine with this now.

@matt335672 Any thoughts?

matt335672:
Yes, I can see the point of this. I'm happy with the change as suggested. We could log the message from the stack, but I can't see there's really any point.

Issue #1683, opened 2020-09-08
migrate travis CI builds from .org to .com


aquesnel:
Problem:

The Travis CI in-progress builds don't get displayed on github commit or pull request page until the build is complete. This has changed in recent months. I believe that this is because the nutrinolabs/xrdp repository is using the legacy travis-ci.org integration. 

There is [TravisCI documentation for migrating](https://docs.travis-ci.com/user/migrate/open-source-repository-migration) open source projects from travis-ci.org to travis-ci.com and integrating with github via GitHub actions.

The TravisCI migration docs says that it's still in Beta, but the migration seems to have been in progress since 2018, so it's likely safe to migrate.
I have followed the instructions for migrating the `aquesnel/xrdp` repository to travis-ci.com and it solved the github integration problem. See example build: https://github.com/aquesnel/xrdp/runs/1088065752

metalefty:
Looks nice. I didn't know that.

aquesnel:
Apparently I was wrong about the travis CI builds not displaying in github only while the build is in progress. It appears instead that the build never gets linked to triggering commit or pull request even tough the build is still triggered.

Example commit that triggered a travis CI build but is not displayed on the github UI
* github commit: https://github.com/neutrinolabs/xrdp/pull/1633/commits/8c5d4c4337b91debf4f11a9d5de202603a72b6c0
* travis CI build: https://travis-ci.org/github/neutrinolabs/xrdp/builds/726144015

Also, I had forgotten to mention in my original comment, that the migration of the `aquesnel/xrdp` repository from travis-ci.org to travis-ci.com did not require any code changes. The migration does require that an administrator of the `neutrinolabs/xrdp` repository connect the repository to a travis-ci.com (free open-source) account.

matt335672:
I'll close this one now we've moved to GitHub actions - it's been somewhat overtaken by events.

PR #1728, opened 2020-11-17
Add GitHub actions

aquesnel:
This change adds CI builds using the GitHub actions infrastructure. The builds performed in GitHub actions is the same as the builds currently being done in Travis CI.

Example build: https://github.com/aquesnel/xrdp/actions/runs/367490998

Adding CI builds in GitHub actions because:
* GitHub actions allows more concurrent builds than TravisCI
* GitHub actions does not require additional configuration by forks of xrdp to enable CI builds
* The [announcement by TravisCI to change their open source offering](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) and their [FAQ](https://docs.travis-ci.com/user/billing-faq#what-if-i-am-building-open-source) says that TravicCI has reduced the default build limits per open source repository to 1000 minutes (lifetime limit) with the possibility to ask for recurring credits.
* the TravisCI open source build queue has recently consistently had a [backlog of builds](https://www.traviscistatus.com/#month) and the queue size seems to be slowly decreasing
* As mentioned in #1683, the [TravisCI documentation](https://docs.travis-ci.com/user/migrate/open-source-repository-migration) says that all open source projects need to migrate from travis-ci.org to travis-ci.com

If this pull request is accepted, please let me know if you think that the TravisCI builds should be disabled.


aquesnel:
Note: the SSL compiler warning is needed for this change because the GitHub actions version of ubuntu-16.04 uses a different version of the OpenSSL package than the TravisCI version of ubuntu-16.04.

The different version of the OpenSSL package is because the GitHub actions version of ubuntu-16.04 includes a pre-installed PPA repository for php which has a newer version of OpenSSL than the official Ubuntu 16.04 release and updates.
I was not able to resolve / downgrade the OpenSSL package on the GitHub actions version of ubuntu-16.04, so I fixed the compiler warning instead.

metalefty:
I was also thinking to switch to GitHub Actions from Travis CI. Good work.

matt335672:
I'd not been aware of the impending changes to travis-ci.com. This is great work @aquesnel

I've done a bit more reading. It looks like the date for travis-ci.org closing down is [31st December 2020](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom).

It also looks like when the 1000 minutes on travis-ci.com expire we could ask for an allotment of OSS minutes. It's far from certain however that we'd get those minutes.

Looking at the [GitHub billing](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) page, it looks like we'd get 2000 minutes per month using actions. Given the test build above took less that 5 minutes, this would be plenty.

So based on the information we have available at the moment, and because of the excellent work above, I'm very happy for this to be merged.

I can't see why we need both Travis and Actions enabled together. Is there a good reason? If not, I suggest we update this PR to remove the travis.yml file.

metalefty:
Just for reference, an article written in Japanese.

"Say thank you to Travis CI and good-bye Travis CI".
https://xuwei-k.hatenablog.com/entry/2020/11/04/161400

metalefty:
Let's have a trial.

metalefty:
Now it's working. https://github.com/neutrinolabs/xrdp/actions

PR #1732, opened 2020-11-23
Add caching of cppcheck to github action CI builds

aquesnel:
This pull request reduces the build time of xrdp in github actions by about 2min 30sec per build by caching the cppcheck build artifacts. Caching the cppcheck build artifacts is safe because cppcheck is a a dependency and the workflow always builds cppcheck from a git tag from the cppcheck repo.

This caching is a 10% reduction in build time across all builds (current build time is approx. 21min CPU time)

Example builds:
* without caching: https://github.com/neutrinolabs/xrdp/runs/1428776621
* with caching cppcheck build output: https://github.com/aquesnel/xrdp/runs/1440410754

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,6 +131,13 @@ jobs:
       CPPCHECK_REPO: https://github.com/danmar/cppcheck.git
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


matt335672:
Excellent work - well worth pulling in I think.

PR #1738, opened 2020-11-30
Unify logging in libxrdp

aquesnel:
Follow-on pull request to #1633 which migrates all logging in the **libxrdp** directory to use the LOG() and LOG_DEVEL() macros.

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -121,7 +121,7 @@ libxrdp_get_pdu_bytes(const char *aheader)
 /******************************************************************************/
 /* only used during connection */
 struct stream *
-libxrdp_force_read(struct trans* trans)
+libxrdp_force_read(struct trans *trans)
 {
     int bytes;
     struct stream *s;
@@ -246,9 +246,9 @@ libxrdp_process_data(struct xrdp_session *session, struct stream *s)
             case 2: /* FASTPATH_INPUT_EVENT */
                 if (xrdp_fastpath_process_input_event(rdp->sec_layer->fastpath_layer, s) != 0)
                 {
-                     LOG_DEVEL(LOG_LEVEL_TRACE, "libxrdp_process_data returned non zero");
-                     cont = 0;
-                     term = 1;
+                    LOG_DEVEL(LOG_LEVEL_TRACE, "libxrdp_process_data returned non zero");
+                    cont = 0;
+                    term = 1;
                 }
                 break;
             default:
@@ -307,7 +307,8 @@ libxrdp_send_palette(struct xrdp_session *session, int *palette)
             return 1;
         }
     }
-    else {
+    else
+    {
         LOG_DEVEL(LOG_LEVEL_DEBUG, "libxrdp_send_palette: slowpath");
         xrdp_rdp_init_data((struct xrdp_rdp *)session->rdp, s);
     }
@@ -329,16 +330,16 @@ libxrdp_send_palette(struct xrdp_session *session, int *palette)
     s_mark_end(s);
     if (session->client_info->use_fast_path & 1) /* fastpath output supported */
     {
-       if (xrdp_rdp_send_fastpath((struct xrdp_rdp *)session->rdp, s,
+        if (xrdp_rdp_send_fastpath((struct xrdp_rdp *)session->rdp, s,
                                    FASTPATH_UPDATETYPE_PALETTE) != 0)
-       {
-           free_stream(s);
-           return 1;
-       }
+        {
+            free_stream(s);
+            return 1;
+        }
     }
     else
     {
-       xrdp_rdp_send_data((struct xrdp_rdp *)session->rdp, s,
+        xrdp_rdp_send_data((struct xrdp_rdp *)session->rdp, s,
                            RDP_DATA_PDU_UPDATE);
     }
     free_stream(s);
@@ -434,7 +435,7 @@ libxrdp_send_bitmap(struct xrdp_session *session, int width, int height,
     line_pad_bytes = line_bytes + e * Bpp;
 
     LOG_DEVEL(LOG_LEVEL_DEBUG, "libxrdp_send_bitmap: bpp %d Bpp %d line_bytes %d "
-           "server_line_bytes %d", bpp, Bpp, line_bytes, server_line_bytes);
+              "server_line_bytes %d", bpp, Bpp, line_bytes, server_line_bytes);
     make_stream(s);
     init_stream(s, MAX_BITMAP_BUF_SIZE);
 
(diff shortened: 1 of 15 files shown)


matt335672:
Nice work!

I'm not hugely familiar with this area of the code, but looking at your PR makes me realise this is well worth doing.

The only comment I've got is similar to one I made on #1738. Where there's currently an output produced by `g_writeln` and friends, I'd rather see that replaced in the short term at least by a `LOG(` macro rather than a `LOG_DEVEL(` macro. I appreciate that many of these may not make a lot of sense to a sysadmin, but at the moment there's a chance of a silent fail which is harder for us to diagnose if it happens.

Please come back to me on that if you've got a strong opinion on this. Also there are a few commented out `g_writeln()` calls and I absolutely think you've done the right thing there.

This looks like pretty much a straight text subs to me. I can see why you've introduced `UNUSED_VAR`, and I think you've got the best solution to that problem. You could use compiler variable attributes but these aren't necessarily portable.

As ever, thanks very much for your contributions.

aquesnel:
ok, I'll review the log changes to avoid silent failures.

aquesnel:
Hi @matt335672 ,

I've modified the log messages in error code paths to use `LOG` instead of `LOG_DEVEL`. Please let me know if there is additional logging lines that you think should have different log levels or log macros.

Since we are discussing log messages, I also want to mention that I've made pull request #1742 which adds on top of this pull request more detailed and clear logging messages in the libxrdp code. I created it as a separate pull request to separate the text substitution from improving the log messages.

libxrdp/xrdp_caps.c
```diff
@@ -492,59 +490,59 @@ xrdp_caps_process_codecs(struct xrdp_rdp *self, struct stream *s, int len)
...
+                codec_id, codec_properties_length);
             self->client_info.jpeg_codec_id = codec_id;
             i1 = MIN(64, codec_properties_length);
             g_memcpy(self->client_info.jpeg_prop, s->p, i1);
             self->client_info.jpeg_prop_len = i1;
             /* make sure that requested quality is  between 0 to 100 */
             if (self->client_info.jpeg_prop[0] < 0 || self->client_info.jpeg_prop[0] > 100)
             {
-                g_writeln("  Warning: the requested jpeg quality (%d) is invalid,"
+                LOG_DEVEL(LOG_LEVEL_WARNING, "  Warning: the requested jpeg quality (%d) is invalid,"
```
matt335672, line 528: Should this be LOG() ?

libxrdp/xrdp_rdp.c
```diff
@@ -457,14 +453,14 @@ xrdp_rdp_recv(struct xrdp_rdp *self, struct stream *s, int *code)
...
+            LOG_DEVEL(LOG_LEVEL_TRACE, "out (1) xrdp_rdp_recv");
             return 0;
         }
 
         if (error != 0)
         {
-            DEBUG(("out xrdp_rdp_recv error"));
-            g_writeln("xrdp_rdp_recv: xrdp_sec_recv failed");
+            LOG(LOG_LEVEL_ERROR, "out xrdp_rdp_recv error");
+            LOG(LOG_LEVEL_ERROR, "xrdp_rdp_recv: xrdp_sec_recv failed");
```
matt335672, line 463: Probably can remove one of these

libxrdp/xrdp_sec.c
```diff
@@ -685,48 +679,48 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
...
 
     /* this is the first test that the decrypt is working */
     if ((flags & RDP_LOGON_NORMAL) != RDP_LOGON_NORMAL) /* 0x33 */
     {
         /* must be or error */
-        DEBUG(("xrdp_sec_process_logon_info: flags wrong, major error"));
-        LLOGLN(0, ("xrdp_sec_process_logon_info: flags wrong, likely decrypt "
-               "not working"));
+        LOG(LOG_LEVEL_ERROR, "xrdp_sec_process_logon_info: flags wrong, major error");
+        LOG(LOG_LEVEL_ERROR, "xrdp_sec_process_logon_info: flags wrong, likely decrypt "
```
matt335672, line 689: Probably can remove one of these

libxrdp/xrdp_sec.c
```diff
@@ -1247,12 +1248,12 @@ xrdp_sec_recv(struct xrdp_sec *self, struct stream *s, int *chan)
...
 
-    DEBUG((" in xrdp_sec_recv"));
+    LOG_DEVEL(LOG_LEVEL_TRACE, " in xrdp_sec_recv");
 
     if (xrdp_mcs_recv(self->mcs_layer, s, chan) != 0)
     {
-        DEBUG((" out xrdp_sec_recv : error"));
-        g_writeln("xrdp_sec_recv: xrdp_mcs_recv failed");
+        LOG(LOG_LEVEL_ERROR, " out xrdp_sec_recv : error");
+        LOG(LOG_LEVEL_ERROR, "xrdp_sec_recv: xrdp_mcs_recv failed");
```
matt335672, line 1256: Probably can remove one of these

matt335672 (commented):
A few minor comments for your consideration

matt335672:
Really good stuff BTW - I can see where some real thought has gone in to some of these.

@metalefty - given this is general substitution (plus of course the re-formatting), I'm happy to merge this. Let me know what you think.

aquesnel:
I've implemented @matt335672 recommendations, and I think this is ready to merge. Please let me know if there is anything else that needs to be cleaned up or squashed before merging.

libxrdp/libxrdp.c
```diff
@@ -135,18 +131,18 @@ libxrdp_force_read(struct trans* trans)
 
     if (trans_force_read(trans, 4) != 0)
     {
-        g_writeln("libxrdp_force_read: header read error");
+        LOG(LOG_LEVEL_WARNING, "libxrdp_force_read: header read error");
```
metalefty, line 134: Thank you very much for your careful work! You did this not just by replacing them with a single log level but considered the log context.

libxrdp/xrdp_iso.c
```diff
@@ -80,11 +74,11 @@ xrdp_iso_negotiate_security(struct xrdp_iso *self)
             if (self->requestedProtocol & PROTOCOL_SSL)
             {
                 if (!g_file_readable(client_info->certificate) ||
-                    !g_file_readable(client_info->key_file))
+                        !g_file_readable(client_info->key_file))
                 {
                     /* certificate or privkey is not readable */
-                    log_message(LOG_LEVEL_DEBUG, "No readable certificates or "
-                                "private keys, cannot accept TLS connections");
+                    LOG(LOG_LEVEL_WARNING, "No readable certificates or "
```
metalefty, line 80: Good catch.

metalefty:
Let's merge.

PR #1753, opened 2020-12-15
Fixing code location log level filtering

aquesnel:
As part of the logging refactoring in #1708 I broke the code location based log level filtering. This pull request fixes the code location log level filtering.

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -523,27 +523,53 @@ internal_log_config_copy(struct log_config *dest, const struct log_config *src)
 
 bool_t
 internal_log_is_enabled_for_level(const enum logLevels log_level,
-                                  const bool_t override_destination_level)
+                                  const bool_t override_destination_level,
+                                  const enum logLevels override_log_level)
 {
     /* Is log initialized? */
     if (g_staticLogConfig == NULL)
     {
         return 0;
     }
-
-    /* Is there at least one log destination which will accept the message based on the log level? */
-    return (g_staticLogConfig->fd >= 0
-            && (override_destination_level || log_level <= g_staticLogConfig->log_level))
-           || (g_staticLogConfig->enable_syslog
-               && (override_destination_level || log_level <= g_staticLogConfig->syslog_level))
-           || (g_staticLogConfig->enable_console
-               && (override_destination_level || log_level <= g_staticLogConfig->console_level));
+    else if (g_staticLogConfig->fd < 0
+             && !g_staticLogConfig->enable_syslog
+             && !g_staticLogConfig->enable_console)
+    {
+        /* all logging outputs are disabled */
+        return 0;
+    }
+    else if (override_destination_level)
+    {
+        /* Override is enabled - should the message should be logged? */
+        return log_level <= override_log_level;
+    }
+    /* Override is disabled - Is there at least one log destination
+     * which will accept the message based on the log level? */
+    else if (g_staticLogConfig->fd >= 0
+             && log_level <= g_staticLogConfig->log_level)
+    {
+        return 1;
+    }
+    else if (g_staticLogConfig->enable_syslog
+             && log_level <= g_staticLogConfig->syslog_level)
+    {
+        return 1;
+    }
+    else if (g_staticLogConfig->enable_console
+             && log_level <= g_staticLogConfig->console_level)
+    {
+        return 1;
+    }
+    else
+    {
+        return 0;
+    }
 }
 
 bool_t
 internal_log_location_overrides_level(const char *function_name,
                                       const char *file_name,
-                                      const enum logLevels log_level)
+                                      enum logLevels *log_level_return)
 {
     struct log_logger_level *logger = NULL;
     int i;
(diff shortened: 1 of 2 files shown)


common/log.c
```diff
@@ -532,17 +533,20 @@ internal_log_is_enabled_for_level(const enum logLevels log_level,
...
+               && ((override_destination_level && log_level <= override_log_level)
+                  || (!override_destination_level && log_level <= g_staticLogConfig->log_level)))
            || (g_staticLogConfig->enable_syslog
-               && (override_destination_level || log_level <= g_staticLogConfig->syslog_level))
+               && ((override_destination_level && log_level <= override_log_level)
+                  || (!override_destination_level && log_level <= g_staticLogConfig->syslog_level)))
            || (g_staticLogConfig->enable_console
-               && (override_destination_level || log_level <= g_staticLogConfig->console_level));
+               && ((override_destination_level && log_level <= override_log_level)
+                  || (!override_destination_level && log_level <= g_staticLogConfig->console_level)));
```
matt335672, line 543: I'm a little concerned with this mega-conditional. It's quite hard to parse mentally (at least it is for me), and there's some repetition in it. Maybe split it up a bit? Here's a suggestion:-

```c
{
    bool_t result;
    /* Is log initialized? */
    if (g_staticLogConfig == NULL)
    {
        result = 0;
    }
    else if (override_destination_level)
    {
        /* Override in place - if the message should be logged, is any
         * destination enabled? */
        if (log_level > override_log_level)
        {
            result = 0;
        }
        else
        {
            result =
                g_staticLogConfig->fd >= 0 ||
                g_staticLogConfig->enable_syslog ||
                g_staticLogConfig->enable_console;
        }
    }
    else
    {
        /* Is there at least one log destination which will accept the
         * message based on the log level? */
        result =
            (g_staticLogConfig->fd >= 0 &&
             log_level <= g_staticLogConfig->log_level) ||
            (g_staticLogConfig->enable_syslog &&
             log_level <= g_staticLogConfig->syslog_level) ||
            (g_staticLogConfig->enable_console &&
             log_level <= g_staticLogConfig->console_level);
    }

    return result;
}
```

common/log.h
```diff
@@ -69,66 +70,67 @@ enum logReturns
 #define LOG_PER_LOGGER_LEVEL
 
 /**
- * @brief Logging macro for messages that are for an XRDP developper to 
+ * @brief Logging macro for messages that are for an XRDP developper to
  * understand and debug XRDP code.
- * 
- * Note: all log levels are relavant to help a developper understand XRDP at 
+ *
+ * Note: all log levels are relavant to help a developper understand XRDP at
```
matt335672, line 76: Suggest replace 'developper' with 'developer'

matt335672 (commented):
Some minor comments

common/log.c
aquesnel: cleaned up to be easier to understand.

common/log.h
aquesnel: fixed

aquesnel:
@matt335672 I've implemented the changes you suggested.

Please let me know if more changes are needed before merging.

metalefty:
Is this should be in v0.9.15?

aquesnel:
@metalefty Yes I think this pull request should be in v0.9.15 because the log filtering by source location in devel is broken without this pull request.

common/log.c
```diff
@@ -530,17 +530,37 @@ internal_log_is_enabled_for_level(const enum logLevels log_level,
...
-               && ((override_destination_level && log_level <= override_log_level)
-                  || (!override_destination_level && log_level <= g_staticLogConfig->console_level)));
+    else if (g_staticLogConfig->fd < 0
+            && !g_staticLogConfig->enable_syslog
+            && !g_staticLogConfig->enable_console)
+    {
+        /* all logging outputs are disabled */
+        return 0;
+    }
+    else if (override_destination_level)
```
matt335672, line 540: I'm happy with either multiple returns in a function, or if / else if with a single return at the end. Here you've mixed styles somewhat. It'll work fine, but I think there's a chance that in the future a static analysis will moan about an unreachable 'else' branch following a return. 

I'd suggest either removing the unnecessary `else` statements, or using a single return - whatever seems more natural to you.

aquesnel: @matt335672 I've implemented this feedback using multiple returns by only have a single level of nesting for the if/else. Please let me know if I correctly interpreted your feedback, or if you'd prefer that I use the single return at the end style.

aquesnel:
CR feedback implemented, branch squashed and rebased to devel.

This should be ready to merge once @matt335672 reviews the changes.

metalefty (approved):

common/log.c
matt335672: Sorry - I've probably not been very clear.

My point was that in this kind of construction:-

<pre>
if (&lt;something&gt;)
{
    return 1;
}
<b>else</b> if (&lt;something else&gt;)
...
</pre>

the `else` in bold is superfluous and not required. At some stage I'd expect a static analysis tool to say something about it. Does that make more sense?

The nesting style you use, or whether you use a single return or multiple returns is completely up to you. I've got my own preferences but they're not relevant at all.

PR #1755, opened 2020-12-18
man page fixes after neutrinolabs/xorgxrdp#175

metalefty:


matt335672 (approved):
That all looks fine to me.

PR #1761, opened 2020-12-23
riscv doesn't require pointers to be aligned

kraj:
Signed-off-by: Khem Raj <raj.khem@gmail.com>


matt335672:
Hi @kraj 

I've taken this as it means that v0.9.15 will now at least compile on RISC-V - thanks very much for your contribution.

Do I take it you've got xrdp running on a RISC-V board at the moment then?

kraj:
> Hi @kraj
> 
> I've taken this as it means that v0.9.15 will now at least compile on RISC-V - thanks very much for your contribution.
> 
> Do I take it you've got xrdp running on a RISC-V board at the moment then?

I have only fixed compile this time and I have board but wemu works too

PR #1778, opened 2021-01-06
Support EC cryptographic keys for TLS

### LINKED ISSUE #1776: Does xrdp support ECC?
I use an ECC-based PKI for whatever I can, and so I attempted to use it for `xrdp`'s TLS encryption. I have a certificate and corresponding EC key (using the `secp384r1` curve, generated by OpenSSL). The key is in `/etc/ssl/` with the correct permissions (644 for the cert and 640 for the key, with `ssl-cert` as the group for the latter and `xrdp` added to the `ssl-cert` group). This is then symlinked to `/etc/xrdp/` in the same way that the default snakeoil certificate is (I've also tried directly pointing to the cert and key). The below configuration is used to enable TLS:

```ini
certificate=/etc/xrdp/cert.pem
key_file=/etc/xrdp/key.pem
ssl_protocols=TLSv1.2, TLSv1.3
tls_ciphers=HIGH
```

However, attempting to connect with both FreeRDP and MSTSC fails, and the latter produces the error `Because of a protocol error detected at the client (code 0x2104), this session will be disconnected`. Capturing via Wireshark, the server behaves seemingly normally (a couple TLS data continuations) and _sometimes_ sends a FIN, but then the client sends a RST. There are also no errors on the `xrdp` side:

```
[20210103-00:03:33] [INFO ] Socket 12: AF_INET6 connection received from ::ffff:192.168.1.32 port 49222
[20210103-00:03:33] [DEBUG] Closed socket 12 (AF_INET6 ::ffff:192.168.1.51 port 3389)
[20210103-00:03:33] [DEBUG] Closed socket 11 (AF_INET6 :: port 3389)
[20210103-00:03:33] [DEBUG] TLSv1.3 enabled
[20210103-00:03:33] [DEBUG] TLSv1.2 enabled
[20210103-00:03:33] [DEBUG] Security layer: requested 1, selected 1
[20210103-00:03:33] [DEBUG] Closed socket 12 (AF_INET6 ::ffff:192.168.1.51 port 3389)
```

Changing to either the snakeoil keys or a self-signed RSA key solves this issue, it's only EC keys that won't work. I've tried both self-signed and CA-signed EC certificates, with a couple different curves. I wasn't able to find anything about whether ECC is supported by `xrdp` so I assumed by the lack of server errors that it has some level of support, but it apparently has issues with that. ECC works just fine with Windows machines so clearly the protocol supports it correctly (of course TLS would support it though), so I'm not sure where the break is.

Running Linux Mint 20 and `xrdp` 0.9.12.

matt335672:
Fixes #1776 

At present the certificate chain presented for TLS needs to be RSA. However, current mstsc.exe and Remmina clients can accept ECC certificates if they are presented with them.

xrdp can be made usable with non-RSA certificates and private keys by changing the call to `SSL_CTX_use_RSAPrivateKey_file()` (which only loads RSA keys) to `SSL_CTX_use_PrivateKey_file()` (which loads any private key type). This call has been present in OpenSSL since at least version 0.9.6.

During the course of investigating #1776, it was found that the error logging in `ssl_calls.c` is simply not working. Much of it goes to stdout which is disabled.

There are three commits in this PR:-
1. Move all logging to user the new macros
2. Run the ssl_calls files through `astyle`
3. Replace the call to load the private key.

Regression testing shows:-
- the new code can read either an RSA key or an secp384r1 key (other types have not been tried).
- Without the change to `SSL_CTX_use_PrivateKey_file` in 849a8e7, the following error is logged when trying to load a non-RSA key:-
   ```
   Error loading TLS RSA private key from /etc/xrdp/ec-key.pem
   ```

### DIFF
--- a/common/ssl_calls.c
+++ b/common/ssl_calls.c
@@ -64,8 +64,8 @@ static inline void
 RSA_get0_key(const RSA *key, const BIGNUM **n, const BIGNUM **e,
              const BIGNUM **d)
 {
-     *n = key->n;
-     *d = key->d;
+    *n = key->n;
+    *d = key->d;
 }
 
 static inline int
@@ -75,7 +75,7 @@ DH_set0_pqg(DH *dh, BIGNUM *p, BIGNUM *q, BIGNUM *g)
      * parameters MUST be non-NULL.  q may remain NULL.
      */
     if ((dh->p == NULL && p == NULL)
-        || (dh->g == NULL && g == NULL))
+            || (dh->g == NULL && g == NULL))
     {
         return 0;
     }
@@ -230,7 +230,7 @@ ssl_md5_complete(void *md5_info, char *data)
 
 /*****************************************************************************/
 void *
-ssl_des3_encrypt_info_create(const char *key, const char* ivec)
+ssl_des3_encrypt_info_create(const char *key, const char *ivec)
 {
     EVP_CIPHER_CTX *des3_ctx;
     const tui8 *lkey;
@@ -246,7 +246,7 @@ ssl_des3_encrypt_info_create(const char *key, const char* ivec)
 
 /*****************************************************************************/
 void *
-ssl_des3_decrypt_info_create(const char *key, const char* ivec)
+ssl_des3_decrypt_info_create(const char *key, const char *ivec)
 {
     EVP_CIPHER_CTX *des3_ctx;
     const tui8 *lkey;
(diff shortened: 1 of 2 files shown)


matt335672:
@aquesnel - if you have the time I'd very much appreciate your comments on the logging changes I've made in this PR.

aquesnel (approved):

aquesnel:
@matt335672 good work. The changes look good to me.

metalefty (approved):
Good work!

matt335672:
Thanks both. I'll squash and merge.

PR #1787, opened 2021-01-20
Minor manpage fixes

matt335672:
Minor manpage fixes and improvements

- `xrdp`, `xrdp-sesman` and `xrdp-chansrv` are in `$(sbindir)` rather than `$(bindir)`. This change makes that clear in the pages.
- The groff macros `.UR` and `.RE` are used for web urls so that HTML manpages (when we generate them) will work properly.
- The repeated string 'http://www.xrdp.org/' for the xrdp homepage has been replaced with a macro.

Any other suggestions welcome.


metalefty:
LGTM.

PR #1793, opened 2021-01-23
Fixing TS_PLAY_SOUND_PDU_DATA to set the correct frequency and duration

aquesnel:
The `libxrdp_send_bell` function does not follow the MS-RDPBCGR spec for the [TS_PLAY_SOUND_PDU_DATA](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/4c1683b2-3a92-4319-8f7e-2318b3546c02) message.

This pull request reconciles the original author's intent based on their comments with the spec.

This pull request has been tested using the VNC backend since the Xorg backend does not seem to support sending bell messages.

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -376,9 +376,11 @@ libxrdp_send_bell(struct xrdp_session *session)
         return 1;
     }
 
-    out_uint32_le(s, 440); /* frequency */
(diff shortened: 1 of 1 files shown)


matt335672:
Agreed - this seems like an obvious fix. Good spot!

PR #1800, opened 2021-02-05
Rename USE_NOPAM as USE_PAM

matt335672:
This is addressing a minor annoyance of mine.

I had a conversation with @metalefty a while ago about avoiding double negatives where possible (i.e. replace `if (!no_daemon)` with `if (daemon)`) to improve readability for developers whose first language isn't English.

This PR replaces `#ifdef USE_NOPAM` with `#ifdef USE_PAM` in the xrdp daemon code. I have to say that even as a native English speaker I find this one needs a couple of tries to parse in my head.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -263,9 +263,9 @@ then
   AC_DEFINE([XRDP_ENABLE_IPV6],1,[Enable IPv6])
 fi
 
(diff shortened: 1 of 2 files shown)


metalefty:
Thank you for the good catch.

metalefty (approved):

PR #1803, opened 2021-02-09
Readability fix for WM states

matt335672:
Readability change for xrdp_wm.

Currently, a `login_mode` member variable is used to keep track of the state of the login box and connection state. It takes the values 0, 1, 2, 3, 10 and 11.

The PR replaces `login_mode` with `login_state` and changes the values assigned to it as follows:-

Was | Now
---|---
0| WMLS_RESET
1| WMLS_USER_PROMPT
2| WMLS_START_CONNECT
3| WMLS_CONNECT_IN_PROGRESS
10| WMLS_CLEANUP
11| WMLS_INACTIVE

These enumeration constants are documented in xrdp_types.h

### DIFF
--- a/xrdp/xrdp.h
+++ b/xrdp/xrdp.h
@@ -150,7 +150,7 @@ xrdp_wm_get_wait_objs(struct xrdp_wm* self, tbus* robjs, int* rc,
 int
 xrdp_wm_check_wait_objs(struct xrdp_wm* self);
 int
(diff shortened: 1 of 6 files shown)


aquesnel:
This is change makes the login window manager code much more understandable. Thank you.

I had done an investigation about the login states when I was improving the logging, and I had come to a similar understanding of the meaning for each of those login state values, and this pull request is better than what I had come up with.

aquesnel (approved):

metalefty (approved):

PR #1812, opened 2021-02-23
Unify logging in xrdpapi/*

aquesnel:
Follow-on pull request to #1633 which migrates all logging in the xrdpapi directory to use the LOG() and LOG_DEVEL() macros.

Changes:

* Migrate logging to the LOG() and LOG_DEVEL() macros (using text transform script)
* error logs use the LOG macro
* update code formatting with astyle for all files in xrdp/*

### DIFF
--- a/xrdpapi/xrdpapi.c
+++ b/xrdpapi/xrdpapi.c
@@ -56,9 +56,9 @@ can_send(int sck, int millis);
 static int
 can_recv(int sck, int millis);
 static int
-mysend(int sck, const void* adata, int bytes);
+mysend(int sck, const void *adata, int bytes);
 static int
-myrecv(int sck, void* adata, int bytes);
+myrecv(int sck, void *adata, int bytes);
 
 /*
  * Opens a handle to the server end of a specified virtual channel - this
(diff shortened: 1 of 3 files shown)


xrdpapi/xrdpapi.c
```diff
@@ -127,15 +122,15 @@ WTSVirtualChannelOpenEx(unsigned int SessionId, const char *pVirtualName,
 
     if (wts->display_num <= 0)
     {
-        LLOGLN(0, ("WTSVirtualChannelOpenEx: fatal error; display is 0"));
+        LOG(LOG_LEVEL_ERROR, "WTSVirtualChannelOpenEx: fatal error; display is 0");
```
matt335672, line 125: Should be <= 0, looking at the test.

matt335672 (commented):
One minor comment.

On the whole I'm happy with this. There are definitely places where the messages themselves could be improved, but this is nothing to do with the PR, and this is in any case a bit step forward.

xrdpapi/xrdpapi.c
aquesnel: done

aquesnel:
@matt335672 I've implemented the feedback, let me know if anything else needs to be changed before this can be merged.

PR #1813, opened 2021-02-26
Add versioning to xrdp_client_info

matt335672:
As discussed in gitter, add a version guard to xrdp_client_info.

There's probably more that can be done here in terms of restructuring this. I notice a lot of the stuff above the line 'Private to xrdp below this line' is not actually used by xorgxrdp. Alternatively, maybe we should split this into two structures so the compiler can pick up differences. Thoughts?

BTW : the include's been run through astyle. All I've actually done is add the 'version' field and the macro and change the comment 'NO CHANGES ABOVE THIS LINE' to the (hopefully) more informative 'Private to xrdp below this line'

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -25,143 +25,155 @@
 
 struct monitor_info
 {
-  int left;
-  int top;
-  int right;
-  int bottom;
-  int is_primary;
+    int left;
+    int top;
+    int right;
+    int bottom;
+    int is_primary;
 };
 
+/**
+ * Information about the xrdp client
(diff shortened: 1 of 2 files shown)


aquesnel:
Since the client info struct is being dumped onto the wire between xrdp and xorgxrdp, it seems sensible to minimize there data that is being sent to just the required fields, as long as backwards comparability can be maintained.

matt335672:
Agreed.

At the moment it's a challenge to change the client_info, as there's more in it that the xorgxrdp driver actually needs.

PR #1817, opened 2021-03-02
Add KRDC to supported client list in README.md

clebig:


matt335672:
Thanks for adding this - more people might find out about KRDC now.

Issue #1773, opened 2020-12-31
Xrdp session Kubuntu 20.04 , how to increase rights?


nikimagic:
When connecting xrdp, it is noticeable that the rights are limited. For example, the timeshift does not start, the restart buttons are not available. And if you go directly, everything works well.

Kubuntu 20.04

matt335672:
This is because of a sub-system called 'polkit'.

Polkit specified policies for sessions. Polkit treats remote sessions differently from local sessions. For example, you might want someone logged in to the console to be able to shut the system down, but not someone remotely.

The system manager configures polkit to suit the system's use-cases. So, using the above example, if the system is a single-user workstation, it might be OK for the user to reboot it remotely. If however the system is a multi-user server, it probably isn't.

You can have a look at previous issues rasied on this [here](https://github.com/neutrinolabs/xrdp/issues?q=is%3Aissue+is%3Aclosed+polkit). I think a good place to start might be #1568.

Let us know how you get on.

nikimagic:
It didn't work.
I need instructions for KDE, Kubuntu 20.04
For about a week I have been trying to do at least something polkit-1

matt335672:
I'm not familiar with KDE at all I'm afraid.

I suggest you enable polkit logging to start with so you can see what errors are being reported. Notmally polkit is run with `--no-debug`, so it doesn't generate any output.

The following works for me on Xubuntu 20.04, so will be fine on Kubuntu as well. It enables logging on polkit until the next reboot:-
1. `sudo systemctl edit --system --runtime polkit.service`
2. In the editor, create the following file. The empty `ExecStart=` line is not a mis-print
   ```
   [Service]
   ExecStart=
   ExecStart=/usr/lib/policykit-1/polkitd
   ```
3. Make a note of the current time in hours minutes and seconds (HH:MM:SS) with the `date` command.
4. `sudo systemctl restart polkit`
5. perform an action in an xrdp session which does not do what you expect.
6. Get the polkit logging output with `sudo journalctl -u polkit -S HH:MM:SS`. Substitute the time in step 3 for HH:MM:SS.

This will tell you what polkit action is failing. You can then create a `.pkla` file to authorize the action. Let us know if you need a hand with this.

nikimagic:
here is such a log:

 system-bus-name::1.3494 is inquiring whether unix-process:54838:7740097 is authorized for org.freedesktop.policykit.exec
[...]

PR #1818, opened 2021-03-03
Add support for multiple currently unsupported architectures

glaubitz:


metalefty:
Please keep in mind that merging this PR doesn't mean we support these architectures. It may fix the build but it just fixes the build.

glaubitz:
Does xrdp have any other architecture-specific code that needs to be taken care of?

matt335672:
Agreed - we're having trouble testing on sparc at the moment, let alone some of these others.

Another way to do this might be to use `endian.h` (Linux) and `sys/endian.h` (FreeBSD) to determine the endian-ness, and then figure out the `NEED_ALIGN`/`NO_NEED_ALIGN` using autoconf. `endian.h` in Linux came with glibc 2.9 (13-Nov-2008) and the FreeBSD equivalent seems to have been around longer than that, looking at https://github.com/freebsd/freebsd-src/commits/master/sys/sys/endian.h

However if we go down this route we'll end up having to add additional exceptions for things like MacOS where we need "machine/endian.h", and we still can't easily test this. We can of course also check for endian-ness using autoconf as well.

To sum up my rather rambling thoughts above, I'm in two minds about this PR. It's harmless enough, but the result could possibly be achieved in other ways which might reduce the overall maintenance burden moving forward.

matt335672:
Hi @glaubitz - not that I'm aware of.

As I said earlier, the biggest problem I think we have with using defines in this way is that it's not testable with the resources we have available to us on the project - there may be automated ways to achieve a more general solution.

glaubitz:
If you need access to a SPARC, you can use the GCC compile farm which hosts a fast SPARC (and other more obscure architectures) machine exactly for these purposes: https://gcc.gnu.org/wiki/CompileFarm

More architectures are being added in the future, I'm currently working on adding a MIPS machine running Debian. Otherwise, I can also help with access to various architectures as I have access to most architectures in Debian (either through my own hardware or hardware hosted by someone else).

Plus, Debian is regularly building the latest versions of xrdp on its build infrastructure, so there is at least a build CI:

> https://buildd.debian.org/status/package.php?p=xrdp&suite=sid

If you add a testsuite, you can also make sure that basic functionality works correctly.

FWIW, we have been shipping xrdp in Debian for years and I don't remember architecture-specific issues from the tip of my head before the 0.9.15 release.

matt335672:
Thanks for that @glaubitz  - that's really useful info.

The architecture-specific code is all in `common/arch.h`. For 0.9.15 we added a PR for RISC-V support but made no other changes in this area. For 0.9.14 we introduced commit 7e58209 which added `-Werror` to the gcc compile flags. I suspect this is behind experiences mentioned in your last sentence.

@metalefty - I'm minded to take this PR now to fix the Debian build farm issues linked above. What do you think?

metalefty (approved):

metalefty:
I'm fine with this PR itself. 

Build farm ensures just xrdp code builds on each architecture. It doesn't detect runtime architecture-specific bugs.
I meant fixing architecture-specific bugs on minor architectures is still contributor-driven.

matt335672:
Thanks @glaubitz - let us know if the Debian CI build needs any more changes.

Is this part of the Bullseye work that's going on right now?

glaubitz:
> Is this part of the Bullseye work that's going on right now?

Yes, as a backported fix. No new upstream version is allowed at this point.

matt335672:
@glaubitz - I have a Bullseye-related question for you.

We occasionally come across issues which affect one particular downstream only (counting Debian/Ubuntu as a single downstream).

One such issue on our side (#1773) resulted in my raising Launchpad bug [1911435](https://bugs.launchpad.net/ubuntu/+source/xrdp/+bug/1911435) which needs a change in the Salsa repo to address. This bug will probably also affect Bullseye.

The bug appears to be getting little attention. It's possible that I've simply reported it in the wrong place. In the past I've also tried raising PRs on Salsa directly (https://salsa.debian.org/debian-remote-team/xrdp/-/merge_requests/2), also with little effect.

If you can give me any detail of a more effective way to communicate with the xrdp packagers for Debian/Ubuntu I'd love to hear it. Or is Launchpad the best way?

glaubitz:
Almost all Ubuntu packages are imported from Debian in source form, so the original maintainer of the package is in Debian.

The rule is that all packages that are in "universe" or "multiverse" in Ubuntu have no maintainer in Ubuntu, so the bug should always reported to Debian. Since **xrdp** is part of the "universe" repository, you always should report issues with that package in Debian as there is simply no one in Ubuntu to answer to your bug reports.

Since Debian itself is a volunteer project, there is unfortunately no guarantee that someone will respond to bugs for leaf packages. But if you report the bug against the xrdp package in Debian, you should get an answer if the package is still maintained.

If not, feel free to contact me via mail (@debian.org) and I will help mediating.

matt335672:
That explains perfectly why the Launchpad fault is going nowhere. I will do as you suggest and see what happens.

Thanks for the lucid explanation.

PR #1820, opened 2021-03-04
Resolution switching without reconnecting.

Nexarian:
Implements https://github.com/neutrinolabs/xrdp/issues/448

- Based on https://github.com/jsorg71/xrdp/tree/dynamic_monitor
- Tested with xorgxrdp
- Tested with vnc
- Only works with single monitor.

Depends on https://github.com/neutrinolabs/xorgxrdp/pull/183

To test either use Microsoft Remote Desktop (From the Windows or Mac OS app store) and configure to resize the screen or FreeRDP with the `/dynamic-resolution` flag set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,15 @@
 ## Overview
 
 **xrdp** provides a graphical login to remote machines using Microsoft
-Remote Desktop Protocol (RDP). xrdp accepts connections from a variety of
-RDP clients: FreeRDP, rdesktop, KRDC, NeutrinoRDP and Microsoft Remote Desktop
-Client (for Windows, Mac OS, iOS and Android).
+Remote Desktop Protocol (RDP). xrdp accepts connections from a variety of RDP clients:
+  - FreeRDP
+  - rdesktop
+  - KRDC
+  - NeutrinoRDP
+  - Windows MSTSC (Microsoft Terminal Services Client)
+  - Microsoft Remote Desktop (which is distinct from MSTSC)
+
+Many of these work on some or all of Windows, Mac OS, iOS, and/or Android.
 
 RDP transport is encrypted using TLS by default.
 
@@ -24,7 +30,7 @@ RDP transport is encrypted using TLS by default.
  * Connect to a Linux desktop using RDP from anywhere (requires
    [xorgxrdp](https://github.com/neutrinolabs/xorgxrdp) Xorg module)
  * Reconnect to an existing session
- * Session resizing
+ * Session resizing on connect and while an existing session is active.
  * RDP/VNC proxy (connect to another RDP/VNC server via xrdp)
 
 ### Access to Remote Resources
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -19,6 +19,7 @@
  */
 
 #include "xrdp_constants.h"
+#include "ms-rdpbcgr.h"
 
 #if !defined(XRDP_CLIENT_INFO_H)
 #define XRDP_CLIENT_INFO_H
@@ -120,8 +121,8 @@ struct xrdp_client_info
     int security_layer; /* 0 = rdp, 1 = tls , 2 = hybrid */
     int multimon; /* 0 = deny , 1 = allow */
     int monitorCount; /* number of monitors detected (max = 16) */
-    struct monitor_info minfo[16]; /* client monitor data */
-    struct monitor_info minfo_wm[16]; /* client monitor data, non-negative values */
+    struct monitor_info minfo[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data */
+    struct monitor_info minfo_wm[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data, non-negative values */
 
     int keyboard_type;
     int keyboard_subtype;
--- a/libxrdp/xrdp_channel.c
+++ b/libxrdp/xrdp_channel.c
@@ -776,7 +776,12 @@ xrdp_channel_drdynvc_start(struct xrdp_channel *self)
     struct mcs_channel_item *ci;
     struct mcs_channel_item *dci;
 
-
+    LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_channel_drdynvc_start: drdynvc_channel_id %d", self->drdynvc_channel_id);
+    if (self->drdynvc_channel_id != -1)
+    {
+        LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_channel_drdynvc_start: already started");
+        return 0;
+    }
     dci = NULL;
     count = self->mcs_layer->channel_list->count;
     for (index = 0; index < count; index++)
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -1224,7 +1224,7 @@ send_update_request_for_resize_status(struct vnc *v)
     make_stream(s);
     init_stream(s, 8192);
 
-    switch (v->initial_resize_status)
+    switch (v->resize_status)
     {
         case VRS_WAITING_FOR_FIRST_UPDATE:
             /*
@@ -1364,14 +1364,14 @@ lib_framebuffer_first_update(struct vnc *v)
             {
                 LOG(LOG_LEVEL_DEBUG, "Server layout is the same "
                     "as the client layout");
-                v->initial_resize_status = VRS_DONE;
+                v->resize_status = VRS_DONE;
             }
             else
             {
                 LOG(LOG_LEVEL_DEBUG, "Server layout differs from "
                     "the client layout. Changing server layout");
                 error = send_set_desktop_size(v, &v->client_layout);
-                v->initial_resize_status = VRS_WAITING_FOR_RESIZE_CONFIRM;
+                v->resize_status = VRS_WAITING_FOR_RESIZE_CONFIRM;
             }
         }
         else
@@ -1382,7 +1382,7 @@ lib_framebuffer_first_update(struct vnc *v)
             LOG(LOG_LEVEL_DEBUG, "Resizing client to server %dx%d",
                 v->server_width, v->server_height);
             error = resize_client(v, 0, v->server_width, v->server_height);
-            v->initial_resize_status = VRS_DONE;
+            v->resize_status = VRS_DONE;
         }
 
         g_free(layout.s);
@@ -1437,7 +1437,7 @@ lib_framebuffer_waiting_for_resize_confirm(struct vnc *v)
                     v->server_width, v->server_height);
                 error = resize_client(v, 0, v->server_width, v->server_height);
             }
-            v->initial_resize_status = VRS_DONE;
+            v->resize_status = VRS_DONE;
         }
 
         g_free(layout.s);
@@ -1777,7 +1777,7 @@ lib_mod_process_message(struct vnc *v, struct stream *s)
     {
         if (type == S2C_FRAMEBUFFER_UPDATE)
         {
-            switch (v->initial_resize_status)
+            switch (v->resize_status)
             {
                 case VRS_WAITING_FOR_FIRST_UPDATE:
                     error = lib_framebuffer_first_update(v);
@@ -2237,7 +2237,7 @@ lib_mod_connect(struct vnc *v)
 
     if (error == 0)
     {
-        v->initial_resize_status = VRS_WAITING_FOR_FIRST_UPDATE;
+        v->resize_status = VRS_WAITING_FOR_FIRST_UPDATE;
         error = send_update_request_for_resize_status(v);
     }
 
@@ -2466,6 +2466,34 @@ lib_mod_suppress_output(struct vnc *v, int suppress,
 }
 
 /******************************************************************************/
+/* return error */
+int
+lib_mod_server_version_message(struct vnc *v)
+{
+    return 0;
+}
+
+/******************************************************************************/
+/* return error */
+int
+lib_mod_server_monitor_resize(struct vnc *v, int width, int height)
+{
+    int error = 0;
+    set_single_screen_layout(&v->client_layout, width, height);
+    v->resize_status = VRS_WAITING_FOR_FIRST_UPDATE;
+    error = send_update_request_for_resize_status(v);
+    return error;
+}
+
+/******************************************************************************/
+/* return error */
+int
+lib_mod_server_monitor_full_invalidate(struct vnc *v, int param1, int param2)
+{
+    return 0;
+}
+
+/******************************************************************************/
 tintptr EXPORT_CC
 mod_init(void)
 {
@@ -2486,6 +2514,9 @@ mod_init(void)
     v->mod_check_wait_objs = lib_mod_check_wait_objs;
     v->mod_frame_ack = lib_mod_frame_ack;
     v->mod_suppress_output = lib_mod_suppress_output;
+    v->mod_server_monitor_resize = lib_mod_server_monitor_resize;
+    v->mod_server_monitor_full_invalidate = lib_mod_server_monitor_full_invalidate;
+    v->mod_server_version_message = lib_mod_server_version_message;
 
     /* Member variables */
     v->enabled_encodings_mask = -1;
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -24,6 +24,8 @@
 #include "xrdp.h"
 #include "log.h"
 #include "string_calls.h"
+#include "ms-rdpedisp.h"
+#include "ms-rdpbcgr.h"
 
 #ifdef USE_PAM
 #if defined(HAVE__PAM_TYPES_H)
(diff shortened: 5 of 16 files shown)


Nexarian:
I anticipate there will be quite a bit of discussion around this one.

@jsorg71 -- Since this is based initially on work you've done, thoughts?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+/******************************************************************************/
+static int
+dynamic_monitor_open_response(intptr_t id, int chan_id, int creation_status)
+{
+    struct xrdp_process *pro;
+    struct xrdp_wm *wm;
+    struct stream *s;
+    int bytes;
+
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_open_response: chan_id %d creation_status 0x%8.8x", chan_id, creation_status);
```
matt335672, line 1003: LOG_DEVEL(LOG_LEVEL_TRACE,...)?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    if (creation_status != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "dynamic_monitor_open_response: error");
```
matt335672, line 1006: Happy for a LOG() here as this is a user error, but shouldn't it be LOG_LEVEL_ERROR ?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    libxrdp_drdynvc_data(wm->session, chan_id, s->data, bytes);
+    free_stream(s);
+    return 0;
+}
+
+/******************************************************************************/
+static int
+dynamic_monitor_close_response(intptr_t id, int chan_id)
+{
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_close_response:");
```
matt335672, line 1029: LOG_DEVEL(LOG_LEVEL_TRACE,...) ?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    return 0;
+}
+
+/******************************************************************************/
+static int
+dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
+                           int total_bytes)
+{
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_data_first:");
```
matt335672, line 1038: LOG_DEVEL(LOG_LEVEL_TRACE,...) ?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    int PhysicalHeight;
+    int Orientation;
+    int DesktopScaleFactor;
+    int DeviceScaleFactor;
+
+    struct xrdp_rect rect;
+    int session_width;
+    int session_height;
+
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_data:");
```
matt335672, line 1072: LOG_DEVEL(LOG_LEVEL_TRACE,...) ?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    in_uint32_le(s, msg_length);
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_data: msg_type %d msg_length %d",
+           msg_type, msg_length);
+
+    rect.left = 8192;
+    rect.top = 8192;
+    rect.right = -8192;
+    rect.bottom = -8192;
+
+    if (msg_type == 2) /* DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT */
```
matt335672, line 1091: It's worth removing these magic numbers.

Could you add a file `common/ms-rdpedisp.h` with these defines in it:-

```
#define DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT 0x00000002
#define DISPLAYCONTROL_PDU_TYPE_CAPS 0x00000005
```

You'll also need a header and a `#ifdef` guard in the file. Use the existing `common/ms-rdperp.h` as a template and modify accordingly.

The naming structure we have for these files makes it easier to cross-reference with the documentation.

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    wm = pro->wm;
+    g_memset(&ls, 0, sizeof(ls));
+    ls.data = data;
+    ls.p = ls.data;
+    ls.size = bytes;
+    ls.end = ls.data + bytes;
+    s = &ls;
+    in_uint32_le(s, msg_type);
+    in_uint32_le(s, msg_length);
+    LOG(LOG_LEVEL_INFO, "dynamic_monitor_data: msg_type %d msg_length %d",
```
matt335672, line 1083: Definitely a LOG_DEVEL(). Possible LOG_LEVEL_DEBUG?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    rect.left = 8192;
+    rect.top = 8192;
+    rect.right = -8192;
+    rect.bottom = -8192;
+
+    if (msg_type == 2) /* DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT */
+    {
+        in_uint32_le(s, MonitorLayoutSize);
+        in_uint32_le(s, NumMonitor);
+        LOG(LOG_LEVEL_INFO, "  MonitorLayoutSize %d NumMonitor %d",
```
matt335672, line 1095: LOG_DEVEL(LOG_LEVEL_DEBUG,...) ?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+            in_uint32_le(s, Left);
+            in_uint32_le(s, Top);
+            in_uint32_le(s, Width);
+            in_uint32_le(s, Height);
+            in_uint32_le(s, PhysicalWidth);
+            in_uint32_le(s, PhysicalHeight);
+            in_uint32_le(s, Orientation);
+            in_uint32_le(s, DesktopScaleFactor);
+            in_uint32_le(s, DeviceScaleFactor);
+            LOG(LOG_LEVEL_INFO, "    Flags 0x%8.8x Left %d Top %d "
```
matt335672, line 1109: LOG_DEVEL(LOG_LEVEL_DEBUG, )

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+        }
+    }
+    return 0;
+}
+
 /******************************************************************************/
 int
 xrdp_mm_drdynvc_up(struct xrdp_mm *self)
 {
     LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_mm_drdynvc_up:");
```
matt335672, line 1153: LOG_LEVEL_TRACE?

xrdp/xrdp_mm.c
```diff
@@ -991,11 +991,185 @@ xrdp_mm_process_rail_drawing_orders(struct xrdp_mm *self, struct stream *s)
...
+    d_procs.data_first = dynamic_monitor_data_first;
+    d_procs.data = dynamic_monitor_data;
+    flags = 0;
+    error = libxrdp_drdynvc_open(self->wm->session,
+                                 "Microsoft::Windows::RDS::DisplayControl",
+                                 flags, &d_procs,
+                                 &(self->dynamic_monitor_chanid));
+    if (error != 0)
+    {
+        LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_mm_drdynvc_up: libxrdp_drdynvc_open failed %d", error);
```
matt335672, line 1170: Definitely needs a LOG() as this is something the user needs to report. Suggest LOG_LEVEL_ERROR.

xup/xup.c
```diff
@@ -1292,6 +1258,119 @@ process_server_paint_rect_shmem_ex(struct mod *amod, struct stream *s)
...
+    out_uint32_le(s, width);
+    out_uint32_le(s, height);
+    out_uint32_le(s, bpp);
+    out_uint32_le(s, 0);
+    s_mark_end(s);
+    int len = (int)(s->end - s->data);
+    s_pop_layer(s, iso_hdr);
+    out_uint32_le(s, len);
+    int rv = lib_send_copy(mod, s);
+    LOG_DEVEL(LOG_LEVEL_INFO, "send_server_monitor_resize: sent resize message with following properties to xorgxrdp backend "
```
matt335672, line 1302: LOG_LEVEL_DEBUG ?

xup/xup.c
```diff
@@ -1292,6 +1258,119 @@ process_server_paint_rect_shmem_ex(struct mod *amod, struct stream *s)
...
+    i = ((width & 0xffff) << 16) | height;
+    out_uint32_le(s, i);
+    out_uint32_le(s, 0);
+    out_uint32_le(s, 0);
+    s_mark_end(s);
+    int len = (int)(s->end - s->data);
+    s_pop_layer(s, iso_hdr);
+    out_uint32_le(s, len);
+    int rv = lib_send_copy(mod, s);
+    LOG_DEVEL(LOG_LEVEL_INFO, "send_server_monitor_full_invalidate: sent invalidate message with following properties to xorgxrdp backend "
```
matt335672, line 1329: LOG_LEVEL_DEBUG ?

matt335672 (commented):
@Nexarian - I've done an initial scan of the code and I have added some comments related primarily to structure and logging. I've not done any testing of my own yet.

I hope these are useful. Feel free to come back at me if I've not explained things well enough.

matt335672:
@Nexarian - something else in the logging which @aquesnel has added is that you can selectively change log levels for LOG_DEVEL() messages per-file or per-function when running. There's a bit more info on this in the description of the LOG_DEVEL() macro. It could be handy for testing/debugging this facility.

bolkedebruin:
@Nexarian what are the conditions to trigger a resize? I'm testing this with a Mac client that has enabled "update session on resize". It does "smart resize" (aka scale on the clients side) but it doesnt do a real resize at the moment.

Nexarian:
@bolkedebruin The option you need to enable looks like this:

image

Smart sizing indicates that you have a fixed resolution selected, which appears to be mutually exclusive with the resize flag. Next, double check you're using the latest version, Microsoft just released a new version, I'm using `10.5.2`

Second, try connecting to a Windows 10 installation and make sure that the dynamic resolution resizing works on that, as we're 100% sure it works there.

The other option is to try FreeRDP with `/dynamic-resolution`

Reach out to me on Gitter, and I'm happy to help more!

README.md
```diff
@@ -9,9 +9,15 @@
...
-Remote Desktop Protocol (RDP). xrdp accepts connections from a variety of
-RDP clients: FreeRDP, rdesktop, KRDC, NeutrinoRDP and Microsoft Remote Desktop
-Client (for Windows, Mac OS, iOS and Android).
+Remote Desktop Protocol (RDP). xrdp accepts connections from a variety of RDP clients:
+  - FreeRDP
+  - rdesktop
+  - KRDC
+  - NeutrinoRDP
+  - Windows MSTSC (Microsoft Terminal Services Client)
+  - Microsoft Remote Desktop (which is distinct from MSTSC and the only Microsoft clients that support resizing without reconnection)
```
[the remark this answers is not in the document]
Nexarian: Updated.

libxrdp/xrdp_channel.c
```diff
@@ -777,6 +777,12 @@ xrdp_channel_drdynvc_start(struct xrdp_channel *self)
     struct mcs_channel_item *dci;
 
 
+    LOG(LOG_LEVEL_INFO, "xrdp_channel_drdynvc_start: drdynvc_channel_id %d", self->drdynvc_channel_id);
```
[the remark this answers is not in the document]
Nexarian: Updated.

libxrdp/xrdp_channel.c
```diff
@@ -777,6 +777,12 @@ xrdp_channel_drdynvc_start(struct xrdp_channel *self)
...
+    if (self->drdynvc_channel_id != -1)
+    {
+        LOG(LOG_LEVEL_INFO, "xrdp_channel_drdynvc_start: already started");
```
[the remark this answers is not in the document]
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xup/xup.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Unfortunately when I make this modification, I get this:

```
xrdp_mm.c:1067:9: error: variable ‘DeviceScaleFactor’ set but not used [-Werror=unused-but-set-variable]
 1067 |     int DeviceScaleFactor;
      |         ^~~~~~~~~~~~~~~~~
xrdp_mm.c:1066:9: error: variable ‘DesktopScaleFactor’ set but not used [-Werror=unused-but-set-variable]
 1066 |     int DesktopScaleFactor;
      |         ^~~~~~~~~~~~~~~~~~
xrdp_mm.c:1065:9: error: variable ‘Orientation’ set but not used [-Werror=unused-but-set-variable]
 1065 |     int Orientation;
      |         ^~~~~~~~~~~
xrdp_mm.c:1064:9: error: variable ‘PhysicalHeight’ set but not used [-Werror=unused-but-set-variable]
 1064 |     int PhysicalHeight;
      |         ^~~~~~~~~~~~~~
xrdp_mm.c:1063:9: error: variable ‘PhysicalWidth’ set but not used [-Werror=unused-but-set-variable]
 1063 |     int PhysicalWidth;
      |         ^~~~~~~~~~~~~
xrdp_mm.c:1058:9: error: variable ‘Flags’ set but not used [-Werror=unused-but-set-variable]
 1058 |     int Flags;
      |         ^~~~~
xrdp_mm.c:1055:9: error: variable ‘MonitorLayoutSize’ set but not used [-Werror=unused-but-set-variable]
 1055 |     int MonitorLayoutSize;
      |         ^~~~~~~~~~~~~~~~~
```

For now, I've moved them to `LOG_LEVEL_DEBUG` and kept the offending log lines as `LOG` -- That may not be the ideal solution we end up with in the end, however.

Nexarian: Ok, I've updated this further so that it passes by moving it into a struct.

Nexarian:
@matt335672 Updated for your comments and tested with https://github.com/Nexarian/xorgxrdp/tree/resize_damage_enhancement

Works fine!

bolkedebruin:
I have tested this and got it working. Firstly, the `drdynvc` channel should be enabled for this to work (we disable channels as much as we can for security reasons). Furthermore, with `max_bpp=16` (we use this for speed reasons) only `freerdp` was working. The w10 clients fails to connect at all and the Mac client does connect but crashes quickly after. Setting it to `32` makes it work with the MS clients.

vnc/vnc.c
```diff
@@ -2465,6 +2465,34 @@ lib_mod_suppress_output(struct vnc *v, int suppress,
...
+int
+lib_mod_server_version_message(struct vnc *v)
+{
+    return 0;
+}
+
+/******************************************************************************/
+/* return error */
+int
+lib_mod_server_monitor_resize(struct vnc *v, int width, int height, int bpp)
```
bolkedebruin, line 2479: `bpp` is ignored here. Why?

matt335672:
@bolkedebruin - I'm connecting to an Ubuntu 20.04 machine with `max_bpp=16` in xrdp.ini from the Windows 10 client (version 10.2.1818.0) on a single-monitor setup, and it seems to be working for me.

I'm assuming you're using Xvnc on the back end(?)

I'm getting :-

```
$ xdpyinfo | grep 'depth of root window'
  depth of root window:    16 planes
$ Xvnc -version

Xvnc TigerVNC 1.10.0 - built Apr  9 2020 06:49:31
Copyright (C) 1999-2019 TigerVNC Team and many others (see README.rst)
See https://www.tigervnc.org for information on TigerVNC.
Underlying X server release 12008000, The X.Org Foundation

```

Can you post some info on your config so we can reproduce your crash? I can't do the Mac client I'm afraid.

As for your question regarding the `bpp` setting, in the vnc module, it isn't used because there's no way to change the monitor depth on-the-fly. The resize messages in MS-RDPEDISP don't allow for a colour depth to be part of the resize message (AFAICT), so I suspect the simplest thing to is to take the `bpp` parameter out of the mod_server_monitor_resize member function in `struct xrdp_mod`. What do you think @Nexarian?

bolkedebruin:
No we use Xorg.

Nexarian:
I just pushed a fix. One of the functions was hard-coding the BPP to 32, even though some of the function calls directly after it were using the user-configured value! Tested with the Mac OS client that it's working.

@matt335672 One of the signals that is sent to the xorg backend that is necessary for the resize handshake includes the bpp. I'm almost certain that this is something that could be dropped in the future, but right now I don't want to go about changing the interface between xrdp and xorg. Probably we should create a backlog item here. The resize code came from a refactor of the session initialization code.

@bolkedebruin Agreed that with MSTSC, this feature is less useful. The good news is that the "Microsoft Remote Desktop" client that comes from the Microsoft App Store does have the same resize features as the Mac client. That's what eventually inspired me to take on this feature, knowing that there was a true Windows 10 client that could indeed do this.

matt335672:
@Nexarian - I've been looking again at the (potentially) confusing use of bpp in the argument list for  the `mod_server_monitor_resize` member function in `struct xrdp_mod`.

I fully get your reasons for needing to send this value to xorgxrdp, but given that the existing  bpp value is already part of the mod data, you don't need the bpp parameter. You can do something llke this:-

```diff
diff --git a/xup/xup.c b/xup/xup.c
index ff66a959..9711dee9 100644
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -1353,7 +1353,7 @@ lib_send_server_monitor_resize(struct mod *mod, int width, int height, int bpp)
     /* send screen size message */
     struct stream *s;
     make_stream(s);
-    int rv = send_server_monitor_resize(mod, s, width, height, bpp);
+    int rv = send_server_monitor_resize(mod, s, width, height, mod->bpp);
     free_stream(s);
     return rv;
 }
```

That then leaves you free to remove the bpp parameter from `mod_server_monitor_resize`.

What do you think?

Nexarian:
@matt335672 Agreed. Pushed an update. Tested to verify it works.

bolkedebruin:
checkingg why we make the change to the bpp settings @matt335672 and @Nexarian. What happens if I change my BPP on the fly or switch to a monitor that has a different bpp setting?

Nexarian:
From what I understand BPP isn't something that can be changed on the fly, only resolution. I think this is a limitation of the Microsoft protocol. @matt335672 -- Can you confirm?

common/xrdp_client_info.h
```diff
@@ -23,6 +23,8 @@
 #if !defined(XRDP_CLIENT_INFO_H)
 #define XRDP_CLIENT_INFO_H
 
+#define XRDP_MAXIMUM_MONITORS 16
```
metalefty, line 26: Whhere does this maximum number of monitors come from?

metalefty: I think this should be defined somewhere else, not in client_info at least.

matt335672:
@bolkedebruin - the dynamic client resize protocol is detailed in [[MS-RDPEDISP]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/). It doesn't contain any provision for changing bpp on the fly.

From the testing I've just done, changing the client resolution on reconnect results in unchanged behaviour (i.e. new session for Xvnc, same sesion for Xorg).

Does that answer your question, or have I missed something?

common/xrdp_client_info.h
matt335672: The max monitors occurs in [[MS-RDPBCGR] section 2.2.1.3.6](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/a8029f8e-01e1-4f19-af67-bcad5bdef624), although it's not given an actual name.

There's some precedent for adding it to `common/ms-rdgbcgr.h` - see `CHANNEL_NAME_LEN` in that file.

matt335672:
Hi @Nexarian 

Just realised as part of another conversation that you'll need to add the extra stub functions to neutrinordp as well. My bad - I should have thought of it before.

I've built neutrinordp and used it to connect to a Win10 box using the Store client. When I resize the client, I get a SEGV here:-

https://github.com/Nexarian/xrdp/blob/52c52055432ef303407653ef352126a1edb75e9f/xrdp/xrdp_mm.c#L1136

common/xrdp_client_info.h
Nexarian: I've added `CLIENT_MONITOR_DATA_MAXIMUM_MONITORS` to ms-rdgbcgr.h and pushed the update. Let me know what you think.

Nexarian:
@matt335672 Good call out! Should NeutrinoRDP also implement these functions, or is stubbing enough until we integrate with FreeRDP 2.0?

I've pushed an update for the stub functions. Since I'm not sure how to test this, if you could test it for me, I'd appreciate it.

matt335672:
Hi @Nexarian 

I've tested the stubbing out, and it's not really satisfactory. The problem is I think that because the dynamic channel's been opened, the client is expecting its DISPLAYCONTROL_MONITOR_LAYOUT_PDU requests to be honoured. So if you start full screen and resize to a smaller screen, you don't get anything useful.

I've had a look through neutrinordp and frankly adding a resize to that is beyond me.

I've tried playing about with not opening the channel at all if the module doesn't support resizing. This doesn't help though. The first call to `xrdp_mm_drdynvc_up()` is before the user has selected a module, so at that point you don't have enough information as to whether to bring the channel up or not. The only way I can think to do that is to delay opening the dynamic channel until the module is loaded. That might work but would  prevent any resize of the login screen being possible.

Does that sound sensible at this point?

Nexarian:
I didn't know how to implement this at all when I first started either. I think we can take a crack at doing some archaeology in the codebase(s) and figure out how to implement this. It will help our understanding of how the entire ecosystem works.

As to the suggestion to delay module loading. I like it, but not as a feature for this issue. I think if we're going to implement that, we should do it right, and add a system to XRDP that determines what channels or features to activate such that they are all only turned on or off based on what module the user has selected at login. We could make this PR dependent on that feature, so as to keep them separate.

The final option is to release dynamic resize as a feature flag/experimental for now, while we work on one or both of the above two.

matt335672:
Another (still temporary) solution is to add some code to `xrdp_mm_drdynvc_up` to allow the user to disable this dynamic channel. At the moment it's enable unconditionally. Being able to turn it off gives a way forward for NeutrinoRDP users that might come across this.

Nexarian:
I'm digging into whether it's possible to add this functionality to NeutrinoRDP. I think it is, but not yet certain. I don't think FreeRDP 1.0.1 supported this functionality, so it might not be possible until it's updated to 2.0.

bolkedebruin:
We have tested this and it works well without regressions. Of course the clients do not enable it automatically so resizing testing was limited.

matt335672:
Thanks @bolkedebruin - that's really useful info with a new release coming up.

Nexarian:
I've updated this PR to include a fix for NeutrinoRDP that allows for an option to disable dynamic resizing in the configuration file.

@matt335672 -- This is largely based on the patch you sent me. I've verified it works, and I didn't see any of the errors in the logs that you referenced.

matt335672 (approved):
I'm happy with this for v0.9.16.

@Nexarian - thanks for all the work you've put into this!

metalefty:
Can you rebase to the latest devel HEAD? Then I'll have a test before the merge.

Nexarian:
What makes you think it isn't rebased? As far as I'm aware, I did that when I force-pushed the update with the NeutrinoRDP fix.

metalefty:
My bad, it is already rebased. It is working quite well. Thank you very much for your hard work!

metalefty:
It would be great if we can see more detailed logs when on-the-fly-resizing happens.
Display number, old size, new size. We can do that later.

## Xorg
```
[20210424-21:52:46] [DEBUG] dynamic_monitor_data: msg_type 2 msg_length 56
[20210424-21:52:46] [DEBUG]   MonitorLayoutSize 40 NumMonitor 1
[20210424-21:52:46] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20210424-21:52:46] [INFO ] xrdp_process_offscreen_bmpcache: support level 0 cache size 0 MB cache entries 0
[20210424-21:52:46] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
```

## VNC
```
[20210424-21:51:31] [DEBUG] dynamic_monitor_data: msg_type 2 msg_length 56
[20210424-21:51:31] [DEBUG]   MonitorLayoutSize 40 NumMonitor 1
[20210424-21:51:31] [DEBUG] Skipping ENC_RAW encoding
[20210424-21:51:31] [DEBUG] Skipping ENC_RAW encoding
[20210424-21:51:31] [DEBUG] VNC server does not support resizing
[20210424-21:51:31] [DEBUG] Resizing client to server 988x961
[20210424-21:51:31] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20210424-21:51:31] [INFO ] xrdp_process_offscreen_bmpcache: support level 0 cache size 0 MB cache entries 0
[20210424-21:51:31] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[20210424-21:51:31] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20210424-21:51:31] [INFO ] xrdp_process_offscreen_bmpcache: support level 0 cache size 0 MB cache entries 0
[20210424-21:51:31] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[20210424-21:51:37] [DEBUG] dynamic_monitor_data: msg_type 2 msg_length 56
[20210424-21:51:37] [DEBUG]   MonitorLayoutSize 40 NumMonitor 1
[20210424-21:51:37] [DEBUG] VNC matched ExtendedDesktopSize rectangle x=0, y=0 geom=988x961
[20210424-21:51:37] [DEBUG] VNC server supports resizing
[20210424-21:51:37] [INFO ] Layout from OldLayout (geom=988x961 #screens=1) : 33613:(988x961+0+0)
[20210424-21:51:37] [DEBUG] VNC setting screen id to 33613 from server
[20210424-21:51:37] [DEBUG] Server layout differs from the client layout. Changing server layout
[20210424-21:51:37] [DEBUG] VNC Sending SetDesktopSize
[20210424-21:51:37] [DEBUG] Skipping ENC_EXTENDED_DESKTOP_SIZE encoding x=0, y=0 geom=988x961
[20210424-21:51:37] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20210424-21:51:37] [INFO ] xrdp_process_offscreen_bmpcache: support level 0 cache size 0 MB cache entries 0
[20210424-21:51:37] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[20210424-21:51:37] [DEBUG] Skipping ENC_EXTENDED_DESKTOP_SIZE encoding x=0, y=0 geom=1172x704
[20210424-21:51:37] [DEBUG] Skipping ENC_EXTENDED_DESKTOP_SIZE encoding x=0, y=0 geom=1172x704
[20210424-21:51:37] [DEBUG] Skipping ENC_EXTENDED_DESKTOP_SIZE encoding x=0, y=0 geom=1172x704
[20210424-21:51:37] [DEBUG] Skipping ENC_EXTENDED_DESKTOP_SIZE encoding x=0, y=0 geom=1172x704
[20210424-21:51:37] [DEBUG] VNC matched ExtendedDesktopSize rectangle x=1, y=0 geom=1172x704
[20210424-21:51:37] [DEBUG] VNC server successfully resized
[20210424-21:51:37] [INFO ] Layout from NewLayout (geom=1172x704 #screens=1) : 33613:(1172x704+0+0)
```

metalefty (approved):

jsorg71:
Thanks everyone who worked on this.
I'll delete my branch.

prg318:
I just really wanted to step in and thank everyone that put effort into this feature.  This is really awesome functionality that I have been wanting to use for quite some time. It's fantastic to see it already working flawlessly on the -devel branches.

This is awesome and you all rule! Thanks again!!

Nexarian:
@jsorg71 @prg318 You're very welcome!

I use XRDP regularly, so this was selfishly for me as much as anyone else, but I am immensely gratified knowing it's improving the experience of everyone who uses it!

And now I want to move on to aid in other feature improvements, like https://github.com/neutrinolabs/xrdp/issues/1422.  Turns out, dynamic resolution for that isn't a walk in the park, either!

matt335672:
@prg318 - thank you for your kind words. It really does make a difference.

PR #1821, opened 2021-03-04
Include OS version in github action cache for cppcheck

matt335672:
This change includes the OS version in the cppcheck cache name for the cppcheck github action.

This ensures that when `ubuntu-latest` is updated, we do not use the old cache - moving from 18.04 to 20.04 broke cppcheck as the symbol `Z3_get_error_msg` was removed from the Z3 library - see https://github.com/Z3Prover/z3/issues/660

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        
+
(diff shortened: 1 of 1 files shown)


aquesnel (approved):
Looks good

Nexarian (approved):

PR #1826, opened 2021-03-06
Unify logging in common/*

aquesnel:
Follow-on pull request to #1633 which migrates all logging in the `common` directory to use the LOG() and LOG_DEVEL() macros.

Changes:

* Migrate logging to the LOG() and LOG_DEVEL() macros (using text transform script)
* error logs use the LOG macro
* update code formatting with astyle for all files in common/*
* remove `DEBUG` macro


matt335672:
Apologies on the collision with `xrdp_client_info.h`. Since my PR also ran astyle on this file, it's a simple fix.

All looks absolutely fine, with just one question on `common/pixman-region.c` which is probably best answered by @metalefty 

This file is originally from  https://cgit.freedesktop.org/pixman/tree/pixman/pixman-region.c, and if I understand it correctly, is intended for systems which do not necessarily have pixman installed.

If we `astyle` this file, it will make it harder to compare it with upstream for regressions. On the other hand, this file is pretty mature now, and if we had serious problems we'd have found them by now.

For reference, the last time this file was modified is got a review comment [here](/neutrinolabs/xrdp/pull/1771#pullrequestreview-560322914).

aquesnel:
I've fixed up the merge conflicts. 

The build is broken because of an issue with the github action runner image (see: https://github.com/actions/virtual-environments/issues/2917 ). I've verified that the root cause for the failed package install is the same for the xrdp build and the build in the issue, namely that the libx32gcc-s1 package is not available from the package archive configured in the VM (see xrdp build with apt-get debugging enabled: https://github.com/aquesnel/xrdp/runs/2109124288?check_suite_focus=true#step:4:296 )

matt335672:
I'm happy with the `common/pixman-region.c` workaround you suggest, so I'll merge this one I think.

BTW, nice bit of detective work on the CI problem. You've made it clear that it's affecting users other than us, so I think we can leave it for a bit and see what happens. I've subscribed to the issue you've linked to above.

PR #1843, opened 2021-03-26
Adding logs to sesman/session.c

aquesnel:
Adding logs to sesman/session.c to help a sys admin understand why a session failed to start.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,7 +6,7 @@ FreeBSD_task:
     freebsd_instance:
       image_family: freebsd-12-2
   prepare_script:
-    - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm fusefs-libs
+    - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm fusefs-libs check
     - git submodule update --init --recursive
   configure_script:
     - ./bootstrap
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2591,16 +2591,41 @@ g_get_errno(void)
 
 /*****************************************************************************/
 /* does not work in win32 */
+
+#define ARGS_STR_LEN 1024
+
 int
 g_execvp(const char *p1, char *args[])
 {
 #if defined(_WIN32)
     return 0;
 #else
+
     int rv;
+    char args_str[ARGS_STR_LEN];
+    int args_len;
+
+    args_len = 0;
+    while (args[args_len] != NULL)
+    {
+        args_len++;
+    }
+
+    g_strnjoin(args_str, ARGS_STR_LEN, " ", (const char **) args, args_len);
+
+    LOG(LOG_LEVEL_DEBUG,
+        "Calling exec (excutable: %s, arguments: %s)",
+        p1, args_str);
 
     g_rm_temp_dir();
     rv = execvp(p1, args);
+
+    /* should not get here */
+    LOG(LOG_LEVEL_ERROR,
+        "Error calling exec (excutable: %s, arguments: %s) "
+        "returned errno: %d, description: %s",
+        p1, args_str, g_get_errno(), g_get_strerror());
+
     g_mk_socket_path(0);
     return rv;
 #endif
@@ -2615,9 +2640,24 @@ g_execlp3(const char *a1, const char *a2, const char *a3)
     return 0;
 #else
     int rv;
+    const char *args[] = {a2, a3, NULL};
+    char args_str[ARGS_STR_LEN];
+
+    g_strnjoin(args_str, ARGS_STR_LEN, " ", args, 2);
+
+    LOG(LOG_LEVEL_DEBUG,
+        "Calling exec (excutable: %s, arguments: %s)",
+        a1, args_str);
 
     g_rm_temp_dir();
     rv = execlp(a1, a2, a3, (void *)0);
+
+    /* should not get here */
+    LOG(LOG_LEVEL_ERROR,
+        "Error calling exec (excutable: %s, arguments: %s) "
+        "returned errno: %d, description: %s",
+        a1, args_str, g_get_errno(), g_get_strerror());
+
     g_mk_socket_path(0);
     return rv;
 #endif
@@ -2715,6 +2755,12 @@ g_fork(void)
     if (rv == 0) /* child */
     {
         g_mk_socket_path(0);
+    }
+    else if (rv == -1) /* error */
+    {
+        LOG(LOG_LEVEL_ERROR,
+            "Process fork failed with errno: %d, description: %s",
+            g_get_errno(), g_get_strerror());
     }
 
     return rv;
@@ -2873,6 +2919,51 @@ g_waitpid(int pid)
     }
 
     return rv;
+#endif
+}
+
+/*****************************************************************************/
+/* does not work in win32
+   returns exit status code of child process with pid */
+struct exit_status
+g_waitpid_status(int pid)
+{
+    struct exit_status exit_status;
+
+#if defined(_WIN32)
+    exit_status.exit_code = -1;
+    exit_status.signal_no = 0;
+    return exit_status;
+#else
+    int rv;
+    int status;
+
+    exit_status.exit_code = -1;
+    exit_status.signal_no = 0;
+
+    if (pid > 0)
+    {
+        LOG(LOG_LEVEL_DEBUG, "waiting for pid %d to exit", pid);
+        rv = waitpid(pid, &status, 0);
+
+        if (rv != -1)
+        {
+            if (WIFEXITED(status))
+            {
+                exit_status.exit_code = WEXITSTATUS(status);
+            }
+            if (WIFSIGNALED(status))
+            {
+                exit_status.signal_no = WTERMSIG(status);
+            }
+        }
+        else
+        {
+            LOG(LOG_LEVEL_WARNING, "wait for pid %d returned unknown result", pid);
+        }
+    }
+
+    return exit_status;
 #endif
 }
 
--- a/common/os_calls.h
+++ b/common/os_calls.h
@@ -23,6 +23,15 @@
 
 #include "arch.h"
 
+struct exit_status
+{
+    /* set to -1 when the process exited via a signal */
+    uint8_t exit_code;
+    
+    /* set to 0 when the process exited normally */
+    uint8_t signal_no;
+};
+
 #define g_tcp_can_recv g_sck_can_recv
 #define g_tcp_can_send g_sck_can_send
 #define g_tcp_recv g_sck_recv
@@ -145,6 +154,7 @@ int      g_getlogin(char *name, unsigned int len);
 int      g_setlogin(const char *name);
 int      g_waitchild(void);
 int      g_waitpid(int pid);
+struct exit_status g_waitpid_status(int pid);
 void     g_clearenv(void);
 int      g_setenv(const char *name, const char *value, int rewrite);
 char    *g_getenv(const char *name);
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -30,6 +30,7 @@
 #include "log.h"
 #include "os_calls.h"
 #include "string_calls.h"
+#include "defines.h"
 
 unsigned int
 g_format_info_string(char *dest, unsigned int len,
@@ -805,4 +806,51 @@ g_strtrim(char *str, int trim_flags)
     free(text);
     free(text1);
     return 0;
+}
+
+/*****************************************************************************/
+char *
+g_strnjoin(char *dest, int dest_len, const char *joiner, const char *src[], int src_len)
+{
+    int len = 0;
+    int joiner_len;
+    int i = 0;
+    int dest_remaining;
+    char *dest_pos = dest;
+    char *dest_end;
+    
+    if (dest == NULL || dest_len < 1)
+    {
+        return dest;
+    }
+    if (src == NULL || src_len < 1)
+    {
+        dest[0] = '\0';
+        return dest;
+    }
+
+    dest[0] = '\0';
+    dest_end = dest + dest_len - 1;
+    joiner_len = g_strlen(joiner);
+    for (i = 0; i < src_len - 1 && dest_pos < dest_end; i++)
+    {
+        len = g_strlen(src[i]);
+        dest_remaining = dest_end - dest_pos;
+        g_strncat(dest_pos, src[i], dest_remaining);
+        dest_pos += MIN(len, dest_remaining);
+
+        if (dest_pos < dest_end)
+        {
+            dest_remaining = dest_end - dest_pos;
+            g_strncat(dest_pos, joiner, dest_remaining);
+            dest_pos += MIN(joiner_len, dest_remaining);
+        }
+    }
+
+    if (i == src_len - 1 && dest_pos < dest_end)
+    {
+        g_strncat(dest_pos, src[i], dest_end - dest_pos);
+    }
+
+    return dest;
 }
--- a/common/string_calls.h
+++ b/common/string_calls.h
@@ -81,6 +81,33 @@ int
 g_text2bool(const char *s);
 
 /**
+ * Joins an array of strings into a single string.
+ *
+ * Note: The joiner is placed between each source string. The joiner is not
+ *      placed after the last source string. If there is only one source string,
+ *      then the result string will be equal to the source string.
+ *
+ * Note: any content that is present in dest will be overwritten with the new
+ *      joined string.
+ *
+ * Note: If the destination array is not large enough to hold the entire
+ *      contents of the joined string, then the joined string will be truncated
+ *      to fit in the destination array.
+ *
+ * @param[out] dest The destination array to write the joined string into.
+ * @param[in] dest_len The max number of characters to write to the destination
+ *      array including the terminating null. Must be > 0
+ * @param[in] joiner The string to concatenate between each source string.
+ *      The joiner string may be NULL which is processed as a zero length string.
+ * @param[in] src An array of strings to join. The array must be non-null.
+ *      Array items may be NULL and are processed as zero length strings.
+ * @param[in] src_len The number of strings to join in the src array. Must be > 0
+ * @return A pointer to the begining of the joined string (ie. returns dest).
+ */
+char *
+g_strnjoin(char *dest, int dest_len, const char *joiner, const char *src[], int src_len);
+
+/**
  * Converts a binary array into a hux dump suitable for displaying to a user.
  *
  * The format of the hex dump is:
--- a/configure.ac
+++ b/configure.ac
@@ -357,6 +357,11 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
 
 CFLAGS="$save_CFLAGS"
 
+# checking for libcheck (used for unit testing)
+PKG_CHECK_MODULES([CHECK], [check >= 0.10.0])
+AC_CHECK_HEADER([check.h], [],
+  [AC_MSG_ERROR([please install check])])
+
 AC_SUBST([moduledir], '${libdir}/xrdp')
 
 AC_ARG_ENABLE([strict-locations],
@@ -398,6 +403,7 @@ AC_CONFIG_FILES([
   sesman/Makefile
   sesman/tools/Makefile
   tests/Makefile
+  tests/common/Makefile
   tests/memtest/Makefile
   tools/Makefile
   tools/devel/Makefile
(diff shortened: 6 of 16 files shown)


sesman/session.c
```diff
@@ -837,12 +889,25 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
                     "auth_stop_session and auth_end from pid %d", g_getpid());
                 auth_stop_session(data);
                 auth_end(data);
+
+                LOG(LOG_LEVEL_INFO, "Cleaning up session: terminating Xserver "
+                    "(pid %d)", display_pid);
```
metalefty, line 894: What about also record the display number?

sesman/session.c
```diff
@@ -837,12 +889,25 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
                 g_sigterm(display_pid);
+
+                LOG(LOG_LEVEL_INFO, "Cleaning up session: terminating "
+                    "xrdp-chansrv (pid %d)", display_pid);
```
metalefty, line 898: I think `display_pid` shuold be `chansrv_pid`.

sesman/session.c
```diff
@@ -173,6 +169,7 @@ x_server_running_check_ports(int display)
 
     if (!x_running)
     {
+        LOG(LOG_LEVEL_INFO, "Did not find a running X server at %s", text);
```
matt335672, line 172: This is going to potentially generate a lot of INFO messages, as `x_server_running_check_ports()` is called in a loop, and a message is logged for each reason why the X server isn't running. You're already logging the reason why the X server isn't running at the end of the function, so I think you could remove the "Did not find..." messages and just beef up the logging at the end (i.e.):-

```c
if (x_running)
{
    if (isdigit(text[0]))
    {
        LOG(LOG_LEVEL_INFO, "Found X server active on display %d (TCP port:%s)", display, text);
    }
    else
    {
        LOG(LOG_LEVEL_INFO, "Found X server active on display %d (file:%s)", display, text);
    }
}
else
{
    LOG(LOG_LEVEL_INFO, "No X server seems to be active on display %d", display);
}
```

That can be somewhat improved on (particularly using `isdigit` to detect a port), but I hope it gives you the idea.

sesman/session.c
```diff
@@ -809,6 +848,13 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
             else
             {
                 int wm_wait_time;
+                int wm_exit_code;
+                int xserver_exit_code;
+                int chansrv_exit_code;
+
+                LOG(LOG_LEVEL_DEBUG, "session_start_fork (pid %d): waiting for "
+                    "the Xserver and the creating the channel server",
```
matt335672, line 856: 'then' rather than 'the'?

sesman/session.c
```diff
@@ -818,14 +864,20 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+                wm_exit_code = g_waitpid_status(window_manager_pid);
                 wm_wait_time = g_time1() - wm_wait_time;
-                if (wm_wait_time < 10)
+                if (wm_exit_code != 0)
+                {
+                    LOG(LOG_LEVEL_WARNING, "window manager (pid %d) exited with "
+                        "non-zero exit code (%d). Window manager config problem?",
+                        window_manager_pid, wm_exit_code);
+                }
+                else if (wm_wait_time < 10)
```
matt335672, line 875: Suggest making the `wm_wait_time < 10` check unconditional, as this was explicitly introduced to catch the common case of the session exiting early. This is something I generally end up triaging, so I was hoping to be able to check logs for 'exited quickly'. Your change makes this harder to spot. How about something like:-

```c
wm_exit_code = g_waitpid_status(window_manager_pid);
wm_wait_time = g_time1() - wm_wait_time;
if (wm_exit_code != 0)
{
    LOG(LOG_LEVEL_WARNING, "window manager (pid %d) exited with "
             "non-zero exit code (%d)",
             window_manager_pid, wm_exit_code);
}
if (wm_wait_time < 10)
{
    /* Common config issue */
    LOG(LOG_LEVEL_WARNING, "window manager (pid %d) exited quickly "
             "(%d secs). Window manager config problem?",
             window_manager_pid, wm_wait_time);
}

```

common/os_calls.c
```diff
@@ -2876,6 +2876,45 @@ g_waitpid(int pid)
...
+        {
+            LOG(LOG_LEVEL_WARNING, "wait for pid %d returned unknown result", pid);
+            rv = -1;
+        }
+    }
+
+    return rv;
+#endif
+}
+
```
matt335672, line 2917: This function doesn't cope with a process exiting after catching a signal. This is relatively common for chansrv. In such cases it looks like rv is set to zero which is misleading.

There's no great answer here I can see without changing the interface, but a better solution might be to include (after the WIFEXITED() test):-

```c
else if (WIFSIGNALED(status))
{
    /* Specify a value not representable in 8 bits */
    rv = -128 - WTERMSIG(status);
}
```

The logic is that `WEXITSTATUS()`returns an 8 bit value (either 0-255 or -128-127), so we can at least see a signal for what it is. This can probably be improved on.

common/string_calls.c
```diff
@@ -139,6 +139,44 @@ g_text2bool(const char *s)
     return 0;
 }
 
+/*****************************************************************************/
+char *
+g_strnjoin(char *dest, int dest_len, const char *joiner, const char *src[], int src_len)
```
matt335672, line 144: src_len == 0 causes src[0]to be dereferenced at line 170. Probably need a guard for this (and a test?)

common/string_calls.h
```diff
@@ -80,6 +80,8 @@ g_bool2text(int value);
 int
 g_text2bool(const char *s);
 
+char    *g_strnjoin(char *dest, int dest_len, const char *joiner, const char *src[], int src_len);
```
matt335672, line 83: Suggest Doxygen header for new funcs in this module.

matt335672 (commented):
A nice illustration of check.

Couple of general points/questions:-
- I take it the plan is to use g_strjoin() in this PR? It doesn't seem to be, unless I'm missing something.
- Need to add package `check` to `.cirrus.yml`.
- I can't see why the CI is failing on Ubuntu. Any ideas??

aquesnel:
@matt335672 , yes the plan is to use g_strnjoin in to implement the feedback I got in the pull request. I broke up my changes into several commits and then I only submitted the first commit to check that this change is working on all platforms.

I did discover that my changes are failing because of the [-Wstringop-truncation](https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/) warning introduced in gcc 8 which is why the build only fails on the newer platforms. The warning is definitely valid and a bug that I'm going to fix. 

What I haven't figured out yet is why the makefile generated by autotools has the gcc output piped to `/dev/null` which seems to be the cause of there being no output in the build log.

sesman/session.c
aquesnel: done

sesman/session.c
aquesnel: fixed

sesman/session.c
aquesnel: The logging at the end of the function only says that last port/socket that was checked for an X server. In my past debugging I have found it very useful in the past to see all of the ports/sockets where xrdp is looking for an X server and in which order the checks are performed. So I've kept all the log messages but lowered the logging level to debug so that it's less noisy under normal circumstances.

sesman/session.c
aquesnel: done

sesman/session.c
aquesnel: done

common/os_calls.c
aquesnel: Since `g_waitpid_status()` is a new function I just changed the interface to return a struct with the return code and signal code.

common/string_calls.c
aquesnel: fixed

common/string_calls.h
aquesnel: done

aquesnel:
@metalefty and @matt335672 I've fixed up the code to address your feedback.

I did find out why the build was failing with no output. It turns out that the gcc `-Wstringop-truncation` check was only finding the issue when the cope was complied without `-fPIC` (position independent code). Libtool is what does the compilation of libraries, and libtool compiles the code both with and without `-fPIC` and pipes the results of compiling without `-fPIC` to `/dev/null` because the expectation is that if there is an errors from compiling the code, then there will be the same errors when compiled with and without `-fPIC`, so by default libtool hides the expected duplicate output. There is the libtool switch `-no-suppress` to change the behaviour and always show the output from both compilations.

I tried to `-no-suppress` to out config and makefiles, but as far as I can tell it can't be done. The `-no-suppress` flag is special and that it needs to be passed to libtool on the command line after the compiler path argument. The problem is that there is autotools variable which is applied to only libtool invocations which is passed after the compiler argument. The existing `AM_LIBTOOLFLAGS` and `LIBTOOLFLAGS` variables are passed to libtool before the compiler argument which makes it so that libtool fails. I also tried passing `-no-suppress` via `AM_CFLAGS` and `CFLAGS` which works for compiling the libraries but fails for  compiling the xrdp binaries (which don't use libtool).

Since this is a weird edge case and there is nothing clean that can be done, I'm just going to leave the compilation flags as is and hope that we don't keep seeing this weird edge case.

matt335672:
Nice bit of detective work on the libtool issue - that's really unexpected.

I had a quick play with this. I can suggest a partial solution, which is to build just the common library with `-no-suppress`:-
```diff
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -32,7 +32,10 @@ if XRDP_DEBUG
 AM_CPPFLAGS += -DXRDP_DEBUG
 endif
 
-AM_CFLAGS = $(OPENSSL_CFLAGS)
+# -no-suppress is an automake-specific flag which is needed
+# to prevent us missing compiler errors in some circumstances
+# (see #1843)
+AM_CFLAGS = -no-suppress $(OPENSSL_CFLAGS)
 
 module_LTLIBRARIES = \
   libcommon.la
```

I agree it's not perfect, but given most of these problems are likely to be located in `string_calls.c` it might be worth considering.

Review to follow...

sesman/session.c
```diff
@@ -396,9 +400,8 @@ session_start_chansrv(char *username, int display)
 
         /* executing chansrv */
         g_execvp(exe_path, (char **) (chansrv_params->items));
+        
         /* should not get here */
-        LOG(LOG_LEVEL_ERROR, "error starting chansrv "
-            "- user %s - pid %d", username, g_getpid());
         list_delete(chansrv_params);
```
matt335672, line 405: Don't understand why the `LOG()` is removed. We should have a log if the `g_evecvp()` fails shouldn't we?

matt335672 (commented):
Thanks for the updates to this. Let me know if I haven't been clear on any of the above comments.

.cirrus.yml
```diff
@@ -6,7 +6,7 @@ FreeBSD_task:
     freebsd_instance:
       image_family: freebsd-12-2
   prepare_script:
-    - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm
+    - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm check
```
metalefty, line 9: Thanks for the catch!

common/string_calls.c
```diff
@@ -651,3 +651,42 @@ g_strtrim(char *str, int trim_flags)
...
+{
+    int len = 0;
+    int joiner_len;
+    int i = 0;
+    char *dest_pos = dest;
+    char *dest_end;
+
+    if (dest == NULL || dest_len < 1 || src == NULL || src_len < 1)
+    {
+        return dest;
```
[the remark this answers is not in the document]
aquesnel: done

common/string_calls.c
```diff
@@ -651,3 +651,42 @@ g_strtrim(char *str, int trim_flags)
...
+    }
+
+    dest[0] = '\0';
+    dest_end = dest + dest_len - 1;
+    joiner_len = g_strlen(joiner);
+    for (i = 0; i < src_len - 1 && dest_pos < dest_end; i++)
+    {
+        len = g_strlen(src[i]);
+        g_strncat(dest_pos, src[i], dest_end - dest_pos);
+        dest_pos += len;
```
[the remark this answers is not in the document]
aquesnel: Thanks, I had no idea about this limitation.

common/os_calls.c
```diff
@@ -2901,17 +2953,21 @@ g_waitpid_status(int pid)
...
             LOG(LOG_LEVEL_WARNING, "wait for pid %d returned unknown result", pid);
-            rv = -1;
+            exit_status.exit_code = -1;
         }
     }
 
-    return rv;
+    return exit_status;
 #endif
 }
```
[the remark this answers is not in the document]
aquesnel: fixed

common/os_calls.h
```diff
@@ -23,6 +23,12 @@
 
 #include "arch.h"
 
+struct exit_status
```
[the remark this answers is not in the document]
aquesnel: good idea, fixed.

scripts/install_xrdp_build_dependencies_with_apt.sh
```diff
@@ -1,20 +1,18 @@
...
-if [ -z "$FEATURE_SET" ]; then
-    FEATURE_SET=min
+if [ $# -ge 1 ]; then
+    FEATURE_SET="$1"
+    shift
 fi
-
-if [ -z "$ARCH" ]; then
-    ARCH=amd64
+if [ $# -ge 2 ]; then
```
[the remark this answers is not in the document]
aquesnel: you are correct, fixed now.

sesman/session.c
```diff
@@ -568,96 +573,84 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
-                        /* logging parameters */
-                        LOG(LOG_LEVEL_ERROR, "errno: %d, "
-                            "description: %s", g_get_errno(), g_get_strerror());
-                        LOG(LOG_LEVEL_ERROR, "execlp3 parameter "
-                            "list:");
-                        LOG(LOG_LEVEL_ERROR, "        argv[0] = %s",
+                    }
+                    else
+                    {
+                        LOG(LOG_LEVEL_DEBUG,
```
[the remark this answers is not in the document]
aquesnel: Here is my thought process:
* the sys admin controls the xrdp config file to enable/disable the `EnableUserWindowManager` setting
* the user controls the `startwm.sh` window manager file in their home directory, including if it exists or not

So I'd expect it to be a normal state for the `startwm.sh` file to be missing from the home directory.

sesman/session.c
```diff
@@ -816,98 +808,105 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
-                    "(pid %d) to exit", window_manager_pid);
+                LOG(LOG_LEVEL_INFO, "Session in progress on display %d, waiting "
+                    "until the window manager (pid %d) exits to end the session", 
+                    display, window_manager_pid);
                 wm_wait_time = g_time1();
-                wm_exit_code = g_waitpid_status(window_manager_pid);
+                wm_exit_status = g_waitpid_status(window_manager_pid);
                 wm_wait_time = g_time1() - wm_wait_time;
-                if (wm_exit_code != 0)
+                if (wm_exit_status.exit_code != 0)
```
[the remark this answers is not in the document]
aquesnel: fixed

sesman/session.c
```diff
@@ -816,98 +808,105 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
                 else
                 {
-                    LOG(LOG_LEVEL_INFO, "window manager (pid %d) was running "
-                        "for approximately %d seconds.",
-                        window_manager_pid, wm_wait_time);
+                    LOG(LOG_LEVEL_DEBUG, "Window manager (pid %d, display %d) "
+                        "was running for %d seconds.",
+                        window_manager_pid, display, wm_wait_time);
                 }
-                LOG(LOG_LEVEL_INFO, "Cleaning up session. Calling "
```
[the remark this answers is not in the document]
aquesnel: good point. fixed

sesman/session.c
```diff
@@ -482,12 +484,16 @@ session_start_fork(tbus data, tui8 type, struct SCP_CONNECTION *c,
...
+        LOG(LOG_LEVEL_ERROR,
+            "[session start] (display %d): Failed to fork for scp with "
+            "errno: %d, description: %s",
+            display, g_get_errno(), g_get_strerror());
     }
     else if (pid == 0)
     {
-        LOG(LOG_LEVEL_INFO, "calling auth_start_session from pid %d",
-            g_getpid());
+        LOG_DEVEL(LOG_LEVEL_DEBUG,
```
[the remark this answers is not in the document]
aquesnel: good point. fixed

sesman/session.c
aquesnel: I moved all of the error logging about failed `exec` calls to the `g_exec*` functions. See [os_calls.c](https://github.com/neutrinolabs/xrdp/pull/1843/commits/b530b9fa5b4d56b12d89fade9c3e2ed25d7d0e86#diff-b04475dad2718b002e94b443cb48a735bae669abe6ab10bdfc10f17f7adf1a20R2616-R2627)

aquesnel:
Hi @matt335672 ,

I've fixed up the code to address the feedback. There are a few of your comments that I responded to, so please let me know if you want me to make additional changes.

I've also rebased and squashed the changes onto devel.

sesman/session.c
matt335672: That's a much better idea.

sesman/session.c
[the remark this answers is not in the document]
matt335672: Agreed - good explanation.

common/string_calls.c
```diff
@@ -775,3 +776,50 @@ g_strtrim(char *str, int trim_flags)
...
+            dest_pos += MIN(joiner_len, dest_remaining);
+        }
+    }
+
+    if (i == src_len - 1 && dest_pos < dest_end)
+    {
+        g_strncat(dest_pos, src[i], dest_end - dest_pos);
+    }
+
+    dest[dest_len] = '\0';
```
matt335672, line 823: Off by one - Should be `dest[dest_len - 1] = '\0`, or `*dest_end = '\0';`

This test catches it:-

```c
START_TEST(test_strnjoin__doesnt_write_off_end)
{
    /* setup */

    const char *src[] = { "a","b","c"};
    char result[5+1+1]; /* a-b-c + term + guard value */

    /* test */

    result[6] = '\x7f';
    g_strnjoin(result, 5 + 1, "-", src, 3);

    /* verify */

    ck_assert_int_eq(result[6], '\x7f');
}
END_TEST
```
Alternatively remove the whole line. You're using `strncat()` so this terminator is guaranteed to be present anyway.

BTW - nice use of MIN().

matt335672 (changes requested):
Good job. Please check my reasoning on the comment in g_strnjoin(). Also you'll need to rebase for which I apologise.

common/string_calls.c
aquesnel: you are right on both points. I added the test case you wrote, and removed the unnecessary writing null after the end of the destination buffer.

I had added the null termination because I misread `strncpy` for `strncat` in the following article:

>An entirely different form of string truncation is one that can result from calls to strncpy. Unlike functions such as snprintf and strncat that always append a terminating NUL, when the source string passed to strncpy is longer than the size specified by the third argument, the function truncates the copy without appending a NUL to the end. 
From [Forming Non-NUL-Terminated Sequences](https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/)

aquesnel:
@matt335672 I think this pull request is ready to merge now. Please let me know if there is anything else that needs to be fixed up.

matt335672 (approved):
Thanks for the swift turnaround.

matt335672:
@aquesnel - I managed to introduce a conflict by merging #1858 before this one! It seemed a bit rotten of me to ask you to rebase yet again, so I've added a merge commit to fix things.

Thanks so much for adding the check stuff - I think that's going to be very useful in the future.

configure.ac
```diff
@@ -357,6 +357,11 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
 
 CFLAGS="$save_CFLAGS"
 
+# checking for libcheck (used for unit testing)
+PKG_CHECK_MODULES([CHECK], [check >= 0.10.0])
+AC_CHECK_HEADER([check.h], [],
+  [AC_MSG_ERROR([please install check])])
```
metalefty, line 363: I noticed that this makes `check` is always required to build xrdp even when users try to install xrdp to their system. I don't think unit testing is always required. Performing unit tests during CI is enough. What do you think?

matt335672: Yes, I think that's a great idea. I see you've raised #1860 - I'll take a look.

Issue #1846, opened 2021-03-30
Shared clipboard does not work if libvnc.so and Vino are used 


yamah-byv:
Hello.
The clipboard are work in normally, if I use XOrg session (libxup.so). But, when I use VPN Session (libvpn,so) and connect to Vino the shared clipboard are not work.  When I connected to Vino by any VNC-client, shared clipboard are work in normally,

my config xrdp.ini
```
[Globals]
ini_version=1

fork=true

port=3389

use_vsock=false

tcp_nodelay=true

tcp_keepalive=true

security_layer=negotiate

crypt_level=high

certificate=
key_file=

ssl_protocols=TLSv1.2, TLSv1.3

autorun=

allow_channels=true
allow_multimon=true
bitmap_cache=true
bitmap_compression=true
bulk_compression=true
max_bpp=32
new_cursors=true
use_fastpath=both
enable_token_login=true

blue=009cb5
grey=dedede

ls_top_window_bg_color=009cb5

ls_width=350
ls_height=430

ls_bg_color=dedede

ls_logo_filename=
ls_logo_x_pos=55
ls_logo_y_pos=50

ls_label_x_pos=30
ls_label_width=65

ls_input_x_pos=110
ls_input_width=210

ls_input_y_pos=220

ls_btn_ok_x_pos=142
ls_btn_ok_y_pos=370
ls_btn_ok_width=85
ls_btn_ok_height=30

ls_btn_cancel_x_pos=237
ls_btn_cancel_y_pos=370
ls_btn_cancel_width=85
ls_btn_cancel_height=30

[Logging]
LogFile=xrdp.log
LogLevel=INFO
EnableSyslog=true

[LoggingPerLogger]

[Channels]
rdpdr=true
rdpsnd=true
drdynvc=true
cliprdr=true
rail=true
xrdpvr=true
tcutils=true

[Xorg]  
name=New Session
lib=libxup.so
username=ask
password=ask
pamsessionmng=127.0.0.1
ip=127.0.0.1 
port=-1
code=20
cliprdr=true

[Xvnc]
name=User Session
lib=libvnc.so
pamusername=ask
pampassword=ask
pamsessionmng=127.0.0.1
ip=127.0.0.1
port=5900
code=20
cliprdr=true
rdpdr=true
delay=1200
```
xrdp - 0.9.15
The connected user include in the groups tsusers,fuse.

What am I doing wrong?

matt335672:
Hi @yamah-byv 

[...]

PR #1849, opened 2021-03-31
Allow DISPLAY=:0 for chansrv

matt335672:
While triaging #1846, I noticed that chansrv won't run if the DISPLAY is :0

This prevents the following use case being possible:-
- User starts console session
- User runs Vino/x11vnc/whatever to expose console session on port 5900 (VNC).
- User runs xrdp-chansrv in console session
- User uses `chansrvport=/path/to/xrdp_chansrv_socket_0`  in `xrdp.ini` to connect to chansrv while connecting to the existing VNC session on port 5900.

This doesn't sound like a bad use-case to me. Am I missing something?

This PR allows chansrv to connect to DISPLAY 0, and also cleans up the function `get_display_num_from_display()`.

Also, shall I add code to `xrdp_mm.c` to default the path to the socks dir, so the user can just specify `xrdp_chansrv_socket_0` rather than needing the socks dir path which is set at compile time?

Thanks.

### DIFF
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -17,8 +17,6 @@
  * limitations under the License.
  */
 
-#include <ctype.h>
-
 #if defined(HAVE_CONFIG_H)
 #include <config_ac.h>
 #endif
@@ -1560,34 +1558,6 @@ x_server_fatal_handler(void)
 }
 
 /*****************************************************************************/
-static int
-get_display_num_from_display(const char *display_text)
-{
-    int rv = -1;
-    const char *p;
-
-    /* Skip over the hostname part of the DISPLAY */
-    if ((p = g_strchr(display_text, ':')) != NULL)
-    {
-        ++p; /* Skip the ':' */
-
-        /* Cater for the (still supported) double-colon. See
-         * https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html */
-        if (*p == ':')
-        {
-            ++p;
-        }
-
-        if (isdigit(*p))
-        {
-            rv = g_atoi(p);
-        }
-    }
-
-    return rv;
-}
-
-/*****************************************************************************/
 int
 main_cleanup(void)
 {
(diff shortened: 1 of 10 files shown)


metalefty:
> This doesn't sound like a bad use-case to me. Am I missing something?

I think that is a valid use case.

sesman/chansrv/chansrv.c
```diff
@@ -1559,61 +1561,30 @@ x_server_fatal_handler(void)
...
+    int rv = -1;
+    const char *p;
+
+    /* Skip over the hostname part of the DISPLAY */
+    if ((p = g_strchr(display_text, ':')) != NULL)
+    {
+        ++p; /* Skip the ':' */
+
+        /* Cater for the (still supported) double-colon. See
+         * https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html */
```
metalefty, line 1575: Good care of the corner case :+1:

metalefty:
> Also, shall I add code to xrdp_mm.c to default the path to the socks dir, so the user can just specify xrdp_chansrv_socket_0 rather than needing the socks dir path which is set at compile time?

I think this is also a good idea.

matt335672:
I've also moved `get_display_num_from_display()` to `string_calls::g_get_display_num_from_display()` as I realised this was common code.

`chansrvport` can now be set to a string (as before), or to `DISPLAY(n)` to connect to `xrdp-chansrv` for `DISPLAY=:n`. Manpages updated.

For testing, I've used this stanza in `xrdp.ini`:-
```
[localvnc]
name=localvnc
lib=libvnc.so
ip=localhost
port=5900
#username=ask
password=
pamusername=mjb
pampassword=ask
pamsessionmng=127.0.0.1
chansrvport=DISPLAY(0)
```
I then ran `x11vnc -N -forever -nopw` and `xrdp-chansrv` from the command line in an active session. I could connect to the session, and `xrdp-chansrv` appeared to be working normally.

One irritation I've found is that if the user ends the session, `xrdp-chansrv` is killed quickly and doesn't get a chance to unmount the shared filesystem. It's recoverable - the user needs to call `fusermount -u` to fix this before starting another session.

I'd appreciate if people think this is good enough to push into devel in this state (once reviews are done). I'll add a Wiki page for it before v0.9.16.

metalefty:
Unmounting of FUSE file systems should be more robust but I think this is not a regression of this PR. I'm fine with shipping this to v0.9.16.

PR #1856, opened 2021-04-09
Enable FUSE in FreeBSD cirrus build

matt335672:
Enables FUSE for the Cirrus FreeBSD CI build.

Also partially addresses the aspects of Debian bug [#983843](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=983843) involving their CI build on the kfreebsd branches.

I haven't tested this code on FreeBSD yet given where it is on the [platform-support-tier](https://github.com/neutrinolabs/xrdp/wiki/Platform-Support-Tier). Despite that, is this worth adding in to the CI system, do people think?


metalefty:
I'm fine with enabling FUSE in CI build. However, I confirmed FUSE is not working on FreeBSD. I still have some more things to work on.  So please keep in mind this doesn't directly mean FUSE is available on FreeBSD.

metalefty (approved):
I approve this PR itself.

matt335672:
I suspected it might need a bit of work to fix it - thanks for the confirmation.

matt335672:
@metalefty - could you try fuse on FreeBSD again, but run this script first:-

```sh
#!/bin/sh

status=0

# Checvk the kernel module is loaded
if [ ! -c /dev/fuse ]; then
    kldload fusefs || {
	status=$?
        echo "** Unable to load kernel module 'fusefs'" >&2
    }
fi

# Users allowed to mount file systems?
if [ $status -eq 0 -a \
     $(sysctl -n vfs.usermount) -eq 0 ]; then
    sysctl vfs.usermount=1 >/dev/null || {
	status=$?
        echo "** Unable to load set vfs.usermount=1" >&2
    }
fi

exit $status
```

At least on FreeBSD 12.2, it seems to be working quite well.

I'm not able to run the xrdp configure script on it for some reason, but other than that it looks reasonably useful.

PR #1860, opened 2021-04-15
Relax dependency checking on check and miscellaneous fixes for pkg-config usage

metalefty:
Fix for pkg-config usage:
* rely on pkg-config when finding check 
* add missing CFLAGS when using check obtained via pkg-config

Adjust unit tests:
* relax dependency checking on check (check is not necessary if unit tests are not run)
* `make distcheck` should only succeed if check is installed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,8 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_DISTCHECK_CONFIGURE_FLAGS = \
   --without-systemdsystemunitdir \
-  --enable-strict-locations
+  --enable-strict-locations \
+  --enable-tests
 
 EXTRA_DIST = \
   COPYING \
--- a/Makefile.am
+++ b/Makefile.am
@@ -67,5 +67,9 @@ SUBDIRS = \
   tests \
   tools
 
+if PERFORM_UNIT_TESTS
+SUBDIRS += tests
+endif
+
 distclean-local:
 	-rm -f xrdp_configure_options.h
--- a/Makefile.am
+++ b/Makefile.am
@@ -68,9 +68,5 @@ SUBDIRS = \
   tests \
   tools
 
-if PERFORM_UNIT_TESTS
-SUBDIRS += tests
-endif
-
 distclean-local:
 	-rm -f xrdp_configure_options.h
--- a/configure.ac
+++ b/configure.ac
@@ -358,8 +358,7 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
 CFLAGS="$save_CFLAGS"
 
 # checking for libcheck (used for unit testing)
-PKG_CHECK_MODULES([CHECK], [check >= 0.10.0])
-AC_CHECK_HEADER([check.h], [],
+PKG_CHECK_MODULES([CHECK], [check >= 0.10.0], [],
   [AC_MSG_ERROR([please install check])])
 
 AC_SUBST([moduledir], '${libdir}/xrdp')
--- a/configure.ac
+++ b/configure.ac
@@ -377,9 +377,6 @@ if test "x$perform_unit_tests" == "xyes"; then
 else
     AC_MSG_NOTICE([libcheck not found, unit tests will be skipped])
 fi
-
-AM_CONDITIONAL(PERFORM_UNIT_TESTS,
-               [test "x$perform_unit_tests" = "xyes"])
 # -- end perform unit tests
 
 AC_SUBST([moduledir], '${libdir}/xrdp')
--- a/configure.ac
+++ b/configure.ac
@@ -357,9 +357,20 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
 
 CFLAGS="$save_CFLAGS"
 
-# checking for libcheck (used for unit testing)
-PKG_CHECK_MODULES([CHECK], [check >= 0.10.0], [],
-  [AC_MSG_ERROR([please install check])])
+# perform unit tests if libcheck found
+PKG_CHECK_MODULES([CHECK], [check >= 0.10.0],
+    [perform_unit_tests=yes],
+    [perform_unit_tests=no])
+
+if test "x$perform_unit_tests" == "xyes"; then
+    AC_MSG_NOTICE([libcheck found, unit tests will be performed])
+else
+    AC_MSG_NOTICE([libcheck not found, unit tests will be skipped])
+fi
+
+AM_CONDITIONAL(PERFORM_UNIT_TESTS,
+               [test "x$perform_unit_tests" = "xyes"])
+# -- end perform unit tests
 
 AC_SUBST([moduledir], '${libdir}/xrdp')
 
@@ -442,6 +453,8 @@ echo "  exec_prefix             $exec_prefix"
 echo "  libdir                  $libdir"
 echo "  bindir                  $bindir"
 echo "  sysconfdir              $sysconfdir"
+echo ""
+echo "  perform unit tests      $perform_unit_tests"
 echo ""
 echo "  CFLAGS = $CFLAGS"
 echo "  LDFLAGS = $LDFLAGS"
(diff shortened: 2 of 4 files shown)


matt335672:
I think this is generally a good idea, just in case some platform doesn't have `check` available.

The CI problem seems to be that `make distcheck` recurses into `DIST_SUBDIRS` and not just `SUBDIRS`. Because it runs configure with default args, and because this doesn't initialise `CHECK_LIBS`, the link fails.

It's a tricky one. Defaulting `--enable-tests` to true should fix it, as would skipping the recursion in `tests/Makefile.am`. I'd personally prefer the first, as `make distcheck` should always run the checks. What do you think?

matt335672:
There's some more useful info [here](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html) [gnu.org] which give us a few more options.

aquesnel:
I also think that running the unit tests should be the default. To make it easier on people trying to install xrdp without libcheck I would suggest adding to the error message for a failure of `PKG_CHECK_MODULES([Check])` to say that the dependency on libcheck can be removed by adding `--disable-unittests` to configure.

metalefty:
@aquesnel That sounds good.

aquesnel:
To check that installing xrdp works without the check library installed, I suggest that  distcheck should be enabled for the min builds with the unit tests disabled and the check library not installed. I'm not sure why the decision was made to have the CI min builds not run `make distcheck` .

metalefty:
This also will work. My position is default off. I think unit tests are developers' stuff. It would be enough for developers to turn on unit tests intentionally.  That will not put users in unnecessary trouble.  However, I'm not in a strong position on this. To guide users to add `--disable-config` in configure error message is also a preferable alternative.

```diff
diff --git a/Makefile.am b/Makefile.am
index 242615d1..1d58faa9 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,8 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_DISTCHECK_CONFIGURE_FLAGS = \
   --without-systemdsystemunitdir \
-  --enable-strict-locations
+  --enable-strict-locations \
+  --enable-tests

 EXTRA_DIST = \
   COPYING \
diff --git a/libpainter b/libpainter
```

>  I'm not sure why the decision was made to have the CI min builds not run make distcheck .

I'm fine with turning on make disstcheck on min builds.

matt335672:
The GNU guidelines for [make distcheck](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html) suggest that a `make check` should be included. I don't think we should deviate from this unless we have a good reason.

Another solution is to make `--enable-tests` be `on` `off` or `auto`, with the default being `auto`. If set to `auto` we enable tests only if we find that `check` is installed. That would work fine for CI builds and `make distcheck`. I've got no idea how easy this would be to set up in the auto tools however.

aquesnel:
@metalefty my thought process for defaulting the tests to enabled is the following:

The tests are not just a developer facing tool to ensure that new commits don't break functionality, the tests are also a package maintainer tool in the sense of "user acceptance tests". For package maintainers/publishers they want to know that the package they are building will work for their users. We should endeavor to simplify the job of the package maintainer by reducing the special knowledge needed to meet their goal of delivering a working package to their users. In this case the special knowledge is the custom `--enable-tests` configure option. Therefore I think that the tests should be enabled by default. In the case where a user is compiling xrdp for themselves, I think they should be treated as a package maintainer with an audience of one user.

Now that I've written out my argument, I realized that I'm more concerned about something else than the default state of compiling and running the tests. I'm concerned that running `make check` will succeed when the tests are not compiled. The reason I'm concerned is when `make check` succeeds then the user should be confident that as best as we know, xrdp will work as expected. So `make check` not running the tests is a silent failure and there is no feedback to the user that the tests did not run, and that they should take some action to get the tests to run (eg. configure with tests enabled). I believe that we should think about the `--enable-tests` flag as a way to enable compilation without the libcheck dependency, but not that it will cause `make check` to succeed.

@matt335672 l like the idea of an `auto` value for the tests flag, and I think that with this auto value the issue of the tests silently not running becomes a bigger issue because of the lack of feedback to the user.

metalefty:
@aquesnel Thanks for the follow-up. I'll go on with "default on".

matt335672:
I absolutely agree that `make check` should only succeed if tests are run, and if they all complete successfully.

If we did go down the `auto` route, this should still be the case. So in this instance `make distcheck` will only succeed if `check` is installed.

metalefty:
@matt335672 @aquesnel
Thank you both. I've updated it to `auto`. And `make distcheck` are always to be run with unit tests.

However,  I'm not sure whether `make check` only succeeds if tests are run, and if they all complete successfully.

Makefile.am
```diff
@@ -67,5 +68,9 @@ SUBDIRS = \
   tests \
   tools
 
+if PERFORM_UNIT_TESTS
+SUBDIRS += tests
```
metalefty, line 72: I think this is much simpler so also other parts can be written like this.

matt335672: Hi @metalefty - I've just had a test of this, and it's working but I think the reasons it's working are quite subtle and not entirely what we expected.

SUBDIRS still contains the `tests` directory anyway (see line 68 above your addition), so lines 71 and 72 are just adding tests to SUBDIRS for a second time (i.e. effectively a no-operation).

If line 68 is removed, SUBDIRS will only contain `tests` if we're running unit tests which is the intention of lines 71 and 72. However this has the problem that `make check` succeeds if `check` isn't installed which isn't what we want. However, if line 68 isn't removed, `make check` fails if `check` isn't installed.

So I took out PERFORM_UNIT_TESTS entirely and I think that's what we want.

The reason it's OK is a subtlety in `tests/common/Makefile.am` and `test/memtest/Makefile.am` which I hadn't realised. These makefiles use `check_PROGRAMS` to define the targets, and these are only built if running `make check` - they are not built for `make all`. So we don't need to cater for the unit tests being built or not built as automake already does it for us.

There's a further complication(!) in that `tests/memtest/Makefile.am` contains a `noinst_PROGRAMS` which is built by default. However, `memtest` doesn't use the `check` package so that's OK. I suggest this is removed so that `memtest` is only built for `make check`.

I hope that makes sense - I know I've learned a lot looking into this.

Here are the changes I made which I think give us what we want:-

```diff
diff --git a/Makefile.am b/Makefile.am
index 13795164..0f7b5d3e 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -68,9 +68,5 @@ SUBDIRS = \
   tests \
   tools
 
-if PERFORM_UNIT_TESTS
-SUBDIRS += tests
-endif
-
 distclean-local:
        -rm -f xrdp_configure_options.h
diff --git a/configure.ac b/configure.ac
index 00de8f5c..b513536a 100644
--- a/configure.ac
+++ b/configure.ac
@@ -378,8 +378,6 @@ else
     AC_MSG_NOTICE([libcheck not found, unit tests will be skipped])
 fi
 
-AM_CONDITIONAL(PERFORM_UNIT_TESTS,
-               [test "x$perform_unit_tests" = "xyes"])
 # -- end perform unit tests
 
 AC_SUBST([moduledir], '${libdir}/xrdp')
diff --git a/tests/memtest/Makefile.am b/tests/memtest/Makefile.am
index 8901722f..890f2716 100644
--- a/tests/memtest/Makefile.am
+++ b/tests/memtest/Makefile.am
@@ -6,9 +6,6 @@ if XRDP_DEBUG
 AM_CPPFLAGS += -DXRDP_DEBUG
 endif
 
-noinst_PROGRAMS = \
-  memtest
-
 check_PROGRAMS = \
   memtest
 
```

metalefty: Thanks for the investigation. I confirmed the change causes `make check` failure if the `check` library is not installed and the ordinary `make install` succeeds without building unit tests.  

Then, is the `--enable-tests`  flag that enforces unit tests still necessary?

BTW, `--enable-tests`, "enforce unit tests", "skip unit tests" sound a little bit odd now. The first two enable strict dependency checking on `check` but not enforcing unit tests. The last one doesn't skip unit tests actually.  The flag does not change the behaviour during the build at all.

matt335672: I think the `--enable-tests` flags is still useful if only for `make distcheck`. In this case, if `check` isn't installed the configure fails with an informative message rather than just failing when it gets to `make check`.

So the meaning of the flag is more along the lines of 'ensure dependencies for the tests are installed' (I think).

metalefty:
@matt335672 Updated the description of `--enable-tests`.  Anything else for this PR before the release?

PR #1872, opened 2021-04-29
neutrinordp: avoid pampassword leakage

metalefty:
The leakage does not occur in the most usual use case of xrdp.
It occurs in NeutrinoRDP proxy mode with PAM authentication enabled.

Reported by @TOMATO-ONE

---

Note: It shouldn't be a problem in the most use case of xrdp and it is not a vulnerability such as remote code execution. So I've raised the fix as a usual public PR.

### DIFF
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -443,7 +443,14 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
 {
     rdpSettings *settings;
 
-    LOG_DEVEL(LOG_LEVEL_DEBUG, "lxrdp_set_param: name [%s] value [%s]", name, value);
+    if (g_strcmp(name, "password") == 0 || g_strcmp(name, "pampassword") == 0)
+    {
+        LOG_DEVEL(LOG_LEVEL_DEBUG, "lxrdp_set_param: name [%s] value [******]", name);
+    }
+    else
+    {
+        LOG_DEVEL(LOG_LEVEL_DEBUG, "lxrdp_set_param: name [%s] value [%s]", name, value);
+    }
     settings = mod->inst->settings;
 
     if (g_strcmp(name, "hostname") == 0)
@@ -495,6 +502,10 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     else if (g_strcmp(name, "enable_dynamic_resizing") == 0)
     {
         settings->desktop_resize = g_text2bool(value);
+    }
+    else if (g_strcmp(name, "pampassword") == 0)
+    {
+        LOG(LOG_LEVEL_WARNING, "lxrdp_set_param: unknown name [%s] value [******]", name);
     }
     else
     {
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -451,6 +451,7 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     {
         LOG_DEVEL(LOG_LEVEL_DEBUG, "lxrdp_set_param: name [%s] value [%s]", name, value);
     }
+
     settings = mod->inst->settings;
 
     if (g_strcmp(name, "hostname") == 0)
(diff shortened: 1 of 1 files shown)


metalefty:
CC: @aquesnel

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -496,6 +503,10 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     {
         settings->desktop_resize = g_text2bool(value);
     }
+    else if (g_strcmp(name, "pampassword") == 0)
+    {
+        LOG(LOG_LEVEL_WARNING, "lxrdp_set_param: unknown name [%s] value [******]", name);
```
matt335672, line 508: Do we need to log this? It's a valid parameter in some contexts. I'd got for logging nothing at all for `pamusername`, `pampassword` and `pamsessionmng`

matt335672 (approved):
Looks fine. I've got a query on the logging, but I'm happy with whatever you decide on that.

neutrinordp/xrdp-neutrinordp.c
metalefty: Indeed. What about doing nothing?

```
    else if (g_strcmp(name, "pampassword") == 0)
    {
    }
```

metalefty: Also shouldn't `pampassword` and `pamsessionmng` reach `lxrdp_set_param: unknown name` at the bottom?

matt335672: Well  `pamusername`, `pampassword` and `pamsessionmng` are all valid settings for this module even if they're not used by it, so I'd vote for not logging them at all:-

```c
    else if (g_strcmp(name, "pamusername") == 0 ||
             g_strcmp(name, "pampassword") == 0 ||
             g_strcmp(name, "pamsessionmng") == 0)
    {
        /* Valid (but unused) parameters not logged */
    }
    else
    {
        LOG(LOG_LEVEL_WARNING, "lxrdp_set_param: unknown name [%s] value [%s]", name, value);
    }
   . . .
```

The counter-argument is that keeping the above list maintained is not going to happen, so maybe we should just stick with omitting the security-sensitive values.

metalefty: This is the suggested change by the reporter @TOMATO-ONE

```diff
--- xrdp/neutrinordp/xrdp-neutrinordp.c.orig
+++ xrdp/neutrinordp/xrdp-neutrinordp.c
@@ -502,6 +502,10 @@
     {
         settings->nla_security = g_text2bool(value);
     }
+    else if (g_strcmp(name, "pampassword") == 0)
+    {
+       LLOGLN(0, ("lxrdp_set_param: unknown name [%s] value [%s]", name, "*****" ));
+    }
     else
     {
         LLOGLN(0, ("lxrdp_set_param: unknown name [%s] value [%s]", name, value));
```

And this is what I'm thinking of:
```diff
+    else if (g_strcmp(name, "pampassword") == 0 ||
+             g_strcmp(name, "pamusername") == 0 ||
+             g_strcmp(name, "pammsessionmng") == 0)
+    {
+        /*
+         * Doing nothing on purpose
+         *
+         * `pampassword will contain raw password
+         * `pamusername` and `pamsessionmng` are also valid parameters in some contexts
+         *  so not to log "unknown name"
+         *
+         */
+    }
```

aquesnel: I agree with both of  @matt335672 points. I also prefer your change @metalefty rather than the reporter's change.

aquesnel (approved):

aquesnel:
Fyi: as I was making the longing changes I tried to omit explicit security sensitive data from the logs, but when logging unknown data I decided that logging with `LOG_DEVEL` and `TRACE` it was ok to log sensitive data since these settings should not be used in a production setting.

If you disagree with decision then maybe we should open an issue to revive logging for all unknown fields for potentially sensitive data.

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you.
My patch was tentative.
I also think your patch is better.

metalefty:
Thanks for the careful work on logging change.

>  I decided that logging with LOG_DEVEL and TRACE it was ok to log sensitive data since these settings should not be used in a production setting.

I'm happy with this method. It would be much better if we add a note that LOG_DEVEL and TRACE level might dump sensitive datum and should not be used in a production setting.

metalefty:
Merged, thanks @TOMATO-ONE ! 

The fix will be included in the next release 0.9.16, will be released soon. Also Thank you for the correct method of reporting security issues.

PR #1879, opened 2021-05-08
Add checking the code formatting with astyle during CI builds

aquesnel:
Add checking the code formatting with astyle during CI builds

Example build with the code formatting failed: https://github.com/aquesnel/xrdp/runs/2535294459
Example build with the code formatting succeeded: https://github.com/neutrinolabs/xrdp/pull/1879/checks?check_run_id=2535309754

Note: the i386 builds are likely failing because of a change to the github actions runner images: https://github.com/actions/virtual-environments/issues/3339 That issue says that it should be resolved next week but the github actions team.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,3 +146,32 @@ jobs:
       - run: ./bootstrap
       - run: scripts/install_cppcheck.sh $CPPCHECK_REPO $CPPCHECK_VER
       - run: scripts/run_cppcheck.sh -v $CPPCHECK_VER
+
+  code_formatting_check:
+    name: code formatting check
+    runs-on: ubuntu-latest
+    env:
+      CC: gcc
+      # This is required to use a version of astyle other than that
+      # supplied with the operating system
+      ASTYLE_VER: 3.1
+      ASTYLE_REPO: https://svn.code.sf.net/p/astyle/code/tags
+    steps:
+      # This is currently the only way to get a version into
+      # the cache tag name - see https://github.com/actions/cache/issues/543
+      - run: |
+          echo "OS_VERSION=`lsb_release -sr`" >> $GITHUB_ENV
+      - uses: actions/checkout@v2
+      - name: Cache astyle
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-astyle
+        with:
+          path: ~/astyle.local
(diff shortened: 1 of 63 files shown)


matt335672:
Well this looks pretty reasonable to me and all looks fine. The version of astyle is controlled, so we won't be blind-sided by changes in the way astyle functions. Also, just after a release (i.e. now) is a really good time to merge this.

The one fairly big concern I'd have is this recent announcement, which I've just found:-

https://sourceforge.net/p/astyle/bugs/548/

There are alternatives to `astyle`, with the obvious being `clang-format`. That's not however a self-contained option in the same way. I haven't tried it yet myself, but there appears to be a bit of impetus behind it - it's used by the [linux kernel](https://www.kernel.org/doc/html/latest/process/clang-format.html) for example.

Given the existence of clang-format, I think it unlikely that anyone else will pick up the maintenance of astyle.

Personally I'm happy to merge this as-is despite the above, as it give us definite benefits despite the odd formatting oddity. If for some reason it stops working we're no worse off anyway. I think this needs a bit of discussion, and then @metalefty will need to make a decision.

Nexarian:
I'm going to wait until this is merged to merge my change here: https://github.com/neutrinolabs/xrdp/pull/1895

I'd prefer that we use clang-format if astyle is being deprecated. We really don't want to bet the farm on things that aren't supported.

However, regardless, I have a recommendation on how to make sweeping changes like this in the future. Check in the linter and make it work for one file, and then blacklist all other files. Then slowly, one by one, submit smaller PRs that un-black-list each file.

.github/workflows/build.yml
```diff
@@ -146,3 +146,32 @@ jobs:
       - run: ./bootstrap
       - run: scripts/install_cppcheck.sh $CPPCHECK_REPO $CPPCHECK_VER
       - run: scripts/run_cppcheck.sh -v $CPPCHECK_VER
+
+  code_formatting_check:
+    name: code formatting check
+    runs-on: ubuntu-latest
```
metalefty, line 152: This might be unstable because the "latest" version will change as time flows but not a big problem.

metalefty (approved):
It is a pretty good idea and LGTM. Let's merge it. We can switch to `clang-format` later.

PR #1883, opened 2021-05-11
PAM module minor improvements

### LINKED ISSUE #1882: PAM_USER enviroment variable is not set when using pam_exec module
## Description

Trying to integrate xrdp with a custom auth method using a script launched by the` pam_exec.so` module, I've found that `PAM_USER` enviroment variable is not set when launching a executable with `pam_exec`.

`PAM_RHOST` and `PAM_RUSER` enviroment variables of `pam_exec.so` module are not set neither.

The variables set in the script enviroment are:

```
*** Tue May 11 08:29:27 2021
Enviroment variables:
PAM_TTY=xrdp-sesman
PWD=/
PAM_SERVICE=xrdp-sesman
SHLVL=1
PAM_TYPE=auth
_=/usr/bin/env
```

## Solution for PAM_USER env var

Digging a bit in the source code, I've found that the user is [not being sent to the pam_start function](https://github.com/neutrinolabs/xrdp/blob/devel/sesman/verify_user_pam.c#L129) as the documentation of pam define, a '0' is sent instead.

You can fix this issue changing the line 129 of `sesman/verify_user_pam.c` file:
```
--- error = pam_start(service_name, 0, &(auth_info->pamc), &(auth_info->ph));
+++ error = pam_start(service_name, user, &(auth_info->pamc), &(auth_info->ph));
```

#### 2. Configure PAM
2.1 Add the following line into the file `/etc/pam.d/common-auth`
```
auth sufficient pam_exec.so log=/var/log/pam.log /usr/local/bin/print-env.sh
```

2.2 Create script `/usr/local/bin/print-env.sh` with the following content:
```
#!/bin/bash
#

echo "Enviroment variables:"
env
```

  2.3 Give executable permissions to the script
```
chmod +x /usr/local/bin/print-env.sh
```

## Resources

- http://www.linux-pam.org/Linux-PAM-html/sag-pam_exec.html

matt335672:
This PR makes minor improvements to the PAM processing.

# General
- `g_printf()` calls for error conditions are replaced with `LOG()` calls.
- `PAM_USER` is now set for (e.g.) `pam_exec.so` - this addresses part of #1882 

# Conversion function
- More error logging/checking is added to the PAM conversion function
- The `PAM_ERROR_MSG` style (see [pam_conv(3)](https://linux.die.net/man/3/pam_conv) is now supported
- The `PAM_TEXT_INFO` style now logs the message rather than ignoring it completely.
- some potential memory leaks if the conv function fails have been removed.

The PAM_USER functionality was tested by adding this line to `/etc/pam.d/xrdp-sesman`:-

```
auth optional pam_exec.so log=/tmp/pam.log /usr/bin/env
```

After logging in, `/tmp/pam.log` contained:-
```
PAM_SERVICE=xrdp-sesman
PAM_USER=testuser
PAM_TTY=xrdp-sesman
PAM_TYPE=auth
```

Although the `PAM_ERROR_MSG` style could not easily be tested directly without writing a PAM module, the `pam_echo.so` was used to exercise the very similar `PAM_TEXT_INFO` style which now outputs to the sesman log.

### DIFF
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -93,7 +93,30 @@ msg_style_to_str(int msg_style, char *buff, unsigned int bufflen)
     return result;
 }
 
-/******************************************************************************/
+/***************************************************************************//**
+ * Provides the PAM conversation callback function
+ *
+ * At present, the main purpose of this function is to supply the
+ * user's password to the PAM stack, although some module logging is
+ * implemented here.
+ *
+ * @param[in] num_msg Count of messages in the msg array
+ * @param[in] msg Messages from the PAM stack to the application
+ * @param[out] resp Message replies from the application to the PAM stack
+ * @param[in] appdata_ptr Used to pass in a struct t_user_pass pointer
+ *
+ * @result PAM_SUCCESS if the messages were all processed successfully.
+ *
+ * @post If PAM_SUCCESS is returned, resp and its contents are allocated here
+ *       and must be freed by the caller
+ * @post If PAM_SUCCESS is not returned, resp is not allocated and must not
+ *       be not freed by the caller
+ *
+ * @note See pam_conv(3) for more information
+ * @note A basic example conversation function can be found in OSF RFC
+         86.0 (1995)
+ */
+
 static int
 verify_pam_conv(int num_msg, const struct pam_message **msg,
                 struct pam_response **resp, void *appdata_ptr)
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -86,7 +86,7 @@ msg_style_to_str(int msg_style, char *buff, unsigned int bufflen)
             break;
 
         default:
-            snprintf(buff, bufflen, "UNKNOWN_0x%x", msg_style);
+            g_snprintf(buff, bufflen, "UNKNOWN_0x%x", msg_style);
             result = buff;
     }
 
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -30,6 +30,7 @@
 
 #include "arch.h"
 #include "os_calls.h"
+#include "log.h"
 #include "string_calls.h"
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


sesman/verify_user_pam.c
```diff
@@ -53,43 +54,117 @@ struct t_auth_info
...
+            {
+                case PAM_PROMPT_ECHO_OFF: /* password */
+                    user_pass = (struct t_user_pass *) appdata_ptr;
+                    reply[i].resp = g_strdup(user_pass->pass);
+                    break;
 
-    for (i = 0; i < num_msg; i++)
+                case PAM_ERROR_MSG:
+                    LOG(LOG_LEVEL_ERROR, "PAM: %s", msg[i]->msg);
+                    break;
```
aquesnel, line 156: It's not clear to me why this error from pam only gets logged and does not propagate to sesman. If this is a recoverable error, I suggest that the log level be decreased to warning instead of error.

aquesnel (commented):
Looks good to me.

sesman/verify_user_pam.c
matt335672: That's a really good point which has led me to question a few assumptions I've made here.

This code is reached via callbacks from [pam_conv(3)](https://linux.die.net/man/3/pam_conv). Here's an extract:-

```
       Each message can have one of four types, specified by the msg_style member of struct pam_message:

       PAM_PROMPT_ECHO_OFF
           Obtain a string without echoing any text.

       PAM_PROMPT_ECHO_ON
           Obtain a string whilst echoing text.

       PAM_ERROR_MSG
           Display an error message.

       PAM_TEXT_INFO
           Display some text.
```
I've taken the simplistic approach of using LOG_LEVEL_ERROR for PAM_ERROR_MSG and LOG_LEVEL_INFO for PAM_TEXT_INFO.

My understanding is that the PAM library itself will return an error to sesman, so there's no need to do it here. Your question however prompted me to check that assumption so I had a look at the source for linux-pam.

There's a generic callback implementation here:-

https://github.com/linux-pam/linux-pam/blob/d62120ed7cfa99dec0e66e562f20fb32987c5980/libpam_misc/misc_conv.c#L276-L401

That sends PAM_ERROR_MSG to stderr, and returns PAM_SUCCESS in the absence of any other errors which is what I'm doing above.

The other calls which mention PAM_ERROR_MSG are all returning errors to the application with the exception of one in pam_selinux which allows for a retry before the application is notified.

TLDR; it's pretty certain that logging an error here will be followed by an error being returned to the application.

If we get an error back from PAM sesman will log it with LOG_LEVEL_ERROR at line 234 or 254 in my modified file.

So after some time looking into this, I'm reasonably sure that the behaviour I've implemented is correct but please come back to me if you disagree.

aquesnel: I don't know anything about pam, and I've only quickly read some parts of http://www.linux-pam.org/Linux-PAM-html/Linux-PAM_ADG.html

My take away from that guide is that you are correct, and the error will be returned from one of the other pam_* functions.

From my reading it seems that the return value of pam_conv callback represents only the answer to "was the text displayed to the user, or was the input received from the user?" and the error status of the authentication is tracked separately.

Can you please add to the function doc what the return value means, and any relevant references you used in your research.

matt335672: Will do - thanks.

aquesnel (approved):

matt335672:
Was about to merge this and a final eyeball check showed a call to `snprintf()` which was better made `g_snprintf()`

Will squash and merge tomorrow now unless I get any objections.

metalefty:
Overall, LGTM.

PR #1895, opened 2021-05-25
Unify monitor processing logic.

Nexarian:
There are two places where monitor descriptions are passed through the
RDP protocol:

- TS_UD_CS_MONITOR
- DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT

The processing logic for both of them is similar enough that they should
be unified.

Note that this is also the first step to making resizing work with the
Extension GFX channel.


libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    int monitorCount;
-    int flags;
-    int x1;
-    int y1;
-    int x2;
-    int y2;
-    int got_primary;
-    struct xrdp_client_info *client_info = (struct xrdp_client_info *)NULL;
+struct display_size_description*
+process_monitor_stream(struct stream *s, int full_parameters) {
```
matt335672: [Coding standard](https://github.com/neutrinolabs/xrdp/wiki/Coding-Style) : Opening brace on separate line. Also, recommend using a name like xrdp_sec_process_monitor_stream so it's easy to find the function.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+    struct display_size_description *description = (struct display_size_description *)
+        g_malloc(sizeof(struct display_size_description), 1);
 
-    client_info = &(self->rdp_layer->client_info);
+    int NumMonitor;
```
matt335672: Coding standard : Use '_' in preference to camel-case for new variables

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    if ((x2 > x1) && (y2 > y1))
+    if ((rect.right > rect.left) && (rect.bottom > rect.top))
     {
-        client_info->width = (x2 - x1) + 1;
-        client_info->height = (y2 - y1) + 1;
+        description->session_width = rect.right - rect.left;
+        description->session_height = rect.bottom - rect.top;
     }
     /* make sure virtual desktop size is ok */
-    if (client_info->width > 0x7FFE || client_info->width < 0xC8 ||
```
matt335672, line 2462: I appreciate that these values aren't given a name in MS-RDBBCGR, but is it worth adding defines for them for the scope of this file to improve readability and add a reference (e.g.):-

```c
/* Desktop limits [MS-RDPBCGR] 2.2.1.3.6 */
#define MIN_VIRTUAL_DESKTOP_WIDTH 200
(etc)
```

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+    {
+        LOG(LOG_LEVEL_ERROR,
+            "[MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR flags MUST be zero, "
+            "received: 0x%8.8x", flags);
+        return 1;
+    }
+
+    struct display_size_description *description = process_monitor_stream(s, 0);
+
+    client_info->monitorCount = description->monitorCount;
```
matt335672: Needs test for NULL return here if an invalid PDU is received.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+        description->session_height = rect.bottom - rect.top;
     }
     /* make sure virtual desktop size is ok */
-    if (client_info->width > 0x7FFE || client_info->width < 0xC8 ||
-            client_info->height > 0x7FFE || client_info->height < 0xC8)
+    if (description->session_width > 0x7FFE || description->session_width < 0xC8 ||
+            description->session_height > 0x7FFE || description->session_height < 0xC8)
     {
-        LOG(LOG_LEVEL_ERROR,
+        LOG(LOG_LEVEL_INFO,
```
matt335672: I don't understand the severity change of this message. An error seems appropriate here. Am I missing something?

xrdp/xrdp_mm.c
```diff
@@ -1042,27 +1042,110 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
     return 0;
 }
 
+/******************************************************************************/
+static int
+process_dynamic_monitor_description(struct xrdp_wm *wm, struct display_size_description *description) {
```
matt335672: Coding standard again, also a few lines below.

xrdp/xrdp_mm.c
```diff
@@ -1078,69 +1161,21 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
     }
-    return 0;
+    in_uint32_le(s, MonitorLayoutSize);
+    if (MonitorLayoutSize != 40)
+    {
+        /* 2.2.2.2 DISPLAYCONTROL_MONITOR_LAYOUT_PDU */
+        LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: MonitorLayoutSize is %d. Per spec it must be 40.", MonitorLayoutSize);
+        return 1;
+    }
+    struct display_size_description *description = process_monitor_stream(s, 1);
```
matt335672: This can return NULL, which will break `process_dynamic_monitor_description()`. Suggest adding a guard here, or in that function.

matt335672:
Hi @Nexarian 

Thanks for this.

I've added a few review comments for you above. If there's anything you disagree with please come back at me.

As regards the coding standard, and particularly given @aquesnel's #1879, can I suggest you run the files you've modified through astyle? There's some instructions on the wiki. 

Personally, all I've done in my dev environment is define `ARTISTIC_STYLE_PROJECT_OPTIONS=astyle_config.as` and then I run astyle on individual files. There's a bit of a trick to it in that if you don't wildcard an argument it grumbles like this:-

```sh
$ astyle os_calls.h
Recursive option with no wildcard
Did you intend quote the filename
Artistic Style has terminated
```

Fix by wildcarding your arg:-
```sh
$ astyle os_calls.h\*
------------------------------------------------------------
Directory  /home/mjb/xrdp/common/os_calls.h*
------------------------------------------------------------
Formatted  os_calls.h
```

Hope that's useful.

Nexarian:
Lets wait until https://github.com/neutrinolabs/xrdp/pull/1879 to merge this then so that my change has no risk of breaking the style change.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    got_primary = 0;
-    /* Add client_monitor_data to client_info struct, will later pass to X11rdp */
-    for (index = 0; index < monitorCount; index++)
+    for (int monitor_index = 0; monitor_index < NumMonitor; ++monitor_index)
     {
-        if (!s_check_rem_and_log(s, 20, "Parsing [MS-RDPBCGR] TS_UD_CS_MONITOR.TS_MONITOR_DEF"))
+        if (!s_check_rem_and_log(s, full_parameters == 0 ? 20 : 40, "Parsing monitor definitions."))
         {
-            return 1;
+            goto exit_error;
```
aquesnel: I'm very much against using `goto` in new code, since it is prone to misuse. Even though it's usage here seems reasonable, I'd still prefer replacing it with a structured programing construct.

One option for removing the goto's is to allocate the description struct in the `xrdp_sec_process_mcs_data_monitors` function, since you already unconditionally free the description struct there currently.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    y1 = 0;
-    x2 = 0;
-    y2 = 0;
-    got_primary = 0;
-    /* Add client_monitor_data to client_info struct, will later pass to X11rdp */
-    for (index = 0; index < monitorCount; index++)
+    for (int monitor_index = 0; monitor_index < NumMonitor; ++monitor_index)
     {
-        if (!s_check_rem_and_log(s, 20, "Parsing [MS-RDPBCGR] TS_UD_CS_MONITOR.TS_MONITOR_DEF"))
+        if (!s_check_rem_and_log(s, full_parameters == 0 ? 20 : 40, "Parsing monitor definitions."))
```
aquesnel: I find it very useful to have the spec and struct refence in the error message when there is a parsing failure. Can you please add the spec and struct reference in the error message that matches the struct that is being parsed.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
         }
         else
         {
-            x1 = MIN(x1, client_info->minfo[index].left);
-            y1 = MIN(y1, client_info->minfo[index].top);
-            x2 = MAX(x2, client_info->minfo[index].right);
-            y2 = MAX(y2, client_info->minfo[index].bottom);
+            LOG_DEVEL(LOG_LEVEL_TRACE, "Received [MS-RDPBCGR] "
+                "TS_UD_CS_MONITOR.TS_MONITOR_DEF %d "
+                "left %d, top %d, right %d, bottom %d, flags 0x%8.8x",
```
aquesnel: why are these two log messages at different levels and using different formats? 
can you please make them consistent, and include the spec and struct reference in the messages.

libxrdp/xrdp_sec.c
```diff
@@ -2298,159 +2298,190 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-        return 0;
-    }
-    if (!s_check_rem_and_log(s, 8, "Parsing [MS-RDPBCGR] TS_UD_CS_MONITOR"))
-    {
-        return 1;
-    }
-    in_uint32_le(s, flags); /* flags */
-    in_uint32_le(s, monitorCount);
-    LOG_DEVEL(LOG_LEVEL_TRACE, "Received [MS-RDPBCGR] TS_UD_CS_MONITOR "
-              "flags 0x%8.8x, monitorCount %d", flags, monitorCount);
```
aquesnel, line 2361: This trace log message seems to have been removed. Can you please add it back in the `xrdp_sec_process_mcs_data_monitors` function?

xrdp/xrdp_mm.c
```diff
@@ -1042,27 +1042,110 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
+    if (description->session_width <= 0 || description->session_height <= 0) {
+        LOG(LOG_LEVEL_INFO, "process_dynamic_monitor_description: Not allowing resize due to invalid dimensions (w: %d x h: %d)", description->session_width, description->session_height);
+        return 0;
+    }
+
+    // TODO: Unify this logic with server_reset
+    error = libxrdp_reset(wm->session, description->session_width, description->session_height, wm->screen->bpp);
+    if (error != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "dynamic_monitor_data: libxrdp_reset failed %d", error);
```
aquesnel: this seem like it warrants an error message (as well as all of the other error cases in this function). The rest of the code use the error log level for these situations.

libxrdp/xrdp_sec.c
Nexarian: Fixed after merged with astyle.

xrdp/xrdp_mm.c
Nexarian: Fixed with merge to astyle.

libxrdp/xrdp_sec.c
Nexarian: Fixed.

libxrdp/xrdp_sec.c
Nexarian: I think this was leftover from my debugging while developing the change. Reverted.

libxrdp/xrdp_sec.c
Nexarian: Good catch. Fixed.

xrdp/xrdp_mm.c
Nexarian: Added null check to process_dynamic_monitor_description.

xrdp/xrdp_mm.c
Nexarian: I think I fixed this?

libxrdp/xrdp_sec.c
Nexarian: The flow of the logic has changed with this PR; making it difficult to put this somewhere elegant, but I've done my best to add it back

libxrdp/xrdp_sec.c
Nexarian: This way of exiting a method is actually what a lot of C programmers at Microsoft use (I used to work there), and other than exiting nested loops this is the only legitimate use of goto I'm aware of.

Still, I hear you. I've refactored it so that `process_monitor_stream` accepts a `display_size_description` structure  that needs to be pre-allocated by the calling functions, and since there are only two places where it's used/allocated/destroyed unconditionally, this shouldn't be a problem.

libxrdp/xrdp_sec.c
Nexarian: Those values aren't directly specified in the Microsoft specification, they're derived. However, I've made an attempt to clarify where they come from and make the error messages clearer.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Updated by adding these to `ms-rdpbcgr.h`

Works great.

Nexarian:
Alright, I've addressed all of the comments and then verified that the update PR works correctly and doesn't break dynamic resizing.

Take another look @metalefty @aquesnel @matt335672

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    if (client_info->multimon != 1) /* are multi-monitors allowed ? */
-    {
-        LOG(LOG_LEVEL_INFO, "Multi-monitor is disabled by server config");
-        return 0;
-    }
-    if (!s_check_rem_and_log(s, 8, "Parsing [MS-RDPBCGR] TS_UD_CS_MONITOR"))
+    LOG_DEVEL(LOG_LEVEL_TRACE, "process_monitor_stream:");
+    if (description == NULL)
     {
+        LOG_DEVEL(LOG_LEVEL_DEBUG, "\tprocess_monitor_stream: description was null. Valid pointer to allocated description expected.");
```
aquesnel: If we get here it should indicate a bug in the code. I think this should be logged as an error.
Also, @matt335672 suggested that we log these error conditions using `LOG` instead of `LOG_DEVEL` so that we see this error message in big reports from users.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+
+    in_uint32_le(s, num_monitor);
+    LOG(LOG_LEVEL_DEBUG, "  num_monitor %d", num_monitor);
+
+    if (num_monitor >= CLIENT_MONITOR_DATA_MAXIMUM_MONITORS)
     {
         LOG(LOG_LEVEL_ERROR,
-            "[MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR flags MUST be zero, "
-            "received: 0x%8.8x", flags);
+            "\tprocess_monitor_stream: [MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR monitorCount "
```
aquesnel: FYI You can remove the function name orchid from the log message since they are added automatically when using a debug build.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-            "[MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR monitorCount "
-            "MUST be less than 16, received: %d", monitorCount);
-        return 1;
+        monitor_struct_stream_check_bytes = 20;
+        monitor_struct_stream_check_message = "\tprocess_monitor_stream: Parsing monitor definitions from 2.2.1.3.6.1 Monitor Definition (TS_MONITOR_DEF).";
+    }
+    else
+    {
+        monitor_struct_stream_check_bytes = 40;
+        monitor_struct_stream_check_message = "\tprocess_monitor_stream: Parsing monitor definitions from 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT.";
```
aquesnel: Since the RDP spec has many extension specs, can you please include the spec name abbreviation in the message. Eg. [MS-RDPBCGR]

aquesnel (commented):
I've done a partial code review and left a few comments. I'll complete the review when I have more time later.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
 
     if (!got_primary)
     {
         /* no primary monitor was set, choose the leftmost monitor as primary */
-        for (index = 0; index < monitorCount; index++)
+        for (int monitor_index = 0; monitor_index < num_monitor; ++monitor_index)
         {
-            if (client_info->minfo[index].left == x1 &&
-                    client_info->minfo[index].top == y1)
+            monitor_layout = description->minfo + monitor_index;
```
aquesnel: I'm curious as to why pointer arithmetic is used here instead of array indexing?

common/xrdp_client_info.h
```diff
@@ -24,15 +24,42 @@
...
+
+    /* Derived setting */
     int is_primary;
 };
 
+struct display_size_description
+{
+    int monitorCount; /* number of monitors detected (max = 16) */
+    struct monitor_info minfo[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data */
+    struct monitor_info minfo_wm[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data, non-negative values */
```
aquesnel, line 64: Can you clarify what the difference between what these two fields represent?
it seems like `minfo` is the client side representation of the monitor layout, and `minfo_wm` is the server side representation of the same monitor layout. Is that correct?

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    in_uint32_le(s, flags); /* flags */
-    in_uint32_le(s, monitorCount);
-    LOG_DEVEL(LOG_LEVEL_TRACE, "Received [MS-RDPBCGR] TS_UD_CS_MONITOR "
-              "flags 0x%8.8x, monitorCount %d", flags, monitorCount);
 
-    //verify flags - must be 0x0
-    if (flags != 0)
+    int num_monitor;
+    struct monitor_info *monitor_layout;
+    struct xrdp_rect rect = {8192, 8192, -8192, -8192};
```
aquesnel: Where do these initial rectangle bounds come from? 
it looks like these initial value get overwritten on line 2481, am I missing something that makes these initial values important?

Also, the variable name `rect` isn't very descriptive, and I want to make sure I understand what's it's calculating correctly: `rect`'s field values are in coordinate space, and `rect` represents the minimum sized bounding box which encloses all monitors. Is this correct?

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+    int got_primary = 0;
+    int monitor_struct_stream_check_bytes;
+    const char *monitor_struct_stream_check_message;
+
+    in_uint32_le(s, num_monitor);
+    LOG(LOG_LEVEL_DEBUG, "  num_monitor %d", num_monitor);
```
aquesnel: Since this log message is for sys admins can you please add more context and expand or avoid abbreviations.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+
+    if (num_monitor >= CLIENT_MONITOR_DATA_MAXIMUM_MONITORS)
     {
         LOG(LOG_LEVEL_ERROR,
-            "[MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR flags MUST be zero, "
-            "received: 0x%8.8x", flags);
+            "\tprocess_monitor_stream: [MS-RDPBCGR] Protocol error: TS_UD_CS_MONITOR monitorCount "
+            "MUST be less than 16, received: %d", num_monitor);
```
aquesnel: Can you please use string formatting with the constant `CLIENT_MONITOR_DATA_MAXIMUM_MONITORS` instead of hard coding the value in the error message so that the error message doesn't get out-of-sync withthe constant value.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+        {
+            in_uint32_le(s, monitor_layout->physical_width);
+            in_uint32_le(s, monitor_layout->physical_height);
+            in_uint32_le(s, monitor_layout->orientation);
+            in_uint32_le(s, monitor_layout->desktop_scale_factor);
+            in_uint32_le(s, monitor_layout->device_scale_factor);
+        }
+
+        monitor_layout->right = monitor_layout->left + monitor_layout->width;
+        monitor_layout->bottom = monitor_layout->top + monitor_layout->height;
```
aquesnel: I think there is a few problem here based on my reading of the spec for [TS_MONITOR_DEF](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/c3964b39-3d54-4ae1-a84a-ceaed311e0f6) and [DISPLAYCONTROL_MONITOR_LAYOUT](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/ea2de591-9203-42cd-9908-be7a55237d1c)

1. The spec says that a `TS_MONITOR_DEF` has a flags field after the height, and I don't see where those bytes are consumed from the stream. Since this is in a loop, when the flags field doesn't get consumed in the first iteration, it means that the second iteration of the loop will be off by the width of the flags field (4 bytes).

2. The `TS_MONITOR_DEF` right/bottom fields are absolute coordinates of the monitor, where as the `DISPLAYCONTROL_MONITOR_LAYOUT` width/height fields are size values and are relative to the coordinates in the left/top fields.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
     return 0;
 }
 
-/*****************************************************************************/
-/* Process a [MS-RDPBCGR] TS_UD_CS_MONITOR message.
-   reads the client monitors data */
-static int
-xrdp_sec_process_mcs_data_monitors(struct xrdp_sec *self, struct stream *s)
+int
+process_monitor_stream(struct stream *s, struct display_size_description *description, int full_parameters)
```
aquesnel: Can you please add the new function documentation, and clarify when the `full_parameters` value should be true vs false.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+
+        monitor_layout->right = monitor_layout->left + monitor_layout->width;
+        monitor_layout->bottom = monitor_layout->top + monitor_layout->height;
+
+        if (full_parameters != 0)
+        {
+            /*
+             * 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT
+             */
+            LOG(LOG_LEVEL_TRACE, "\tprocess_monitor_stream: "
```
aquesnel: I have been putting the trace messages as `LOG_DEVEL` since these messages are only useful for developers and not sys admins.

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+                "Received [MS-RDPBCGR] 2.2.1.3.6.1 TS_UD_CS_MONITOR.TS_MONITOR_DEF "
+                "Index: %d, Left %d, Top %d, Right %d, Bottom %d, Flags 0x%8.8x",
+                monitor_index,
+                monitor_layout->left,
+                monitor_layout->top,
+                monitor_layout->right,
+                monitor_layout->bottom,
+                monitor_layout->is_primary);
```
aquesnel: `monitor_layout->is_primary` isn't assigned anywhere in the new version, but was assigned in the old version. Is this intentional?

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
 
     /* set wm geometry */
-    if ((x2 > x1) && (y2 > y1))
+    if ((rect.right > rect.left) && (rect.bottom > rect.top))
     {
-        client_info->width = (x2 - x1) + 1;
-        client_info->height = (y2 - y1) + 1;
+        description->session_width = rect.right - rect.left;
+        description->session_height = rect.bottom - rect.top;
     }
```
aquesnel: why do we only set the session's width and height when `rect` is well-formed? it seems like if `rect` is not well-formed then xrdp should log and return an error. 

Also, why was the `+1` from `(x2 - x1) + 1` removed?

libxrdp/xrdp_sec.c
```diff
@@ -2302,159 +2302,235 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    /* make sure virtual desktop size is ok */
-    if (client_info->width > 0x7FFE || client_info->width < 0xC8 ||
-            client_info->height > 0x7FFE || client_info->height < 0xC8)
+
+    /* make sure virtual desktop size is ok - 2.2.1.3.6 Client Monitor Data (TS_UD_CS_MONITOR) */
+    if (description->session_width > CLIENT_MONITOR_DATA_MAXIMUM_VIRTUAL_DESKTOP_WIDTH || description->session_width < CLIENT_MONITOR_DATA_MINIMUM_VIRTUAL_DESKTOP_WIDTH ||
+            description->session_height > CLIENT_MONITOR_DATA_MAXIMUM_VIRTUAL_DESKTOP_HEIGHT || description->session_height < CLIENT_MONITOR_DATA_MINIMUM_VIRTUAL_DESKTOP_HEIGHT)
```
aquesnel: The `description->session_width` and `description->session_height` are not initialized by this function, and only conditionally set if the `rect` variable is well-formed. This means that this function could be logging values in the error message that were not sent by the client. Can you please fix this so that the error message always shows what was received from the client.

common/xrdp_client_info.h
```diff
@@ -24,15 +24,42 @@
...
+    int height;
+    int physical_width;
+    int physical_height;
+    int orientation;
+    int desktop_scale_factor;
+    int device_scale_factor;
+
+    /* Derived setting */
     int is_primary;
 };
```
aquesnel: The `width` and `height` fields are redundant with the `right` and `bottom` fields. To avoid having out-of-sync fields, I think we should only keep the `right` and `bottom` fields, and calculate `width` and `height` when needed.

Also, all of the `DISPLAYCONTROL_MONITOR_LAYOUT` are optional, but there is no way to determine of the values in those fields are valid or not. Can we make them always valid, and just put in some sensible default values?

xrdp/xrdp_mm.c
```diff
@@ -1075,73 +1169,31 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    if (msg_type != DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT)
+    {
+        return 0;
     }
-    return 0;
+    in_uint32_le(s, monitor_layout_size);
+    if (monitor_layout_size != 40)
+    {
+        /* 2.2.2.2 DISPLAYCONTROL_MONITOR_LAYOUT_PDU */
+        LOG(LOG_LEVEL_ERROR, "\tdynamic_monitor_data: monitor_layout_size is %d. Per spec it must be 40.", monitor_layout_size);
```
aquesnel: > Per spec it must be 40.

Which spec? when I find this message in the log file, where will I find where in the spec this value is defined?
I recommend moving the code comment into the error message, so that the error message has the full context of the problem.

aquesnel (changes requested):
I've finished reviewing the code. Please let me know if you disagree with my suggestions.

libxrdp/xrdp_sec.c
matt335672: Yes - I appreciate that a 'null pointer' error may not mean much to a user, but if it's not easy or possible to find a user-focussed answer it's better for them to have something to report than nothing at all.

common/xrdp_client_info.h
matt335672: Pardon me for jumping in, but since I'm here and I've used these values I can say that that's essentially correct The _wm values are normalised for a top-left co-ordinate of (0,0), which is useful for some bits of the software, namely the VNC multimon stuff and the login window.

common/xrdp_client_info.h
```diff
@@ -24,15 +24,42 @@
...
 
+struct display_size_description
+{
+    int monitorCount; /* number of monitors detected (max = 16) */
+    struct monitor_info minfo[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data */
+    struct monitor_info minfo_wm[CLIENT_MONITOR_DATA_MAXIMUM_MONITORS]; /* client monitor data, non-negative values */
+    int session_width;
+    int session_height;
+};
+
```
matt335672, line 68: There's some duplication here - the new `struct display_size_description` contains fields which are already present in this file [further down](https://github.com/neutrinolabs/xrdp/blob/d126a31067135a0620c541ee24649bef0e6de2dc/common/xrdp_client_info.h#L123-L125).

xrdp/xrdp_mm.c
```diff
@@ -1042,27 +1042,121 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
+    wm->client_info->width = description->session_width;
+    wm->client_info->height = description->session_height;
+    g_memcpy(wm->client_info->minfo, description->minfo, sizeof(struct monitor_info) * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
+    g_memcpy(wm->client_info->minfo_wm, description->minfo_wm, sizeof(struct monitor_info) * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
+    return 0;
+}
+
+/******************************************************************************/
+int
+process_monitor_stream(struct stream *s, struct display_size_description *description, int full_parameters);
```
matt335672: See comments for `xrdp_sec.c` - delete this function declaration.

libxrdp/xrdp_sec.c
matt335672: To add to @aquesnel's comment - there's another more significant issue here. This function is defined and used here, but also used in `xrdp_mm.c`. At the moment there isn't a common declaration for this function in an include file, and so the compiler isn't able to check that the function signature is the same between `xrdp_mm.c` and this file. So if you added an extra parameter to this function and forgot to update `xrdp_mm.c` we'd end up with a difficult-to-debug situation.

The function is defined in libxrdp, and as far as I can tell, the naming convention here is to prefix global functions from this library with `libxrdp_` and declare them in `libxrdp/libxrdpinc.h`. I don't think this is written down anywhere, and I've only just worked it out myself, but if you look at the calls that xrdp makes to libxrdp, you'll see what I mean.

 So I think this function should be called `libxrdp_process_monitor_stream()`

Also, we're trying for new stuff to add proper function documentation headers using Doxygen-style comments. At some stage we can then start generating development docs from the code using Doxygen like [FreeRDP](https://pub.freerdp.com/api/).

My suggestion here is:-
- Rename this function as `libxrdp_process_monitor_stream()`
- Add a declaration for the function to `libxrdp/libxrdpinc.h`
- Add function documentation to the declaration using a Doxygen-style header. Examples are in `common/log.h` or `common/string_calls.h`. Both of these are fairly new.

There's the small matter of `struct display_size_description` which isn't specified in `libxrdp/libxrdpinc.h`. One solution to this would be to move the whole struct to `libxrdp/libxrdpinc.h` and call it `struct xrdp_display_size_description`. Personally I don't think this is necessary. Since you're just using a pointer to the structure in your function prototype you can use an incomplete type declaration for the struct in `libxrdp/libxrdpinc.h` like this:-

```c
/* Defined in xrdp_client_info.h */
struct display_size_description;

<stuff snipped>

/**
 * Doxygen function comment
 * @param s . . . 
 * @param description . . .
 * @param full_parameters . . .
 * @return . . .
 */
int
libxrdp_process_monitor_stream(struct stream *s,
                               struct display_size_description *description,
                               int full_parameters);
```

Let me know if that doesn't makes sense.

matt335672 (changes requested):
Hi @Nexarian - thanks again for all the work you're putting into this. Please come back to me if you disagree with anything I've said below.

common/xrdp_client_info.h
Nexarian: I know. I didn't want to change xrdp_client_info since that will incur a revision of the header file and also affect xorgxrdp. I want to make the change to unify these two, but not in this PR... I want to do it for a PR that's focused on unifying these two data structures and leave the ones that change xorgxrdp more focused and linked. Make sense?

libxrdp/xrdp_sec.c
Nexarian: I'm happy to do it either way, I liked the way pointer arithmetic worked better here, but if you'd prefer indexing I'll change it.

common/xrdp_client_info.h
Nexarian: Exactly. Does this need further documentation in the code?

libxrdp/xrdp_sec.c
Nexarian: Updated rect to `all_monitors_encompassing_bounds` -- You're right the initial values are from an older version of the code, so I've removed them.

libxrdp/xrdp_sec.c
Nexarian: You're right. Fixed.

libxrdp/xrdp_sec.c
Nexarian: Good catch. Updated.

common/xrdp_client_info.h
Nexarian: I don't see where in https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/ea2de591-9203-42cd-9908-be7a55237d1c the values in the structure are optional...

I was trying to keep the data in the monitor info structure to be representative of the data sent from the client, but I agree data consistency is more important. Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Good point -- I added the error.

The +1 was removed because it wasn't present in the similar code that came from the resizing code, but it probably should be there as it's actually more correct.

libxrdp/xrdp_sec.c
Nexarian: Makes sense! Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: I think I addressed this? Let me know.

libxrdp/xrdp_sec.c
Nexarian: Done.

libxrdp/xrdp_sec.c
Nexarian: Fixed.

libxrdp/xrdp_sec.c
Nexarian: You're right. My alignment was off. I believe I've now fixed it.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: But this isn't a debug log (DEVEL) -- If it's not a debug build, does the orchid still appear?

Nexarian:
Addressed comments, give it another look!

common/xrdp_client_info.h
matt335672: Personally I'm happy with the existing comment

common/xrdp_client_info.h
matt335672: Yes that's fine- you can raise a PR now and mark it as draft if you like to keep the book-keeping simple.

libxrdp/xrdp_sec.c
matt335672: Also, I agree this should be LOG_LEVEL_ERROR

libxrdp/xrdp_sec.c
matt335672: Good point - the orchid won't appear in a non-debug build.

libxrdp/xrdp_sec.c
matt335672: Another also - please split the string into multiple string literals `"..." "..."` and reformat to 80 chars (or a little bit over) max line width as per the coding standard. There are quite a few other instances of this further below too. This bit of shell magic picks out the culprits:-

```
grep -n '' libxrdp/xrdp_sec.c | sed -ne '2307,2505p' | grep -E ':.{80}'
```

I appreciate that other bits of this file are also breaking the standard here, but for new code I think we should try to avoid this where possible.

libxrdp/xrdp_sec.c
```diff
@@ -2302,160 +2303,256 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+    description->monitorCount = num_monitor;
 
-    x1 = 0;
-    y1 = 0;
-    x2 = 0;
-    y2 = 0;
-    got_primary = 0;
-    /* Add client_monitor_data to client_info struct, will later pass to X11rdp */
-    for (index = 0; index < monitorCount; index++)
+    for (int monitor_index = 0; monitor_index < num_monitor; ++monitor_index)
```
matt335672: This is perfectly valid C99, but you'll have to move the loop variable declaration out of the loop. See the [coding standard](https://github.com/neutrinolabs/xrdp/wiki/Coding-Style#for-statements) and #1722 for the history of this restriction.

libxrdp/xrdp_sec.c
```diff
@@ -2302,160 +2303,256 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+        {
+            got_primary = 1;
+        }
     }
 
     if (!got_primary)
     {
         /* no primary monitor was set, choose the leftmost monitor as primary */
-        for (index = 0; index < monitorCount; index++)
+        for (int monitor_index = 0; monitor_index < num_monitor; ++monitor_index)
```
matt335672: See comment above.

libxrdp/xrdp_sec.c
```diff
@@ -2302,160 +2303,256 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+
+    LOG_DEVEL(LOG_LEVEL_TRACE, "xrdp_sec_process_mcs_data_monitors:");
+
+    /* this is an option set in xrdp.ini */
+    if (client_info->multimon != 1) /* are multi-monitors allowed ? */
+    {
+        LOG(LOG_LEVEL_INFO, "\txrdp_sec_process_mcs_data_monitors: Multi-monitor is disabled by server config");
+        return 0;
+    }
+    if (!s_check_rem_and_log(s, 8, "\txrdp_sec_process_mcs_data_monitors: Parsing [MS-RDPBCGR] TS_UD_CS_MONITOR"))
```
matt335672: You're checking for 8 bytes on the stream here, of which 4 are used for the flags, and 4 are used for the monitor_count in `libxrdp_process_monitor_stream()`.  That works here, but in `xrdp_mm.c` your call to `libxrdp_process_monitor_stream()` is reading the stream without a check that the data is available (if I'm reading it right). Suggest you make this check for 4 bytes for the flags, and add a 4-byte check to `libxrdp_process_monitor_stream()` before reading the monitor count.

Hope that makes sense

matt335672 (commented):
Hi @Nexarian - thanks for the swift turnaround. No major comments this time. As ever, come back to me if you disagree with anything I've raised here.

common/xrdp_client_info.h
aquesnel: Sorry my comment was not clear. You are correct that all of the fields in the spec are mandatory.

What I meant by optional fields is that with the changes to the xrdp monitor_info structure, now the physical size and scaling fields are are not populated when parsing a TS_MONITOR_DEF.

I suggest that when the code parses a TS_MONITOR_DEF then all of the monitor_info field are populated with reasonable defaults, so that we don't have any undefined data in those fields under some conditions.

aquesnel:
Hi @Nexarian 

Thanks for addressing my comments. I haven't been able to take a look at the most recent changes yet. I'll let you know when I do.

Separately, since there is now the infrastructure for unit tests, and units tests are good for proving that refactoring hasn't broken anything, is there anything in this pr that you think would benefit from adding some unit tests?

Nexarian:
@aquesnel If we want to add unit tests to this I propose we do one of two things:
1. Add unit tests for the functions that are in master BEFORE we merge in this PR, so we can lock-in the behavior that's already there, and then make sure this PR doesn't break them.
2. Check this code in and then add unit tests in a separate PR.

I don't want to add unit tests (or at least, unit tests for new functionality) to this PR because it's already pretty big. I'm in favor of (1) but that'll take more time as I have to learn the unit test system and write them. Probably another month given the bandwidth I currently have, though that's probably OK. We want HiDPI and multi-mon done right.

Nexarian:
@matt335672 I'll get to your comments sometime this week. I spent the weekend playing Factorio as I needed a break from manic coding :)

Nexarian:
Take a look at https://github.com/neutrinolabs/xrdp/pull/1932. Let's get that checked in before I finish this one, that way we can make sure that this PR doesn't break any functionality with the existing monitor processing.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Done

Nexarian:
@matt335672 @aquesnel I know it's been a long time, but I've finished updating this and incorporating all the comments and rebasing off of devel. Take another look, we still need this!

libxrdp/xrdp_sec.c
```diff
@@ -2328,162 +2329,352 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
 /*****************************************************************************/
-/* Process a [MS-RDPBCGR] TS_UD_CS_MONITOR message.
-   reads the client monitors data */
+/*
+   Process a [MS-RDPBCGR] TS_UD_CS_MONITOR message.
+   reads the client monitors data
+*/
 int
-xrdp_sec_process_mcs_data_monitors(struct xrdp_sec *self, struct stream *s)
+libxrdp_process_monitor_stream(struct stream *s,
```
aquesnel: Now that this function is named to be a part of the libxrdp, should it be moved to one of the libxrdp files?

matt335672:
Just back from a week away today. I'll get to this later in the week. Thanks for sticking with it.

Nexarian:
@matt335672 This is still an important next step to merging in the resizing stability changes that I've made to be compatible with the EGFX layer. I haven't been idle, I continue to work on tweaking the branch that incorporates Jay's changes and merges in a lot of the resizing and stability fixes, I go in and out of motivation :)

libxrdp/libxrdp.h
```diff
@@ -359,6 +359,10 @@ int
 xrdp_mcs_disconnect(struct xrdp_mcs *self);
 
 /* xrdp_sec.c */
+#define SEC_PROCESS_MONITORS_ERR 1
+#define SEC_PROCESS_MONITORS_ERR_TOO_MANY_MONITORS 2
+#define SEC_PROCESS_MONITORS_ERR_INVALID_DESKTOP 3
```
matt335672, line 371: These values are (I believe) restricted to return values from `xrdp_sec_process_mcs_data_monitors()`. Suggest adding a comment here to make this clear. Alternatively, use an `enum`, although with X that's little more than a stylistic preference.

libxrdp/xrdp_sec.c
matt335672: That's a fair point - the libxrdp_* functions we already have are defined in `libxrdp/libxrdp.c`. At least, according to `grep ^libxrdp_ libxrdp/*.c` they are.

There doesn't seem to be anything in `libxrdp_process_monitor_stream()` which requires it to be in `xrdp_sec.c`, unless I'm missing something(?)

libxrdp/xrdp_sec.c
```diff
@@ -2328,162 +2329,352 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
-    LOG_DEVEL(LOG_LEVEL_TRACE, "Received [MS-RDPBCGR] TS_UD_CS_MONITOR "
-              "flags 0x%8.8x, monitorCount %d", flags, monitorCount);
 
-    //verify flags - must be 0x0
-    if (flags != 0)
+    in_uint32_le(s, num_monitor);
+    LOG(LOG_LEVEL_DEBUG, "  The number of monitors received is: %d",
+        num_monitor);
+
+    if (num_monitor >= CLIENT_MONITOR_DATA_MAXIMUM_MONITORS)
```
matt335672: Strictly speaking, `num_monitor` (and `monitor_index`) should be unsigned to prevent surprises if a malicious client sends a large number here which would overflow a 32-bit signed int (that's the world we're living in these days). Alternatively, add a `num_monitor < 0` clause to this test.

libxrdp/xrdp_sec.c
```diff
@@ -2328,162 +2329,352 @@ xrdp_sec_process_mcs_data_channels(struct xrdp_sec *self, struct stream *s)
...
+             * Ex: in_uint32_le(s, monitor_layout->height);
+             */
+            in_uint32_le(s, monitor_layout->bottom);
+
+            in_uint32_le(s, monitor_layout->physical_width);
+            in_uint32_le(s, monitor_layout->physical_height);
+            in_uint32_le(s, monitor_layout->orientation);
+            in_uint32_le(s, monitor_layout->desktop_scale_factor);
+            in_uint32_le(s, monitor_layout->device_scale_factor);
+
```
matt335672: 2.2.2.2.1 of [MS-EDISP] has some MUST clauses that apply to validating the above data items.

xrdp/xrdp_mm.c
```diff
@@ -1030,27 +1030,154 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
+    int error = 0;
+    struct xrdp_mm *mm = wm->mm;
+    struct xrdp_mod *module = mm->mod;
+
+    LOG_DEVEL(LOG_LEVEL_TRACE, "process_dynamic_monitor_description:");
+
+    if (wm->client_info->suppress_output == 1)
+    {
+        LOG(LOG_LEVEL_ERROR,
+            "\tprocess_dynamic_monitor_description:"
```
matt335672: I'm curious as to why you've used a tab `\t` at the start of a lot of your log messages(?)

matt335672 (commented):
Looking good - a few comments/questions

On a more practical side, @metalefty is looking to release v0.9.19 in the reasonably near future.  The next release is going to be a v0.10.x or v1.0.x. As yet, we've not decided on a development time for it, or what we might want to get in to it. I'll raise a discussion item after v0.9.19 is out.

xrdp/xrdp_mm.c
Nexarian: The idea was to have something like this:

```
process_dynamic_monitor_description:
    process_dynamic_monitor_description: Foobar foobar foobar
```

Anyway, I recognize this is not standard practice throughout XRDP, so I've removed it.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Moved to libxrdp.c

libxrdp/libxrdp.h
Nexarian: Comment added.

I'll stick with defines for now. I think this is a good topic of discussion to have with XRDP, however. Should error codes be a list of defines or enums? We should decide as a project which we prefer, and I'm happy to go with the consensus.

common/xrdp_client_info.h
Nexarian: I decided to actually just do it now. It's not as terrible as I thought.

libxrdp/xrdp_sec.c
Nexarian: I'm not sure what to do with the "This value MUST be ignored if..." -- That maybe LOG a warning and do nothing else? Don't populate the structure with the loaded value after first checking for the range?

I've updated all the fields that have a "THIS [value] MUST BE" with a check, however.

aquesnel (approved):

Nexarian:
@aquesnel Thanks for the approve! Please take a look at https://github.com/neutrinolabs/xorgxrdp/pull/212 as well!

Nexarian:
@matt335672 @metalefty I've updated to include most of the comments above. What could be discussed is:
1. I took the liberty of redoing xrdp_client_info to both be more compliant with the spec and to prevent duplication for monitor processing. This is an important win, but I separated it out into a separate commit for clarity.
2. This could probably use more unit tests to make absolutely sure I'm not regressing anything, but given that multi-mon is already in flux, that could also be done later and we can take this an important change that will help inform other important changes going forward such as https://github.com/neutrinolabs/xorgxrdp/pull/172
3.  I'm not sure what to do with the "This value MUST be ignored if..." -- That maybe LOG a warning and do nothing else? Don't populate the structure with the loaded value after first checking for the range? Right now, I simply load the value without checking anything... What is our standard here?

Please take another look! In concert with https://github.com/neutrinolabs/xorgxrdp/pull/212, this works on my Ubuntu VM and I verified resizing still works.

xrdp/xrdp_mm.c
matt335672: I'm happy with that personally, provided the log levels are the same.  Thanks for the explanation.

libxrdp/libxrdp.h
matt335672: Personally, I'm happy with either. BTW, where I said 'X' above, I meant 'C'! The advantage of enums is that the debugger understands them, and it's an indicator for the maintainer. On the flip side they can't be really used for bitmasks. At runtime, there's no difference I'm aware of.

matt335672:
As regards 'MUST be ignored' values, there aren't many of them in [MS-RDPEDISP]. My suggestion is:-
- PhysicalWidth/PhysicalHeight - these aren't currently used. Set to zero, and add a comment to xrdp.h that they may be zero. When/if we use them for scaling we'll have to cope with that. An alternative is to assume a particular DPI to set them, but I think that's dangerous in the long run.
- Orientation - assume ORIENTATION_LANDSCAPE
- DeviceScaleFactor/DesktopScaleFactor - assume 100%

That's the ignored values. Out-of-range values can just be set to the corresponding min/max value as appropriate.

How does that sound? It's only a suggestion.

libxrdp/libxrdp.c
```diff
@@ -1766,3 +1764,324 @@ libxrdp_send_session_info(struct xrdp_session *session, const char *data,
...
+            if (monitor_layout->bottom
+                    > CLIENT_MONITOR_DATA_MAXIMUM_VIRTUAL_MONITOR_HEIGHT
+                    || monitor_layout->bottom
+                    < CLIENT_MONITOR_DATA_MINIMUM_VIRTUAL_MONITOR_HEIGHT)
+            {
+                return SEC_PROCESS_MONITORS_ERR_INVALID_MONITOR;
+            }
+
+            in_uint32_le(s, monitor_layout->physical_width);
+            in_uint32_le(s, monitor_layout->physical_height);
```
matt335672, line 1909: Note to implement whatever we decide to do for ignored values.

libxrdp/libxrdp.c
```diff
@@ -1766,3 +1764,324 @@ libxrdp_send_session_info(struct xrdp_session *session, const char *data,
...
+                default:
+                    LOG(LOG_LEVEL_WARNING, "libxrdp_process_monitor_stream:"
+                        " Orientation is not one of %d, %d, %d, or %d."
+                        " Value was %d and ignored.",
+                        ORIENTATION_LANDSCAPE,
+                        ORIENTATION_PORTRAIT,
+                        ORIENTATION_LANDSCAPE_FLIPPED,
+                        ORIENTATION_PORTRAIT_FLIPPED,
+                        monitor_layout->orientation);
+            }
```
matt335672, line 1959: Ditto

libxrdp/libxrdp.c
```diff
@@ -1766,3 +1764,324 @@ libxrdp_send_session_info(struct xrdp_session *session, const char *data,
...
+
+            in_uint32_le(s, monitor_layout->desktop_scale_factor);
+            in_uint32_le(s, monitor_layout->device_scale_factor);
```
matt335672, line 1975: Ditto

matt335672 (commented):
Looking great - thanks @Nexarian 

From my perspective, we can merge this once we've decided on the 'must ignore' actions.

Nexarian:
I like the idea of adding defaults in for "must ignore if" parameters. I'm working on it, I've also decided to add more unit tests. Should be finished in a day or so more.

libxrdp/libxrdp.c
```diff
@@ -1766,3 +1764,325 @@ libxrdp_send_session_info(struct xrdp_session *session, const char *data,
...
+                    monitor_layout->orientation = ORIENTATION_LANDSCAPE;
+            }
+
+            in_uint32_le(s, monitor_layout->desktop_scale_factor);
+            in_uint32_le(s, monitor_layout->device_scale_factor);
+
+            /*
+             * 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT
+             */
+            LOG_DEVEL(LOG_LEVEL_TRACE, "libxrdp_process_monitor_stream:"
```
metalefty: I think the level of this log should be promoted. This is informative. What about DEVEL-INFO?

metalefty:
I'm fine with other parts 'cause Matt and Aaquesnel have already reviewed well.

libxrdp/libxrdp.c
Nexarian: Done.

libxrdp/libxrdp.c
Nexarian: Done

libxrdp/libxrdp.c
Nexarian: Done

matt335672 (approved):
I'm OK with this now - thanks @Nexarian

Nexarian:
@matt335672 I've updated https://github.com/neutrinolabs/xorgxrdp/pull/212. Let's get these checked in, and then we can move on to my state machine changes here: https://github.com/neutrinolabs/xrdp/pull/2175

matt335672:
See also #2201

PR #1899, opened 2021-05-26
Fixed the ability to overwrite common channel settings.

TOMATO-ONE:
Hi.

I have found that I cannot override the common channel settings when using the NeutrinoRDP Proxy.

I tried to write the following in `xrdp.ini`.
```
[cliprdr-ture]
name=cliprdr-true
lib=libxrdpneutrinordp.so
ip=cliprdr-true-Windows
<snip>
channel.cliprdr=true

[cliprdr-false]
name=cliprdr-false
lib=libxrdpneutrinordp.so
ip=cliprdr-false-Windows
<snip>
channel.cliprdr=false
```

I did some trial and error and added three lines to `xrdp_mm.c`.
I also added the same process as #1872 to` xrdp-neutrinordp.c `because the log shows the error `lxrdp_set_param: unknown name`.

I tested the behavior and confirmed that `channel.cliprdr`, `channel.rdpsnd `and `channel.rdpdr `work fine.
However, I did not know how to test `channel.drdynvc`.

Best regards.

### DIFF
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -535,6 +535,13 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     {
         /* Valid (but unused) parameters not logged */
     }
+    else if (g_strcmp(name, "channel.rdpdr") == 0 ||
+             g_strcmp(name, "channel.rdpsnd") == 0 ||
+             g_strcmp(name, "channel.cliprdr") == 0 ||
+             g_strcmp(name, "channel.drdynvc") == 0)
+    {
+        /* Valid (but unused) parameters not logged */
+    }
     else
     {
         LOG(LOG_LEVEL_WARNING, "lxrdp_set_param: unknown name [%s] value [%s]", name, value);
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -1797,7 +1797,6 @@ xrdp_mm_process_login_response(struct xrdp_mm *self, struct stream *s)
                 {
                     g_snprintf(port, 255, "%d", 7200 + display);
                 }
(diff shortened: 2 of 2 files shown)


matt335672:
Nice find.

I've had a look at the code, and I think all that needs doing in `xrdp_mm.c` is that the existing call to `xrdp_mm_update_allowed_channels()` at line 1802 (or thereabouts) should be moved to line 2496 (or thereabouts) in `xrdp_mm_connect()`. Whereas your code looks perfectly functional to me, it does mean that `xrdp_mm_update_allowed_channels()` can be called twice when chansrv is started, and this will produce two lots of logging. You won't have seen this as you're not using chansrv in your configuration.

Does that make sense? Could you give that a go and report back?

TOMATO-ONE:
Thanks! @matt335672 

I have moved the function `xrdp_mm_update_allowed_channels()` on line 1800 of `xrdp_mm.c` to line 2494 as you suggested.
604fa30

I have tested the operation.
It worked fine.

This fix is better.

matt335672:
Thanks for the swift reply.

I'll do some checking with chansrv and then merge this.

Also, before I do, I'll get this annoying CI problem reported - I'd hoped it was some sort of transient thing, but it appears not.

matt335672:
Looks OK to me.

I ran up with `channel.cliprdr=false` in a standard XRDP stanza (i.e. not NeutrinoRDP) and got the following message in the xrdp.log:-

``
[20210528-11:16:42] [INFO ] [xrdp_mm_update_allowed_channels(xrdp_mm.c:1741)] xrdp_mm_update_allowed_channels: channel cliprdr channel id 2 is disabled
```

And indeed the clip channel didn't work.

@TOMATO-ONE - thanks again for your contribution.

TOMATO-ONE:
Thank you too.

PR #1900, opened 2021-05-27
Rework VNC text clipboard interface

### LINKED ISSUE #1616: debian 10 <-> windows 10: clipboard works one way only
on a fresh debian 10 xrdp install, at the moment, can only copy from debian (xrdp) to windows (mstsc.exe), but not the other way around. What i have tried:
**autocutsel** -> no changes
**xrdp-chansrv** not running by default, when starting it manually, can see no changes, also in the logs
**added** CHANSRV_LOG_LEVEL=LOG_LEVEL_DEBUG to /etc/xrdp/sesman.ini, but didn't notice any changes in the log files /var/log/xrdp.log and /var/log/xrdp-sesman.log as well.

This install is not strictly default - i am using xrdp to share several xvfb sessions running in parallel on different virtual screens. Hence, the sessions are on arbitrary displays, :66, :98 and so on...

One more thing, strings "[DEBUG] CLIPRDR_FORMAT_ANNOUNCE" appear when i try to copy from windows, so apparently it gets through? I suspect the problem is indeed in xrdp-chansrv not running? How to make xrdp-chansrv up and running in this multi-display setup?

**Here's an example log:**
```
[20200624-02:27:57] [INFO ] starting xrdp with pid 31729
[20200624-02:27:57] [INFO ] listening to port 3389 on 0.0.0.0
[20200624-02:27:57] [INFO ] Socket 12: AF_INET6 connection received from ::ffff:192.168.111.135 port 63632
[20200624-02:27:57] [DEBUG] Closed socket 12 (AF_INET6 ::ffff:192.168.111.1 port 3389)
[20200624-02:27:57] [DEBUG] Closed socket 11 (AF_INET6 :: port 3389)
[20200624-02:27:57] [INFO ] Using default X.509 certificate: /etc/xrdp/cert.pem
[20200624-02:27:57] [INFO ] Using default X.509 key file: /etc/xrdp/key.pem
[20200624-02:27:57] [ERROR] Cannot read private key file /etc/xrdp/key.pem: Permission denied
[20200624-02:27:57] [DEBUG] TLSv1.3 enabled
[20200624-02:27:57] [DEBUG] TLSv1.2 enabled
[20200624-02:27:57] [DEBUG] Security layer: requested 11, selected 0
[20200624-02:27:57] [DEBUG] Closed socket 12 (AF_INET6 ::ffff:192.168.111.1 port 3389)
[20200624-02:27:57] [INFO ] Socket 12: AF_INET6 connection received from ::ffff:192.168.111.135 port 63633
[20200624-02:27:57] [DEBUG] Closed socket 12 (AF_INET6 ::ffff:192.168.111.1 port 3389)
[20200624-02:27:57] [DEBUG] Closed socket 11 (AF_INET6 :: port 3389)
[20200624-02:27:57] [INFO ] Using default X.509 certificate: /etc/xrdp/cert.pem
[20200624-02:27:57] [INFO ] Using default X.509 key file: /etc/xrdp/key.pem
[20200624-02:27:57] [ERROR] Cannot read private key file /etc/xrdp/key.pem: Permission denied
[20200624-02:27:57] [DEBUG] TLSv1.3 enabled
[20200624-02:27:57] [DEBUG] TLSv1.2 enabled
[20200624-02:27:57] [DEBUG] Security layer: requested 0, selected 0
[20200624-02:27:57] [INFO ] connected client computer name: TEST3
[20200624-02:27:57] [INFO ] adding channel item name rdpdr chan_id 1004 flags 0x80800000
[20200624-02:27:57] [INFO ] adding channel item name rdpsnd chan_id 1005 flags 0xc0000000
[20200624-02:27:57] [INFO ] adding channel item name cliprdr chan_id 1006 flags 0xc0a00000
[20200624-02:27:57] [INFO ] adding channel item name drdynvc chan_id 1007 flags 0xc0800000
[20200624-02:27:58] [INFO ] Non-TLS connection established from ::ffff:192.168.111.135 port 63633: encrypted with standard RDP security
[20200624-02:27:58] [DEBUG] xrdp_00007c05_wm_login_mode_event_00000001
[20200624-02:27:58] [INFO ] Loading keymap file /etc/xrdp/km-00000409.ini
[20200624-02:27:58] [WARN ] local keymap file for 0x00000409 found and doesn't match built in keymap, using local keymap file
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC started connecting
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC connecting to 127.0.0.1 5901
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC tcp connected
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC security level is 1 (1 = none, 2 = standard)
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC sending share flag
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC receiving server init
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC receiving pixel format
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC receiving name length
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC receiving name
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC sending pixel format
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC sending encodings
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC sending framebuffer update request
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC sending cursor
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: VNC connection complete, connected ok
[20200624-02:28:02] [DEBUG] xrdp_wm_log_msg: connected ok
[20200624-02:28:02] [DEBUG] return value from xrdp_mm_connect 0
[20200624-02:28:02] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 144
[20200624-02:28:09] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 180
[20200624-02:28:12] [DEBUG] VNC got clip data
[20200624-02:28:12] [DEBUG] CLIPRDR_FORMAT_ACK - status 1 length 0
[20200624-02:28:13] [DEBUG] VNC got clip data
[20200624-02:28:13] [DEBUG] CLIPRDR_FORMAT_ACK - status 1 length 0
[20200624-02:28:14] [DEBUG] CLIPRDR_DATA_REQUEST - status 0 length 4
[20200624-02:31:41] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 180
[20200624-02:32:43] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 144
[20200624-02:32:53] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 180
[20200624-02:34:50] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 144
[20200624-02:35:04] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 180
[20200624-02:35:32] [DEBUG] CLIPRDR_FORMAT_ANNOUNCE - status 0 length 180
```

**When running xrdp-chansrv manually, the log shows this and never more:**
```
root@localhost:~# CHANSRV_LOG_LEVEL=LOG_LEVEL_DEBUG DISPLAY=:98 xrdp-chansrv
chansrv::main: using log file [/root/.local/share/xrdp/xrdp-chansrv.98.log]
[20200624-03:00:13] [CORE ] main: app started pid 32387(0x00007e83)
[20200624-03:00:13] [INFO ] main: DISPLAY env var set to :98
[20200624-03:00:13] [INFO ] main: using DISPLAY 98
[20200624-03:00:13] [INFO ] channel_thread_loop: thread start
```

**xrdp info**
```
root@localhost:~# xrdp -v
xrdp 0.9.9
  Copyright (C) 2004-2018 Jay Sorg, Neutrino Labs, and all contributors.

  Configure options:
      --enable-ipv6
      --enable-jpeg
      --enable-fuse
      --enable-rfxcodec
      --enable-opus
      --enable-painter
      --enable-vsock
      --build=x86_64-linux-gnu
      --prefix=/usr
      --includedir=${prefix}/include
      --mandir=${prefix}/share/man
      --infodir=${prefix}/share/info
      --sysconfdir=/etc
      --localstatedir=/var
      --disable-silent-rules
      --libdir=${prefix}/lib/x86_64-linux-gnu
      --libexecdir=${prefix}/lib/x86_64-linux-gnu
      --disable-maintainer-mode
      --disable-dependency-tracking
      --with-socketdir=/var/run/xrdp/sockdir
      build_alias=x86_64-linux-gnu
      CFLAGS=-g -O2 -fdebug-prefix-map=/build/xrdp-GQ2R2u/xrdp-0.9.9=. -fstack-protector-strong -Wformat -Werror=format-security
      LDFLAGS=-Wl,-z,relro -Wl,-z,now -Wl,--as-needed
      CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
      PKG_CONFIG_PATH=/build/xrdp-GQ2R2u/xrdp-0.9.9/pkgconfig

  Compiled with OpenSSL 1.1.1d  10 Sep 2019
```

### LINKED ISSUE #1885: Disconnect after selecting large amount of text (Ubuntu 21.04, xrdp, x11vnc)

matt335672:
This is no longer in draft.

The built-in VNC clipboard text-interface has been largely superseded by the `xrdp-chansrv` clipboard function. However there are still a few edge cases where this is required. For example, the user may be trying to connect to a VNC server on a platform where `xrdp_chansrv` is unavailable

There are a few problems with the interface at present, as reported in #1885 and #1616. To summarise:-
- The clip text on the virtual channel isn't chunked into smaller virtual channel PDUs as detailed in MS-RDPBCGR 3.1.5.2. This means only a few kilobytes can be copied at most.
- At times the clipboard only works one way
- (Related to the above) The interface startup sequence doesn't conform to that details in MS-RDPECLIP.
- Carriage returns are not always handled consistently.

On an engineering level, we're missing an ms-rdpeclip.h header for the constants. Also the modules are unaware of whether chansrv is running or not, which makes the logging difficult and means we're missing some optimisations. The actual clip functions in `vnc.c` use a lot of magic numbers which could be replaced with proper defines.

There are currently 5 commits in this PR:-
1) Add a function `server_chansrv_in_use()` to the module interface so modules know if chansrv is active.
2), Add a new string function `g_strnchr()` to look for a character in a string bounded by length.
3) Create an ms-rdpeclip.h include, using defines already present in the chansrv clipboard feature
4) Split the VNC module up, with the clip functionality in `vnc_clip.h`
5) Update `vnc_clip.c` to prevent refreshed RFB clip data which isn't actually a change from being communicated to the RDP client. This was useful when running `autocutsel` to prevent unnecessary messages to the RDP client.

### DIFF
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -31,56 +31,13 @@
 #endif
 
 #include "vnc.h"
+#include "vnc_clip.h"
+#include "rfb.h"
 #include "log.h"
 #include "trans.h"
 #include "ssl_calls.h"
 #include "string_calls.h"
 #include "xrdp_client_info.h"
-
-/* Client-to-server messages */
-enum c2s
-{
-    C2S_SET_PIXEL_FORMAT = 0,
-    C2S_SET_ENCODINGS = 2,
-    C2S_FRAMEBUFFER_UPDATE_REQUEST = 3,
-    C2S_KEY_EVENT = 4,
-    C2S_POINTER_EVENT = 5,
-    C2S_CLIENT_CUT_TEXT = 6,
-};
-
-/* Server to client messages */
-enum s2c
-{
-    S2C_FRAMEBUFFER_UPDATE = 0,
-    S2C_SET_COLOUR_MAP_ENTRIES = 1,
-    S2C_BELL = 2,
-    S2C_SERVER_CUT_TEXT = 3
-};
-
-/* Encodings and pseudo-encodings
- *
- * The RFC uses a signed type for these. We use an unsigned type as the
- * binary representation for the negative values is standardised in C
- * (which it wouldn't be for an enum value)
- */
-
-typedef uint32_t encoding_type;
-
-#define ENC_RAW                   (encoding_type)0
-#define ENC_COPY_RECT             (encoding_type)1
-#define ENC_CURSOR                (encoding_type)-239
-#define ENC_DESKTOP_SIZE          (encoding_type)-223
-#define ENC_EXTENDED_DESKTOP_SIZE (encoding_type)-308
-
-/* Messages for ExtendedDesktopSize status code */
-static const char *eds_status_msg[] =
-{
-    /* 0 */ "No error",
-    /* 1 */ "Resize is administratively prohibited",
-    /* 2 */ "Out of resources",
-    /* 3 */ "Invalid screen layout",
-    /* others */ "Unknown code"
-};
 
 /* elements in above list */
 #define EDS_STATUS_MSG_COUNT \
@@ -92,11 +49,8 @@ enum
     MSK_EXTENDED_DESKTOP_SIZE = (1 << 0)
 };
 
-static int
-lib_mod_process_message(struct vnc *v, struct stream *s);
-
 /******************************************************************************/
-static int
+int
 lib_send_copy(struct vnc *v, struct stream *s)
 {
     return trans_write_copy_s(v->trans, s);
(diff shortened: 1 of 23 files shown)


vnc/vnc.c
```diff
@@ -1054,40 +866,40 @@ skip_encoding(struct vnc *v, int x, int y, int cx, int cy,
 
     switch (encoding)
     {
-        case ENC_RAW:
+        case RFB_ENC_RAW:
         {
             int need_size = cx * cy * get_bytes_per_pixel(v->server_bpp);
-            LOG(LOG_LEVEL_DEBUG, "Skipping ENC_RAW encoding");
+            LOG(LOG_LEVEL_DEBUG, "Skipping RFB_ENC_RAW encoding");
```
aquesnel, line 872: Are all of the encoding that are skipped just not supported by xrdp with libvnc? Should the user get a warning message about unsupported encodings?

vnc/vnc_clip.c
```diff
@@ -0,0 +1,1102 @@
...
+ *
+ * @param v VNC object
+ * @param s stream buffer
+ *
+ * @pre stream buffer must have been initialised with a call to
+ *      out_cliprdr_header()
+ * @pre Stream is terminated with s_mark_end()
+ */
+static int
+send_stream_to_clip_channel(struct vnc *v, struct stream *s)
```
aquesnel, line 162: This chunking is specific to the RDP protocol and not the RFB protocol. Since you are changing the interface for the clipboard, why not move this RDP specific functionality to libxrdp, and have the converted but un-chunked stream as an argument to a new libxrdp "virtual channel send with chunking" function?

vnc/vnc_clip.c
```diff
@@ -0,0 +1,1102 @@
...
+
+    free_stream(s);
+    return rv;
+}
+
+/******************************************************************************/
+int
+vnc_clip_open_clip_channel(struct vnc *v)
+{
+    v->clip_chanid = v->server_get_channel_id(v, "cliprdr");
```
aquesnel: I think "cliprdr" should be a constant in ms-rdpbcgr.h since it is a well-known value for the channel name.

xrdp/xrdp_mm.c
```diff
@@ -3610,14 +3621,11 @@ server_send_to_channel(struct xrdp_mod *mod, int channel_id,
...
     if (wm->mm->usechansrv)
     {
-        /*
-         * Xvnc backend reaches here
-         * should not return 1 as this case is not an error
-         */
-        return 0;
+        /* Modules should not be calling this if chansrv is running -
+         * they can use server_chansrv_in_use() to avoid doing this */
+        return 1;
```
aquesnel: I think we should add a `LOG_DEVEL` warning message to notify devs of the programming error.

aquesnel (commented):
@matt335672 reviewed with a few comments.

vnc/vnc_clip.c
matt335672: It's a good point.

I actually started down this road, and then realised I was going to get in a bit of a mess with the dechunking coming in, as this needs to be done somewhere.

The thing that decided me on this route was that most of the channel data is handled by chansrv, not XRDP modules. If we do the chunking/dechunking in libxrdp, we'll end up sending giant messages between chansrv and xrdp which will result in noticeable pauses if (for example) the user decides to paste a few MB of data in one go. One of the advantages of chunking is it allows PDUs to be prioritised if necessary.

We could maybe provide a chunking/dechunking facility as part of libcommon, or even provide another library. What do you think?

vnc/vnc.c
matt335672: The `skip_encoding()` function is used when we're trying to negotiate a desktop size with the user. We send an update request for a single pixel bit of the screen. If the server responds with an ExtendedDesktopSize rectangle we know the server supports resizing. If it doesn't, we know it doesn't support resizing. In both instances we're not interested in the rest of the data from the server at that time because we're going to either resize the server or ask for a full frame update. `skip_encoding()` is used to pass over data in the stream which we're expecting but which we don't care about.

The VNC protocol prevents the server from sending encodings the client hasn't said it supports. If it does, that's an error.

vnc/vnc_clip.c
matt335672: Agreed. There are probably a few other names I can find scattered about in the sources.

xrdp/xrdp_mm.c
matt335672: Agreed.

matt335672:
@aquesnel - thanks for that. I've addressed your minor review comments with the discussion about chunking still outstanding.

For the 32-bit compile fails, see #1916

vnc/vnc_clip.c
aquesnel: I agree that pipelining chunked messages between chansrv and the client is good for latency since the messages have to do an extra inter process communication hop.

I'm less convinced that there is a latency gain for chunking messages to the modules loaded in xrdp's memory space. It seems like the modules always need all chunks reassembled before any application logic can be run, and there is no cost for transferring a large stream to a loaded module.

If you think it's worth putting the chunking and de chunking logic into a library, I'd add it to libxrdp as it's own facility, and not integrated into the send/receive path.

Since you've already explored putting the chunking into libxrdp and determined that it's not the right path, I trust your judgement and we should go with the solution you've got.

matt335672: Thanks for that.

I agree that there's no great advantage for modules in re-assembling the chunks themselves. The SVC handler code in xrdp is pretty basic though, and if I'm reading it right (quite a big 'if') it just passes all SVC chunks for all channels through to a module if the module registers a handler in `mod_event`

https://github.com/neutrinolabs/xrdp/blob/96052ba6e3d97cfa6c0d41b039171d1db1a1eb59/xrdp/xrdp_wm.c#L1838-L1864

If we did the dechunking at that level we'd have to make the code more sophisticated to ignore chunks for unused channels, or we'd be re-assembling stuff just to bin it.

I'll raise an issue (tomorrow now) about this and identify all the chunking/dechunking code we're currently using.

matt335672: @aquesnel - I've raised #1920 as a separate issue regarding the chunking/dechunking.

Are you happy that satisfies the immediate requirement?

Feel free to add some detail to #1920 if you think it will be useful or necessary.

aquesnel: That separate issue looks go to me, I think this PR is ready to ship.

aquesnel (approved):

PR #1925, opened 2021-06-18
Move scp to standard transport

matt335672:
The PR is based on part of #1602.

#1602 is in two parts:-
1. Move SCP away from using `struct SCP_CONNECTION` as a connection type to the more standard `struct trans` transport type used everywhere else
2. Add the capability to use UNIX domain sockets for SCP

Both of these are necessary moving forwards to modernise SCP. This PR focusses on the first half only.

I've taken @jsorg71's original commits and rebased them. I've then completely removed the `struct SCP_CONNECTION` to avoid confusion, and to ensure that the transition to `struct trans` is complete. This has entailed updating the utilities.

There's still a lot to do for SCP. This is only a start.

### DIFF
--- a/sesman/libscp/libscp_v1s.c
+++ b/sesman/libscp/libscp_v1s.c
@@ -95,7 +95,7 @@ int in_string8(struct stream *s, char str[], const char *param, int line)
  * @return SCP_SERVER_STATE_OK for success
  */
 static enum SCP_SERVER_STATES_E
-scp_v1s_init_session(struct SCP_CONNECTION *c, struct SCP_SESSION *session)
+scp_v1s_init_session(struct trans *t, struct SCP_SESSION *session)
 {
     tui8 type;
     tui16 height;
@@ -103,12 +103,13 @@ scp_v1s_init_session(struct SCP_CONNECTION *c, struct SCP_SESSION *session)
     tui8 bpp;
     tui8 sz;
     char buf[256];
+    struct stream *in_s = t->in_s;
 
     scp_session_set_version(session, 1);
 
     /* Check there's data for the session type, the height, the width, the
      * bpp, the resource sharing indicator and the locale */
-    if (!s_check_rem(c->in_s, 1 + 2 + 2 + 1 + 1 + 17))
+    if (!s_check_rem(in_s, 1 + 2 + 2 + 1 + 1 + 17))
     {
         LOG(LOG_LEVEL_WARNING,
             "[v1s:%d] connection aborted: short packet",
@@ -116,7 +117,7 @@ scp_v1s_init_session(struct SCP_CONNECTION *c, struct SCP_SESSION *session)
         return SCP_SERVER_STATE_SIZE_ERR;
     }
 
-    in_uint8(c->in_s, type);
+    in_uint8(in_s, type);
 
     if ((type != SCP_SESSION_TYPE_XVNC) && (type != SCP_SESSION_TYPE_XRDP))
     {
@@ -126,11 +127,11 @@ scp_v1s_init_session(struct SCP_CONNECTION *c, struct SCP_SESSION *session)
 
     scp_session_set_type(session, type);
 
-    in_uint16_be(c->in_s, height);
+    in_uint16_be(in_s, height);
     scp_session_set_height(session, height);
-    in_uint16_be(c->in_s, width);
+    in_uint16_be(in_s, width);
     scp_session_set_width(session, width);
-    in_uint8(c->in_s, bpp);
+    in_uint8(in_s, bpp);
     if (0 != scp_session_set_bpp(session, bpp))
     {
         LOG(LOG_LEVEL_WARNING,
@@ -138,14 +139,14 @@ scp_v1s_init_session(struct SCP_CONNECTION *c, struct SCP_SESSION *session)
             __LINE__, bpp);
         return SCP_SERVER_STATE_INTERNAL_ERR;
     }
-    in_uint8(c->in_s, sz);
+    in_uint8(in_s, sz);
     scp_session_set_rsr(session, sz);
-    in_uint8a(c->in_s, buf, 17);
+    in_uint8a(in_s, buf, 17);
     buf[17] = '\0';
     scp_session_set_locale(session, buf);
 
     /* Check there's enough data left for at least an IPv4 address (+len) */
-    if (!s_check_rem(c->in_s, 1 + 4))
+    if (!s_check_rem(in_s, 1 + 4))
     {
         LOG(LOG_LEVEL_WARNING,
             "[v1s:%d] connection aborted: IP addr len missing",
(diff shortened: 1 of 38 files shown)


sesman/libscp/libscp_v1s.c
```diff
@@ -284,186 +252,122 @@ enum SCP_SERVER_STATES_E scp_v1s_accept(struct SCP_CONNECTION *c, struct SCP_SES
...
-    if (cmd != 1)
+    switch (cmd)
     {
-        LOG(LOG_LEVEL_WARNING, "[v1s:%d] connection aborted: sequence error", __LINE__);
-        return SCP_SERVER_STATE_SEQUENCE_ERR;
-    }
+        case 1:
+            s->current_cmd = cmd;
+            result = scp_v1s_init_session(t, s);
+            break;
```
aquesnel: is it worth it to change these magic numbers to constants now or will all of this code be refactored with your next change?

sesman/notes.txt
```diff
@@ -0,0 +1,58 @@
+
```
aquesnel, line 9: Assuming that committing this file was intentional, can you please add a header, or change the file name to clarify that this file is documenting the SCP protocol messages.

sesman/scp_v1.c
```diff
@@ -156,66 +151,126 @@ scp_v1_process(struct SCP_CONNECTION *c, struct SCP_SESSION *s)
...
+    e = scp_v1s_list_sessions42(t, scount, slist);
+    if (SCP_SERVER_STATE_OK != e)
     {
-        /* here goes scp resource sharing code */
+        LOG(LOG_LEVEL_ERROR, "scp_v1s_list_sessions42 failed");
     }
 
-    /* cleanup */
-    if (do_auth_end)
+    return SCP_SERVER_STATE_OK;
```
aquesnel: the log message here is at the error level, but the connection is not disconnected. I think this should be lowered to Warning.

sesman/scp_v1.c
```diff
@@ -156,66 +151,126 @@ scp_v1_process(struct SCP_CONNECTION *c, struct SCP_SESSION *s)
...
+    struct session_item *sitem;
+    enum SCP_SERVER_STATES_E e;
+    int display;
+
+    sitem = session_get_bypid(s->return_sid);
+    if (0 == sitem)
     {
-        auth_end(data);
+        e = scp_v1s_connection_error(t, "Internal error");
+        LOG(LOG_LEVEL_INFO, "Cannot find session item on the chain");
```
aquesnel: What's the "chain"? can you please make this log message more informative, like "The server does not have a session with id %s"?

sesman/sesman.c
```diff
@@ -187,6 +246,90 @@ static int sesman_listen_test(struct config_sesman *cfg)
...
+        self->in_s->p = self->in_s->data;
+        if (scp_process(self, sc->s) != SCP_SERVER_STATE_OK)
+        {
+            LOG(LOG_LEVEL_ERROR, "sesman_data_in: scp_process_msg failed");
+            return 1;
+        }
+        /* reset for next message */
+        self->header_size = 8;
+        self->extra_flags = 0;
+        init_stream(self->in_s, 0);
```
aquesnel: setting the transport input stream to size zero doesn't feel right. Can you please clarify this?

sesman/sesman.c
```diff
@@ -187,6 +246,90 @@ static int sesman_listen_test(struct config_sesman *cfg)
...
+    }
+    return 0;
+}
+
+/******************************************************************************/
+static int
+sesman_listen_conn_in(struct trans *self, struct trans *new_self)
+{
+    struct sesman_con *sc;
+    if (g_con_list->count >= 16)
```
aquesnel: why a maximum of 16 connections to sesman? are the connections to sesman short lived or do they last the duration of the client rdp connection?

sesman/libscp/libscp_v1s.c
matt335672: I was going to replace all of this, but intentions are not the same as reality! I'll replace the magic numbers.

sesman/notes.txt
matt335672: Agreed.

sesman/scp_v1.c
matt335672: Agreed

sesman/scp_v1.c
matt335672: Agreed.

sesman/sesman.c
matt335672: I'll add a comment. The effect is to reset the stream pointers, but leave the buffer size unchanged.

sesman/sesman.c
matt335672: The '16' is somewhat inherited. Your question is however, like most of your questions, very incisive.

At the moment, connections to sesman are short lived, and last for the duration of making a request and receiving a response.

This may, however, have to change in the future. There are two reasons I'm aware of for this:-
- To support reconnections, sesman is going to need to keep a copy of a secret which is shared with the RDP client. This secret is regenerated hourly - see [MS-RDPBCGR] 5.5 which implies either a permanent connection between sesman and the xrdp instance, or xrdp reconnecting to sesman at regular intervals.
- The `xrdp-sesadmin` utility should support more functionality that it currently does. Implentation of (e.g.) a 'disconnect session' command implies sesman will ned to talk to the relevant xrdp instance.

For now I'll add a define for this with a short comment.

matt335672:
Review comments addressed.

aquesnel (approved):

jsorg71:
Thanks for this, it will be great to removing this unnecessary TCP listener.

metalefty:
There's 1 month until the next release. This is a breaking change so we need to triage if shipping this to August release or December release. Do you think 1 month is enough to test and debug this?

matt335672:
I should point out that a fair bit of this code isn't actually used at present.

The server-side SCP V0 code is used by xrdp and sesrun, and even they aren't using the SCP client-side library at the moment. The SCP V1 admin code is used by sesadmin, and the other SCP V1 code is used by sestest which isn't shipped by default, primarily as it isn't completed yet.

I've tested all the interfaces I can. xrdp/sesrun appear to work well, as does what there is of sesadmin. The sestest code run OK, but has other (existing) problems which I'm not planning on addressing right now.

In other words, I think an August release for this is realistic, but happy to discuss this further. I'm about for most of August apart from a few days from the 15th onwards.

If we need to bounce it until December that's fine, but I'd suggest a merge straight after v0.9.17 early in September if we do that. There are other changes to make in this area, but they're dependent on this one going in.

metalefty:
Then, I'm fine with shipping this to next August release.
Just to be sure, does this PR switches connection between xrdp and sesman from tcp to uds?

matt335672:
No it doesn't.

That's a subsequent part of Jay's original #1602. Because that's a significantly more user-visible change however, it will be good to get some more preparation done before we make that switch, so it's clearer what's going on.

In particular, at the moment, xrdp and sesrun don't use libscp to talk to sesman. This PR doesn't address that, but I've got another one planned that will.

After what we should be able to move to UDS with a relatively simple PR.

metalefty (approved):

metalefty:
Okay, let's ship this to the next release.

PR #1934, opened 2021-06-28
neutrinordp: Allow keyboard layout information to be sent to remote. (#1933)

TOMATO-ONE:
This is a  PR for Discussions #1933.

The goal is to resolve the keyboard layout mismatch when connecting to remote Windows via NeutrinoRDP Proxy from an RDP Client with a Japanese keyboard connected.
A lot of changes are needed to make this functionality possible.
Therefore, this PR will make the following two steps possible.

1st one: NeutrinoRDP module pass such information received from the client to another server as-is. 
2nd one: Allow xrdp to send fixed keyboard information to another server.

I welcome suggestions for corrections to redundant code or confusing comments.

### DIFF
--- a/common/string_calls.h
+++ b/common/string_calls.h
@@ -155,6 +155,15 @@ int      g_strncmp_d(const char *c1, const char *c2, const char delim, int len);
 int      g_strcasecmp(const char *c1, const char *c2);
 int      g_strncasecmp(const char *c1, const char *c2, int len);
 int      g_atoi(const char *str);
+/**
+ * Extends g_atoi(), Converts decimal and hexadecimal number String to integer
+ *
+ * Prefix hexadecimal numbers with '0x'
+ *
+ * @param str String to convert to an integer
+ * @return int Integer expression of a string
+ */
+int      g_atoix(const char *str);
 int      g_htoi(char *str);
 int      g_bytes_to_hexstr(const void *bytes, int num_bytes, char *out_str,
                            int bytes_out_str);
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -94,10 +94,50 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
 
 /******************************************************************************/
 /* return error */
+static void
+set_keyboard_overrides(struct mod *mod)
+{
+    rdpSettings *settings = mod->inst->settings;
+
+    if (mod->allow_client_kbd_settings)
+    {
+        settings->kbd_type = mod->client_info.keyboard_type;
+        settings->kbd_subtype = mod->client_info.keyboard_subtype;
+        /* Define the most common number of function keys, 12.
+           because we can't get it from client. */
+        settings->kbd_fn_keys = 12;
+        settings->kbd_layout = mod->client_info.keylayout;
+
+        /* Exception processing for each RDP Keyboard type */
+        if (mod->client_info.keyboard_type == 0x00)
+        {
+            /* 0x00000000 : Set on Server */
+            LOG(LOG_LEVEL_WARNING, "keyboard_type:[0x%02x] ,Set on Server",
+                mod->client_info.keyboard_type);
+        }
+        else if (mod->client_info.keyboard_type == 0x04)
+        {
+            /* 0x00000004 : IBM enhanced (101- or 102-key) keyboard */
+            /* Nothing to do. */
+        }
+        else if (mod->client_info.keyboard_type == 0x07)
+        {
+            /* 0x00000007 : Japanese keyboard */
+            /* Nothing to do. */
+        }
+    }
+    LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy remote keyboard settings, "
+        "kbd_type:[0x%02X], kbd_subtype:[0x%02X], "
+        "kbd_fn_keys:[%02d], kbd_layout:[0x%08X]",
+        settings->kbd_type, settings->kbd_subtype,
+        settings->kbd_fn_keys, settings->kbd_layout);
+}
+
 static int
 lxrdp_connect(struct mod *mod)
 {
     boolean ok;
+    set_keyboard_overrides(mod);
 
     LOG_DEVEL(LOG_LEVEL_TRACE, "lxrdp_connect:");
 
@@ -609,6 +649,10 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
         {
             mod->perf_settings_values_mask |= PERF_DISABLE_CURSOR_SHADOW;
         }
+    }
+    else if (g_strcmp(name, "neutrinordp.allow_client_keyboardLayout") == 0)
+    {
+        mod->allow_client_kbd_settings = g_text2bool(value);
     }
     else
     {
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -97,6 +97,7 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
 static void
 set_keyboard_overrides(struct mod *mod)
 {
+    const struct kbd_overrides *ko = &mod->kbd_overrides;
     rdpSettings *settings = mod->inst->settings;
 
     if (mod->allow_client_kbd_settings)
@@ -126,6 +127,42 @@ set_keyboard_overrides(struct mod *mod)
             /* Nothing to do. */
         }
     }
+
+    if (ko->type != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "overrode kbd_type 0x%02X with 0x%02X",
+            settings->kbd_type, ko->type);
+        settings->kbd_type = ko->type;
+    }
+
+    if (ko->subtype != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "overrode kbd_subtype 0x%02X with 0x%02X",
+            settings->kbd_subtype, ko->subtype);
+        settings->kbd_subtype = ko->subtype;
+    }
+
+    if (ko->fn_keys != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "overrode kbd_fn_keys %d with %d",
+            settings->kbd_fn_keys, ko->fn_keys);
+        settings->kbd_fn_keys = ko->fn_keys;
+    }
+
+    if (ko->layout != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "overrode kbd_layout 0x%08X with 0x%08X",
+            settings->kbd_layout, ko->layout);
+        settings->kbd_layout = ko->layout;
+    }
+
+    if (ko->layout_mask != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "Masked kbd_layout 0x%08X to 0x%08X",
+            settings->kbd_layout, settings->kbd_layout & ko->layout_mask);
+        settings->kbd_layout &= ko->layout_mask;
+    }
+
     LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy remote keyboard settings, "
         "kbd_type:[0x%02X], kbd_subtype:[0x%02X], "
         "kbd_fn_keys:[%02d], kbd_layout:[0x%08X]",
@@ -529,6 +566,7 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     }
     else if (g_strcmp(name, "keylayout") == 0)
     {
+        LOG(LOG_LEVEL_DEBUG, "%s:[0x%08X]", name, g_atoi(value));
     }
     else if (g_strcmp(name, "name") == 0)
     {
@@ -2236,6 +2274,31 @@ lfreerdp_receive_channel_data(freerdp *instance, int channelId, uint8 *data,
         {
             LOG(LOG_LEVEL_ERROR, "lfreerdp_receive_channel_data: error %d", error);
         }
+    }
+    else if (g_strcmp(name, "neutrinordp.allow_client_keyboardLayout") == 0)
+    {
+        mod->allow_client_kbd_settings = g_text2bool(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_keyboardLayout_mask") == 0)
+    {
+        /* Keyboard values are stored for later processing */
+        mod->kbd_overrides.layout_mask = g_atoix(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_kbd_type") == 0)
+    {
+        mod->kbd_overrides.type = g_atoix(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_kbd_subtype") == 0)
+    {
+        mod->kbd_overrides.subtype = g_atoix(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_kbd_fn_keys") == 0)
+    {
+        mod->kbd_overrides.fn_keys = g_atoix(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_kbd_layout") == 0)
+    {
+        mod->kbd_overrides.layout = g_atoix(value);
     }
     else
     {
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -408,6 +408,30 @@ g_atoi(const char *str)
 }
 
 /*****************************************************************************/
+/* As g_atoi() but allows for hexadecimal too */
+int
+g_atoix(const char *str)
+{
+    int base = 10;
+    if (str == NULL)
+    {
+        str = "0";
+    }
+
+    while (isspace(*str))
+    {
+        ++str;
+    }
+
+    if (*str == '0' && tolower(*(str + 1)) == 'x')
+    {
+        str += 2;
+        base = 16;
+    }
+    return strtol(str, NULL, base);
+}
+
+/*****************************************************************************/
 int
 g_htoi(char *str)
 {
--- a/neutrinordp/xrdp-neutrinordp.h
+++ b/neutrinordp/xrdp-neutrinordp.h
@@ -222,4 +222,5 @@ struct mod
     int allow_client_experiencesettings;
     int perf_settings_override_mask; /* Performance bits overridden in ini file */
     int perf_settings_values_mask; /* Values of overridden performance bits */
+    int allow_client_kbd_settings;
 };
--- a/neutrinordp/xrdp-neutrinordp.h
+++ b/neutrinordp/xrdp-neutrinordp.h
@@ -63,6 +63,15 @@ struct pointer_item
 
 struct source_info;
 
+struct kbd_overrides
+{
+    int type;
+    int subtype;
+    int fn_keys;
+    int layout;
+    int layout_mask;
+};
+
 struct mod
 {
     int size; /* size of this struct */
@@ -223,4 +232,5 @@ struct mod
     int perf_settings_override_mask; /* Performance bits overridden in ini file */
     int perf_settings_values_mask; /* Values of overridden performance bits */
     int allow_client_kbd_settings;
+    struct kbd_overrides kbd_overrides; /* neutrinordp.overide_kbd_* values */
 };
--- a/xrdp/xrdp.ini.in
+++ b/xrdp/xrdp.ini.in
@@ -279,6 +279,13 @@ password=ask
 ; If you want to tell the remote the keyboard layout of the RDP Client,
 ; by uncommenting the following line.
 #neutrinordp.allow_client_keyboardLayout=true
+; The following options will override the remote keyboard layout settings.
+; These options are for DEBUG and are not recommended for regular use.
+#neutrinordp.override_keyboardLayout_mask=0x0000FFFF
+#neutrinordp.override_kbd_type=0x04
+#neutrinordp.override_kbd_subtype=0x01
+#neutrinordp.override_kbd_fn_keys=12
+#neutrinordp.override_kbd_layout=0x00000409
 
 ; You can override the common channel settings for each session type
 #channel.rdpdr=true
--- a/xrdp/xrdp.ini.in
+++ b/xrdp/xrdp.ini.in
@@ -275,6 +275,10 @@ password=ask
 ; you problems (e.g. cursor is a black rectangle) try disabling cursor
 ; shadows by uncommenting the following line.
 #perf.cursor_shadow=false
+; By default, NeutrinoRDP uses the keyboard layout of the remote RDP Server.
+; If you want to tell the remote the keyboard layout of the RDP Client,
+; by uncommenting the following line.
+#neutrinordp.allow_client_keyboardLayout=true
 
 ; You can override the common channel settings for each session type
 #channel.rdpdr=true


neutrinordp/xrdp-neutrinordp.c
```diff
@@ -609,6 +610,77 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
...
             mod->perf_settings_values_mask |= PERF_DISABLE_CURSOR_SHADOW;
         }
+    }
+        else if (g_strcmp(name, "neutrinordp.allow_client_keyboardLayout") == 0)
+    {
+        if (g_text2bool(value))
+        {
+            if (settings->kbd_type == 0)
+            {
+                settings->kbd_type = mod->client_info.keyboard_type;
```
matt335672: `mod->client_info.keyboard_type` might be zero (see [MS-RDPBCGR] [2.2.1.3.2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/00f1da4a-ee9c-421a-852f-c19f92343d73). If this is the case, the layout will be set on the server. You might want to log a warning in this case, or the user will wonder why the layout is not being transferred.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -609,6 +610,77 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
...
+            }
+            LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy remote keyboard settings, "
+                                "kbd_type:[0x%02X], kbd_subtype:[0x%02X], "
+                                "kbd_fn_keys:[%02d], kbd_layout:[0x%08X]",
+                                settings->kbd_type, settings->kbd_subtype,
+                                settings->kbd_fn_keys, settings->kbd_layout);
+        }
+    }
+    /* The following options are for DEBUG and workaround.*/
+    else if (g_strcmp(name, "neutrinordp.override_keyboardLayout_masks") == 0)
```
matt335672: I think this code may behave differently depending on the order that these variables are encountered in the .ini file. If for example I put this:-

```
neutrinordp.allow_client_keyboardLayout=true
neutrinordp.override_kbd_type=0
```

I'll get a `settings->kbd_type` of zero.

However if I put this:-
```
neutrinordp.override_kbd_type=0
neutrinordp.allow_client_keyboardLayout=true
```

I'll get the client keyboard type.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -609,6 +610,77 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
...
+    {
+            settings->kbd_layout =
+                          mod->client_info.keylayout & strtol(value, NULL, 16);
+            LOG(LOG_LEVEL_WARNING,
+                "%s:[%s],overrided masked kbd_layout:[0x%08X]",
+                name, value, settings->kbd_layout);
+    }
+    else if (g_strcmp(name, "neutrinordp.override_kbd_type") == 0)
+    {
+            settings->kbd_type = strtod(value, NULL);
```
matt335672: Suggest `g_atoi()` rather than `strtod()` for this as it's used everywhere else. Or am I missing something? That applies to the following sections too.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -609,6 +610,77 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
...
+            LOG(LOG_LEVEL_WARNING, "%s:[%s],overrided kbd_type:[0x%02X]",
```
matt335672: Suggest 'overrode' instead of 'overrided'. Also applies to following sections.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -609,6 +610,77 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
...
+            }
+            if (settings->kbd_fn_keys == 0)
+            {
+                /* Define the most common number of function keys, 12. 
+                   because we can't get it from client. */
+                settings->kbd_fn_keys = 12;
+            }
+            if (settings->kbd_layout == 0)
+            {
+                settings->kbd_layout = mod->client_info.keylayout;
```
matt335672: I don't understand the need for this mask, so I've had a look around.

I'm still a bit confused, but it seems that some keyboards may need this mask and some may not.

The keyboard layouts are currently [here](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/windows-language-pack-default-values) on the MS website. Or at least some of them are. There's only one Japanese one listed, and it's `0x00000411`. The US ones however are:-

Layout | Identifier
-- | --
United States - English | 0x00000409
United States - Dvorak | 0x00010409
United States - International | 0x00020409
United States-Dvorak for left hand | 0x00030409
United States-Dvorak for right hand | 0x00040409

So now I understand your question about US Dvorak keyboards, although I'm not qualified to answer it.

Is it the case that Japanese keyboards will always need this mask?

matt335672 (commented):
Hi @TOMATO-ONE 

I've got a few comments/questions for you. Hope it's useful.

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thanks.
I'll rewrite it with g_atoi().

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you.
I'll ovverode it.

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you.
I forgot that there could be a case of zero.

I would use the following code to display the warning.
```
if (mod->client_info.keyboard_type == 0)
{
    LOG(LOG_LEVEL_WARNING, "keyboard_type:[0x%02x] ,Set on Server",
           mod->client_info.keyboard_type  );
}
```

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you.

I forgot the case of zero, so the logic is broken....

We wanted to deal with this in a simple way, so as not to add more member variables to the mod structure than necessary.
For this reason, we thought of handling it within the `lxrdp_set_param()` function.
On the other hand, the behavior may change depending on the order in which the parameters are written.

Giving up on processing within the `lxrdp_set_param()` function, the `lfreerdp_pre_connect()` function?

TOMATO-ONE: I have a problem here.

I have tested the behavior with a local build.
The `neutrinordp.allow_client_keyboardLayout` part works fine.
The Japanese keyboard layout was correctly reflected in the remote Windows.

However, this override option may or may not work.
Depending on the keyboard layout that `xrdp_wm` is aware of, it may not work as intended. Sometimes it works.
The override option is I think it is only for debugging use, where someone who has read the source code and understands how it works can change the parameters and see the change in the remote Windows behavior.

remove it from the `xrdp.ini.in` example and
Leave it as an undocumented option for debug that can be used in any order of description.

Or should I just delete the code for the `neutrinordp.overrode_*` option?

Perhaps I could use the `IFDEF` syntax to make it only work with `--enable- devel-all`?

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you for all your research.

Yes, it's true.
That's the complication with Japanese keyboards.

@metalefty  is the leading expert on this issue in xrdp.
Thanks to his efforts over the years, xrdp is now more than usable in a Japanese environment.

@metalefty has published his work in Japanese.
[osc2016-tokyo-spring-xrdp-keylayout. pdf](https://xrdp.vmeta.jp/%E6%B4%BB%E5%8B%95%E5%AE%9F%E7%B8%BE?action=AttachFile&do=view&target=osc2016-tokyo-spring-xrdp-keylayout.pdf)

I have referred to this.

TOMATO-ONE: The reason for masking is as follows.

Japanese input uses an input method called IME.
(ASCII characters are entered with IME turned off, and IME is turned on when entering Japanese multibyte strings.)
The upper 16bit of KyeboadLayout in Japanese environment is set to the value which indicates this IME.
For example, I am using the standard Windows Microsoft IME, and the value of `keyboardLayout` is `0xE0010411`.
`E001` indicates the IME and `0411` means the Japanese keyboard layout.

According to @metalefty's document, if you install additional 3rd party IME, the upper 16bit value will be changed. And the value is not fixed.
This makes it difficult to have a table with all the patterns.

On the other hand, Remote Windows seems to ignore the upper 16 bits and recognize only the lower 16 bits as the Japanese keyboard layout. 
`0xE0010411` and `0x00000411` are both recognized as the Japanese keyboard layout.

If so, there is no need to mask them, is there?

I'm assuming that NeutrinoRDP is the constraint.
https://github.com/neutrinolabs/NeutrinoRDP/blob/f7832d6ce8f12f4b999a2a4edc6cc72373bc5168/include/freerdp/kbd/layouts.h#L161-L164
there are no definitions except `0xE0010411`.

For this reason, if you have installed a third party IME, we believe that NeutrinoRDP will not work due to an unexpected key layout.

This is why you have to mask the upper 16 bits of keyboardLayout.

TOMATO-ONE: @metalefty  Could you please add more information?
Also, could you please point out if I'm wrong?

neutrinordp/xrdp-neutrinordp.c
matt335672: I think getting it working is the best solution, and I also think it can be got working with a bit of effort.

I think the problem is that you're trying to do all the work while the parameters are being processed in `lxrdp_set_param()`. The overrides need to be moved to after the loop where the parameters are processed. You can do this by setting the keyboard parameters before the connection is made.
*EDIT* I just saw your comment above about `lfreerdp_pre_connect()`. I'm not sure this routine is even called. You can use `lxrdp_connect()` though.

Here's a rough example. Please tell me if I'm not being clear. Also I haven't tested this! I've made some simplifications to your code which may not work well, but hopefully will make the idea clearer.

- In xrdp-neutrinordp.h
```c
/* Values in the ini file used to override keyboard settings */
struct kbd_overrides
{
    int type;
    int subtype;
    int fn_keys;
    int layout;
    int layout_mask;
};
. . .
struct mod
{
. . .
    int allow_client_kbd_settings;
    struct kbd_overrides kbd_overrides; /* neutrinordp.override_kbd_* values */
};
```

- Then in xrdp-neutrinordp.c:-
    - in `lxrdp_set_params()`, just record values. Don't change anything.
    - add a new function called `set_keyboard_overrides()`
    - call `set_keyboard_overrides()` from `lxrdp_connect()` after all the parameters have been read.

So the code in `lxrdp_set_params()` will look a bit like this:-
```c
    else if (g_strcmp(name, "neutrinordp.allow_client_keyboardLayout") == 0)
    {
        /* Keyboard values are stored for later processing */
        mod->allow_client_kbd_settings = g_text2bool(value);
    }
    else if (g_strcmp(name, "neutrinordp.override_keyboardLayout_mask") == 0)
    {
        mod->kbd_overrides.layout_mask = g_atoi(value);
    }
    else if (g_strcmp(name, "neutrinordp.override_kbd_type") == 0)
    {
        mod->kbd_overrides.type = g_atoi(value);
    }
    else if (g_strcmp(name, "neutrinordp.override_kbd_subtype") == 0)
    {
        mod->kbd_overrides.subtype = g_atoi(value);
    }
    else if (g_strcmp(name, "neutrinordp.override_kbd_fn_keys") == 0)
    {
        mod->kbd_overrides.fn_keys = g_atoi(value);
    }
    else if (g_strcmp(name, "neutrinordp.override_kbd_layout") == 0)
    {
        mod->kbd_overrides.layout = g_atoi(value);
    }
```

And the function `set_keyboard_overrides()` will look a bit like this.

```c
set_keyboard_overrides(struct mod *mod)
{
    const struct kbd_overrides *ko = &mod->kbd_overrides;
    rdpSettings *settings  = settings = mod->inst->settings;

    if (mod->allow_client_kbd_settings)
    {
        settings->kbd_type = mod->client_info.keyboard_type;
        settings->kbd_subtype = mod->client_info.keyboard_subtype;
        /* Define the most common number of function keys, 12. 
           because we can't get it from client. */
        settings->kbd_fn_keys = 12;

        settings->kbd_layout = mod->client_info.keylayout;
        /* When the keyboard type is 0x07 (Japanese keyboard), 
           this function masks the upper 16 bits.*/
        if (mod->client_info.keyboard_type == 0x07)
        {
            LOG(LOG_LEVEL_DEBUG,
            "keyboardLayout mask for Japanese PC106/109 keyboard:"
            "[0x%08X]", 0x0000FFFF);
            settings->kbd_layout &= 0x0000FFFF;
        }

        LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy remote keyboard settings, "
                            "kbd_type:[0x%02X], kbd_subtype:[0x%02X], "
                            "kbd_fn_keys:[%02d], kbd_layout:[0x%08X]",
                            settings->kbd_type, settings->kbd_subtype,
                            settings->kbd_fn_keys, settings->kbd_layout);
    }


    if (ko->type != 0)
    {
        LOG(LOG_LEVEL_INFO, "overrode kbd type 0x%02X with 0x%02X",
            settings->kbd_type, ko->type);
        settings->kbd_type = ko->type;
    }

    if (ko->subtype != 0)
    {
        LOG(LOG_LEVEL_INFO, "overrode kbd subtype 0x%X with 0x%X",
            settings->kbd_subtype, ko->subtype);
        settings->kbd_subtype = ko->subtype;
    }

    if (ko->fn_keys != 0)
    {
        LOG(LOG_LEVEL_INFO, "overrode kbd fn keys 0x%X with 0x%X",
            settings->kbd_fn_keys, ko->fn_keys);
        settings->kbd_fn_keys = ko->fn_keys;
    }

    if (ko->layout != 0)
    {
        LOG(LOG_LEVEL_INFO, "overrode kbd layout 0x%X with 0x%X",
            settings->kbd_layout, ko->layout);
        settings->kbd_layout = ko->layout;
    }

    if (ko->layout_mask != 0)
    {
        LOG(LOG_LEVEL_INFO, "Masked kbd layout 0x%X to 0x%X",
            settings->kbd_layout, settings->kbd_layout & ko->layout_mask);
        settings->kbd_layout &= ko->layout_mask;
    }
}
```

Hope that's useful.

TOMATO-ONE: @matt335672 Thank you.
I will code and test it based on your suggestions.

Due to my work, please wait until next week for the result.

neutrinordp/xrdp-neutrinordp.c
metalefty: Regarding the US keylayout value, the lower 16 bits imply the language and the upper 16 bits imply layout variants like QWERTY or Dvorak, etc. 

The Japanese keylayout is a little bit different. Japanese keyboard has a single layout but we Japanese use several kinds of IMEs. Ordinary people only use Microsoft's one but some people install and use the third-party one. I see different upper 16 bits are sent from client to server depending on the active IME.
But regardless of the upper 16 bits value, keyboard layouts are same. 

Layout | Identifier
-- | --
Japanese - Direct input | 0x00000411
Japanese - Microsoft IME | 0x00010411
Japanese - Third-party IME | 0xE02n0411

`n` will be incremented by installed order of third-party IMEs.

The mask is a hack to treat as same with the representative keylayout `0x00000411` if any different upper 16 bits are sent from the client.

> I'm still a bit confused, but it seems that some keyboards may need this mask and some may not.

I'm also confused but actually, some keyboards may need this mask. I couldn't find any official MS documents about the upper 16 bits of keylayout value. I'm still looking for that.

matt335672: Thanks for that - it helps a bit.

In that case, I can't see any reason for now why we shouldn't hard-code the mask for a keyboard type of 7. There will be an override anyway, so should we come across another type which needs a mask (probably 0xffff again) we can add it after client testing.

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: I tried to use `g_atoi()`. This is a proprietary function of xrdp. 
Because it uses `atoi()` inside the function, it seems that it can only support decimal numbers.
https://github.com/neutrinolabs/xrdp/blob/e7fdb45744c2e8548545594042ae804dff9435ca/common/string_calls.c#L386-L395

I would like to make sure that both decimal and hexadecimal numbers can be written here.

Is there a problem with using `strtod()` ?
Would it be better to create and use the following function `g_strtod()` in `string_calls.c` ?
```
 int 
 g_strtod(const char *str) 
 { 
     if (str == 0) 
     { 
         return 0; 
     } 
       return strtod(str, NULL); 
 } 
```

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: Thank you, @matt335672  @metalefty !

@metalefty ,Thank you for supplementing my poor explanation.

As @matt335672  said, I thought hardcoding `keyboardLayout` value as `0x00000411` for Japanese keyboard (`kbd_type:0x07`) would make my intention clearer.
Once I've done that, I'm coding with that policy.

neutrinordp/xrdp-neutrinordp.c
matt335672: `strtod()` parses floating-point numbers, so I don't think it's what you want here. Callers will expect it to return a `double` which it doesn't. Also, callers may expect to be able to pass in the 2nd argument to `strtod()`.

A couple of options for you:-
1) Have `g_strtod()` return a `double` and also allow the second argument:-
   ```c
   double
   g_strtod(const char *str, char **endptr) 
   {
       if (str == 0) 
       { 
           str = "0";
       } 
       return strtod(str, endptr); 
   } 
   ```
   Call with

   ```c
    /* Use g_strtod() to allow for hexadecimal as well as decimal */
   settings->kbd_type = g_strtod(value, NULL);
   ```

2) Do something similar with `strtol()` instead. This allows for octal numbers as well.

3) Write a special `g_atoix()` in string_calls.c which parses decimal or hexadecimal integers only. Here's an example:-

   ```c
   /* As g_atoi() but allows for hexadecimal too */
   int g_atoix(const char *str)
   {
       int base = 10;
       if (str == NULL)
       {
           str = "0";
       }

       while (isspace(*str))
       {
           ++str;
       }

       if (*str == '0' && tolower(*(str + 1)) == 'x')
       {
           str += 2;
           base = 16;
       }
       return strtol(str, NULL, base);
   }
   ```

Hope that makes sense.

TOMATO-ONE: Thank you, @matt335672 , for all your kind suggestions.

I think I will adopt your third suggestion.

TOMATO-ONE:
I rewrote the code.
Based on @matt335672 's suggestion.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -94,10 +94,107 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
...
+        }
+        else if (mod->client_info.keyboard_type == 0x04)
+        {
+            /* When the keyboard type is 0x04 (PC101/102 keyboard) and the lower 16
+               bits are a value of 0x0411(Japanese KeyLayout), the kbd_layout is set
+               to 0x00000411 no matter what the upper 16 bits are. */
+            if ((settings->kbd_layout & 0x0000FFFF) == 0x00000411)
+            {
+                settings->kbd_layout = 0x00000411;
+            }
```
TOMATO-ONE: Here is a supplement about this code.

I also tested it on Japanese Windows10 with an English keyboard(`kbd_type:0x04`) connected.
For some reason, the `client_info.keyLayout` was `0xE0010411`.

I'm confused.

I've added some hardcoding to make sure that `0x00000411` can still be sent to remote Windows in this case.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -94,10 +94,107 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
...
+        }
+        else if (mod->client_info.keyboard_type == 0x07)
+        {
+            /* When the keyboard type is 0x07 (Japanese keyboard) and the lower 16
+               bits are a value of 0x0411(Japanese KeyLayout), the kbd_layout is set
+               to 0x00000411 no matter what the upper 16 bits are. */
+            if ((settings->kbd_layout & 0x0000FFFF) == 0x00000411)
+            {
+                settings->kbd_layout = 0x00000411;
+            }
```
TOMATO-ONE: I realized that `kbd_type: 0x07` could also result in a `keyLayout` other than `0xnnnn0411`.

So I decided not to simply set `kbd_layout = 0x00000411` for `kbd_type: 0x07`.
Instead, I set `0x00000411` only when the lower 16 bits of `keyboardLayout` are `0x0411`.

matt335672: Hi @TOMATO-ONE 

Thanks for those changes.

I don't feel qualified to comment on the Japanese keyboard changes. Not only do I not have suitable keyboard(s) available, but I wouldn't be able to use them even if I did.

@metalefty - how does this look to you?

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -94,10 +94,107 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
 
 /******************************************************************************/
 /* return error */
+static void
+set_keyboard_overrides(struct mod *mod)
+{
+    const struct kbd_overrides *ko = &mod->kbd_overrides;
+    rdpSettings *settings  = settings = mod->inst->settings;
```
matt335672: You'll have seen that the CI code check is complaining about this line. It should be :-

```
rdpSettings *settings  = mod->inst->settings;
```

Looking back at the conversation, I can see that I suggested it in the first place. Apologies.

common/string_calls.h
```diff
@@ -154,6 +154,7 @@ int      g_strncmp_d(const char *c1, const char *c2, const char delim, int len);
 int      g_strcasecmp(const char *c1, const char *c2);
 int      g_strncasecmp(const char *c1, const char *c2, int len);
 int      g_atoi(const char *str);
+int      g_atoix(const char *str);
```
matt335672, line 166: Suggest adding a Doxygen-compatible function comment here rather than in `string_calls.c`. Here's an example. Feel free to change it:-

```c
/**
 * As `g_atoi()`, but also allows hexadecimal numbers
 *
 * Prefix hexadecimal numbers with '0x' 
 * 
 * @param str String to convert to an integer
 * @return Integer representation of string
 */
int      g_atoix(const char *str);
```

I'm hopeful at some stage we'll start to use Doxygen to read the source code and give us some useful documentation.

matt335672 (commented):
Starting to look OK to me. Do you want to remove the 'Draft' label?

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: @metalefty  I would like to ask you to do the same.
The `keyLayout`==`0xnnnn0411` at `kbd_Type`==`0x04` causes a keyboard layout mismatch problem in xrdp itself.
I would like to address only the NeutrinoRDP Proxy module issue in this PR, and consider the xrdp side in another PR.

neutrinordp/xrdp-neutrinordp.c
TOMATO-ONE: It's not your fault.
I had pasted too easily.
I should have considered it carefully.

I will correct it.

common/string_calls.h
TOMATO-ONE: thanks, @matt335672 
I'll add a Doxygen-compatible function comment.

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -132,7 +132,7 @@ set_keyboard_overrides(struct mod *mod)
             /* When the keyboard type is 0x07 (Japanese keyboard) and the lower 16
                bits are a value of 0x0411(Japanese KeyLayout), the kbd_layout is set
                to 0x00000411 no matter what the upper 16 bits are. */
-            if ((settings->kbd_layout & 0x0000FFFF) == 0x00000411)
+            if ((mod->client_info.keylayout & 0x0000FFFF) == 0x00000411)
```
TOMATO-ONE: I have changed the structure member variables to be checked, to make the intention clearer.

TOMATO-ONE:
I have removed the 'draft' label.
Please review.

Best regards

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -94,10 +94,107 @@ lxrdp_start(struct mod *mod, int w, int h, int bpp)
...
+        else if (mod->client_info.keyboard_type == 0x04)
+        {
+            /* When the keyboard type is 0x04 (PC101/102 keyboard) and the lower 16
+               bits are a value of 0x0411(Japanese KeyLayout), the kbd_layout is set
+               to 0x00000411 no matter what the upper 16 bits are. */
+            if ((mod->client_info.keylayout & 0x0000FFFF) == 0x00000411)
+            {
+                LOG(LOG_LEVEL_INFO, "overrode kbd_layout 0x%08X with 0x%08X"
+                    " for Japanese environment", settings->kbd_layout,
+                    0x00000411);
```
TOMATO-ONE: Added logging when kbd_layout is overridden.
The same applies to the following IF statement

neutrinordp/xrdp-neutrinordp.c
```diff
@@ -610,6 +708,31 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
             mod->perf_settings_values_mask |= PERF_DISABLE_CURSOR_SHADOW;
         }
     }
+    else if (g_strcmp(name, "neutrinordp.allow_client_keyboardLayout") == 0)
+    {
+        /* Keyboard values are stored for later processing */
+        mod->allow_client_kbd_settings = g_text2bool(value);
+    }
+    else if (g_strcmp(name, "neutrinordp.overrode_keyboardLayout_mask") == 0)
```
matt335672: Again, I suspect I wasn't being clear in a previous comment.

For the values in xrdp.ini, use `override` instead of `overrode`.

`Override` uses the English present tense, whereas 'overrode' is the past tense.

It's more usual to use the present tense for these settings.

matt335672 (commented):
Hi @TOMATO-ONE

Just the one comment below. Thanks for the work you're putting into this.

@metalefty - can you comment on the keyboard usage here? Are you happy with it?

metalefty:
No, we had a discussion in our native language but I don't think the `set_keyboard_overrides` is necessary. 

He thinks that it is not easy switching `keylayout` on the Windows client-side using mstsc.exe. This is the reason he's trying to implement the feature to override the client's keylayout. But actually, it is quite easy so I don't think the feature is necessary.


If we choose English here, `0x00000409` will be sent to server. 
image

matt335672:
Thanks for that, @metalefty 

@TOMATO-ONE - I was unaware of the Windows language selection widget which @metalefty has pointed out above. Although I don't understand Japanese, I've installed a Dutch language pack (I can speak Dutch), and now I can set English with a UK keyboard, or Dutch with a US keyboard, or Dutch with a Dutch keyboard:-

kbd

Is there anything in this dialog box which can't be done without the mods above, or am I misunderstanding the problem?

TOMATO-ONE:
@metalefty 
`set_keyboard_overrides` itself is necessary.
Without it, we can't achieve the goal of this PR.

We are only [discussing](https://groups.google.com/g/xrdp-jp/c/pYgcAiAYl_g) the following parts for US keyboards in Japanese.
```
   else if (mod->client_info.keyboard_type == 0x04)
        {
            /* When the keyboard type is 0x04 (PC101/102 keyboard) and the lower 16
               bits are a value of 0x0411(Japanese KeyLayout), the kbd_layout is set
               to 0x00000411 no matter what the upper 16 bits are. */
            if ((mod->client_info.keylayout & 0x0000FFFF) == 0x00000411)
            {
                LOG(LOG_LEVEL_INFO, "overrode kbd_layout 0x%08X with 0x%08X"
                    " for Japanese environment", settings->kbd_layout,
                    0x00000411);
                settings->kbd_layout = 0x00000411;
            }
            LOG(LOG_LEVEL_INFO,
                "keyboardLayout for PC101/102 keyboard:"
                "[0x%08X]", settings->kbd_layout);
        }
```
Please review again, including the other parts.

TOMATO-ONE:
@matt335672 Thank you for everything.

It's true that we are discussing in Japanese.
We are discussing on GoogleGroups of **[The "xrdp" Users' Group Japan](https://groups.google.com/g/xrdp-jp)**.

I'm sorry that my way of discussing confused him.

We will discuss more.

metalefty:
Guys, I think we're mixing two or three separate issues. Let me know if I'm misunderstanding something. 

First one: As far as I understand, NeutrinoRDP module is not sending keyboard info that appeared in [TS_INPUT_CAPABILITYSET PDU](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/b3bc76ae-9ee5-454f-b197-ede845ca69cc) to another server. This should be fixed. NeutrinoRDP module should pass such information received from the client to another server as-is.  Let's focus on this first.

Second one: Sometimes users might want to ignore the client's keyboard information. Allow xrdp to send fixed keyboard information to another server. 

The last one: Japanese hack.

So what about separating PRs into two or three? It would make it easier to review.

TOMATO-ONE:
@metalefty 
I agree with the division of PR#1934.
Since parts 1) and 2) have been reviewed to some extent by @matt335672
#1934 and leave it at that.

The special support for Japanese  hack has been dropped from #1934.
I would like to propose this as another PR.

matt335672:
That sounds like a good idea to me, particularly as I can't contribute much to the Japanese discussion. Thanks both.

TOMATO-ONE:
I take full responsibility for this slight confusion.
I was trying to address multiple issues in one PR.
I will endeavour to keep the PR simple to help reviewers understand.

TOMATO-ONE:
I have removed the Japanese keyboard hack from code so far.
I will propose the removed code in another PR.

matt335672:
@TOMATO-ONE 

Thanks for that.

I still don't think we're fully agreed on what should be in this PR. You have included allowing keyboard overrides in `xrdp.ini`, but if I'm understanding @metalefty correctly, he doesn't think this should be necessary.

Am I understanding you correctly @metalefty?

If so, can we reach agreement on this before we proceed?

Thanks both.

metalefty:
Can you separate the commits into at least 2,  passing keyboard layout and keyboard overrides?
Then I'm fine with doing both in this PR.

TOMATO-ONE:
@metalefty 
Thank you.
I wanted to respect the arguments made so far.
Thank you for your advice.

I will try to separate the commit into two.

metalefty:
ありがとうございます。なんども手間をとらせてすみません。

TOMATO-ONE:
#e0a482f 
This code is intended to

>1st one: NeutrinoRDP module pass such information received from the client to another server as-is.

TOMATO-ONE:
#e3ffad7
This code is intended to
>2nd one: Allow xrdp to send fixed keyboard information to another server.

matt335672:
Thanks again both! Apologies for causing some confusion earlier.

TOMATO-ONE:
@metalefty @matt335672 
I'm sorry for the confusion.
Thank you!

matt335672:
@TOMATO-ONE - I had a small issue with your last commit in that it didn't compile when merged with the devel version.

Apologies for not catching it sooner.

Could you re-test against the latest devel branch please and make sure everything works for your use-case?

Thanks.

TOMATO-ONE:
@matt335672 
Sorry
I made a mistake with the patch when separating the code.
Tested and no problem.

Thanks again for your help.

matt335672:
Great - thanks. I really should have spotted the problem with the patch, but it's all worked out OK.

Thanks again for your contributions.

PR #1943, opened 2021-07-14
Remove tcutils

matt335672:
Resulting from discussions in #1938

I can widen the scope of this if people think it's a good idea.


aquesnel (approved):

metalefty (approved):

jsorg71:
Thanks, this indeed should be removed.  I'll create a separate project for this if needed.  vrplayer is another but that can be another PR.

Issue #909, opened 2017-10-13
sesman should return auth result/reason to xrdp


metalefty:
Currently, xrdp can know if sesman could create a new session or find an existing session but cannot know the reason  Due to this limitation,  #642 will occur. Returning the result or reason to xrdp is very good to provide the exact reason of failure to users. It also helps to implement IP ban (fail2ban, blacklistd).

Maybe the following reasons are required at least.

* authentication failed
* user password expired
* exhausted sessions count
* user not in the `TerminalServerUsers` group
* failed to execute backend 
  * ex. Xvnc, X11rdp, Xorg executable doesn't exist

PR #1947, opened 2021-07-19
Log a message for failed logins

### LINKED ISSUE #1724: Log of failed logins
In the file scp_v0.c, the allowed logins are logged with the code  

`log_message(LOG_LEVEL_INFO, "++ created session (access granted): username %s, ip %s", s->username, s->client_ip);
`

and the username and client_ip are logged.

It is possible the log the failed logins? I think it will be useful.

Regards,
Rafel Amer
Technic University of Catalonia

### LINKED ISSUE #1946: /var/log/xrdp.log has no failed login such as ssh failed log.
Hi,

Now I am using xrdp in public network for desktop remote. But for security, I need to know who try to login my xrdp and to forbiden it. Now I am using fail2ban to ban these ip; But when I check the xrdp log in /var/log/xrdp.log, it just looks like this:

```
[20210719-14:17:51] [INFO ] Socket 12: AF_INET6 connection received from ::ffff:80.66.76.27 port 33250
[20210719-14:17:51] [INFO ] Using default X.509 certificate: /etc/xrdp/cert.pem
[20210719-14:17:51] [INFO ] Using default X.509 key file: /etc/xrdp/key.pem
[20210719-14:17:51] [ERROR] SSL_accept: I/O error
[20210719-14:17:51] [ERROR] trans_set_tls_mode: ssl_tls_accept failed
[20210719-14:17:51] [ERROR] xrdp_sec_incoming: trans_set_tls_mode failed
[20210719-14:17:51] [ERROR] xrdp_rdp_incoming: xrdp_sec_incoming failed
[20210719-14:17:51] [ERROR] xrdp_process_main_loop: libxrdp_process_incoming failed
[20210719-14:17:51] [ERROR] SSL_shutdown: Failure in SSL library (protocol error?)
```

It's so complex to check if this ip failed to login for what reason.

Here is ssh login log:

```
Failed password for invalid user backups from 176.111.173.156 port 9292 ssh2
```

Can I obtain the failed log in my xrdp?

matt335672:
Fixes #1724 and #1946

Apologies for dropping this one. I was hoping to pick it up as a rework of the SCP code, but this is a much larger job than I first realised.

The following message is now logged as in sesman.log for unrecognised users or invalid passwords:-

```
Username or password error for user: <user>
```

This is very much a quick fix. If preferred. I can add more detailed logging to the PAM code, or I could pull the existing PAM detail from `xrdp/xrdp_mm.c` into some sort of library module and use it to generate extra info in sesman.

Thoughts?

### DIFF
--- a/sesman/scp_v0.c
+++ b/sesman/scp_v0.c
@@ -157,6 +157,8 @@ scp_v0_process(struct SCP_CONNECTION *c, struct SCP_SESSION *s)
     }
     else
     {
+        LOG(LOG_LEVEL_INFO, "Username or password error for user: %s",
+            s->username);
         scp_v0s_deny_connection(c);
     }
     if (do_auth_end)


metalefty:
Maybe related to #909.

matt335672:
Also #1921 on the discussions page.

I think the approach should be:-
- provide detailed information in sesman log
- more information to users via the xrdp login dialog, but not enough to compromise security.

We've currently got some PAM code in `xrdp_mm.c` which is used for logging in gateway situations (i.e. using `pamusername=` and `pampassword=` in `xrdp.ini`). That code could be moved to sesman, so we can get more detailed info in the sesman log. sesman can then return a simpler code to xrdp (and the user) for display purposes.

Does that sound reasonable?

metalefty:
Yes, it looks reasonable to me. Thanks!

Also, the PR is definitely simple, and there's no reason to object.

metalefty (approved):

MikoyChinese:
Why not append the client ip to log? `s->client_ip`, It will be useful to ban ip from attack.

```
    {
        LOG(LOG_LEVEL_INFO, "Username or password error for user: %s from %s",
            s->username, s->client_ip);
        scp_v0s_deny_connection(c);
    }
```

matt335672:
I did think about this, and was very much in two minds.

On the positive side, it's useful information for log scraping utilities.
On the negative side, this string may well change in the future, so I didn't want to introduce user dependencies on it.

The string is created by a call to `g_write_ip_address()` from the xrdp daemon and then sent to sesman. I'm hoping to rework that part of the protocol between xrdp and sesman to just send the actual connection information which will then get formatted by sesman. At the moment sesman has to jump through some rather ugly hoops to parse the data from xrdp.

What's your take on it? I can see you being rather annoyed if you're provided information which could change between releases. On the other hand you may prefer that to nothing at all. If you've got a strong preference to have the info I'll happily add it in.

timriker:
On servers with public IPs, it would be a very good idea to setup things like fail2ban that recognize failed logins and then ban the source IP for a reasonable time. This would need a message like the above in https://github.com/neutrinolabs/xrdp/pull/1947#issuecomment-884767405
that includes the failure state, and the IP. It would be nice to have the user they tried as well to correlate with pam logs, but for fail2ban the date, failure state, and IP are the critical pieces.
Pick a format and stick to it. With the rework, the data can come through different paths, or even be output from different code, but keep the format the same. Something like:
"Login failure for %s from %s"
would be great.

matt335672:
@timriker - could you take a look at #1976 for me, and comment?

Thanks.

timriker:
> @timriker - could you take a look at #1976 for me, and comment?

@matt335672 - commented there. That's looking great.

Issue #1944, opened 2021-07-15
XRDP - 0.9.10 | keyborad layout changes after if the same session


AbinayaSandhiyaM:
Hi Team,

Below are the our environment details

OS - RHEL7
XRDP Version - 0.9.10
Client to connect XRDP server is - RDP [windows 10]

As per our client request we chaned default keyboard layout to German and it been working when we connect the fresh session.

the issue is keyboard is changing to US after reconnect the same session.

could you please let us know the fix/suggestion.

Thanks !

matt335672:
Hi @AbinayaSandhiyaM 

You're running a fairly old version of XRDP there, but I think you can get logging for the keyboard code from `journalctl`

Can you post the output of this command after reconnecting and getting a US keyboard?

```
sudo journalctl -u xrdp -S -5m
```

Thanks.

haridasa-alt:
[root@ulegcprclin tmp]$ sudo journalctl -u xrdp -S -5m
-- No entries --
[root@ulegcprclin tmp]$

haridasa-alt:
We have upgraded the xrdp version to 0.9.16 but observing the same issue

metalefty:
@haridasa-alt You're not the original reporter, who are you? @AbinayaSandhiyaM Are you guys co-workers or something? Or just an unrelated person who has the same issue and is hi-jacking @AbinayaSandhiyaM 's question?

BTW, You should get logs not executing the given example as-is. `-S -5m` means to show the log since 5 minutes ago. It is natural there's no entry if you didn't execute `journalctl` cmd immediately after the issue happens.

haridasa-alt:
Me and @AbinayaSandhiyaM  is co-workers and working in same issue.

 ```
sudo journalctl -u xrdp
Jul 16 17:08:34 ulegcprclin.emea.nsn-net.net systemd[1]: Started xrdp daemon.
Jul 16 17:08:34 ulegcprclin.emea.nsn-net.net xrdp[1945]: (1945)(140003063015488)[INFO ] starting xrdp with pid 1945
Jul 16 17:08:34 ulegcprclin.emea.nsn-net.net xrdp[1945]: (1945)(140003063015488)[INFO ] listening to port 3389 on 0.0.0.0
Jul 16 17:16:20 ulegcprclin.emea.nsn-net.net xrdp[1945]: (1945)(140003063015488)[INFO ] Socket 12: AF_INET connection received from 10.150.44.33 port 51248
[...]

Issue #1576, opened 2020-05-08
Keyboard uses incorrect keymap in russian layout


Mingun:
My setup: I try to connect to my working machine with Ubuntu 18.04 from my laptop on Windows 10 with default windows RDP client mstsc.exe.

**xrdp** from Ubuntu repositories: version **0.9.5**:
```sh
$ xrdp --version

xrdp: A Remote Desktop Protocol server.
Copyright (C) Jay Sorg 2004-2014
See http://www.xrdp.org for more information.
Version 0.9.5
```

I connect via **vnc-any**, which only one allows me to connect to my existing local session

My Ubuntu machine have 2 keyboard layouts: `us` and `ru`.
Laptop keyboard layout:
![HP Pavilion Keyboard](https://cache3.youla.io/files/images/780_780/5d/77/5d77851ca0995dbba375b3b2.jpg)

When I connect via RDP, on russian layout I can't enter some symbols, instead corresponding symbol from US layout is printed. I have no such problem, when I use keyboard, directly connected to my Ubuntu machine. These symbols are:
|I try to type|I get|
|-------------|----|
|`ж`|`;`|
|`б`|`,`|
|`ю`|`.`|
|`.`|`/`|
|`Ж`|`:`|
|`Э`|`"`|
|`Б`|`/`|
|`,`|`?`|

I googled the problem, almost everywhere recommend run `xrdp-genkeymap /etc/xrdp/km-00000419.ini`. I tried this, but without a visible effect, although the file generated by the utility is different from the one delivered with the xrdp package (file extension renamed to `.txt`, because GitHub do not support files with `.ini` extension):
- file from package: [km-00000419.ini.txt](https://github.com/neutrinolabs/xrdp/files/4599785/km-00000419.ini.txt)
- generated by `xrdp-genkeymap /etc/xrdp/km-00000419.ini` (current layout do not matter): 
[xrdp-genkeymap.ini.txt](https://github.com/neutrinolabs/xrdp/files/4599789/xrdp-genkeymap.ini.txt)
- generated by `xrdp-genkeymap`, but after `setxkbmap -layout 'ru,us'` (current layout do not matter): [xrdp-genkeymap-setxkbmap.ini.txt](https://github.com/neutrinolabs/xrdp/files/4599791/xrdp-genkeymap-setxkbmap.ini.txt)

On another site, someone advised me to run `xrdp-genkeymap` after running the `setxkbmap` command, so I run
```sh
[...]

PR #1952, opened 2021-07-23
override keyboard infomation #1950 (1st)

TOMATO-ONE:
This is a PR for Discussions #1950.
In this discussion we are considering several issues.

This is the first one.

add an option to force the RDP Client to overwrite any keyboard information sent to it with the specified value for debugging purposes.

I welcome suggestions for corrections to redundant code or confusing comments.

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -33,6 +33,14 @@ struct monitor_info
     int is_primary;
 };
 
+/* xrdp keyboard overrids */
+struct xrdp_keyboard_overrides
+{
+    int type;
+    int subtype;
+    int layout;
+};
+
 /**
  * Information about the xrdp client
  *
@@ -172,9 +180,12 @@ struct xrdp_client_info
 
     int enable_token_login;
     char domain_user_separator[16];
+
+    /* xrdp.override_* values */
+    struct xrdp_keyboard_overrides xrdp_keyboard_overrides;
 };
 
 /* yyyymmdd of last incompatible change to xrdp_client_info */
-#define CLIENT_INFO_CURRENT_VERSION 20210225
+#define CLIENT_INFO_CURRENT_VERSION 20210723
 
 #endif
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -383,20 +383,20 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
         client_info->keyboard_type, client_info->keyboard_subtype,
         client_info->keylayout);
 
-    if (ko->type != 0)
+    if (ko->type != -1)
     {
         LOG(LOG_LEVEL_INFO, "overrode keyboard_type 0x%02X"
             " with 0x%02X", client_info->keyboard_type, ko->type);
         client_info->keyboard_type = ko->type;
     }
-    if (ko->subtype != 0)
+    if (ko->subtype != -1)
     {
         LOG(LOG_LEVEL_INFO, "overrode keyboard_subtype 0x%02X"
             " with 0x%02X", client_info->keyboard_subtype,
             ko->subtype);
         client_info->keyboard_subtype = ko->subtype;
     }
-    if (ko->layout != 0)
+    if (ko->layout != -1)
     {
         LOG(LOG_LEVEL_INFO, "overrode keylayout 0x%08X"
             " with 0x%08X", client_info->keylayout, ko->layout);
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -374,9 +374,34 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
     char keyboard_cfg_file[256] = { 0 };
     char rdp_layout[256] = { 0 };
 
-    LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: keyboard_type [%d] keyboard_subtype [%d]",
-        client_info->keyboard_type, client_info->keyboard_subtype);
+    const struct xrdp_keyboard_overrides *ko =
+            &client_info->xrdp_keyboard_overrides;
 
+    LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: Keyboard information sent"
+        " by the RDP client, keyboard_type:[0x%02X], keyboard_subtype:[0x%02X],"
+        " keylayout:[0x%08X]",
+        client_info->keyboard_type, client_info->keyboard_subtype,
+        client_info->keylayout);
+
+    if (ko->type != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "overrode keyboard_type 0x%02X"
+            " with 0x%02X", client_info->keyboard_type, ko->type);
+        client_info->keyboard_type = ko->type;
+    }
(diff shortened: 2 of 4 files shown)


TOMATO-ONE:
Sorry.
It should have been a draft PR, but it turned out to be a normal PR...

libxrdp/xrdp_sec.c
```diff
@@ -374,9 +374,42 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
     char keyboard_cfg_file[256] = { 0 };
     char rdp_layout[256] = { 0 };
 
-    LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: keyboard_type [%d] keyboard_subtype [%d]",
-        client_info->keyboard_type, client_info->keyboard_subtype);
+    const struct xrdp_keyboard_overrides *ko =
+            &client_info->xrdp_keyboard_overrides;
 
+    LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: Keyboard information sent"
+        " by the RDP client, keyboard_type:[0x%02X], keyboard_subtype:[0x%02X],"
```
TOMATO-ONE, line 381: If there is a clearer and more appropriate wording, please suggest it.

matt335672: Looks fine to me.

common/xrdp_client_info.h
```diff
@@ -172,9 +183,12 @@ struct xrdp_client_info
 
     int enable_token_login;
     char domain_user_separator[16];
+
+    /* xrdp.overrode_* values */
```
matt335672: Small point. `xrdp.override_*` rather than `xrdp.overrode_*`.

libxrdp/xrdp_sec.c
```diff
@@ -374,9 +374,47 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
...
+        client_info->keylayout = ko->layout;
+    }
+    if ((ko->type | ko->subtype | ko->layout) != 0)
+    {
+        LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: Overrode keyboard"
+            " infomation, keyboard_type:[0x%02X], keyboard_subtype:[0x%02X],"
+            " keylayout:[0x%08X]",
+            client_info->keyboard_type, client_info->keyboard_subtype,
+            client_info->keylayout);
+    }
```
matt335672: Suggest removing 410-417 entirely, and making the preceding messages `LOG_LEVEL_INFO`. You've already logged the client info at line 380-384. The override messages should also be at `LOG_LEVEL_INFO` level so we get a complete picture of the keyboard layout variables when running at `LOG_LEVEL_INFO`.

Otherwise we're just duplicating information when running at `LOG_LEVEL_DEBUG`.

common/xrdp_client_info.h
```diff
@@ -33,6 +33,17 @@ struct monitor_info
...
 
+/* xrdp keyboard overrids */
+struct xrdp_keyboard_overrides
+{
+    int type;
+    int subtype;
+    int layout;
+    int original_type;
+    int original_subtype;
+    int original_layout;
```
matt335672: I don't think the `original_*` values are needed(?)

common/xrdp_client_info.h
```diff
@@ -172,9 +183,12 @@ struct xrdp_client_info
...
     int enable_token_login;
     char domain_user_separator[16];
+
+    /* xrdp.overrode_* values */
+    struct xrdp_keyboard_overrides xrdp_keyboard_overrides;
 };
 
 /* yyyymmdd of last incompatible change to xrdp_client_info */
-#define CLIENT_INFO_CURRENT_VERSION 20210225
+#define CLIENT_INFO_CURRENT_VERSION 20210723
```
matt335672, line 189: Good catch.

matt335672 (commented):
Hi @TOMATO-ONE. Thanks for this. Sorry it took a while to get to it.

common/xrdp_client_info.h
TOMATO-ONE: It was a typo. Sorry.
I will correct it.

libxrdp/xrdp_sec.c
TOMATO-ONE: Thank you.
I will leave this text as it is.

libxrdp/xrdp_sec.c
TOMATO-ONE: Thank you.
I was adding this for future code additions.
I've changed my mind and will remove lines 410-417.

I will change the override message to LOG_LEVEL_INFO.

common/xrdp_client_info.h
TOMATO-ONE: I had added it in case the original client information was needed after the override.
This is not needed for this PR and will be removed.

matt335672 (approved):
@TOMATO-ONE - thank you. That looks fine to me now.

matt335672:
@metalefty - when you get a moment can you look at this? It looks like it should be useful for debugging keyboard issues.

One thing which occurs to me is that tf we merge it for 0.9.17 we'll need to update xorgxrdp too, as the client_info has changed. At the moment that isn't necessary.

libxrdp/xrdp_sec.c
```diff
@@ -374,9 +374,34 @@ xrdp_load_keyboard_layout(struct xrdp_client_info *client_info)
...
+    const struct xrdp_keyboard_overrides *ko =
+            &client_info->xrdp_keyboard_overrides;
 
+    LOG(LOG_LEVEL_INFO, "xrdp_load_keyboard_layout: Keyboard information sent"
+        " by the RDP client, keyboard_type:[0x%02X], keyboard_subtype:[0x%02X],"
+        " keylayout:[0x%08X]",
+        client_info->keyboard_type, client_info->keyboard_subtype,
+        client_info->keylayout);
+
+    if (ko->type != 0)
```
TOMATO-ONE: I have found a bug in this part.

In xrdp.ini, put the following.
`xrdp.override_keyboard_type.type = 0x00`

In this case, for this if statement
It will not override `client_info->keyboard_type` .

The same is true for the next two.

TOMATO-ONE:
I set the initial value of the variable `client_info->xrdp_keyboard_overrides.*`  to `-1`.

Please advise if there is a better way.

matt335672:
I think that's a good a way as any other.

libxrdp/xrdp_sec.c
metalefty: keylayout value is not logged here. Is this enough?

metalefty: No, it is logged. Sorry for the noise.

metalefty:
LGTM. Needs to be triaged whether to include it in 0.9.17 or not.

TOMATO-ONE:
@matt335672 @metalefty 
Thank you for being so busy with the release!

This PR is part of the discussion #1950.
So I understand that this PR will be triaged.
On the other hand, I think this PR could stand alone to help resolve issues like issue #1944 and #1576.

Either way, I respect the thoughts of the team.

matt335672:
I'm happy for this to go in to 0.9.17. As it stands, it doesn't cause any changes in behaviour that I can determine, so I see it as very low risk.

PR #1954, opened 2021-07-27
clipboard: log file transfer for the purpose of audit

metalefty:
Some customers want to record who took out which file from the xrdp server to the client via clipboard for the purpose of audit.

Note: chansrv logs are replaced with an empty file every time when chansrv is restarted on the same display number.  Syslog should be used instead to make the log non-volatile.

### DIFF
--- a/sesman/chansrv/clipboard_file.c
+++ b/sesman/chansrv/clipboard_file.c
@@ -445,6 +445,10 @@ clipboard_send_file_data(int streamId, int lindex,
     rv = send_channel_data(g_cliprdr_chan_id, s->data, size);
     free_stream(s);
     g_file_close(fd);
+
+    /* Log who transferred which file via clipboard for the purpose of audit */
+    LOG(LOG_LEVEL_INFO, "S2C: Transfered a file: filename=%s, uid=%d", full_fn, g_getuid());
+
     return rv;
 }
 


sesman/chansrv/clipboard_file.c
```diff
@@ -445,6 +445,10 @@ clipboard_send_file_data(int streamId, int lindex,
     rv = send_channel_data(g_cliprdr_chan_id, s->data, size);
     free_stream(s);
     g_file_close(fd);
+
+    /* Log who transferred which file via clipboard for the purpose of audit */
+    LOG(LOG_LEVEL_INFO, "Transfered a file: filename=%s, uid=%d", full_fn, g_getuid());
```
matt335672: Rather than just saying the file is transferred, is it better to mention the direction as well, so there's no ambiguity in which way the transfer went. For Example:-

```c
"uid=%d downloaded file %s via clipboard", g_getuid(), full_fn
```

metalefty: Indeed.

danielperna84:
Could this also be extended (maybe with an option that can be turned on and off) to also cover transferred strings (not the content, just the event)? In terms of auditing it could be interesting if the user has transferred a password into the session.

And while at the topic of logging: has it been discussed before to use a standardized format that's easy to parse by tools like Graylog? Here's a list of a few formats Graylog supports: https://www.graylog.org/post/log-formats-a-complete-guide

I'd open an issue for that if that's of interest.

metalefty:
> Could this also be extended (maybe with an option that can be turned on and off) to also cover transferred strings (not the content, just the event)? In terms of auditing it could be interesting if the user has transferred a password into the session.

Yes, but in another PR.

> And while at the topic of logging: has it been discussed before to use a standardized format that's easy to parse by tools like Graylog? Here's a list of a few formats Graylog supports: https://www.graylog.org/post/log-formats-a-complete-guide

 We've never discussed Graylog but have discussed logging improvements. Thanks for letting us know another option.

sesman/chansrv/clipboard_file.c
```diff
@@ -445,6 +445,10 @@ clipboard_send_file_data(int streamId, int lindex,
     rv = send_channel_data(g_cliprdr_chan_id, s->data, size);
     free_stream(s);
     g_file_close(fd);
+
+    /* Log who transferred which file via clipboard for the purpose of audit */
+    LOG(LOG_LEVEL_INFO, "S2C: Transfered a file: filename=%s, uid=%d", full_fn, g_getuid());
```
metalefty, line 450: Added file transfer direction. S2C means server to client.

metalefty:
I would like to ship this to the next release.

PR #1962, opened 2021-08-05
Add imlib2 support for login screen customisation

### LINKED ISSUE #1819: Custom banner bitmap must be multiple of 4 pixels wide
Custom bitmap is rendered incorrectly.

_Bitmap was created in Microsoft Paint (standard 24bit bmp)._

Bitmap and the result:
<a href="https://imgbb.com/">banner1</a>
<a href="https://ibb.co/q9gcbpr">bitmapinc</a>

### LINKED ISSUE #1929: Feature-Request: PNG support for background images and logos
We have a customised version of XRDP where a previous employee enabled png support for images. Also enabled is the ability to scale center the background image. Compared to bitmap 24 the file size of 30MB is reduced to just 5MB. 

Unfortunately we no longer have the knowledge or skills in our team to customise the latest releases to include this and think that smaller images with the support for transparency would be beneficial to others. I believe @matt335672 may have some knowledge on how this can be achieved already.

matt335672:
Fixes #1929 #1819

See also discussion #1931 

This PR allows for more customization of the xrdp login screen. An example customised screen from an Ubuntu 20.04 VM is:-

Screenshot_2021-08-05_12-24-10

This is achieved by adding the following parameters to `xrdp.ini`:-

```ini
ls_background_image=/usr/share/backgrounds/ryan-stone-skykomish-river.jpg
ls_background_transform=zoom

ls_logo_filename=/usr/share/icons/hicolor/256x256/apps/ubuntu-logo-icon.png
ls_logo_transform=zoom
ls_logo_width=140
ls_logo_height=140
ls_logo_x_pos=105
ls_logo_y_pos=50
```

File types other than bmp files are supported by way of the imlib2 library which is available on all our currently supported platforms. Imlib2 support is optional. Without it, we fall back to the existing bmp only code. The existing code has been reworked to (hopefully) make it simpler to understand, and also to fix #1819.

Alpha blending of images onto the background colour behind them is supported - this can be seen on the Ubuntu logo in the screenshot above. On multi monitor setups, the background image is displayed on the primary monitor only.

A test suite has been added for the extended `xrdp_bitmap_load()` call, so that the correct operation of the external library can be checked.

This is a fairly significant PR, and so I suggest we hold off merging it until after the next release.


matt335672:
Just re-read #1931. I've added an extra check so that wallpaper isn't loaded if the client has disabled wallpaper in the performance settings.

metalefty:
> I've added an extra check so that wallpaper isn't loaded if the client has disabled wallpaper in the performance settings.

Awesome.

metalefty:
Overall, LGTM.

configure.ac
```diff
@@ -166,6 +166,8 @@ AC_ARG_ENABLE(rdpsndaudin, AS_HELP_STRING([--enable-rdpsndaudin],
               [], [enable_rdpsndaudin=no])
 AM_CONDITIONAL(XRDP_RDPSNDAUDIN, [test x$enable_rdpsndaudin = xyes])
 
+AC_ARG_WITH(imlib2, AC_HELP_STRING([--with-imlib2=ARG], [imlib2 library to use (yes/no/auto/<path>)]),,)
```
metalefty: It is not clear using imlib2 enables what.

configure.ac
```diff
@@ -216,6 +218,48 @@ AC_CHECK_HEADER([security/_pam_types.h],
...
+        else
+            AC_MSG_RESULT([no])
+            AC_MSG_ERROR([Can't find $with_imlib2/include/Imlib2.h])
+        fi
+        AC_MSG_RESULT([yes])
+        AC_SUBST([IMLIB2_LIBS])
+        AC_SUBST([IMLIB2_CFLAGS])
+        use_imlib2=yes
+esac
+
```
metalefty, line 254: Auto-detection is generally very nice but autodetected libraries will not appear in `xrdp -v`. I want an easy way to tell users if your compiled xrdp supports png/jpg images on the login screen. `ldd /usr/sbin/xrdp` sounds a little bit guru's way.

If we introduce more auto-detection, we might need to invent a new way to show which libraries/features compiled with.

matt335672: That's a really good point - I hadn't thought of that but it's a feature I use a lot when triaging problems. I'll remove the auto-detect.

configure.ac
matt335672: I'll update this string.

matt335672:
`xrdp -v` now works OK for imlib2.

I also noticed I'd not added CI support for i386 legacy. It's a small thing, but it's in now.

metalefty (approved):

metalefty:
> so I suggest we hold off merging it until after the next release. 

I agree.

PR #1965, opened 2021-08-06
sesman/startwm.sh : add support for Alpine Linux

fcolista:
Hello. 
This patch adds support for Alpine Linux in order to correctly find the .xinitrc.
This patch follows up an issue in alpinelinux gitlab : https://gitlab.alpinelinux.org/alpine/aports/-/issues/12896

Credits to https://gitlab.alpinelinux.org/raidenii

Thank you!

.: Francesco

### DIFF
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -90,6 +90,14 @@ wm_start()
     exit 0
   fi
 
+  # alpine
+  if [ -r /etc/X11/xinit/xinitrc ]; then
+    pre_start
+    . /etc/X11/xinit/xinitrc
+    post_start
+    exit 0
+  fi
+
   pre_start
   xterm
   post_start


metalefty:
LGTM.

metalefty (approved):

matt335672:
Hi @fcolista 

Thanks very much for this.

Could I trouble you to move your added stanza from the front of the existing list to the end (e.g. after the SuSE section)?

There's a chance that an existing platform has both `/etc/X11/xinit/xinitrc` and another file in place, and on these platforms your change could introduce a change in existing behaviour. Ubuntu 20.04 has both `/etc/X11/xinit/xinitrc` and `/etc/X11/Xsession`. On that platform I don't think it makes a difference, but I'm not in a position to check every single platform out there.

Apart from that I see no reason why this shouldn't be merged at all.

fcolista:
Hi @matt335672,
done as you requested. Your suggestion makes sense.

Thanks.

.: Francesco

matt335672:
Thanks for your contribution, @fcolista

pelepelin:
@fcolista @matt335672 Sorry for being late for the party, with the new block added to the end it won't be executed on Alpine because Alpine also has `/etc/X11/xinit/Xsession` file (which is unfortunately not suitable for the purpose of `startwm.sh`)

matt335672:
That's a little awkward, for the reasons posted above.

Which apk provides `/etc/X11/xinit/Xsession` on Alpine? I'll have a look at what it does.

pelepelin:
Provided by [xinit](https://pkgs.alpinelinux.org/contents?branch=v3.14&name=xinit&arch=x86_64&repo=community)
Unlike Ubuntu where `/etc/X11/Xsession` finds and executes session manager itself, in Alpine `/etc/X11/xinit/Xsession` just sets up some env and ends with `exec $@` expecting chained executable in args.

matt335672:
Yes - I've just configured an Alpine XFCE desktop so I can have a look at this.

I installed the xrdp apk. The best solution I found is this:-

```diff
--- startwm.sh.old
+++ startwm.sh
@@ -70,6 +70,18 @@
     post_start
     exit 0
   fi
+                                          
+  # Alpine                                                              
+  # Don't use /etc/X11/xinit/Xsession - it doesn't work
+  if [ -f /etc/alpine-release ]; then     
+    if [ -f /etc/X11/xinit/xinitrc ]; then
+        pre_start                         
+        /etc/X11/xinit/xinitrc                          
+        post_start                                                      
+    else                                                
+        echo "** xinit package isn't installed" >&2     
+    fi                                                  
+  fi                                                    
 
   # el
```

I can add this in if you like, and revert the previous patch. If I do so I'll ask you to review it so you can test it.

One other thing;  If we go down this route, you might want to add a couple of dependencies to the `apk`? I've got this:-

```
$ apk info -R xrdp
xrdp-0.9.15-r1 depends on:
/bin/sh
so:libX11.so.6
so:libXfixes.so.3
so:libXrandr.so.2
so:libc.musl-x86_64.so.1
so:libcrypto.so.1.1
so:libfuse.so.2
so:libssl.so.1.1
so:libturbojpeg.so.0
```

However, if used as-is, the script introduces a dependency on bash, and xinit. So if we go ahead with the above, should you not have a `depends="bash xinit"` in your [APKBUILD file](https://git.alpinelinux.org/aports/tree/community/xrdp/APKBUILD)?

matt335672:
@pelepelin, @fcolista - I need some feedback on this. At the moment I suspect we don't want this (merged) patch at all, so it needs to be removed or replaced. I don't know which way you want to go. If I don't hear from you in a week or so, I'll unmerge this so at least we're in a sane state.

I'm happy to work with you on this.

fcolista:
@matt335672 I'm fine with adding xinit to the APKBUILD. I'm not sure on why you want bash. That /bin/sh is busybox ash.

fcolista:
@matt335672  ok, done. The ```xinit``` dependency has been added:  http://dup.pw/alpine/aports/e855ad6c00fd
This commit has done in edge. So if you want to test it in your environment, you should add the following community/edge repository in ```/etc/apk/repositories``` : http://dl-cdn.alpinelinux.org/alpine/edge/community

Thanks.

.: Francesco

matt335672:
@fcolista - the reason we might want to add bash is that startwm.sh as supplied by xrdp [uses bash](https://github.com/neutrinolabs/xrdp/blob/devel/sesman/startwm.sh).

So as I see it, you either need to add bash as a dependency, or provide your own startwm.sh (as Fedora and Ubuntu do).

fcolista:
Ah ok, gotcha. I missed that part.
Now Alpine package has bash as dep:  https://git.alpinelinux.org/aports/commit/?id=c7a048bc510c

Thanks.

.: Francesco

matt335672:
@fcolista - I'm having testing problems at the moment with xorgxrdp:-

```
$ apk list -I \*xrdp
xorgxrdp-0.2.17-r0 x86_64 {xorgxrdp} (X11) [installed]
xrdp-0.9.17-r2 x86_64 {xrdp} (Apache-2.0) [installed]
```
xorgxrdp builds against xrdp, as they share a common file. I'm getting this error in the X server log from the xorgxrdp module:-

```
[   451.543] expected xrdp client_info version 20210225, got 20210723
[   451.543] (EE)                                                           
Fatal server error:                                        
[   451.543] (EE) Incompatible xrdp version detected  - please recompile(EE)
[   451.543] (EE)                             
```

I suspect it's because xorgxrdp 0.2.17-r0 is being built on a machine with an older version of `xrdp-dev` on it. Is that possible?

fcolista:
You should find xorgxrdp-0.2.17-r1 built against new xrdp:

https://git.alpinelinux.org/aports/commit/?id=33a600962b8b

This will took some time before mirrors got syncd after package build.

.: Francesco

matt335672:
Thanks @fcolista - that's working now.

I've raised #2005 to cover this. Feel free to comment!

PR #1975, opened 2021-08-26
Use text strings for security level logging at ISO layer

### LINKED ISSUE #1974: Improve log message: "Security layer: requested 3, selected 1"
Xrdp's log (`0.9.12-1` from _Debian 11_ `stable`) often has entries like this

```
Security layer: requested 0, selected 1

Security layer: requested 3, selected 1
```

It is unclear what it means. We do not need the answer here in this Issue but in the log itself.

Please clarify the log message that it can be understood by the user/admin who reads the log messages.

matt335672:
Fixes #1974 

Can be merged after v0.9.17 (if people are happy with the code, that is).

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -891,3 +891,69 @@ g_strnjoin(char *dest, int dest_len, const char *joiner, const char *src[], int 
 
     return dest;
 }
+
+int
+g_bitmask_to_str(int bitmask, const struct bitmask_string bitdefs[],
+                 char delim, char *buff, int bufflen)
+{
+    int rlen = 0; /* Returned length */
+
+    if (bufflen <= 0) /* Caller error */
+    {
+        rlen = -1;
+    }
+    else
+    {
+        char *p = buff;
+        /* Find the last writeable character in the buffer */
+        const char *last = buff + (bufflen - 1);
+
+        const struct bitmask_string *b;
+
+        for (b = &bitdefs[0] ; b->mask != 0; ++b)
+        {
+            if ((bitmask & b->mask) != 0)
+            {
+                if (p > buff)
+                {
+                    /* Not first item - append separator */
+                    if (p < last)
+                    {
+                        *p++ = delim;
+                    }
+                    ++rlen;
+                }
+
+                int slen = g_strlen(b->str);
+                int copylen = MIN(slen, last - p);
+                g_memcpy(p, b->str, copylen);
+                p += copylen;
+                rlen += slen;
+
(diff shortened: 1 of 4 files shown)


matt335672:
I've had a think about this overnight and it makes sense (to me at least) to move the logic to convert a bitmask to a string to the string calls module for general logging.

Thoughts?

metalefty:
It makes sense to me, too.

matt335672:
I've moved the mask output stuff to the string calls module and added some unit tests.

matt335672:
Example messages:-

```
[20210901-12:12:56] [DEBUG] Security layer: requested RDP, selected RDP
```

```
[20210826-15:29:39] [DEBUG] Security layer: requested SSL|HYBRID|HYBRID_EX, selected SSL
```

metalefty:
At least this solves the original issue.

Thoughts:

1. It would be very nice if also capable protocols on the server-side are logged. 

```
requested=SSL|HYBRID|HYBRID_EX, capable=RDP|SSL, selected=SSL
```

2. We've been using the word "security layer" but it might not be a standard word in MS spec. "requested protocol" or "security protocol" appears in [MS-RDPBCGR 2.2.1.1.1](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/902b090b-9cb3-4efc-92bf-ee13373371e3). Let's switch to such standard words for easy reference.

3. The current log format at least includes necessary information but I think there's some room to pursuit easy-to-read format. Can you think of anything else?

```
[20210826-15:29:39] [DEBUG] Security layer: requested=[SSL|HYBRID|HYBRID_EX] selected=[SSL]
```

matt335672:
Example messages are now as follows. I've bumped the level to INFO, as this may be useful for users:-

```
[20210902-16:01:28] [INFO ] Security protocol: configured [SSL], requested [SSL|HYBRID|HYBRID_EX|RDP], selected [SSL]
```

```
[20210902-16:02:35] [INFO ] Security protocol: configured [RDP], requested [SSL|HYBRID|HYBRID_EX|RDP], selected [RDP]
```

```
[20210902-16:03:16] [INFO ] Security protocol: configured [SSL|RDP], requested [SSL|HYBRID|HYBRID_EX|RDP], selected [SSL]
```

I've also added a warning about the TLS certificate being unreadable if the security level is set to 'negotiate':-

```
[20210902-16:04:04] [WARN ] Cannot accept TLS connections because certificate or private key file is not readable. certificate file: [], private key file: []
```

metalefty (approved):
Much improved and informative for Debian users.

PR #1985, opened 2021-09-03
32-bit CI build process : Remove some 64-bit packages

matt335672:
The 32-bit CI build process is [broken again](https://github.com/neutrinolabs/xrdp/actions/runs/1197593050).

The reason this time appears to be that the libtiff-dev:i386 package conflicts with the libtiff-dev:amd64 package. The exact reason for this isn't clear to me. At the time of writing, both i386 and amd64 packages appear to have the same version, namely `4.0.9-5ubuntu0.4`.

I've tried to come up with a more pro-active solution this time. What I've done, is for i386 builds where `libxxx-dev:i386` is installed, I've removed `libxxx-dev:amd64` beforehand. We shouldn't need any of these packages for a 32-bit build and removing them should greatly decrease the chances of a package version clash if the repos are out-of sync.

### DIFF
--- a/scripts/install_xrdp_build_dependencies_with_apt.sh
+++ b/scripts/install_xrdp_build_dependencies_with_apt.sh
@@ -4,6 +4,39 @@ set -eufx
 FEATURE_SET=min
 ARCH=amd64
 
+# See apt.conf(5)
+DEBUG_OPTS=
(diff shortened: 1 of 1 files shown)


matt335672:
I've merged this PR without review to get the CI working again.

I'm not completely happy with doing this.

If anyone has got any better ideas on improving the 32-bit solution, or can improve my shell scripting please feel free to jump in.

PR #1988, opened 2021-09-06
tests: add test on g_file_get_size()

metalefty:
See also #1987.

### DIFF
--- a/tests/common/test_os_calls.c
+++ b/tests/common/test_os_calls.c
@@ -83,6 +83,8 @@ START_TEST(test_g_file_get_size__just_less_than_2GiB)
 }
 END_TEST
 
+/* skip these tests until g_file_get_size() supports large files*/
+#if 0
 START_TEST(test_g_file_get_size__2GiB)
 {
     const char *file = "./file_2GiB.dat";
@@ -140,6 +142,7 @@ START_TEST(test_g_file_get_size__5GiB)
     ck_assert_int_eq(size, expected_size);
 }
 END_TEST
+#endif
 
 /******************************************************************************/
 Suite *
@@ -155,8 +158,10 @@ make_suite_test_os_calls(void)
     tcase_add_test(tc_os_calls, test_g_file_get_size__returns_file_size);
     tcase_add_test(tc_os_calls, test_g_file_get_size__1GiB);
     tcase_add_test(tc_os_calls, test_g_file_get_size__just_less_than_2GiB);
+#if 0
     tcase_add_test(tc_os_calls, test_g_file_get_size__2GiB);
     tcase_add_test(tc_os_calls, test_g_file_get_size__5GiB);
+#endif
 
     return s;
 }
--- a/tests/common/test_os_calls.c
+++ b/tests/common/test_os_calls.c
@@ -44,7 +44,7 @@ START_TEST(test_g_file_get_size__1GiB)
                "dd if=/dev/zero of=%s bs=%u seek=%u count=1",
                file,
                block_size,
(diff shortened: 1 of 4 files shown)


tests/common/test_os_calls.c
```diff
@@ -0,0 +1,51 @@
...
+    ck_assert_int_eq(size, 19766);
+
+}
+END_TEST
+
+START_TEST(test_g_file_get_size__more_than_4GiB)
+{
+    unsigned long long size;
+
+    system("dd if=/dev/zero of=./file_5GiB.dat bs=4k count=1310720");
```
metalefty: I'm not sure this is the correct way to create files required to perform tests. Let me know if there's an appropriate method.

tests/common/test_os_calls.c
```diff
@@ -0,0 +1,51 @@
...
+
+    size = g_file_get_size("./file_5GiB.dat");
+    ck_assert_int_eq(size, 5368709120);
```
metalefty: This fails as assumed until `g_file_get_size()` gets fixed to be able to handle more than 4GiB files.

tests/common/test_os_calls.c
matt335672: On my system I'm getting:-

```
FAIL: test_common 18 - test_os_calls.c:oscalls-file:test_g_file_get_size__more_than_4GiB: Test timeout expired
```

but, then my test system is using a mirror of hard drives.

Also, you're leaving the file on the system after the test.

Using a sparse file works well on most modern filesystems:-

```c
START_TEST(test_g_file_get_size__more_than_4GiB)
{
    const char *file = "./file_5GiB.dat";
    unsigned int block_size = 4096;
    unsigned int block_count = 1310720;
    /* C90 5.2.4.2.1 guarantees long long is at least 64 bits */
    const long long expected_size =
        (long long)block_size * block_count;
    long long size;

    char cmd[256];

    /* Create a sparse file of the expected size with one block of data
     * at the end */
    g_snprintf(cmd, sizeof(cmd),
               "dd if=/dev/zero of=%s bs=%u seek=%u count=1",
               file,
               block_size,
               block_count -1);

    size = g_file_get_size(file);
    g_file_delete(file);
    ck_assert_int_eq(size, expected_size);
}
END_TEST
```

metalefty: Thanks. My file system creates a sparse file automatically when writing out consecutive zero fills.
Creating sparse files explicitly will work well.

tests/common/test_os_calls.c
```diff
@@ -0,0 +1,162 @@
...
+
+    s = suite_create("OS-Calls");
+
+    tc_os_calls = tcase_create("oscalls-file");
+    suite_add_tcase(s, tc_os_calls);
+    tcase_add_test(tc_os_calls, test_g_file_get_size__returns_file_size);
+    tcase_add_test(tc_os_calls, test_g_file_get_size__1GiB);
+    tcase_add_test(tc_os_calls, test_g_file_get_size__just_less_than_2GiB);
+    //tcase_add_test(tc_os_calls, test_g_file_get_size__2GiB);
+    //tcase_add_test(tc_os_calls, test_g_file_get_size__5GiB);
```
metalefty: These 2 cases are disabled intentionally because it `g_file_get_size()` needs to be fixed to return file size as `off_t`.

metalefty:
@matt335672 I've finished this. I will fix `g_file_get_size()` in another PR.

matt335672:
I think this is a good approach to get the basic tests in. I'll merge it.

PR #1989, opened 2021-09-06
Rename misnamed function get_eds_status_msg

### LINKED ISSUE #1986: Missing rfb_get_eds_status_msg() function in vnc/vnc.c in 0.9.17
This is preventing modern Fedora version from loading libvnc.so, because of the missing symbol in the shared library. I trivially patched that in the RPMs, which I am building now, but the implementation of the function was likely missed in the release.

matt335672:
Fixes #1986

### DIFF
--- a/vnc/rfb.c
+++ b/vnc/rfb.c
@@ -42,7 +42,7 @@ static const char *eds_status_msg[] =
  * Returns an error string for an ExtendedDesktopSize status code
  */
 const char *
-get_eds_status_msg(unsigned int response_code)
+rfb_get_eds_status_msg(unsigned int response_code)
 {
     if (response_code >= EDS_STATUS_MSG_COUNT)
     {


metalefty (approved):

matt335672:
Hi @avolkov-astra 

@bsmojver (from Fedora) was involved in raising this and we've communicated exactly what's going on to him. For v0.9.18 he will simply remove the patch.

This is a particular issue for Fedora as that distro disables lazy binding by default. I'm not aware of any distros apart from Fedora that do this. Is this missing symbol causing you particular trouble for another distro?

avolkov-astra:
Hi,
I made a package of 0.9.17 for AstraLinux and then got error messages from users:
 [ERROR] xrdp_wm_log_msg: error loading libvnc.so specified in xrdp.ini, please add a valid entry like lib=libxrdp-vnc.so or similar

Perhaps someone else might have encountered this error.

matt335672:
It sounds like it might be the same problem. You can find out with this command:-

```
readelf -d /path/to/libvnc.so | grep NOW
```

If the shared library was built with `-z now` (which is what Fedora Hardening does), you'll have `BIND_NOW` in the output.

No-one else has mentioned this, and v0.9.18 will be out end of December which will incorporate this. I'd have though the simplest thing for you to do, if this is indeed your problem, is to incorporate ce27e25dcc2d9a647d41e3609211dca2fe4fb136 directly into your package. It's a simpler fix than the Fedora one above.

avolkov-astra:
Thanks. I fixed it by adding rfb.c (it is missing in the release tarball) and then applying ce27e25dcc2d9a647d41e3609211dca2fe4fb136 and ae45344a45cca5e5f6282cb5bdd9e5e2001e9dcf

Just decided to warn you about the problem, cause it seems to be rather important.

matt335672:
Thanks for the heads up - it's appreciated.

I'm in two minds about this one. I didn't catch it during module testing, as I didn't build with `-z now` by default (I do now!), and it's really unlikely that the user will encounter the missing codepath. I'm also aware that Debian has taken 0.9.17 into testing now (Debian repo is [here](https://salsa.debian.org/debian-remote-team/xrdp)). This may be of interest to you.

avolkov-astra:
Thanks for the tip.
Merge request for Debian: https://salsa.debian.org/debian-remote-team/xrdp/-/merge_requests/5

PR #2005, opened 2021-09-27
Rework alpine support

matt335672:
Following comments raised against #1965, this additional PR has been raised.

This patch was tested against the following Alpine packages:-

```
$ apk list -I \*xrdp
xorgxrdp-0.2.17-r1 x86_64 {xorgxrdp} (X11) [installed]
xrdp-0.9.17-r2 x86_64 {xrdp} (Apache-2.0) [installed]
```

I've added two commits here to make it clear the previous commit has been backed out.

### DIFF
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -90,14 +90,6 @@ wm_start()
     exit 0
   fi
 
-  # alpine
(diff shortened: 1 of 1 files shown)


metalefty:
Let's merge if it is fine for Alpine users.

PR #2014, opened 2021-10-11
Fix definitely lost memory leaks

### LINKED ISSUE #1562: Memory leaks in xrdp
Leaks was shown by valgrind

- [ ] - 
```
==2582== 80 bytes in 1 blocks are still reachable in loss record 1 of 3
==2582==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2582==    by 0x4855C46: internalInitAndAllocStruct (log.c:397)
==2582==    by 0x4855CA8: internalReadConfiguration (log.c:293)
==2582==    by 0x4855D67: log_start (log.c:488)
==2582==    by 0x10CD9D: main (xrdp.c:518)
```

- [ ] - 
```
==2582== 512 bytes in 1 blocks are still reachable in loss record 2 of 3
==2582==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2582==    by 0x4855ADA: internal_config_read_logging (log.c:353)
==2582==    by 0x4855CE4: internalReadConfiguration (log.c:307)
==2582==    by 0x4855D67: log_start (log.c:488)
==2582==    by 0x10CD9D: main (xrdp.c:518)

```
- [ ] - 
```
==2582== 1,064 bytes in 1 blocks are still reachable in loss record 3 of 3
==2582==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==2582==    by 0x10CCBD: main (xrdp.c:441)

```
- [ ] - and leak with ssl_init() which needs to be finished with ssl_finish()(I lost log)

matt335672:
Fixes #1562, although it seems a lot of this has been done already.

Remove some minor 'definitely lost' memory leaks which are generating noise while testing #2011 

The first of these is in `common/log.c`, and occurs when `log_config_free()` is called here as part of `log_start()`:-

https://github.com/neutrinolabs/xrdp/blob/c11d0e3d2efd0d2df3ee6cce03cca76508f2c29a/common/log.c#L735

Memory is allocated here:-
https://github.com/neutrinolabs/xrdp/blob/c11d0e3d2efd0d2df3ee6cce03cca76508f2c29a/common/log.c#L319-L330

The second change is to the VNC module where a stream is not freed after use.

The third change is simply to close the log properly in `xrdp/xrdp.c` if error conditions are detected during startup.

Leaks were tracked down by running xrdp as follows:-

```
sudo valgrind --leak-check=full --keep-debuginfo=yes xrdp -n
```

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -667,6 +667,13 @@ log_config_free(struct log_config *config)
             list_delete(config->per_logger_level);
             config->per_logger_level = NULL;
         }
+
+        if (0 != config->log_file)
+        {
+            g_free(config->log_file);
+            config->log_file = 0;
+        }
+
         g_free(config);
     }
 
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -1006,6 +1006,8 @@ find_matching_extended_rect(struct vnc *v,
         }
     }
 
+    free_stream(s);
+
     return error;
 }
 
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -572,6 +572,7 @@ main(int argc, char **argv)
     {
         g_writeln("It looks like xrdp is already running.");
         g_writeln("If not, delete %s and try again.", pid_file);
+        log_end();
         g_deinit();
         g_exit(0);
     }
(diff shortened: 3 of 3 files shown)


metalefty (approved):
Thanks for the catch!

PR #2015, opened 2021-10-12
tests: add tests on g_strtrim()

metalefty:
I was just learning how to write tests using libcheck. Just to share the output.

### DIFF
--- a/tests/common/test_string_calls.c
+++ b/tests/common/test_string_calls.c
@@ -328,8 +328,7 @@ END_TEST
 START_TEST(test_strtrim__trim_left)
 {
     /* setup */
-    const char *input = "\t\t    \tDone is better than perfect.\t\t    \n\n";
-    char *output = g_strdup(input);
+    char output[] = "\t\t    \tDone is better than perfect.\t\t    \n\n";
 
     /* test */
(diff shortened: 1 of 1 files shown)


tests/common/test_string_calls.c
```diff
@@ -324,12 +324,72 @@ START_TEST(test_bm2str__overflow_some_bits_undefined)
 END_TEST
 
 /******************************************************************************/
+
+START_TEST(test_strtrim__trim_left)
+{
+    /* setup */
+    const char *input = "\t\t    \tDone is better than perfect.\t\t    \n\n";
+    char *output = g_strdup(input);
```
matt335672: Can avoid the string copy by replacing

```c
    const char *input = "\t\t    \tDone is better than perfect.\t\t    \n\n";
    char *output = g_strdup(input);
```

with:-

```c
char output[] = "\t\t    \tDone is better than perfect.\t\t    \n\n";
```

matt335672 (commented):
Looks good. Minor comment.

tests/common/test_string_calls.c
metalefty: Thanks, fixed!

PR #2030, opened 2021-10-22
Cleanup unnecessary files 

metalefty:
Cleanup obsoleted, not up-to-date, not in-use files.

Most files are created in the SourceForge era. Now we are living GitHub era. Supplemental information can be provided by Wiki. If the deletion is too aggressive, please let me know.

To make it easier to revert, I separated deletion to multiple commits for each file.

- design.txt
    - not updated, Architecture Overview on the wiki covers it
- file-loc.txt
    -  not updated 
- faq-compile.txt
    - covered by Wiki
- faq-general.txt
    -   (will) covered by Wiki
 - install.txt
   -  not updated, covered by Wiki
- *-pak
   - created for packaging but not used actually by distro
   - we should create `debian/` directory for example if we add such files


matt335672:
Personally I'm happy with this approach. I can't remember the last time I looked in a tarball for information, rather than at a project website or wiki.

The files also need removing from the top-level `Makefile.am` to fix the CI build:-

```diff
--- a/Makefile.am
+++ b/Makefile.am
@@ -11,15 +11,8 @@ EXTRA_DIST = \
   astyle_config.as \
   bootstrap \
   coding_style.md \
-  description-pak \
-  design.txt \
-  faq-compile.txt \
-  faq-general.txt \
-  file-loc.txt \
   fontdump \
-  install.txt \
   m4 \
-  postinstall-pak \
   vrplayer
 
 if XRDP_NEUTRINORDP
```

A couple of suggestions/questions:-
- How about coding_style.md?
- Is it worth adding a link to the Wiki to README.md?

metalefty:
Ah, I forgot to edit Makefile.am. Thanks for the catch.

> How about coding_style.md?

I think astyle_config.as is the master of the coding style and coding_style.md is a human-friendly translation of astyle_config.as. So it is a supplemental document but not strongly needed because we do coding style checks during CI now.

> Is it worth adding a link to the Wiki to README.md?

Yes, quick guide should be written in README.md but detailed information can be put on Wiki and linked from README.md.

I personally think the "Compiling" section can be minimized and replaced with a link to Wiki.

metalefty:
BTW,  the filename of coding_style.md appears odd a bit for me. Because other markdown files are named uppercase. 

Personally, I would like to remove coding_style.md and add CONTRIBUTING.md for a quick guide of new contributions. Then I would like to mention running code formatter before making a pull request in that file.

matt335672:
Agreed about the CONTRIBUTING.md. That's more useful I think than what we've got at the moment.

metalefty:
Let's merge this and work on CONTRIBUTING.md later.

PR #2054, opened 2021-11-16
Set loongarch no-need-align-macro.

liuxiang88:
This commit is adding support for LoongArch architecture.

### DIFF
--- a/common/arch.h
+++ b/common/arch.h
@@ -88,7 +88,8 @@ typedef int bool_t;
       defined(__i386__) || defined(__aarch64__) || \
       defined(__PPC__) || defined(__LITTLE_ENDIAN__) || \
       defined(__s390__) || defined (__s390x__) || \
-      defined(__riscv)
+      defined(__riscv) || \
+      defined(__loongarch__)
 #define NO_NEED_ALIGN
 #else
 #warning unknown arch


matt335672:
Thanks for this @liuxiang88.

I have a query. I've had a look at the loongarch reference manual regarding unaligned memory access here:-

https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#unaligned-memory-access

My reading of this paragraph is that some implementations on loongarch may generate an address fault on unaligned access, and if so, the default implementation may be better being NEED_ALIGN.

The default for the MIPS architecture is NEED_ALIGN, and I understand loongarch is somewhat based on that.

Have I misunderstood something?

liuxiang88:
> Thanks for this @liuxiang88.
> 
> I have a query. I've had a look at the loongarch reference manual regarding unaligned memory access here:-
> 
> https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#unaligned-memory-access
> 
> My reading of this paragraph is that some implementations on loongarch may generate an address fault on unaligned access, and if so, the default implementation may be better being NEED_ALIGN.
> 
> The default for the MIPS architecture is NEED_ALIGN, and I understand loongarch is somewhat based on that.
> 
> Have I misunderstood something?

After I communicated with the chip developer, I got a reply to prove that you are right.It can be NO_NEED_ALIGN，but should be better to being NEED_ALIGN.
I'll close this request,then make a new request.
Thank you for yourtime.

PR #2066, opened 2021-11-30
Set stream size correctly for TS_SURFCMD_STREAM_SURF_BITS header

matt335672:
Without this PR, the fastpath code aborts when xrdp is build with `--enable-devel-streamcheck` :-

```
[20211130-11:13:04] [CORE ] [parser_stream_overflow_check(parse.c:49)] libxrdp.c:1678 Stream output buffer overflow. Size=0, pos=7, requested=2
```


metalefty (approved):
Good catch.

PR #2067, opened 2021-12-01
Replace `which` with `command -v`

metalefty:
as well as neutrinolabs/xorgxrdp#204.

Reported by: @drwilly

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,27 +1,27 @@
 #!/bin/sh
 
-which autoconf
+command -v autoconf
 if ! test $? -eq 0
 then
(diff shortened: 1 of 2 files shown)


matt335672:
Good idea.

I found these ones too:-

- genkeymap/dump-keymaps.sh
- libpainter/bootstrap
- librfxcodec/bootstrap

matt335672:
@metalefty - I'll have a bit of time tomorrow to do the libpainter and librfxcodec changes.

Shall I pick that up, or will you do it?

metalefty:
@matt335672 Can you do that?

PR #2074, opened 2021-12-06
Fix wrong debug level log message for g_file_atom2

kenhys:
As g_file_atom2 is x-special/gnome-copied-files
(See g_file_atom2 definition in sesman/chansrv/clipboard.c),
it should be "x-special/gnome-copied-files" in this context.

https://github.com/neutrinolabs/xrdp/blob/b557de48885337357a709e630a83902bc251e663/sesman/chansrv/clipboard.c#L2062-L2066


metalefty:
Indeed. Thanks for the catch.

metalefty (approved):

Issue #1762, opened 2020-12-24
Do not spit on the console 2nd stage (inspired by Debian)


metalefty:
Thank you for all of your contributions, @aquesnel, especially logging improvement.

I would like to share some basic directions about logging on the console.

As Debian's downstream patch expresses, they don't like xrdp spit logs on the console when it runs as daemons. I agree with Debian guys at this point. 
* https://salsa.debian.org/debian-remote-team/xrdp/-/blob/master/debian/patches/shutup-daemon.diff

It is partially done in #1142. I would like to complete this. Here's my thoughts about logging:

1. `xrdp` and `xrdp-sesman` should not write anything on stdout/stderr when they run as daemons (except fatal errors)
2. information such as configuration summaries can be dumped when it runs in the foreground

Any feedback is welcome, thanks.

Thoughts:

I personally would like to pick downstream patches up and merge proactively if they're useful to everyone. Because I'm an upstream developer and also a distro package maintainer. Upstream first is a good strategy in most cases. Less downstream patches should make both upstream and downstream happy.

aquesnel:
hi @metalefty ,

Thanks for asking for my opinion.
I agree with you about only logging fatal messages to the console when running as a daemon. I also agree that configuration being dumped to the console when not running as a daemon is a good idea.

I looked at the Debian patch that you linked to, and I don't think it should be applyed it as is because it unconditionally removes useful debugging information. I think that xrdp should conditionally output that information if the user asks for it or create a config that downstream maintainers can define a default for their distribution to disable the information.

I think that a solution based on the logging config will have the biggest impact because as I continue to migrate to the logging to the unified macros across the whole code base then not writing to the console will just follow the configuration.

My ideas for addressing this issue are:
* set the default log level for the console to error or fatal 
* or, when running as a daemon conditionally set the default log level for the console to error or fatal, and if the not running as a daemon then leave the console default to info.
* or, add a command line parameter to enable dumping the config to the console on start up

matt335672:
Hi both,

I added an additional member `dump_on_start` to `struct log_config` as part of #1711. At the moment this is used to suppress dumping the logging output to stdout when running a console utility like `sesrun` which is (I think) exactly what you're trying to do here. It doesn't currently have an easy way to set it in other circumstances.

I was going to suggest using this, but on the other hand I think it might be best to remove it entirely.

If I take @aquesnel's suggestions above as a starting point, this gives the following outline implementation:-

[...]

PR #2078, opened 2021-12-08
xrdp: don’t spit on the console when starting

metalefty:
Picked from Debian's local patch [1].

We have worked on reducing Debian's local patches in #1762 #1765.
This is the last one left and helps to  remove shutup-daemon.diff
completely. I don't think this log to stdout is necessary.

[1] https://salsa.debian.org/debian-remote-team/xrdp/-/blob/debian/0.9.17-2/debian/patches/shutup-daemon.diff

### DIFF
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -635,7 +635,6 @@ main(int argc, char **argv)
 
         if (0 != pid)
         {
-            g_writeln("daemon process %d started ok", pid);
             /* exit, this is the main process */
             log_end();
             g_deinit();


matt335672 (approved):
This is a good idea.

I've got one observation - there are quite a few uses of `g_writeln()` in xrdp/xrdp.c after line 570 (when logging has been started) which are used for fatal errors only. These could be replaced with calls to `LOG(LOG_LEVEL_ALWAYS,`

What do you think? Is this a good time to do this, or would you rather leave it for now?

metalefty:
Yes, lets' replace them now.

matt335672:
I know you're busy these days, so let me know if you need a hand.

metalefty:
@matt335672 Thanks. Let's merge this and can you make another PR on the point?

PR #2088, opened 2021-12-16
Use LOG instead of LOG_DEVEL

kenhys:
According to https://github.com/neutrinolabs/xrdp/wiki/Logging,
it may be better to emit this log message because this log is
useful for system administrator to know whether RestrictOutboundClipboard
configuration works or not

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -2533,7 +2533,7 @@ clipboard_xevent(void *xevent)
             }
             else
             {
-                LOG_DEVEL(LOG_LEVEL_DEBUG, "outbound clipboard is restricted because of config");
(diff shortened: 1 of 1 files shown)


metalefty:
Agreed.

metalefty (approved):
This can be included in the next release.

metalefty:
BTW, you can also promote the log level if needed.

matt335672:
I agree - a level of DEBUG seems to low to me. INFO would be better I think.

metalefty:
@kenhys Can you promote the log level to INFO? Then we'll merge this PR soon.

kenhys:
@metalefty 

Fixed.

metalefty (approved):

PR #2105, opened 2022-01-04
Fix gcc11.1 uninitialised var warning

### LINKED ISSUE #2103: Call to find_matching_extended_rect at vnc.c:1211 looks like a bug
https://github.com/neutrinolabs/xrdp/blob/0c605a3cc488a5ebe977dabf5d7e932a0d080696/vnc/vnc.c#L1215

the parameter `response_code`
```
    error = find_matching_extended_rect(v,
                                        rect_is_reply_to_us,
                                        NULL,
                                        &response_code,
                                        &layout);
```
is passed as a `match_y` argument to
```
find_matching_extended_rect(struct vnc *v,
                            int (*match)(int x, int y, int cx, int cy),
                            int *match_x,
                            int *match_y,
                            struct vnc_screen_layout *match_layout)
```
by quick review, it looks like a bug. Please verify the call.

The main source of the finding is compilation with `-O3` that results with:
```
  CC       vnc.lo
  CC       vnc_clip.lo
In file included from ../common/parse.h:29,
                 from vnc.h:26,
                 from vnc.c:33:
vnc.c: In function ‘lib_data_in’:
../common/log.h:153:33: error: ‘response_code’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  153 | #define LOG(log_level, args...) log_message(log_level, args)
      |                                 ^~~~~~~~~~~
vnc.c:1211:9: note: ‘response_code’ was declared here
 1211 |     int response_code;
      |         ^~~~~~~~~~~~~
```

matt335672:
Fixes #2103 

For a discussion, see #2103

### DIFF
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -1210,7 +1210,7 @@ lib_framebuffer_waiting_for_resize_confirm(struct vnc *v)
 {
     int error;
     struct vnc_screen_layout layout = {0};
-    int response_code;
+    int response_code = 0;
 
     error = find_matching_extended_rect(v,
                                         rect_is_reply_to_us,


metalefty:
Let's ship this quite simple fix to the v0.9.18 release.

metalefty (approved):

PR #2118, opened 2022-01-12
Support imlib2 on RHEL 7

matt335672:
Slightly late here, but after [this comment](/neutrinolabs/xrdp/commit/8b9b22c773495b4d1f3ae98198a31ec905a8cd69#commitcomment-63453607) from @bensilea, I realised I hadn't considered the still active RHEL 7 for imlib2 support.

This PR fixes that. As an example, with these changes to `/etc/xrdp/xrdp.ini` on a stock CentOS 7 install:-

```
ls_background_image=/usr/share/backgrounds/default.jpg
ls_background_transform=zoom

ls_logo_filename=/usr/share/pixmaps/fedora-logo-sprite.png
ls_logo_transform=scale
ls_logo_width=140
ls_logo_height=140
```

I can get this login screen:-
Capture

As I remarked to @bensilea, you'll need to compile with `CFLAGS="-Wno-error=missing-braces"` on RHEL 7 which is something I picked up from @bsmojver

Bojan - sorry this is late. Any good to you?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -224,7 +224,7 @@ case "$with_imlib2" in
         use_imlib2=no
         ;;
     yes)
-        PKG_CHECK_MODULES([IMLIB2], [imlib2 >= 1.4.10],
+        PKG_CHECK_MODULES([IMLIB2], [imlib2 >= 1.4.5],
             [use_imlib2=yes],
             [AC_MSG_ERROR([please install libimlib2-dev or imlib2-devel])])
         ;;


matt335672:
As a postscript, the imlib2 unit tests run on CentOS 7, but a lot of massaging is required to get them working with the old version of the check library.

bsmojver:
I added that option to rhel7 build, because there seems to be a gcc bug present around that. I also disabled imlib2 on rhel7/8, because it was not getting detected correctly. I didn't really have time to find out why when I was building stuff.

matt335672:
That's my mistake, for which I apologise. I should have tested with RHEL 7 as well as some ancient Debian versions. I'm pretty sure the other problem is a gcc bug which I've looked to in the conversation with @bensilea.

I tested on RHEL 7 by adding patch8 (i.e. this change) to your new spec file and running an autoreconf after the patches were applied. I removed the checks to disable imlib2 as well.

metalefty (approved):

PR #2144, opened 2022-02-09
Remove s_check() macro

matt335672:
The s_check() macro in parse.h serves no useful purpose, and its two uses are actually erroneous. The PR removes the macro.

The macro is defined as:-

```c
#define s_check(s) s_check_rem(s, 0)

#define s_check_rem(s, n) ((s)->p + (n) <= (s)->end)
```
In other words, it's checking that 0 or more bytes are available in the input stream. It does not promise that the input stream can be read.

Use 1) is in sesman/chansrv/clipboard.c:-

```c
    while (s_check(s))
    {
        in_uint16_le(s, wchr);
        wtext[index] = wchr;
        if (wchr == 0)
        {
            break;
        }
        index++;
    }
    wtext[index] = 0;
```

In normal use, the string sent over by the client is null-terminated, and so problems do not arise. To prevent a read of uninitialised memory if the string is not terminated, the loop should be protected with `while(s_check_rem(s, 2))`

Use 2 is in `vnc/vnc_clip.c`:-

```c
                case CF_TEXT:
                    lastc = '\0';
                    while (s_check(s))
                    {
                        in_uint8(s, c);
                        if (c == '\n' && lastc == '\r')
                        {
                            /* Overwrite the `\r' */
                            --vc->rfb_clip_s->p;
                        }
                        out_uint8(vc->rfb_clip_s, c);
                        lastc = c;
                    }
```

The reason this does not cause a read of uninitialised memory is because the code is almost never used. Most clients ask for `CF_UNICODETEXT` rather than `CF_TEXT`. In any case, the loop should be protrected with `while (s_check_rem(s, 1))`


metalefty (approved):
Thanks for the housekeeping.

PR #2146, opened 2022-02-10
Fixed possible SIGCHILD signal lost

zbstao:
When multiple(eg. 20) xrdp connections are disconnected at the same time(eg.  close all rdp client at the same time), zombie process will be spawned. 
use xorgrdp session and set the parameter KillDisconnected to true.

I found similar code in file chansrv.c
```
void
child_signal_handler(int sig)
{
    int pid;

    LOG_DEVEL(LOG_LEVEL_INFO, "child_signal_handler:");
    do
    {
        pid = g_waitchild();
        LOG_DEVEL(LOG_LEVEL_INFO, "child_signal_handler: child pid %d", pid);
        if ((pid == g_exec_pid) && (pid > 0))
        {
            LOG_DEVEL(LOG_LEVEL_INFO, "child_signal_handler: found pid %d", pid);
            //shutdownx();
        }
    }
    while (pid >= 0);
}
```

### DIFF
--- a/sesman/sig.c
+++ b/sesman/sig.c
@@ -122,12 +122,16 @@ sig_sesman_session_end(int sig)
         return;
     }
 
-    pid = g_waitchild();
-
-    if (pid > 0)
+    do
     {
-        session_kill(pid);
+        pid = g_waitchild();
+
+        if (pid > 0)
+        {
+            session_kill(pid);
+        }
     }
+    while (pid >= 0);
 }
 
 /******************************************************************************/
--- a/xrdp/xrdp.c
+++ b/xrdp/xrdp.c
@@ -163,9 +163,7 @@ xrdp_shutdown(int sig)
 static void
 xrdp_child(int sig)
 {
-    int safety;
-
-    for (safety = 0; (g_waitchild() >= 0) && (safety <= 10); safety++)
+    while (g_waitchild() >= 0)
     {
     }
 }


xrdp/xrdp.c
```diff
@@ -163,11 +163,8 @@ xrdp_shutdown(int sig)
 static void
 xrdp_child(int sig)
 {
-    int safety;
+	while (g_waitchild() >= 0) ;
```
matt335672: The [coding style](/neutrinolabs/xrdp/wiki/Coding-Style) requires spaces for indents rather than tabs, and also needs braces for the empty loop (e.g.):-

```c
while(...)
{
}
```

I'm happy with the removal of the safety variable. It was introduced in f763cb37885b02c310aa906147669a29b99405dc but doesn't seem to be necessary.

sesman/sig.c
```diff
@@ -122,12 +122,15 @@ sig_sesman_session_end(int sig)
...
     {
-        session_kill(pid);
-    }
+        pid = g_waitchild();
+
+        if (pid > 0)
+        {
+            session_kill(pid);
+        }
+    }while (pid >= 0);
```
matt335672: The `while` on the same line matches the [coding style](/neutrinolabs/xrdp/wiki/Coding-Style), but will be rejected by our use of `astyle` in the CI pipeline. Please can you place the `while` on a separate line?

matt335672 (changes requested):
Hi @zbstao,

Thanks for your useful contribution. I've got a couple of comments relating to the coding style we use, but the functionality looks fine to me.

If you address the comments I'll enable the CI pipeline for the PR.

Thanks.

xrdp/xrdp.c
zbstao: sorry didn't say the coding style page, the code needs to be modified.

sesman/sig.c
zbstao: yes

matt335672 (changes requested):
Nearly there, but you've used a tab instead of 4 spaces in both locations. The CI still won't like that.

Apologies if I wasn't clear.

matt335672 (approved):

matt335672:
Looks great.

Can you squash your commits? I'll merge it in then.

zbstao:
hi @matt335672 
Done. In addition I will read the coding style asap.

matt335672:
Thanks for your contribution @zbstao

PR #2149, opened 2022-02-14
Fix NPEs in log.c

Nexarian:
Multiple NPEs can happen in the internal_log_config_copy procedure, and we need to address this before we merge in changes for egfx.


matt335672:
LGTM

PR #2151, opened 2022-02-15
Fixed possible infinite loop

zbstao:
Fixed possible infinite loop, sorry for the PR #2146 @matt335672.
If "KillDisconnected=true" and "Policy=UBI", then login same user from same IP will cause xrdp/xrdp-sesman process cost 100% CPU, xrdp server can't log in.


matt335672:
Hi @zbstao 

Thanks for that.

I think there's some confusing code in the `g_waitchild()` function in `os_calls.c`. The `g_waitchild()` function looks like this:-

https://github.com/neutrinolabs/xrdp/blob/371c0dc8738e9a3246863354d7e773e0e8e5818f/common/os_calls.c#L2949-L2972

Reading this code suggests that 0 can be legitimately returned if a signal occurs, which is why some of these loops have used '>= 0'.

I don't think however that this is the case. Because `WNOHANG` is specified, the operating of `waitpid` will not be interrupted by a signal delivery, so we don't need to test for this case.

Does that sound right to you?

If so, can I also ask you to also remove these lines from `g_waitchild()`:-

```c
    if (rv == -1)
    {
        if (errno == EINTR) /* signal occurred */
        {
            rv = 0;
        }
    }

```

Also, the comment above the function should not mention signals, but should say something like:-

```
/* does not work in win32
   returns pid of process that exits or <= 0 if no process was found */
```

If you don't agree, please come back to me.

zbstao:
hi @matt335672 
In `man waitpid`, it says "waitpid(): on success, returns the process ID of the child whose state has changed; if WNOHANG was specified and one or more child(ren) specified by pid exist, **but have not yet changed state, then 0 is returned.** On error, -1 is returned."
So g_waitchild will always return 0 if no child process exits, this is the problem.

matt335672:
Yes, I think that's it more or less. My reading of the page is that if no children exist, -1 is returned and ECHILD is in errno. If however children exist (which haven't exited) 0 is returned as you say.

Also further down the man page we get:-

```
ERRORS
       . . .
       EINTR  WNOHANG  was  not  set  and an unblocked signal or a SIGCHLD was
              caught; see signal(7).

```

So do you agree that `g_waitchild()` should be modified as I suggest?

Thanks.

zbstao:
I agree.

matt335672:
Thanks @zbstao

PR #2152, opened 2022-02-15
Minor logging fixes in xrdp_iso.c

Nexarian:
Two logging errors found while working in these files.

### DIFF
--- a/libxrdp/xrdp_iso.c
+++ b/libxrdp/xrdp_iso.c
@@ -428,7 +428,8 @@ xrdp_iso_send_cc(struct xrdp_iso *self)
             out_uint16_le(s, 8);                          /* length (must be 8) */
             out_uint32_le(s, self->selectedProtocol);     /* selectedProtocol */
             LOG_DEVEL(LOG_LEVEL_TRACE, "Adding structure [MS-RDPBCGR] RDP_NEG_RSP "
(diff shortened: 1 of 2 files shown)


libxrdp/xrdp_iso.c
```diff
@@ -428,7 +428,7 @@ xrdp_iso_send_cc(struct xrdp_iso *self)
             out_uint16_le(s, 8);                          /* length (must be 8) */
             out_uint32_le(s, self->selectedProtocol);     /* selectedProtocol */
             LOG_DEVEL(LOG_LEVEL_TRACE, "Adding structure [MS-RDPBCGR] RDP_NEG_RSP "
-                      "flags 0, length 8, selectedProtocol 0x%8.8x",
+                      "flags 0x01, length 8, selectedProtocol 0x%8.8x",
```
matt335672: Wouldn't it be better to use `EXTENDED_CLIENT_DATA_SUPPORTED` here as that's what's in the code (e.g.):-

```c
                      "flags 0x%02x, length 8, selectedProtocol 0x%8.8x",
                      EXTENDED_DATA_SUPPORTED,
                      self->selectedProtocol);
```

Nexarian: Yes, updated!

matt335672:
Thanks @Nexarian!

PR #2163, opened 2022-03-01
Replace SCP V0 and V1 with new IPC mechanism

matt335672:
This PR is part of the implementation of #1961 

It's also prompted by [CVE-2020-4044](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-4044) and [CVE-2022-23613](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2022-23613). The idea is to move most of the sensitive code into a single layer which shouldn't need to be changed a lot, and which (importantly) can also be subjected to automated testing.

The current complicated SCP architecture with V0 and V1 is replaced with a much simpler SCP V2 which sits on top of a new inter-process messaging library `libipm`.

The new SCP implementation can be found in `libipm/scp.h` / `libipm/scp.c`. It implements all of the currently used functionality in SCP V0 and V1 (i.e. starting sessions, PAM authentication and listing sessions). No additional features are provided yet.

`libipm` is responsible for :-
- marshalling and de-marshalling messages
- checking types
- checking for buffer overflows
- erasing sensitive data (e.g. passwords) from transports after use.

The library is layered on top of the standard `struct trans` mechanism, and uses the standard calls in `parse.h` to create and parse messages. It's inspired by the SD-Bus library for D-Bus messages, and re-uses the D-Bus type system (although there are no D-Bus dependencies).

Running `doxygen` in the `libipm` library will (or should) generate complete local HTML documentation for the library calls.

A comprehensive test suite is included for libipm.

There are a total of 11 commits in this PR
1) Add the SIZEOF_INT autotools macro. This allows the compiler to determine how big an int is at compile time.
2) Adds a sub-class capability to `struct trans` so we can provide libipm-specific data for a transport.
3) Adds libipm (and SCP V2)
4) Add the test suite
5) Move sesman to the new interface
6) Move xrdp to the new interface
7) Tidy up sesman tools
8) Ditto
9) Move the tools to the new interface
10) Remove SCP V0 and V1 files
11) As an illustration of the new flexibility in the library, an extra parameter is added to an SCP message to allow fail2ban logging to be implemented for the PAM authentication mechanism.

Apart from 11), there are no changes in functionality to the product.

I couldn't really think of a better way to present this. I appreciate reviewing will be tricky. Should I add a new Wiki page on using libipm?


metalefty:
First off, thanks for the great work. 

I would like to release v0.9.19 before merging this because some users & customers want new clipboard restriction feature implemented at#2082 but I don't want to include other big changes.

Furthermore, when this new IPC mechanism merged, there was a big leap. So the new version will be v1.0 (or v0.10).

matt335672:
OK - I'm fine with that.

Since we're going to change the major version, shall I also implement a move to Unix Domain Sockets? There's not a lot left to do on that when this is merged.

matt335672:
Minor change - libipm sockets are now always non-blocking, so that a `trans_force_write()` does not block indefinitely.

matt335672:
Merged now v0.9.19 is released.

Issue #2021, opened 2021-10-17
reconnect after connection drop hangs


vixie:
connecting to suse leap 15.3 inside hyper-v under windows 10 (21H2) using windows rdp client. any network failure including suspend/resume results in an unkillable and nonreconnectable session. searched the web, found this old bug report:

https://sourceforge.net/p/xrdp/bugs/54/

that fix, modified for changes made since then to the surrounding source code, is still necessary, and still sufficient.

```
--- xrdp-0.9.13.1/sesman/session.c.orig 2021-10-17 21:26:45.386189301 +0000
+++ xrdp-0.9.13.1/sesman/session.c      2021-10-17 21:26:26.638741197 +0000
@@ -833,6 +833,7 @@
     }
     else
     {
+        g_signal_child_stop(sig_sesman_session_end); /* SIGCHLD */
         temp->item->pid = pid;
         temp->item->display = display;
         temp->item->width = s->width;
```

matt335672:
Thanks for that, @vixie

I've had a look at the existing code, and as Paul van Tilburg says on the original thread this shouldn't be necessary on the face of it, although I'm quite prepared to believe it as the signal handling needs some updating.

In the existing code, the SIGCHLD handler is set once, whereas the patch causes the SIGCHLD handler to be reset after every session is created. That may be significant.

I don't suppose you've got any sesman logs from the hangs have you? It would be good to get some more understanding of exactly what codepath is being taken to cause the SIGCHLD handler to be lost.

matt335672:
I'd like to try to reproduce this here if possible, but I'm a little confused from your description and the link to https://sourceforge.net/p/xrdp/bugs/54/.

When you say 'network failure', this wouldn't normally cause a session to exit, and so SIGCHLD wouldn't come into this. Are you setting `KillDisconnected` and `DisconnectedTimeLimit` in `/etc/xrdp/sesman.ini`?

Thanks.

vixie:
at the moment i've broken _everything_ and i need to set up a clean 
environment, so that i can do A/B testing, with the diags you've asked 
for. sorry for any delay.

matt335672:
I sympathise with you on 'breaking everything' - it's something I seem to have a talent for myself.

[...]

PR #2168, opened 2022-03-03
Fix sesman signal processing

### LINKED ISSUE #1729: session end code is not signal-safe
Currently a new session is added in to the session list with these two line of code:-

https://github.com/neutrinolabs/xrdp/blob/0d8e4146a2f5a614090be6f0db72c95e80a88572/sesman/session.c#L854-L855

If a `SIGCHLD` is received between these two lines, `session_kill()` is called. If this removes the first item in the chain, the result will be `g_sessions->next` pointing to free'd memory.

Potential solutions:-
- On suitable platforms, replace the existing signal code which uses `signal()` with `sigaction()` and use `sigprocmask()` to block delivery of `SIGCHLD` during this critical section.
- Use a different method to detect session termination (e.g. a closed file handle to the session leader). See [this comment](1684#issuecomment-731153999) for one possibility.

matt335672:
**edit 2022-03-04:** No longer draft.

This series of commits fixes #1729 and may also address reports of other session ending problems (e.g. #2021).

The "self-pipe trick" [Stevens 1992] is used to keep the processing performed in all signal handlers to a minimum. This is really simple to do in xrdp by using `g_set_wait_obj()` and related calls.

As a result, SIGPIPE and SIGHUP are now processed from the sesman main loop, like all other events.

Other problems with SIGHUP processing have been addressed:-

1) The log file was closed and re-opened. Since the log file is shared with the child processes, this will cause the parent and the children to have a different file table entries. Writes at the end of the file will not be synchronised.
2) The existing code does not allow for the listening port to be changed, which is a major reason why the user may have edited the config in the first place.

This series of 4 commits addresses these issues as follows:-
1) [tidy-up] Some of the global variables defined in sesman.c are declared in other c files as `extern`. This commit moves these declarations into `sesman.h`
2) [tidy-up] `struct log_config` contains a few members which are not always needed. For example, the `per_logger_level` member is only used for debug builds. This commit makes these members (and associated code) conditional.
3) The 3rd parameter of the `log_start()` call is change from a boolean `dump_on_start` to a more generic `flags` field. One of the new flags is `LOG_START_DUMP_CONFIG`, which provides the existing `dump_on_start` functionality. The other one, `LOG_START_RESTART` allows for logging to be restarted smoothly. Existing log files are left untouched by a restart, even if the name changes.
4) Finally, sesman is updated as follows:-
   - Signal handlers now just call `g_set_wait_obj()`. No significant processing is done in the signal handlers.
   - In the main loop, the wait objects are handled in the same way as file descriptors. All actions resulting from signals are synchronised with actions resulting from file activity.
   - The restart code now calls `log_start()` with `LOG_START_RESTART`, and allows the user to change the listening port dynamically.

### DIFF
--- a/common/log.c
+++ b/common/log.c
@@ -224,7 +224,7 @@ internal_log_end(struct log_config *l_cfg)
 }
 
 /**
- * Converts a string representing th log level to a value
+ * Converts a string representing the log level to a value
  * @param buf
  * @return
  */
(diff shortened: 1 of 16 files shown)


matt335672:
As another note, when developing with sesman, it's now really easy to change the logging level, or add entries to `[LoggingPerLogger]`. After editing the config, a `sudo pkill -HUP xrdp-sesman` causes the file to be re-read without needing to stop and start sesman.

PR #2175, opened 2022-03-15
Refactoring resizing into state machine.

### LINKED ISSUE #1928: MSTSC crashes when resolution is changed by maximizing on a different monitor
I have two monitors with 2560x1440 and 1920x1080 resolution.

and setting mstsc to use one monitor for full screen mode.

when I drag mstsc from a monitor to another, error occurred as below.
image
It means internal error.

and will got log 
```

==> /home/.xorgxrdp.12.log <==
[1135171.145] rdpClientConProcessMsgClientInput: invalidate x 0 y 0 cx 2560 cy 1440
[1135171.145] rdpClientConProcessMsgVersion: version 0 0 0 1
[1135171.145] rdpClientConProcessScreenSizeMsg: set width 2560 height 1440 bpp 32
[1135171.146] rdpClientConProcessScreenSizeMsg: shmemid 254443553 shmemptr 0x7f5f87b3a000
[1135171.146] rdpRRScreenSetSize: width 2560 height 1440 mmWidth 677 mmHeight 381
[1135171.151] rdpRRGetInfo:
[1135171.151]   screen resized to 2560x1440
[1135171.157] rdpClientConProcessScreenSizeMsg: RRScreenSizeSet ok=[1]
[1135171.157] rdpClientConProcessMsgClientInput: invalidate x 0 y 0 cx 2560 cy 1440
[1135171.157] rdpClientConProcessMsgClientInput: invalidate x 0 y 0 cx 2560 cy 1440
[1135171.158] rdpClientConProcessMsgVersion: version 0 0 0 1
[1135171.158] rdpClientConProcessScreenSizeMsg: set width 2560 height 1440 bpp 32
[1135171.158] rdpClientConProcessScreenSizeMsg: shmemid 254476321 shmemptr 0x7f5f87b3a000
[1135171.158] rdpClientConProcessMsgClientInput: invalidate x 0 y 0 cx 2560 cy 1440
[1135171.158] rdpClientConProcessMsgClientInfo:
[1135171.158]   got client info bytes 7064
[1135171.158]   jpeg support 0
[1135171.158]   offscreen support 1
[1135171.158]   offscreen size 10485760
[1135171.158]   offscreen entries 100
[1135171.158] rdpClientConProcessMsgClientInfo: got RFX capture
[1135171.158]   cap_width 2560 cap_height 1472
[1135171.158] rdpClientConProcessMsgClientInfo: shmemid 254509089 shmemptr 0x7f5f87aea000 bytes 15073280
[1135171.158]   client can not do offscreen to offscreen blits
[1135171.158]   client can do new(color) cursor
[1135171.158]   client can not do multimon
[1135171.158] rdpRRSetRdpOutputs: numCrtcs 1 numOutputs 1 monitorCount 0
[1135171.158] rdpRRSetRdpOutputs: update output 0 left 0 top 0 width 2560 height 1440
[1135171.158] rdpRRUpdateOutput:
[1135171.159] rdpLoadLayout: keylayout 0x00000804 variant  display 12
[1135171.159] rdpkeybChangeKeyboardControl:
[1135171.159] rdpkeybChangeKeyboardControl: autoRepeat on
[1135171.159] rdpkeybChangeKeyboardControl:
[1135171.160] rdpkeybChangeKeyboardControl: autoRepeat on
[1135171.160] rdpClientConProcessMsgClientInfo:
[1135171.160]   got client info bytes 7064
[1135171.160]   jpeg support 0
[1135171.160]   offscreen support 1
[1135171.160]   offscreen size 10485760
[1135171.160]   offscreen entries 100
[1135171.160] rdpClientConProcessMsgClientInfo: got RFX capture
[1135171.160]   cap_width 2560 cap_height 1472
[1135171.160] rdpClientConProcessMsgClientInfo: shmemid 254541857 shmemptr 0x7f5f87aea000 bytes 15073280
[1135171.160]   client can not do offscreen to offscreen blits
[1135171.161]   client can do new(color) cursor
[1135171.161]   client can not do multimon
[1135171.161] rdpRRSetRdpOutputs: numCrtcs 1 numOutputs 1 monitorCount 0
[1135171.161] rdpRRSetRdpOutputs: update output 0 left 0 top 0 width 2560 height 1440
[1135171.161] rdpRRUpdateOutput:
[1135171.161] rdpLoadLayout: keylayout 0x00000804 variant  display 12
[1135171.161] rdpkeybChangeKeyboardControl:
[1135171.161] rdpkeybChangeKeyboardControl: autoRepeat on
[1135171.161] rdpkeybChangeKeyboardControl:
[1135171.161] rdpkeybChangeKeyboardControl: autoRepeat on
[1135171.240] rdpRRGetInfo:
[1135171.259] rdpInDeferredRepeatCallback:
[1135171.259] rdpkeybChangeKeyboardControl:
[1135171.259] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.260] rdpRRGetInfo:
[1135171.260] rdpInDeferredRepeatCallback:
[1135171.260] rdpkeybChangeKeyboardControl:
[1135171.260] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.268] rdpInDeferredRepeatCallback:
[1135171.268] rdpkeybChangeKeyboardControl:
[1135171.268] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.268] rdpInDeferredRepeatCallback:
[1135171.268] rdpkeybChangeKeyboardControl:
[1135171.268] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.273] rdpRRGetInfo:

==> /var/log/xrdp.log <==
[20210623-15:44:38] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20210623-15:44:38] [INFO ] xrdp_process_offscreen_bmpcache: support level 1 cache size 10485760 MB cache entries 100
[20210623-15:44:38] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[20210623-15:44:38] [WARN ] xrdp_caps_process_codecs: unknown codec id 5
[20210623-15:44:38] [INFO ] xrdp_caps_process_codecs: RemoteFX, codec id 3, properties len 49

==> /home/.xorgxrdp.12.log <==
[1135171.458] KbdSync: toggling num lock
[1135171.458] rdpkeybChangeKeyboardControl:
[1135171.458] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.458] rdpkeybChangeKeyboardControl:
[1135171.458] rdpkeybChangeKeyboardControl: autoRepeat off
[1135171.486] rdpClientConProcessMsgClientInput: invalidate x 0 y 0 cx 2559 cy 1439

==> /var/log/xrdp.log <==
[20210623-15:44:38] [ERROR] SSL_read: I/O error
[20210623-15:44:38] [ERROR] xrdp_iso_send: trans_write_copy_s failed
[20210623-15:44:38] [ERROR] SSL_shutdown: I/O error
[20210623-15:44:38] [ERROR] Sending [ITU T.125] DisconnectProviderUltimatum failed

==> /home/.xorgxrdp.12.log <==
[1135171.521] rdpClientConRecv: g_sck_recv failed(returned -1)
[1135171.521] rdpClientConRecvMsg: error
[1135171.521] rdpClientConCheck: rdpClientConGotData failed
[1135171.525] rdpClientConDisconnect:
[1135171.525] rdpRemoveClientConFromDev: removing clientCon 0x55644bb73110
```

Nexarian:
Switch to a state machine that is processed at its fastest once per XRDP main loop cycle. This allows for for more flexibility in how resizing is managed, and also allows for preparation for resizing to work with the EGFX changes that require this.

This also is a stability enhancement in that by queueing up the resize changes, they are processed one at a time, and if multiple resizes were to come in at once, or duplicates were to come in, this handles them elegantly. With the older version, XRDP was more likely to crash or cause a client disconnect, especially if there was a duplicate resize that came in as it sometimes does from MSTSC.

Fixes: https://github.com/neutrinolabs/xrdp/issues/1928

### DIFF
--- a/xrdp/xrdp.h
+++ b/xrdp/xrdp.h
@@ -147,8 +147,12 @@ xrdp_wm_get_wait_objs(struct xrdp_wm *self, tbus *robjs, int *rc,
                       tbus *wobjs, int *wc, int *timeout);
 int
 xrdp_wm_check_wait_objs(struct xrdp_wm *self);
+const char *
+xrdp_wm_login_state_to_str(enum wm_login_state login_state);
 int
 xrdp_wm_set_login_state(struct xrdp_wm *self, enum wm_login_state login_state);
+int
+xrdp_wm_can_resize(struct xrdp_wm *self);
 void
 xrdp_wm_mod_connect_done(struct xrdp_wm *self, int status);
 
@@ -397,18 +401,12 @@ xrdp_bitmap_compress(char *in_data, int width, int height,
 
 /* xrdp_mm.c */
 
-struct dynamic_monitor_layout
+struct display_control_monitor_layout_data
 {
-    int flags;
-    int left;
-    int top;
-    int width;
-    int height;
-    int physical_width;
-    int physical_height;
-    int orientation;
-    int desktop_scale_factor;
-    int device_scale_factor;
+    struct display_size_description description;
+    enum display_resize_state state;
+    int last_state_update_timestamp;
+    int start_time;
 };
 
 int
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -43,6 +43,7 @@
 
 #include "xrdp_encoder.h"
 #include "xrdp_sockets.h"
+#include <limits.h>
 
 
 /* Forward declarations */
@@ -65,6 +66,8 @@ struct xrdp_mm *
 xrdp_mm_create(struct xrdp_wm *owner)
 {
     struct xrdp_mm *self;
+    char buf[1024];
+    int pid;
 
     self = (struct xrdp_mm *)g_malloc(sizeof(struct xrdp_mm), 1);
     self->wm = owner;
@@ -72,6 +75,14 @@ xrdp_mm_create(struct xrdp_wm *owner)
     self->login_names->auto_free = 1;
     self->login_values = list_create();
     self->login_values->auto_free = 1;
+    self->resize_queue = list_create();
+    self->resize_queue->auto_free = 1;
+
+    pid = g_getpid();
+    /* setup wait objects for signalling */
+    g_snprintf(buf, sizeof(buf), "xrdp_%8.8x_resize_ready", pid);
+    self->resize_ready = g_create_wait_obj(buf);
+    self->resize_data = NULL;
 
     LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_mm_create: bpp %d mcs_connection_type %d "
               "jpeg_codec_id %d v3_codec_id %d rfx_codec_id %d "
@@ -165,6 +176,9 @@ xrdp_mm_delete(struct xrdp_mm *self)
     self->sesman_trans = 0;
     list_delete(self->login_names);
     list_delete(self->login_values);
+    list_delete(self->resize_queue);
+    g_free(self->resize_data);
+    g_delete_wait_obj(self->resize_ready);
     g_free(self);
 }
 
@@ -969,7 +983,8 @@ dynamic_monitor_open_response(intptr_t id, int chan_id, int creation_status)
     struct stream *s;
     int bytes;
 
-    LOG_DEVEL(LOG_LEVEL_TRACE, "dynamic_monitor_open_response: chan_id %d creation_status 0x%8.8x", chan_id, creation_status);
+    LOG_DEVEL(LOG_LEVEL_TRACE, "dynamic_monitor_open_response: chan_id %d "
+              "creation_status 0x%8.8x", chan_id, creation_status);
     if (creation_status != 0)
     {
         LOG(LOG_LEVEL_ERROR, "dynamic_monitor_open_response: error");
@@ -1031,152 +1046,24 @@ sync_dynamic_monitor_data(struct xrdp_wm *wm,
              MAXIMUM_MONITOR_SIZE);
 }
 
-/******************************************************************************/
-static int
-process_dynamic_monitor_description(struct xrdp_wm *wm,
-                                    struct display_size_description *description)
+int
+advance_resize_state_machine(struct xrdp_mm *mm,
+                             enum display_resize_state new_state)
 {
-    int error = 0;
-    struct xrdp_mm *mm = wm->mm;
-    struct xrdp_mod *module = mm->mod;
-
-    LOG_DEVEL(LOG_LEVEL_TRACE, "process_dynamic_monitor_description:");
-
-    if (wm->client_info->suppress_output == 1)
-    {
-        LOG(LOG_LEVEL_ERROR,
-            "process_dynamic_monitor_description:"
-            " Not allowing resize. Suppress output is active.");
-        return 0;
-    }
-    if (description == NULL)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR,
-                  "process_dynamic_monitor_description:"
-                  " description is null.");
-        return 0;
-    }
-    if (description->session_width <= 0 || description->session_height <= 0)
-    {
-        LOG(LOG_LEVEL_ERROR,
-            "process_dynamic_monitor_description: Not allowing resize due to"
-            " invalid dimensions (w: %d x h: %d)",
-            description->session_width, description->session_height);
-        return 0;
-    }
-    if (description->session_width
-            == wm->client_info->display_sizes.session_width
-            && description->session_height
-            == wm->client_info->display_sizes.session_height)
-    {
-        LOG(LOG_LEVEL_WARNING, "process_dynamic_monitor_description:"
-            " Not resizing. Already this size. (w: %d x h: %d)",
-            description->session_width,
-            description->session_height);
-        sync_dynamic_monitor_data(wm, description);
-        return 0;
-    }
-
-    // TODO: Unify this logic with server_reset
-    error = libxrdp_reset(wm->session,
-                          description->session_width,
-                          description->session_height,
-                          wm->screen->bpp);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR,
-                  "process_dynamic_monitor_description:"
-                  " libxrdp_reset failed %d", error);
-        return error;
-    }
-    /* reset cache */
-    error = xrdp_cache_reset(wm->cache, wm->client_info);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR,
-                  "process_dynamic_monitor_description: xrdp_cache_reset"
-                  " failed %d", error);
-        return error;
-    }
-    /* load some stuff */
-    error = xrdp_wm_load_static_colors_plus(wm, 0);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                  " xrdp_wm_load_static_colors_plus failed %d", error);
-        return error;
-    }
-
-    error = xrdp_wm_load_static_pointers(wm);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                  " xrdp_wm_load_static_pointers failed %d", error);
-        return error;
-    }
-    /* resize the main window */
-    error = xrdp_bitmap_resize(wm->screen,
-                               description->session_width,
-                               description->session_height);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                  " xrdp_bitmap_resize failed %d", error);
-        return error;
-    }
-    /* redraw */
-    error = xrdp_bitmap_invalidate(wm->screen, 0);
-    if (error != 0)
-    {
-        LOG_DEVEL(LOG_LEVEL_ERROR,
-                  "process_dynamic_monitor_description:"
-                  " xrdp_bitmap_invalidate failed %d", error);
-        return error;
-    }
-
-    if (module != 0)
-    {
-        error = module->mod_server_version_message(module);
-        if (error != 0)
-        {
-            LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                      " mod_server_version_message failed %d", error);
-            return error;
-        }
-        error = module->mod_server_monitor_resize(
-                    module,
-                    description->session_width,
-                    description->session_height);
-        if (error != 0)
-        {
-            LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                      "mod_server_monitor_resize failed %d", error);
-            return error;
-        }
-        error = module->mod_server_monitor_full_invalidate(
-                    module,
-                    description->session_width,
-                    description->session_height);
-        if (error != 0)
-        {
-            LOG_DEVEL(LOG_LEVEL_ERROR, "process_dynamic_monitor_description:"
-                      "mod_server_monitor_full_invalidate failed %d", error);
-            return error;
-        }
-    }
-
-    // Need to recreate the encoder for connections that use it.
-    if (mm->encoder != NULL)
-    {
-        xrdp_encoder_delete(mm->encoder);
-        mm->encoder = NULL;
-    }
-    if (mm->encoder == NULL)
-    {
-        mm->encoder = xrdp_encoder_create(mm);
-    }
-
-    sync_dynamic_monitor_data(wm, description);
+    struct display_control_monitor_layout_data *description = mm->resize_data;
+    LOG_DEVEL(LOG_LEVEL_INFO,
+              "advance_resize_state_machine:"
+              " Processing resize to: %d x %d."
+              " Advancing state from %s to %s."
+              " Previous state took %d MS.",
+              description->description.session_width,
+              description->description.session_height,
+              XRDP_DISPLAY_RESIZE_STATE_TO_STR(description->state),
+              XRDP_DISPLAY_RESIZE_STATE_TO_STR(new_state),
+              g_time3() - description->last_state_update_timestamp);
+    description->state = new_state;
+    description->last_state_update_timestamp = g_time3();
+    g_set_wait_obj(mm->resize_ready);
     return 0;
 }
 
@@ -1192,10 +1079,19 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
     struct xrdp_process *pro;
     struct xrdp_wm *wm;
     int monitor_layout_size;
+    struct display_size_description *display_size_data;
 
     LOG_DEVEL(LOG_LEVEL_TRACE, "dynamic_monitor_data:");
     pro = (struct xrdp_process *) id;
     wm = pro->wm;
+
+    if (wm->client_info->suppress_output == 1)
+    {
+        LOG(LOG_LEVEL_INFO, "dynamic_monitor_data: Not allowing resize."
+            " Suppress output is active.");
+        return error;
+    }
+
     g_memset(&ls, 0, sizeof(ls));
     ls.data = data;
     ls.p = ls.data;
@@ -1204,16 +1100,33 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
     s = &ls;
     in_uint32_le(s, msg_type);
     in_uint32_le(s, msg_length);
-    LOG(LOG_LEVEL_DEBUG, "dynamic_monitor_data: msg_type %d msg_length %d",
-        msg_type, msg_length);
+    LOG_DEVEL(LOG_LEVEL_DEBUG,
+              "dynamic_monitor_data: msg_type %d msg_length %d",
+              msg_type, msg_length);
 
-    if (msg_type != DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT)
+    if (msg_type != DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT
+            && msg_length >= 0)
     {
+        // Unsupported message types
+        switch (msg_type)
+        {
+            case DISPLAYCONTROL_PDU_TYPE_CAPS:
+                LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: Received"
+                    " DISPLAYCONTROL_PDU_TYPE_CAPS. Per MS-RDPEDISP 2.2.2.1,"
+                    " this is a server-to-client message, client should never"
+                    " send this. Ignoring message");
+                break;
+            default:
+                LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: Received neither"
+                    " nor DISPLAYCONTROL_PDU_TYPE_CAPS. Ignoring message.");
+                break;
+        }
         return 0;
     }
     in_uint32_le(s, monitor_layout_size);
     if (monitor_layout_size != 40)
     {
+        /* 2.2.2.2 DISPLAYCONTROL_MONITOR_LAYOUT_PDU */
         LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: monitor_layout_size"
             " is %d. Per spec ([MS-RDPEDISP] 2.2.2.2"
             " DISPLAYCONTROL_MONITOR_LAYOUT_PDU) it must be 40.",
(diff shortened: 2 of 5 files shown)


Nexarian:
@matt335672 Strangely, this does not fully fix https://github.com/neutrinolabs/xrdp/issues/1928. In the logs I can literally see the double change request being rejected:

```
[20220330-02:15:51] [INFO ] connected ok
[20220330-02:16:04] [INFO ] dynamic_monitor_data:
[20220330-02:16:04] [INFO ] dynamic_monitor_data: msg_type 2 msg_length 56
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from QUEUED to QUEUED. Previous state took -646477620 MS.
[20220330-02:16:04] [INFO ] dynamic_monitor_process_queue: Not allowing resize due to invalid dimensions (w: 0 x h: 0)
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from QUEUED to SERVER_MONITOR_RESIZE. Previous state took 1 MS.
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from SERVER_MONITOR_RESIZE to SERVER_VERSION_MESSAGE. Previous state took 0 MS.
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from SERVER_VERSION_MESSAGE to XRDP_CORE_RESIZE. Previous state took 0 MS.
[20220330-02:16:04] [INFO ] dynamic_monitor_data:
[20220330-02:16:04] [INFO ] dynamic_monitor_data: msg_type 2 msg_length 56
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from QUEUED to QUEUED. Previous state took -646477616 MS.
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from XRDP_CORE_RESIZE to SERVER_INVALIDATE. Previous state took 8 MS.
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from SERVER_INVALIDATE to COMPLETE. Previous state took 0 MS.
[20220330-02:16:04] [INFO ] dynamic_monitor_process_queue: Clearing completed resize (w: 1921 x h: 1081). It took 12 MS.
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Not resizing. Already this size. (w: 1921 x h: 1081)
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from QUEUED to COMPLETE. Previous state took 8 MS.
[20220330-02:16:04] [INFO ] dynamic_monitor_process_queue: Clearing completed resize (w: 1921 x h: 1081). It took 9 MS.
```

The relevant lines are:

```
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Not resizing. Already this size. (w: 1921 x h: 1081)
[20220330-02:16:04] [INFO ] process_dynamic_monitor_description: Processing resize to: 1921 x 1081. Advancing state from QUEUED to COMPLETE. Previous state took 8 MS.
```

So we may need both this and your xorgxrdp change to fix it.

Nexarian:
@matt335672 @metalefty @aquesnel Please take a look!

matt335672:
Good stuff - thanks @Nexarian.

I'll take a detailed look after the weekend.

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
         return 1;
     }
 
-    struct display_size_description *description =
-        (struct display_size_description *)
-        g_malloc(sizeof(struct display_size_description), 1);
+    dynamic_description = (struct dynamic_monitor_description *)
+                          g_malloc(sizeof(struct dynamic_monitor_description), 1);
+
+    error = libxrdp_process_monitor_stream(s, &(dynamic_description->description), 1);
```
aquesnel: Why is this error code being ignored?

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    switch (description->state)
+    {
+        case WMRZ_QUEUED:
+            advance_resize_state_machine(mm->resize_state_machine, description, WMRZ_ENCODER_DELETE);
+            break;
+        case WMRZ_ENCODER_DELETE:
+            // Disable the encoder until the resize is complete.
+            if (mm->encoder != NULL)
+            {
+                xrdp_encoder_delete(mm->encoder);
```
aquesnel, line 1209: I didn't look at the module manager that uses the encoder code, so just looking at this code deleting the encoder feels like we are putting the module manager into an invalid or inconsistent state. Can you please explain why a single step of the state machine doesn't move the system completely from the current screen size to the new screen size?

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    while (self->resize_queue->count > 0)
+    {
+        description = (struct dynamic_monitor_description *)list_get_item(self->resize_queue, 0);
+        if (description->state == WMRZ_COMPLETE)
+        {
+            LOG(LOG_LEVEL_INFO, "dynamic_monitor_process_queue: Clearing completed resize (w: %d x h: %d). It took %d MS.", description->description.session_width, description->description.session_height, g_time3() - description->start_time);
+        }
+        else if (description->state == WMRZ_ERROR)
+        {
+            LOG(LOG_LEVEL_INFO, "dynamic_monitor_process_queue: Clearing completed resize (w: %d x h: %d)", description->description.session_width, description->description.session_height);
```
aquesnel: Typo: s/clearing completed/clearing errored/

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+        }
+        else if (description->description.session_width <= 0 && description->description.session_height <= 0)
+        {
+            LOG(LOG_LEVEL_INFO, "dynamic_monitor_process_queue: Not allowing resize due to invalid dimensions (w: %d x h: %d)", description->description.session_width, description->description.session_height);
```
aquesnel: Why is this condition of invalid bounds handled in both this function and the process_dynamic_monitor_description function? Can we simplify this by handling this condition only in process_dynamic_monitor_description and seeing the state to WMRZ_ERROR?

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+        return 0;
+    }
+    if (description->state == WMRZ_QUEUED && description->description.session_width == wm->client_info->display_sizes.session_width && description->description.session_height == wm->client_info->display_sizes.session_height)
+    {
+        LOG(LOG_LEVEL_INFO, "process_dynamic_monitor_description: Not resizing. Already this size. (w: %d x h: %d)", description->description.session_width, description->description.session_height);
+        advance_resize_state_machine(mm->resize_state_machine, description, WMRZ_COMPLETE);
+        return 0;
+    }
+
+    switch (description->state)
```
aquesnel, line 1203: When I read the issue and commit description about a state machine to coordinate server screen resizes, I assumed that there would be a single resize state for the screen. Having the resize state stored in each resize request description is surprising since it means that the actual state of the screen (and supporting components) can be out of sync with the resize state enum in the resize description request.
Can you please explain why you went with sorting the resize state enum in the description  instead of with the screen that is being resized?

aquesnel (commented):
@Nexarian thanks for this stability fix. There are a few things I didn't understand about the design choices in this change, and I've left comments for the things that I don't understand. Can you please help me to understand the choices made in this change.

Nexarian:
@aquesnel Extremely busy this weekend, but I should start to be able to respond to your comments around  Tuesday!

xrdp/xrdp.h
```diff
@@ -397,18 +397,38 @@ xrdp_bitmap_compress(char *in_data, int width, int height,
...
+    WMRZ_XRDP_CORE_RESIZE,
+    WMRZ_ENCODER_CREATE,
+    WMRZ_SERVER_INVALIDATE,
+    WMRZ_COMPLETE,
+    WMRZ_ERROR
+};
+
+#define XRDP_DISPLAY_RESIZE_STATE_TO_STR(status) \
+    ((status) == WMRZ_QUEUED ? "QUEUED" : \
+     (status) == WMRZ_ENCODER_DELETE ? "WMRZ_ENCODER_DELETE" : \
```
matt335672: Small point - this string is incompatible with the others in that it has a WMRZ_ prefix

xrdp/xrdp_mm.c
```diff
@@ -1023,138 +1034,21 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
-    wm->client_info->display_sizes.session_height = description->session_height;
-    g_memcpy(wm->client_info->display_sizes.minfo,
-             description->minfo,
-             sizeof(struct monitor_info)
-             * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
-    g_memcpy(wm->client_info->display_sizes.minfo_wm,
-             description->minfo_wm,
-             sizeof(struct monitor_info)
-             * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
+    LOG(LOG_LEVEL_INFO,
```
matt335672: LOG_DEVEL() ? This doesn't look like user-specific logging.

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+
+    error = libxrdp_process_monitor_stream(s, &(dynamic_description->description), 1);
+    list_add_item(wm->mm->resize_queue, (tintptr)dynamic_description);
+    dynamic_description->start_time = g_time3();
+    advance_resize_state_machine(wm->mm->resize_state_machine, dynamic_description, WMRZ_QUEUED);
+    return 0;
+}
+
+const char *
+wm_login_state_to_str(enum wm_login_state login_state);
```
matt335672: A function prototype not in a header is rarely a good idea. You could rename this function as `xrdp_wm_login_state_to_str()` and add the prototype to xrdp/xrdp.h along with all the other wm functions.

A better idea might be to replace this code which occurs twice with minor variations:-

```c
    if (wm->login_state != WMLS_CLEANUP && wm->login_state != WMLS_INACTIVE)
    {
        LOG(LOG_LEVEL_INFO, "process_dynamic_monitor_description: Not trying to resize while logging in! State was %s", wm_login_state_to_str(wm->login_state));
        return 0;
    }
```

with a call to a function like `xrdp_wm_can_resize()`:-

```c
int
xrdp_wm_can_resize(struct xrdp_wm *self)
{
    int rv = (self->login_state != WMLS_CLEANUP && self->login_state != WMLS_INACTIVE);
    if (!rv)
    {
        LOG(LOG_LEVEL_INFO, "Not trying to resize while logging in!");
        LOG_DEVEL(LOG_LEVEL_INFO, "State is %s", wm_login_state_to_str(self->login_state));
    }
    return rv;
}
```

That better keeps the encapsulation of the `login_state` member. The above is off the top of my head, but I'm sure you get the idea. Prototype for `xrdp_wm_can_resize()` lives in xrdp/xrdp.h.

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    if (module == NULL)
+    {
+        return 1;
+    }
+
+    if (wm->login_state != WMLS_CLEANUP && wm->login_state != WMLS_INACTIVE)
+    {
+        LOG(LOG_LEVEL_INFO, "process_dynamic_monitor_description: Not trying to resize while logging in! State was %s", wm_login_state_to_str(wm->login_state));
+        return 0;
+    }
```
matt335672, line 1196: See above.

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+{
+    struct dynamic_monitor_description *description;
+
+    LOG_DEVEL(LOG_LEVEL_TRACE, "dynamic_monitor_process_queue:");
+
+    if (self == 0)
+    {
+        return 0;
+    }
+    if (self->wm->login_state != WMLS_CLEANUP && self->wm->login_state != WMLS_INACTIVE)
```
matt335672: See above

xrdp/xrdp_mm.c
```diff
@@ -1257,11 +1369,13 @@ xrdp_mm_drdynvc_up(struct xrdp_mm *self)
...
         LOG(LOG_LEVEL_INFO, "User has disabled dynamic resizing.");
+        return error;
     }
-    else
-    {
-        error = dynamic_monitor_initialize(self);
-    }
+    ignore_marker = (struct display_size_description *)
+                    g_malloc(sizeof(struct display_size_description), 1);
+    g_memset(ignore_marker, 0, sizeof(struct display_size_description));
```
matt335672: Unnecessary call to `g_memset()`. Passing `1` as the second parameter of `g_malloc()` [does this for you](/neutrinolabs/xrdp/blob/cd1af4772cb6223c1348b0c6af953e45a533c2d1/common/os_calls.c#L177-L193)

xrdp/xrdp_wm.c
```diff
@@ -2231,8 +2231,7 @@ xrdp_wm_check_wait_objs(struct xrdp_wm *self)
 }
 
 /*****************************************************************************/
-
-static const char *
+const char *
```
matt335672, line 2233: See above.

Nexarian:
@matt335672 Thanks for the review, while I work on addressing the comments from both of you, would you be able to test and verify that you don't see the MSTSC crash with this as well? I want to make sure someone else other than me can verify that it does fix what's intended here.

xrdp/xrdp_mm.c
```diff
@@ -1023,138 +1034,21 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
-    g_memcpy(wm->client_info->display_sizes.minfo,
-             description->minfo,
-             sizeof(struct monitor_info)
-             * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
-    g_memcpy(wm->client_info->display_sizes.minfo_wm,
-             description->minfo_wm,
-             sizeof(struct monitor_info)
-             * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS);
+    LOG(LOG_LEVEL_INFO,
+        "process_dynamic_monitor_description:"
```
matt335672: Probably mean `advance_resize_state_machine` here

matt335672 (commented):
On a general point, there are some lines quite a bit over 80 columns here. Can you make an effort to break them where appropriate to resize?

Thanks again for all the work you're putting in to this.

Nexarian:
Also I will fix the "previous state took [very large number] MS" log messages.

Now that you've verified that it actually fixes the issue I think it does, I'll get to work on cleanup!

xrdp/xrdp_mm.c
```diff
@@ -1192,24 +1094,233 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    if (self->resize_queue == 0)
+    {
+        // Potentially the client does not support resizing, so the queue was never created.
+        return 0;
+    }
+    if (self->resize_queue->count <= 0)
+    {
+        return 0;
+    }
+
```
matt335672, line 1335: The above test isn't needed, as it's done by the `while()` loop below.

xrdp/xrdp_mm.c
```diff
@@ -1244,6 +1355,7 @@ dynamic_monitor_initialize(struct xrdp_mm *self)
 int
 xrdp_mm_drdynvc_up(struct xrdp_mm *self)
 {
+    struct display_size_description *ignore_marker;
```
matt335672: Should be `struct dynamic_monitor_description *`

`struct display_size_description` is smaller than `struct dynamic_monitor_description`. Using the smaller structure results in invalid memory accesses in `dynamic_monitor_process_queue()` when the pointer is cast to `struct dynamic_monitor_description *`

Valgrind found this one.

xrdp/xrdp_mm.c
```diff
@@ -1257,11 +1369,13 @@ xrdp_mm_drdynvc_up(struct xrdp_mm *self)
...
     {
         LOG(LOG_LEVEL_INFO, "User has disabled dynamic resizing.");
+        return error;
     }
-    else
-    {
-        error = dynamic_monitor_initialize(self);
-    }
+    ignore_marker = (struct display_size_description *)
+                    g_malloc(sizeof(struct display_size_description), 1);
```
matt335672: The above two should be `struct dynamic_monitor_description` as mentioned above.

matt335672 (commented):
A couple more for you which I found while testing #2219 against #1928

xrdp/xrdp_mm.c
Nexarian: Done.

xrdp/xrdp_mm.c
Nexarian: Fixed.

xrdp/xrdp_mm.c
Nexarian: Yes. Fixed.

xrdp/xrdp_mm.c
Nexarian: Fixed.

xrdp/xrdp_mm.c
Nexarian: The logic here is that when a resize request comes in, we don't process it immediately, we store it in a queue and deal with it later. The data that comes in from a resize request is of the same format as that of any type of display description. So, store the display description that is targeted in the queue, perform the resize, and then when we're done, overwrite whatever the previous config was with the new one.

Maybe the word "description" is misleading? I could change it to something else. Otherwise, I don't fully grok your concern?

xrdp/xrdp_mm.c
Nexarian: There are several reasons:
1. We may or may not have an encoder in the module manager. Some types of XRDP rendering don't use it. Therefore, having no encoder is a valid state. You could argue that deleting an encoder when there was one previously IS invalid, however. But I don't have a ton of choice here because
2. This is engineering for the future when the EGFX code for hardware acceleration gets merged in. For this specific change, yes, I could delete and re-create the encoder in one step, but the way XRDP is architected right now, we can't do that for EGFX, which I have a prototype branch of here: https://github.com/Nexarian/xrdp/tree/mainline_merge/xrdp
3. After extensive testing, the EGFX resizing setup requires that you delete the EGFX connection and reconstitute it. The way XRDP is architected right now, if a frame comes into the encoder, it processes it. Also, the encoder has the width and height of the screen(s?) hard coded at creation, and when it comes to H264, I'm not sure that's dynamically changeable. Deleting the EGFX connection requires a state machine, as you can't re-create one until the old one has been shut down (and the Microsoft Clients get REALLY angry if you try to break this order).

What I could do is architect a way to not delete the encoder, put some sort of pause or disable flag in it so it stops processing new frames, and then delete and re-create it in one go before invalidate is sent to xorgxrdp. However, I'm not sure if a "disabled" encoder is better than a "deleted/NULL" encoder...

xrdp/xrdp_mm.c
Nexarian: Done.

xrdp/xrdp_mm.c
Nexarian: Done.

xrdp/xrdp_mm.c
Nexarian: Done.

xrdp/xrdp_mm.c
Nexarian: Great catch! This code has been in the works for a while so I missed that :)

xrdp/xrdp.h
Nexarian: Fixed.

xrdp/xrdp_mm.c
Nexarian: It isn't anymore! Good catch :)

xrdp/xrdp_mm.c
Nexarian: For more information, see how I discovered this here: https://github.com/neutrinolabs/xrdp/issues/1422#issuecomment-830761288

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/bdc90b21-4b14-43bc-9c03-b7fecbfc6a1f
https://github.com/ogon-project/ogon/blob/master/rdp-server/frontend.c#L579

xrdp/xrdp_mm.c
Nexarian: Fixed.

xrdp/xrdp_wm.c
Nexarian: Fixed.

xrdp/xrdp_mm.c
Nexarian: Fixed.

xrdp/xrdp_mm.c
Nexarian: Fixed.

Nexarian:
@matt335672 @aquesnel Take another look! I've updated and addressed all the comments I could find.

matt335672:
Thanks @Nexarian - I'll take a look at this tomorrow, now.

Sorry about the conflict - I think merging #2233 was to blame.

xrdp/xrdp_mm.c
```diff
@@ -1152,10 +1050,19 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
     struct xrdp_process *pro;
     struct xrdp_wm *wm;
     int monitor_layout_size;
+    struct dynamic_monitor_description *dynamic_description;
 
     LOG_DEVEL(LOG_LEVEL_TRACE, "dynamic_monitor_data:");
     pro = (struct xrdp_process *) id;
     wm = pro->wm;
+
+    if (wm->client_info->suppress_output == 1)
```
aquesnel, line 1088: I'm not familiar with suppress output, so I don't understand why processing a resize message would conflict with not having any output. Can you please explain why this restriction exists.

xrdp/xrdp_mm.c
```diff
@@ -1174,26 +1081,279 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
+    }
+    error = libxrdp_process_monitor_stream(s,
+                                           &(dynamic_description->description),
+                                           1);
+    if (error)
     {
-        error = process_dynamic_monitor_description(wm, description);
+        LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: libxrdp_process_monitor_stream"
+            " failed with error %d.", error);
+        return error;
```
aquesnel: I think the `dynamic_description` struct needs to be freed here to prevent a memory leak.

xrdp/xrdp_mm.c
```diff
@@ -61,13 +62,21 @@ struct xrdp_mm *
...
     self->login_names = list_create();
     self->login_names->auto_free = 1;
     self->login_values = list_create();
     self->login_values->auto_free = 1;
+    self->resize_queue = list_create();
+
+    pid = g_getpid();
+    /* setup wait objects for signalling */
+    g_snprintf(buf, 1024, "xrdp_%8.8x_resize_state_machine", pid);
+    self->resize_state_machine = g_create_wait_obj(buf);
```
aquesnel: `resize_state_machine` as a name for a wait object makes the purpose of the wait object unclear. When this wait object is set, what does it mean about the resize?
I think this wait object is trying to represent that a resize request is ready and waiting to be processed, so I suggest renaming the wait object to  `resize_ready`

xrdp/xrdp.h
```diff
@@ -397,18 +401,38 @@ xrdp_bitmap_compress(char *in_data, int width, int height,
...
-    int physical_width;
-    int physical_height;
-    int orientation;
-    int desktop_scale_factor;
-    int device_scale_factor;
+    struct display_size_description description;
+    enum display_resize_state state;
+    int last_state_update_timestamp;
+    int start_time;
 };
```
aquesnel: The name of this struct is confusing:
* I don't know what "dynamic" means in this context 
* this struct seems to only be used for resize operations

I'd suggest renaming this to `resize_monitor_state` but then there is confusion between this and the `display_resize_state`. To have some clear naming here we will need to put more thought into it.

xrdp/xrdp_wm.c
```diff
@@ -2267,10 +2266,29 @@ int
...
     return 0;
 }
+
+int
+xrdp_wm_can_resize(struct xrdp_wm *self)
+{
+    if (self->login_state != WMLS_CLEANUP
+            && self->login_state != WMLS_INACTIVE)
+    {
+        LOG(LOG_LEVEL_INFO, "Not trying to resize while logging in!");
```
aquesnel: This log message is a bit unclear to a sys admin and is inconsistent with the other error messages in this function, I suggest something more straight forward like "Not allowing resize. login in progress"

xrdp/xrdp_mm.c
aquesnel: My concern is that for the resize to be processed correctly we need to strictly process one resize request at a time to completion before starting the next resize request. With the current code it is possible to represent a resize state like the following: (and this example it would be a bug in the resize code)

```
xrdp_mm = {
    resize_queue = [
        dynamic_monitor_description = { display_resize_state = WMRZ_SERVER_VERSION_MESSAGE, ...},
        dynamic_monitor_description = { display_resize_state = WMRZ_QUEUED, ...},
        dynamic_monitor_description = { display_resize_state = WMRZ_ENCODER_DELETE, ...},
    ],
    ...
}
```

To avoid this situation the code has to be carefully written to ensure the invariants that only one dynamic_monitor_description has a state other than `WMRZ_QUEUED` at a time, and that 

The two ways that I deal with code that must be carefully written to avoid bugs is to encapsulate the critical path code that updates the state, and the other is to make invalid states unrepresentable. Here is how I would change the code to make the invalid state unrepresentable:
* move the `enum display_resize_state state` from the `dynamic_monitor_description` struct to the `xrdp_mm` struct
* add a `display_size_description display_description_resize_in_progress` to the `xrdp_mm` struct
* remove the explicit `WMRZ_QUEUED` state and replace its concept with the `display_size_description` just being in the resize_queue list

This would make the xrdp_mm struct look like:
```
xrdp_mm = {
    display_resize_state = WMRZ_SERVER_VERSION_MESSAGE,
    display_description_resize_in_progress = { /* display_size_description struct */ },

    resize_queue = [
        display_size_description = { /* display_size_description struct */ },
        display_size_description = { /* display_size_description struct */ },
    ],
    ...
}
```

Using this representation of the resize in progress it is much easier to enforce the invariant that there is only 1 resize in progress at a time, and the resize is processed to completion before the next resize is processed.

Nexarian:
@matt335672 The merge conflict was due to the fact that I got a little over-zealous about trying to make more lines conform to the 80 character limit as you had mentioned earlier. I've rebased and force pushed again. Should be good to go!

xrdp/xrdp_mm.c
aquesnel: I think that having a disabled encoder is much safer since you have said that different encoders have different workflows for being cleaned up and deleted. My concern is that deleting the encoder without synchronizing with the code that uses the encoder can leaver the other code in an invalid state and potentially in a race condition or use-after-free situation.

To me this feels like there is an interface that is missing for delegating to the encoder to clean itself up. I'm thinking of an interface like `encoder.try_cleanup()` that can be called repeatedly until it succeeds. For example:

```
resize_display() {
// ...
switch(resize_sate):
    case WMRZ_DELETE_ENCODER:
        if (encoder.try_cleanup() == SUCCESS) {
            advance_resize_state(resize_sate, WMRZ_SERVER_MONITOR_RESIZE);
        } else {
            LOG_DEVEL(LOG_LEVEL_INFO, "encoder cleanup still in progress, waiting before trying the encoder clean up again");
        }
        break;
//...
}
```

This is just my naive thoughts. Since you've been looking into the code quite a bit, what it is about the xrdp architecture that is preventing a solution that would allow just deleting an recreating the encoder safely in one step?

xrdp/xrdp_mm.c
```diff
@@ -61,13 +62,21 @@ struct xrdp_mm *
...
     self->wm = owner;
     self->login_names = list_create();
     self->login_names->auto_free = 1;
     self->login_values = list_create();
     self->login_values->auto_free = 1;
+    self->resize_queue = list_create();
+
+    pid = g_getpid();
+    /* setup wait objects for signalling */
+    g_snprintf(buf, 1024, "xrdp_%8.8x_resize_state_machine", pid);
```
matt335672: Small point - suggest `sizeof(buf)` rather than 1024, for maintainability reasons.

xrdp/xrdp_mm.c
matt335672: I agree that the resize state is better stored in xrdp_mm, as apart from the resize being processed, every other item should be `WMRZ_QUEUED`, and so doesn't need an explicit state, which is another way of saying what @aquesnel has said above.

Using an explicit `display_description_resize_in_progress` is very clear, but you could also just use the head of the queue as the struct being worked on. There are advantages either way.  Using the head of the queue may save you a state, as you can use the queue length being > 0 as an indication that a resize is in progress.

matt335672 (commented):
Thanks for this @Nexarian.

aquesnel:
@Nexarian thanks for the updates

xrdp/xrdp_mm.c
Nexarian: Fixed.

xrdp/xrdp_wm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Here is the documentation on the suppress output functionality. https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/79202daa-76b8-4a09-ac67-fc3cae7d895f

Essentially: The client can send a message to the server to prevent any display output updates from the server. Though thinking on this more, it only says that display output should be suppressed, it's reasonable to assume that we should still process resizing messages during this.

I'll remove.

xrdp/xrdp_mm.c
Nexarian: Correct. Fixed.

xrdp/xrdp_mm.c
Nexarian: Sounds reasonable.

xrdp/xrdp.h
Nexarian: I took a cue from the spec: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/d2954508-f487-48bc-8731-39743e0854a9

How's `display_control_monitor_layout_data`?

Nexarian:
I'm still working on and thinking about how to address some of the feedback here. Standby for a few more days :)

matt335672:
Thanks @Nexarian 

Please let me know if you need a hand looking at the session startup delay you've mentioned on Gitter and I'll give it a priority.

Nexarian:
@aquesnel Your questions were really hard for me to answer! I'm working on a wiki to explain exactly what is going on, so we can review and discuss it.

Nexarian:
Accidentally closed, hit the wrong button :)

Nexarian:
@aquesnel @matt335672 Can you take a look at [this](https://github.com/neutrinolabs/xrdp/wiki/On-EGFX-architecture-and-Dynamic-Monitors), especially [this](https://github.com/neutrinolabs/xrdp/wiki/On-EGFX-architecture-and-Dynamic-Monitors#special-topic-deleting-and-re-creating-the-encoder-in-the-egfx-workflow-for-resizing), and let me know what you think about deleting and re-creating the encoder after understanding my documentation?

matt335672:
@Nexarian - gimme a few days and I'll get to it.

matt335672:
@Nexarian - that looks good to me.I can't say I fully understand it on a first reading, but you've done a good job of making the case for the 'acceleration assistant', and the UML diagram is extremely helpful.

A couple of minor comments:-
- This area changes pretty quickly, so adding some dates in at various places might be useful. At some stage, when the wiki gets revised, it will be useful to know when particular technologies were being used.
- You may want to give some thought into making the source for the UML diagram available somewhere.

Thanks.

xrdp/xrdp_mm.c
Nexarian: @matt335672 @aquesnel I spent a lot of time thinking about the right way to do this. I settled on this:
- Queue up only the `display_size_description` struct, as @aquesnel proposed.
- In xrdp_mm, now have `display_control_monitor_layout_data` which wraps `display_size_description` and contains the state and timestamps for logging. There is only one of these, and it is only ever populated if it is null. If it is not null, the state machine is processed as in the original version. This is a separate commit from the original commit so you can see how I've changed it.

xrdp/xrdp_mm.c
Nexarian: @aquesnel I'm nearly certain that the encoder delete code that already exists [here](https://github.com/neutrinolabs/xrdp/blob/devel/xrdp/xrdp_encoder.c#L142) does exactly what you want. It's already a `try_cleanup`, it blocks until the encoder is finished, deleted, and all state is cleaned up. It's very nice.

Now, the question you asked that took me over a month to answer: `What about the xrdp architecture is preventing a solution that would just allow  deleting and recreating the encoder safely in one step`. I've done my best to answer that in [this wiki section](https://github.com/neutrinolabs/xrdp/wiki/On-EGFX-architecture-and-Dynamic-Monitors#special-topic-deleting-and-re-creating-the-encoder-in-the-egfx-workflow-for-resizing).

I recognize this is a lot to read, but I think it's important that more XRDP contributors start to understand the direction that EGFX is headed and how it affects resizing, as these will become bread-and-butter features in the future.

Nexarian:
@matt335672 I have:
1. Switched the UML diagram to SVG. Now anyone with an SVG editor can change it, and it scales so much better! I used yEd (https://yed.yworks.com/) to create it. The original is in a .graphml file, but I'm not sure how to preserve that longer term. Suggestions appreciated!
2. I went through the wiki and updated major sections to include a date header. I can move the old section into an appendix when it changes appreciably.

Nexarian:
@matt335672 @aquesnel Please re-review the PR. I've updated all the comments, the main one being addressing the enumeration state as part of a queued item. I moved that out. Rebased from devel and tested, made sure the CI checks pass.

If you feel I've properly addressed the question as to why deleting the encoder when I do appropriately, are there any other blockers?

aquesnel:
Hi @Nexarian ,

I read the [wiki page you created](https://github.com/neutrinolabs/xrdp/wiki/On-EGFX-architecture-and-Dynamic-Monitors) and like Matt, I don't fully understand the section about "Special Topic: Deleting and re-creating the encoder in the EGFX workflow for resizing" nor the what is the solution that you used to minimize/avoid memory copies of encoded data.

That being said, the wiki page has convinced me that what you have made is very valuable and seems reasonably tested. I would like to fully understand this, but at this point I think that would be making "perfect the enemy of good". My recommendation is to move forward with the merge and in parallel/after continue to document and discuss the architecture of encoders in XRDP in a new issue/discussion topic.

FYI, I like https://plantuml.com/ for sequence diagrams since it is text based (which is easy source that can be shared/saved) and can produce a very similar diagram to what you have.

xrdp/xrdp_mm.c
```diff
@@ -1008,194 +1021,343 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
+    g_memset(&ls, 0, sizeof(ls));
+    ls.data = data;
+    ls.p = ls.data;
+    ls.size = bytes;
+    ls.end = ls.data + bytes;
+    s = &ls;
+    in_uint32_le(s, msg_type);
+    in_uint32_le(s, msg_length);
+    LOG(LOG_LEVEL_DEBUG, "dynamic_monitor_data: msg_type %d msg_length %d",
+        msg_type, msg_length);
```
aquesnel: This looks like a log message that is only useful for devs and not admins. I suggest changing this to `LOG_DEVL`

xrdp/xrdp_mm.c
```diff
@@ -1008,194 +1021,343 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
...
+
+    if (msg_type != DISPLAYCONTROL_PDU_TYPE_MONITOR_LAYOUT)
     {
-        LOG_DEVEL(LOG_LEVEL_ERROR,
-                  "process_dynamic_monitor_description:"
-                  " description is null.");
         return 0;
```
aquesnel: Are there other message types that this function is expected to process but is not supported?
Can you add a LOG message for the message types that are ignored here.

xrdp/xrdp_mm.c
aquesnel: I'll paraphrase what I put in a comment for the issue:
I've read the wiki you wrote fully but I don't fully understand it. I think that improving that wiki and my understanding of the architecture should be moved to a new issue/discussion so that this pull request can be unblocked.

matt335672:
The `.graphml` file can be embedded in another file (see for example https://github.com/matt335672/xrdp/wiki/Authentication-architecture-redesign), or even just moved into the wiki as is, I suspect. It's worth a play with the wiki tools to find out.

I'll get round to a review later in the week.

Thanks again for all the effort you're putting into this @Nexarian

Nexarian:
@aquesnel In short:
1. We have to re-create the encoder on resize because for at least some encoders, the size of the screen is set read-only on encoder creation.
2. If we didn't have to use a state machine for the resizing workflow, re-creating the encoder in one step is preferred. I just checked that into devel [here](https://github.com/neutrinolabs/xrdp/commit/eeb5daa3382a6640121cb94f580500ab9e611c16)
3. However, due to the impending integration of EGFX work, we need a state machine. That makes life harder, because now resizing happens in multiple asynchronous steps.
4. During that async workflow, there is no easy way to prevent invalid data from being sent to the encoder. We could add some flag/state management to XRDP, but that is an additional opportunity for bugs.
5. Thus, the solution is to shut down and delete the encoder during resize, and bring it back only after everything has settled. A NULL encoder is a supported state, and always has been as not all session types use it.
6. Microsoft clients are super sensitive to the "wrong" data being sent to them, so we have to be extremely careful as to how the state machine is constructed.
7. This involves juggling, with EGFX and HW accel integrated, the states of 3 different processes. It gets a little crazy.

I'll address everyone's feedback later this week.

xrdp/xrdp_mm.c
```diff
@@ -1093,49 +1094,46 @@ dynamic_monitor_data(intptr_t id, int chan_id, char *data, int bytes)
...
-    advance_resize_state_machine(resize_state_machine, description, WMRZ_ERROR);
+    advance_resize_state_machine(mm, description, WMRZ_ERROR);
     return error;
 }
 
 /******************************************************************************/
 static int
-process_dynamic_monitor_description(struct xrdp_wm *wm,
-                                    struct dynamic_monitor_description *description)
+process_display_control_monitor_layout_data(struct xrdp_wm *wm,
```
matt335672: Thanks for sorting out the state variable.

As there's only one `description` struct in place now, it may be clearer to remove the `description` parameter from this function and also `advance_resize_state_machine()` and `advance_error()`. That will also let you remove the double-indirections in `dynamic_monitor_process_queue()`

xrdp/xrdp_mm.c
```diff
@@ -1328,33 +1309,54 @@ dynamic_monitor_process_queue(struct xrdp_mm *self)
...
+
+            g_memcpy(&((*description)->description),
+                     (struct display_size_description *)queue_head,
+                     sizeof(struct display_size_description));
+            (*description)->start_time = g_time3();
+            advance_resize_state_machine(self->wm->mm,
+                                         *description,
+                                         WMRZ_ENCODER_DELETE);
+            list_remove_item(self->resize_queue, 0);
+            g_free((struct display_size_description *)queue_head);
```
matt335672: If you set `auto_free` for the resize_queue, you can remove this `g_free()` call as `list_remove_item()` will do it for you. That will also fix the possible memory leak in `xrdp_mm_delete()` where you are calling `list_delete()` on the `resize_queue` without checking it's empty first.

matt335672 (approved):
Here's my somewhat belated review - looking good.

Nexarian:
Once https://github.com/neutrinolabs/xrdp/pull/2301,  https://github.com/neutrinolabs/xrdp/pull/2300, and https://github.com/neutrinolabs/xrdp/issues/1928 are resolved, I'll update, rebase, and merge this.

Nexarian:
Still working on this. Found some bugs I have to work out before I'm ready to commit this.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Updated.

xrdp/xrdp_mm.c
Nexarian: Added this code block

```
 // Unsupported message types
        switch (msg_type)
        {
            case DISPLAYCONTROL_PDU_TYPE_CAPS:
                LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: Received"
                    " DISPLAYCONTROL_PDU_TYPE_CAPS. Per MS-RDPEDISP 2.2.2.1,"
                    " this is a server-to-client message, client should never"
                    " send this. Ignoring message");
                break;
            default:
                LOG(LOG_LEVEL_ERROR, "dynamic_monitor_data: Received neither"
                    " nor DISPLAYCONTROL_PDU_TYPE_CAPS. Ignoring message.");
                break;
        }
```

Hope that settles this concern.

Nexarian:
@aquesnel Created https://github.com/neutrinolabs/xrdp/issues/2313 to further continue discussions of the encoder.

Nexarian:
The only thing I haven't yet addressed is how to upload the graphml/svg for the state diagram in the wiki. That's out of scope for this PR, so I'll merge this and then we can discuss that separately.

PR #2183, opened 2022-03-20
sesman: Reorder PATH entries for xrdp session

t-a-k:
Hello,

In the xrdp session `PATH` environment variable is currently set to `/sbin:/bin:/usr/bin:/usr/local/bin`, but I think that such ordering is unusual, and it will be better if `/usr/local/bin` is searched before `/usr/bin` by default.

### DIFF
--- a/sesman/env.c
+++ b/sesman/env.c
@@ -131,7 +131,7 @@ env_set_user(const char *username, char **passwd_file, int display,
         {
             g_clearenv();
             g_setenv("SHELL", pw_shell, 1);
-            g_setenv("PATH", "/sbin:/bin:/usr/bin:/usr/local/bin", 1);
+            g_setenv("PATH", "/usr/local/bin:/usr/bin:/bin:/sbin", 1);
             g_setenv("USER", username, 1);
             g_setenv("LOGNAME", username, 1);
             g_sprintf(text, "%d", uid);


matt335672:
Hi @t-a-k,

Thanks for raising this.

I've had a look into the history of this line in git. It's only used for locating utilities which are used by xrdp itself. For most Linux operating systems, utilities will be in "/bin" or "/usr/bin". "/usr/local/bin" is used on FreeBSD for locating ports. /sbin is used on FreeBSD for locating "/sbin/mount_fusefs" (see #387 for the discussion on this).

For a normal user session, this setting is (or should be) irrelevant. Before the user session starts, the PATH will be set by either `pam_env.so`, or by the login shell scripts.

Is there a particular situation you are finding yourself in where the ordering of the components of this line is making a difference?

t-a-k:
> For most Linux operating systems, utilities will be in "/bin" or "/usr/bin". "/usr/local/bin" is used on FreeBSD for locating ports.

As far as I know, `/usr/local/bin` is also used by most Linux distributions for programs local to the site ([Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html) also standardizes this).  And, in order to allow site admins to override standard utilities with their custom versions (or create wrapper scripts), it is important to place `/usr/local/bin` before standard directories like `/usr/bin` in PATH.

> For a normal user session, this setting is (or should be) irrelevant. Before the user session starts, the PATH will be set by either `pam_env.so`, or by the login shell scripts.
> 
> Is there a particular situation you are finding yourself in where the ordering of the components of this line is making a difference?

Some Linux distributions like CentOS Linux 7 don't use `pam_env.so` to initialize PATH, and their user sessions seem to inherit PATH set by session managers like sshd or xrdp-sesman.  (CentOS 7's default login scripts also will not initialize PATH, only prepend or append directories to it.)
So, console logins (and SSH logins) will get PATH like `/usr/local/bin:/usr/bin` while xrdp logins will get `/sbin:/bin:/usr/bin:/usr/local/bin`.  When I locally installed recent version of Git as `/usr/local/bin/git`, plain `git` command is `/usr/local/bin/git` (as expected) on console/ssh sessions, but `/bin/git` on xrdp sessions, which confuses users.

Furthermore, I heard that Solaris has both `/bin/sh` and `/sbin/sh`, so with the current PATH of sesman plain `sh` will be `/sbin/sh` which is not for general use (it is statically linked for emergency shell).  From this as well, I think that `/sbin` should come last in PATH.

matt335672:
There are definitely plusses and minuses to this.

From a developer/maintainer perspective, there are few positives to this PR. We're going to be juggling entries to try to satisfy all uses - there are no guarantees that a single order will satisfy all users. You're also imposing quite a testing burden on us, and also introducing the possibility of calls from Linux users who have overridden legitimate utilities with local 'specials'.

On the other hand, we've had PATH ordering issues in the past, like #1773, where polkit was refusing to authenticate programs picked up from a different place in the PATH - this is a result of [UsrMerge](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/). The Ubuntu case has ultimately been solved by the Debian package moving to `pam_env.so` [ see [Debian BTS #996418](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=996418)

In summary, changing this line will address your problem, but introduces problems for other people. I think the best way forward is to fix this elsewhere, possibly in startwm.sh, where we can make platform-specific fixes with limited change effects.

There's the following line in `/etc/libexec/xrdp/startwm.sh`:-

```
# change PATH to be what your environment needs usually what is in
# /etc/environment
#PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games"
#export PATH=$PATH
```

On my CentOS 7 system, if I set this to :-

```
export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
```

I get this PATH in both an XRDP session and a GNOME session:-

```
/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin:/home/testuser/.local/bin:/home/testuser/bin
```

I can't test easily in lightdm right now.

Isn't this what you want?

t-a-k:
Sorry for delayed response, but that makes sense to me.
Please let me cancel this PR.

Issue #1855, opened 2021-04-08
xrdp_wm_log_msg: Error connecting to sesman: 127.0.0.1 port: 3350


min050820:
I believe this issue is the same as #1777. I opened new issue because I'm okay with installing nmap.

My system is aarch64 (same as that issue), running ubuntu 20.04 with linux 4.9.
_Perhaps this issue has something to do with aarch64?_

xrdp-sesman service is running. I also tried restarting them.
My firewall also is okay. I don't have ufw, and tried flushing iptable.

Result of what @matt335672 suggested:
```
root@odroid:~# ps -e -o pid,netns,command | grep xrdp
   2760 4026531955 /usr/sbin/xrdp-sesman
   2860 4026531955 /usr/sbin/xrdp
  25022 4026531955 grep --color=auto xrdp

root@odroid:~# ss -alp 'sport = 3350'
RTNETLINK answers: Invalid argument
Netid      State       Recv-Q      Send-Q            Local Address:Port             Peer Address:Port      Process
tcp        LISTEN      0           2                         [::1]:3350                     [::]:*          users:(("xrdp-sesman",pid=2760,fd=7))

root@odroid:~# nmap -p 3350 127.0.0.1
Starting Nmap 7.80 ( https://nmap.org ) at 2021-04-08 15:24 KST
Nmap scan report for localhost (127.0.0.1)
Host is up (0.00011s latency).

PORT     STATE  SERVICE
3350/tcp closed findviatv

Nmap done: 1 IP address (1 host up) scanned in 0.17 seconds
```

xrdp.log:
```
[20210408-15:16:08] [INFO ] connected client computer name: DESKTOP-URBOSA
[20210408-15:16:08] [INFO ] adding channel item name rdpdr chan_id 1004 flags 0x80800000
[20210408-15:16:08] [INFO ] adding channel item name rdpsnd chan_id 1005 flags 0xc0000000
[20210408-15:16:08] [INFO ] adding channel item name cliprdr chan_id 1006 flags 0xc0a00000
[20210408-15:16:08] [INFO ] adding channel item name drdynvc chan_id 1007 flags 0xc0800000
[20210408-15:16:08] [INFO ] TLS connection established from ::ffff:218.158.49.161 port 51387: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384
[20210408-15:16:08] [DEBUG] xrdp_00006152_wm_login_mode_event_00000001
[...]

Issue #1805, opened 2021-02-11
xrdp process seems to be using wrong IP stack for connection to sesman


BrianDead:
I'm trying to get to the bottom of a problem where xrdp has stopped working for me recently. The symptom is that when I connect (using Microsoft remote desktop from a Mac) I just get a blank black screen. It was working great in the past.

Looking in the logs, I can see that at the point where xrdp is trying to connect to sesman to initiate an Xorg session, it fails - I see no corresponding connection received in the xrdp-sesman.log. In the xrdp log I see error messages about closed connections, suggesting that the xrdp process is binding to an IPv6 stack on the interface even though it's trying to connect to 127.0.0.1

xrdp-sesman.log
```
Feb 10 19:56:48 plethora xrdp[21818]: (21818)(140681374807872)[INFO ] TLS connection established from 10.46.162.2 port 51147: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384
Feb 10 19:56:48 plethora xrdp[21818]: (21818)(140681374807872)[DEBUG] xrdp_0000553a_wm_login_mode_event_00000001
Feb 10 19:56:48 plethora xrdp[21818]: (21818)(140681374807872)[INFO ] Loading keymap file /etc/xrdp/km-00000809.ini
Feb 10 19:56:48 plethora xrdp[21818]: (21818)(140681374807872)[WARN ] local keymap file for 0x00000809 found and doesn't match built in keymap, using local keymap file
Feb 10 19:56:48 plethora xrdp[21818]: (21818)(140681374807872)[DEBUG] xrdp_wm_log_msg: connecting to sesman ip 127.0.0.1 port 3350
Feb 10 19:56:52 plethora xrdp[21818]: (21818)(140681374807872)[DEBUG] Closed socket 18 (AF_INET6 fe80::cf20:a5a:7fa0:72e3 port 44516)
Feb 10 19:56:56 plethora xrdp[21818]: (21818)(140681374807872)[DEBUG] Closed socket 18 (AF_INET6 fe80::cf20:a5a:7fa0:72e3 port 44518)
Feb 10 19:57:00 plethora xrdp[21818]: (21818)(140681374807872)[DEBUG] Closed socket 18 (AF_INET6 fe80::cf20:a5a:7fa0:72e3 port 44520)
[...]

Issue #1482, opened 2020-01-16
xrdp fails with black screen if pam_namespace is enabled


mjungowski:
Running XRDP 0.9.11 on CentOS 7 I am having issues with pam_namespace.so. Basically, as soon as I enable pam_namespace.so in /etc/pam.d/xrdp-sesman XRDP sessions (both Xorg and XVNC) fail with a black screen. After about 5 minutes or so this is what I see in /var/log/xrdp.log:

> [20200116-22:34:10] [DEBUG] xrdp_wm_log_msg: connection problem, giving up
> [20200116-22:34:10] [DEBUG] Closed socket 19 (AF_UNIX)
> [20200116-22:34:10] [DEBUG] xrdp_wm_log_msg: some problem
> [20200116-22:34:10] [DEBUG] xrdp_mm_module_cleanup
> [20200116-22:34:10] [DEBUG] Closed socket 18 (AF_INET 127.0.0.1:44860)

Uncommenting the pam_namespace.so line in my xrdp-sesman pam.d configfile and restarting XRDP immediately restores full XRDP functionality save for polyinstantiation, which is required on this multi-user Linux terminal server. It is working great when users log in via our thin clients using XDMCP and XDM and it's also working great for SSH logins, so I know for sure it's not a PAM issue per se but must be XRDP-related somehow.

I know that it is at least partly working because I can see the proper .X11-unix and .X10-lock files created in _/tmp-inst/system_u:object_r:tmp_t:s0_martin/_ instead of _/tmp_ with pam_namespace.so enabled. However, when logging in from a thin client via XDM or via SSH from a remote machine the polyinstantiated tmp-directory is a different one: _/tmp/system_u:object_r:tmp_t:s0-s0:c0.c1023_martin_. Not sure if that makes a difference or not but I thought I'd mention it either way.

My configuration:

```
[ root @ ebdlsa001 ] ~ > cat /etc/pam.d/xrdp-sesman 
#%PAM-1.0
# Generic Fedora config
auth       include      password-auth
auth	   required		pam_env.so
account    include      password-auth
password   include      password-auth
session    required     pam_selinux.so close
session    required     pam_selinux.so open
session    required     pam_loginuid.so
session    required     pam_namespace.so
session    include       password-auth
```

```
[ root @ ebdlsa001 ] ~ > ls -Z /usr/sbin/xrdp*
-rwxr-xr-x. root root system_u:object_r:bin_t:s0       /usr/sbin/xrdp
[...]

PR #2207, opened 2022-04-01
Move to Unix Domain Socket for SCP (sesman)

### LINKED ISSUE #1596: xrdp - vnc connection does not work in pure IPv4 scenario
Hi, I have configured xrdp to communicate with a x11vnc server in localhost. I noticed that in some machines that I managed this didn't work. I noticed through tcpdump that xrdp was trying to communicate with vnc using IPv6. The machines where this didn't work have IPv6 disabled through firewall rules.
xrdp.ini:

```
[globals]
ini_version=1
bitmap_cache=yes
bitmap_compression=yes
port=3389
crypt_level=low
channel_code=1
max_bpp=24
ssl_protocols=TLSv1.2, TLSv1.3
allow_channels=true
[Logging]
LogFile=xrdp.log
LogLevel=DEBUG
EnableSyslog=true
SyslogLevel=DEBUG
[Channels]
; Channel names not listed here will be blocked by XRDP.
; You can block any channel by setting its value to false.
; IMPORTANT! All channels are not supported in all use
; cases even if you set all values to true.
; You can override these settings on each session type
; These settings are only used if allow_channels=true
rdpdr=true
rdpsnd=true
drdynvc=true
cliprdr=true
rail=true
xrdpvr=true
tcutils=true
[Console]
name=console
lib=libvnc.so
ip=127.0.0.1
port=5900
password=ask
```

Command used to launch x11vnc:

```
x11vnc -listen 127.0.0.1 -no6 -rfbportv6 -1 -accept popup
```

From my understanding, with these configurations there should be no reason for xrdp to be using IPv6, and still I see this on tcpdump:

```
18:52:58.852891 IP6 ip6-localhost.51474 > ip6-localhost.5900: Flags [S], seq 2933527359, win 65476, options [mss 65476,sackOK,TS val 2368947658 ecr 0,nop,wscale 7], length 0
18:52:58.852925 IP6 ip6-localhost.5900 > ip6-localhost.51474: Flags [R.], seq 0, ack 2933527360, win 0, length 0
18:52:58.953630 IP localhost.54790 > localhost.5900: Flags [S], seq 2933527362, win 65495, options [mss 65495,sackOK,TS val 1807070734 ecr 0,nop,wscale 7], length 0
18:52:58.953663 IP localhost.5900 > localhost.54790: Flags [S.], seq 143156007, ack 2933527363, win 65483, options [mss 65495,sackOK,TS val 1807070734 ecr 1807070734,nop,wscale 7], length 0
18:52:58.953695 IP localhost.54790 > localhost.5900: Flags [.], ack 1, win 512, options [nop,nop,TS val 1807070734 ecr 1807070734], length 0
18:52:59.083123 IP localhost.5900 > localhost.54790: Flags [P.], seq 1:13, ack 1, win 512, options [nop,nop,TS val 1807070863 ecr 1807070734], length 12
18:52:59.083165 IP localhost.54790 > localhost.5900: Flags [.], ack 13, win 512, options [nop,nop,TS val 1807070863 ecr 1807070863], length 0
18:52:59.083580 IP localhost.54790 > localhost.5900: Flags [P.], seq 1:13, ack 13, win 512, options [nop,nop,TS val 1807070864 ecr 1807070863], length 12
18:52:59.083754 IP localhost.5900 > localhost.54790: Flags [.], ack 13, win 512, options [nop,nop,TS val 1807070864 ecr 1807070864], length 0
```

I would like to have xrdp communicate purely through IPv4 with vnc. Is there a way to do this? I haven't find any options to control this.

Thanks in advance

matt335672:
Depends on #2204 

Fixes : #1596 #1805 #1855

This draft PR implements sesman listening on a UNIX domain socket rather than TCP port 3350

It's currently in draft, as more testing is required. I thought however it was worth having visibility of this while I work on it.

Noteworthy features:-
1) The existing sockets directory is unsuitable for the sesman socket, as the file permissions mean anyone could create `sesman.socket` which is essentially what  CVE-2020-4044 was all about. So I've added another configuration value `--with-sesmanruntimedir` which defaults to `/var/run/xrdp-sesman`
2) I've changed the default value of the socketdir location to `/var/run/xrdp` rather than `/tmp/.xrdp` which better fits in with existing filesystem guidelines and also means polyinstantiation of /tmp works out-of-the-box (see #1482)
3) The server parameter has been removed from xrdp-sesrun and xrdp-sesadmin
4) Compatibility with the existing sesman and xrdp configs is maintained for upgrades, but the older settings are warned about and ignored if found.
5) I've needed to create a lock file for the UNIX domain socket, as the listen test which can be used for TCP sockets to enforce a single listener doesn't work with UNIX domain sockets.

No additional features of UNIX domain sockets are yet implemented. That is for later.

I'm not expecting anyone to review the code yet, but comments/questions on the above would be welcome.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2323,6 +2323,55 @@ g_file_lock(int fd, int start, int len)
 }
 
 /*****************************************************************************/
+/* Converts a hex mask to a mode_t value */
+#if !defined(_WIN32)
+static mode_t
+hex_to_mode_t(int hex)
+{
+    mode_t mode = 0;
+
+    mode |= (hex & 0x4000) ? S_ISUID : 0;
+    mode |= (hex & 0x2000) ? S_ISGID : 0;
+    mode |= (hex & 0x1000) ? S_ISVTX : 0;
+    mode |= (hex & 0x0400) ? S_IRUSR : 0;
+    mode |= (hex & 0x0200) ? S_IWUSR : 0;
+    mode |= (hex & 0x0100) ? S_IXUSR : 0;
+    mode |= (hex & 0x0040) ? S_IRGRP : 0;
+    mode |= (hex & 0x0020) ? S_IWGRP : 0;
+    mode |= (hex & 0x0010) ? S_IXGRP : 0;
+    mode |= (hex & 0x0004) ? S_IROTH : 0;
+    mode |= (hex & 0x0002) ? S_IWOTH : 0;
+    mode |= (hex & 0x0001) ? S_IXOTH : 0;
+    return mode;
+}
+#endif
+
+/*****************************************************************************/
+/* Converts a mode_t value to a hex mask */
+#if !defined(_WIN32)
+static int
+mode_t_to_hex(mode_t mode)
+{
+    int hex = 0;
+
+    hex |= (mode & S_ISUID) ?  0x4000 : 0;
+    hex |= (mode & S_ISGID) ?  0x2000 : 0;
+    hex |= (mode & S_ISVTX) ?  0x1000 : 0;
+    hex |= (mode & S_IRUSR) ?  0x0400 : 0;
+    hex |= (mode & S_IWUSR) ?  0x0200 : 0;
+    hex |= (mode & S_IXUSR) ?  0x0100 : 0;
+    hex |= (mode & S_IRGRP) ?  0x0040 : 0;
+    hex |= (mode & S_IWGRP) ?  0x0020 : 0;
+    hex |= (mode & S_IXGRP) ?  0x0010 : 0;
+    hex |= (mode & S_IROTH) ?  0x0004 : 0;
+    hex |= (mode & S_IWOTH) ?  0x0002 : 0;
+    hex |= (mode & S_IXOTH) ?  0x0001 : 0;
+
+    return hex;
+}
+#endif
+
+/*****************************************************************************/
 /* returns error */
 int
 g_chmod_hex(const char *filename, int flags)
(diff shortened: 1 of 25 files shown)


matt335672:
Any comments?

If not I'll merge this so it gets devel testing before our next release.

metalefty (approved):
Let's go ahead.

matt335672:
Thanks.

I've updated the NEWS page with the user-visible changes as a result of this.

jsorg71:
I love it.  UDS is a great security move.
I see this in log.  error - no ip set
patch that seems to fix it
```
index ca36efc7..4ca395b9 100644
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -162,12 +162,6 @@ lib_mod_connect(struct mod *mod)
         return 1;
     }
 
-    if (g_strcmp(mod->ip, "") == 0)
-    {
-        mod->server_msg(mod, "error - no ip set", 0);
-        return 1;
-    }
-
     make_stream(s);
     g_sprintf(con_port, "%s", mod->port);
 
@@ -183,6 +177,12 @@ lib_mod_connect(struct mod *mod)
     {
         socket_mode = TRANS_MODE_TCP;
         LOG(LOG_LEVEL_INFO, "lib_mod_connect: connecting via TCP socket");
+        if (g_strcmp(mod->ip, "") == 0)
+        {
+            mod->server_msg(mod, "error - no ip set", 0);
+            free_stream(s);
+            return 1;
+        }
     }
 
     mod->trans = trans_create(socket_mode, 8 * 8192, 8192);

```

matt335672:
Thanks for that.

Sorry about the brief comment - I've got to catch a bus in a few minutes I'm afraid. I'll check the patch out tomorrow and merge it in.

Nexarian:
Same patch needs to be applied to the VNC backend, as it also has this check.

matt335672:
See #2233 for the regression fix.

@Nexarian - your point about VNC is interesting. As it stands, we need the 'ip' field for 'VNC', as the module only supports TCP connections, and not UDS connections. We could possibly change this - see #2234

Issue #2213, opened 2022-04-06
Timeouts in unit tests - test_gen_key_xrdp1, test_bitmap_load__max_width_{zoom,scale}__ok


elboulangero:
Hi, recently xrdp 0.19 was uploaded to Debian. Some tests timed out on some architectures: https://buildd.debian.org/status/package.php?p=xrdp&suite=sid

## armhf architecture - test_gen_key_xrdp1 timed out

```
test_ssl_calls.c:340:E:ssl_calls:test_gen_key_xrdp1:0: (after this point) Test timeout expired
```

These tests were introduced in `v0.9.19`.

From what I see, this test generates a RSA key, so it's very possible (and legitimate) that it times out, I suppose. What about using [tcase_set_timeout](https://libcheck.github.io/check/doc/check_html/check_4.html#Test-Timeouts) for this test, and set it to something real long, like 1 minute? It seems that by default, the timeout is 4 seconds.

Complete build logs at https://buildd.debian.org/status/fetch.php?pkg=xrdp&arch=armhf&ver=0.9.19-1&stamp=1648811785&raw=0

## alpha architecture -  test_bitmap_load__max_width_{zoom,scale}__ok

```
PASS: test_xrdp 1 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__with_invalid_image__fail: Passed
PASS: test_xrdp 2 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__4_bit__ok: Passed
PASS: test_xrdp 3 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__8_bit__ok: Passed
PASS: test_xrdp 4 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__24_bit__ok: Passed
FAIL: test_xrdp 5 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__max_width_zoom__ok: Test timeout expired
PASS: test_xrdp 6 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__max_height_zoom__ok: Passed
PASS: test_xrdp 7 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__min_zoom__ok: Passed
FAIL: test_xrdp 8 - test_bitmap_load.c:xrdp_bitmap_load:test_bitmap_load__max_width_scale__ok: Test timeout expired
[...]

PR #2214, opened 2022-04-06
Increase timeouts for some tests

elboulangero:


matt335672:
@metalefty - we're waiting for feedback on this. See #2213

metalefty:
I see. Thank you for letting me know the current status.

matt335672:
Some more analysis.

At the moment, all the SSL calls are lumped together in one test suite which has a common timeout.

I can reproduce this by running the test suite under valgrind to simulate a slow machine. What I'm finding is that pretty much all the time is being taken up in `ssl_gen_key_xrdp1()`. Drilling into this, all the time is being generated in the call to `EVP_PKEY_generate()` (i.e. in the SSL library call).

I've got a Xeon E3-1275 v3 which is about a decade old now. On that, the call to `EVP_PKEY_generate()` normally takes 150ms, but amazingly takes 15-30 seconds under valgrind. The next slowest test under valgrind in this suite is `test_sha1_hash_ok()` at 112ms.

I'll put together a more nuanced version of this PR to split the test case up a bit and add an extended timeout to the `ssl_gen_key_xrdp1()` case only.

matt335672:
Closing this - see #2901 which replaces it.

PR #2216, opened 2022-04-06
Fixes a stream overflow in the test suite

matt335672:
An overflow has been found in `test_libxrdp_process_monitor_stream__with_sextuple_monitor_happy_path()`

This is triggered when xrdp is built with `--enable-devel-logging`

Also, the logging sub-system is initialised for libxrdp tests. With this change, the test program generates this line when `--enable-devel-logging` is used:-

```
[20220406-12:03:44] [CORE ] [parser_stream_overflow_check(parse.c:49)] test_libxrdp_process_monitor_stream.c:221 Stream output buffer overflow. Size=233, pos=232, requested=4
```

### DIFF
--- a/tests/libxrdp/test_libxrdp_main.c
+++ b/tests/libxrdp/test_libxrdp_main.c
@@ -4,6 +4,7 @@
 
 #include <stdlib.h>
 #include <check.h>
+#include "log.h"
 #include "test_libxrdp.h"
 
 int main (void)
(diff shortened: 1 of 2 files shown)


matt335672:
@Nexarian - can you take a look at this? I just found it when looking at another issue. For some reason I can't add you as a reviewer.

Thanks. I'll get on to the testing you asked for tomorrow.

PR #2223, opened 2022-04-12
Add PAM_RHOST support

### LINKED ISSUE #392: Pass remote ip address to PAM
Trying to handle RDP logins in fail2ban, I found that the PAM stack apparently does not receive the address of the remote host trying to authenticate. I can imagine that knowledge of that could come in handy in other situations as well.

Is it possible to make xrdp pass the remote address to PAM as rhost?

matt335672:
Fixes #392 

Also, see #2221

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1401,12 +1401,10 @@ g_sck_accept(int sck, char *addr, int addr_bytes, char *port, int port_bytes)
 
 /*****************************************************************************/
 
-void
-g_write_connection_description(int rcv_sck, char *description, int bytes)
+int
+g_get_peer_details(int sck, struct peer *p)
 {
-    char *addr;
-    int port;
-    int ok;
+    int ok = 0;
 
     union
     {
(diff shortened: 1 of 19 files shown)


matt335672:
I've had a think about this PR, and whereas it should work OK, I think this is a good opportunity to remove the need for `g_get_ip_from_description()` in os_calls.c.

Creating a connection description including the IP address, and then sending it over the link where we extract the IP address seems like a bit of a bodge to me.

I think we should rework the existing `connection_description` into a remote IP address and port field and use that instead. It should make the code easier to read. I'll start working on that basis.

matt335672:
I've reworked this to replace the `connection_description` string with `struct peer` which contains the IP address and port of the client.

This allows the `g_get_ip_from_description()` to be replaced with simple struct member accesses.

There are two commits in this PR. The first reworks the connection_description, and once that's in place, adding `PAM_RHOST` support is extremely simple.

matt335672:
I've just realised this breaks xorgxrdp with:-

```
  CC       rdpDraw.lo
In file included from rdp.h:36,
                 from rdpDraw.c:46:
/home/mjb/xrdp/common/xrdp_client_info.h:117:17: error: field ‘peer_details’ has incomplete type
  117 |     struct peer peer_details; /* Details of peer on other end of connection */
```

I'll need to fix this.

matt335672:
This gets worse - I've just realised xrdp_client_info.h has `client_addr` and `client_port` fields which are doing exactly the same thing as the connection description.

I'm going to abandon this PR as it stands and re-submit it when I've cleaned that up too.

PR #2247, opened 2022-05-03
Remove TCP socket support from chansrv

matt335672:
The code in xrdp_mm.c to connect to chansrv over a TCP socket has been removed, with the move to UDS (see 0db849fc5c3f02ee24f10acd9f54a91a96fa2eaa). This PR simply removes the chansrv TCP listening code. Without doing this, some configurations result in a failure of xrdp to connect to chansrv.


Nexarian:
Is this the fix for the startup delay that we were talking about over Gitter?

Nexarian (approved):

Issue #874, opened 2017-09-13
cannot connect from Wyse SX0(S10) client due to some encryption issue


lptadmin:
Trying to connect a WYSE SX0 to Debain 9.1.0.
Connects fine to older xrdp and windows.
Will not connect to 0.9.1from Debian packages or compiled from sources.
Also tried 0.9.3.1 compiled on CentOS7 with same error.
Tried every fix I could find on here and the web.
Any ideas?  I would really like to get this working.
Output of xrdp -ns:
```
[20170913-11:00:09] [INFO ] A connection received from: ::ffff:10.1.1.220 port 2426
[20170913-11:00:09] [DEBUG] Closed socket 11 (AF_INET6 ::ffff:10.1.1.219 port 3389)
[20170913-11:00:09] [DEBUG] Closed socket 10 (AF_INET6 :: port 3389)
[20170913-11:00:09] [DEBUG] Security layer: requested 0, selected 0
colorDepth 0xca01 (0xca00 4bpp 0xca01 8bpp)
postBeta2ColorDepth 0xca01 (0xca00 4bpp 0xca01 8bpp 0xca02 15bpp 0xca03 16bpp 0xca04 24bpp)
highColorDepth 0x000f (0x0004 4bpp 0x0008 8bpp 0x000f 15bpp 0x0010 16 bpp 0x0018 24bpp)
supportedColorDepths 0x0007 (0x0001 24bpp 0x0002 16bpp 0x0004 15bpp 0x0008 32bpp)
earlyCapabilityFlags 0x0001 (0x0002 want32)
got client client connection type 0x00000000
xrdp_sec_process_mcs_data_CS_SECURITY:
  client supports 40 bit encryption
  client supports 128 bit encryption
  client supports 56 bit encryption
xrdp_sec_out_mcs_data: using no security
**xrdp_process_main_loop: libxrdp_process_incoming failed
[20170913-11:00:09] [DEBUG] Closed socket 11 (AF_INET6 ::ffff:10.1.1.219 port 3389)
xrdp_mcs_disconnect - socket closed
[20170913-11:00:09] [ERROR] Listening socket is in wrong state, terminating listener**
```

lptadmin:
Compiled 0.9.4rc1 from source on Debian 9, same error.  :(

metalefty:
Please let me know the "older xrdp" version which you can connect.

lptadmin:
0.6.1 from the epel packages running on CentOS 6.8

[...]

PR #2266, opened 2022-05-16
Improve RDS compatibility

### LINKED ISSUE #2166: Cannot connect from Wyse Thin clients
Hey All,

I have stumbled across this issue https://github.com/neutrinolabs/xrdp/issues/874 while researching an issue I have been having with some Wyse sx0 thin clients and I believe it is the same issue.

The original issue was closed because the developers were unable to get a hold of one of the clients, I have a box of them and would be happy to ship a few out if it could help fix this issue.

matt335672:
Fixes #2166

This PR improves compatibility with RDS in two ways:-
1) Windows 10 (and FreeRDP) set the user channel ID (see [MS-RDPBCGR] [3.3.1.2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/950cd069-252e-4c33-9988-cd5a6d3b5842)) to be one more than the last static channel ID. xrdp v0.9.19 and earlier use 1002 for the user channel ID. This is supposed to be the server channel ID ([MS-RDPBCGR] 3.3.1.5). This PR emulates Windows behaviour.
2) The use of MCS channel join PDUs is documented in [MS-RDPBCGR] [3.3.5.3.8](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/2cd07374-76de-4df9-b999-c796c84b6723) and [3.3.5.3.9](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/c604522b-33cb-4416-9826-71992e51c1d6). These sections do not mention what to do if too few channel join PDUs are received.
- xrdp versions which support TLS (i.e. xrdp v0.9.1 and later) insist on receiving exactly the right number of channel join PDUs. This is because immediately after these, a TLS client hello' is received and the code is much simpler if this can be started cleanly.
- xrdp versions which do not support TLS (i.e. previous to xrdp v0.9.1) simply accept the join requests they are given and acknowledge them. If any are missing, that's fine.
- Experiments with FreeRDP and Windows 10 (see [here](/neutrinolabs/xrdp/issues/2166#issuecomment-1124830344)) show that Windows RDS is fine with too few join requests in both TLS and non-TLS mode.

   The move to xrdp v0.9.x broke compatibility with some older clients (see #874, #2166) as these clients do not send a channel join request for the user channel.

We can't easily make xrdp emulate Windows behaviour fully, as this would really complicate the TLS handshake. What this PR does however, is to re-instate the pre v0.9.1 behaviour for non-TLS clients only. Older clients either can't use TLS, or can only use insecure versions of it, so this at least gives users of those devices a way to keep using them.

### DIFF
--- a/libxrdp/libxrdp.h
+++ b/libxrdp/libxrdp.h
@@ -67,6 +67,9 @@ struct xrdp_mcs
     struct stream *client_mcs_data;
     struct stream *server_mcs_data;
     struct list *channel_list;
+    /* This boolean is set to indicate we're expecing channel join
+     * requests as part of the connect sequence */
+    int expecting_channel_join_requests;
 };
 
 /* fastpath */
--- a/libxrdp/xrdp_mcs.c
+++ b/libxrdp/xrdp_mcs.c
@@ -27,6 +27,11 @@
 #include "ms-rdpbcgr.h"
 #include "log.h"
 
+/* Forward references */
+static int
+handle_non_tls_client_channel_join_requests(struct xrdp_mcs *self,
+        struct stream *s, int *appid);
+
 /*****************************************************************************/
 struct xrdp_mcs *
 xrdp_mcs_create(struct xrdp_sec *owner, struct trans *trans,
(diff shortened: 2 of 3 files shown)


Nexarian:
@matt335672 I have no objection to any of this, but only want to say: Now you understand the pains Microsoft goes through to truly make things backwards compatible. I saw this first hand when I worked there. It's insane.

metalefty:
Thanks for your hard work! This should be shipped to v0.9, too.

PR #2226, opened 2022-04-15
Bump FreeBSD CI version to 12.3

matt335672:
Update to latest FreeBSD 12.x version for CI to prevent these kind of errors:-

```
./bootstrap
/usr/local/bin/autoconf
/usr/local/bin/automake
/usr/local/bin/libtool
/usr/local/bin/pkg-config
autoreconf-2.69: Entering directory `.'
autoreconf-2.69: configure.ac: not using Gettext
autoreconf-2.69: running: aclocal --force -I m4
ld-elf.so.1: /usr/local/lib/perl5/5.32/mach/CORE/libperl.so.5.32: Undefined symbol "strerror_l@FBSD_1.6"
aclocal: error: /usr/local/bin/autom4te-2.69 failed with exit status: 1
autoreconf-2.69: aclocal failed with exit status: 1
ld-elf.so.1: /usr/local/lib/perl5/5.32/mach/CORE/libperl.so.5.32: Undefined symbol "strerror_l@FBSD_1.6"
```

See also #1804

PR #2273, opened 2022-05-18
Improve RDS compatibility

matt335672:
Port of #2266 to the 0.9 branch.

Is this the right place for it @metalefty? If so please merge.


matt335672:
I think we also need #2226 porting over before this one to fix the CI.

metalefty:
Indeed.

matt335672:
OK - for now I've just added this to the v0.9.20 milestone, like #2226. This one is however, based on v0.9.19 rather than devel.

matt335672:
I've updated the NEWS page, and tried to make it clear the next major version is based on v0.9.19

PR #2276, opened 2022-05-19
Update xrdpapi simple example to work with new logging

matt335672:
See discussion #2143 

Worth merging?


metalefty (approved):
Thanks, I belive it's worth.

matt335672:
I'll see if I can get in the CI too, or it will just get broken again.

PR #2278, opened 2022-05-23
Bump cppcheck version to 2.8

matt335672:
Minor change to the (Windows) `fontdump/fontdump.c` was needed to prevent this error:-

```
fontdump/fontdump.c:80:28: error: Uninitialized variable: lpMsgBuf [uninitvar]
                   (LPSTR)&lpMsgBuf, 0, NULL);
                           ^
```
Message is a false positive, possibly caused by the cast and the lack of a prototype.

However, there is an error here, which is if the call to [FormatMesasgeA](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessagea) fails, lpMsgBuf will be unititialised after the call.

Simplest fix is to initialise the variable to NULL which fixes the cppcheck error and check the call result before dereferencing the pointer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,7 @@ jobs:
       CC: gcc
       # This is required to use a version of cppcheck other than that
       # supplied with the operating system
-      CPPCHECK_VER: 2.7
+      CPPCHECK_VER: 2.8
       CPPCHECK_REPO: https://github.com/danmar/cppcheck.git
     steps:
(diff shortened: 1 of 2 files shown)


matt335672:
Minor update. We've used a fairly hacky way to get the OS image type in the YAML file up to now. The latest commit uses the canonical way to do this - see actions/cache#543

metalefty (approved):

PR #2292, opened 2022-06-27
Backport MSTSC resize fix.

Nexarian:
Ports https://github.com/neutrinolabs/xrdp/pull/2291 to v0.9

PR #2300, opened 2022-07-02
Block resize if already this size.

Nexarian:
While this feature is part of other branches for testing EGFX
integration, it somehow never made it into devel. This should fix

https://github.com/neutrinolabs/xrdp/issues/1928, for real this time!

### DIFF
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -1009,6 +1009,29 @@ dynamic_monitor_data_first(intptr_t id, int chan_id, char *data, int bytes,
 }
 
 /******************************************************************************/
+static const int MAXIMUM_MONITOR_SIZE
+    = sizeof(struct monitor_info) * CLIENT_MONITOR_DATA_MAXIMUM_MONITORS;
+
+/******************************************************************************/
+static void
+sync_dynamic_monitor_data(struct xrdp_wm *wm,
+                          struct display_size_description *description)
+{
+    struct display_size_description *display_sizes
+        = &(wm->client_info->display_sizes);
+
+    display_sizes->monitorCount = description->monitorCount;
+    display_sizes->session_width = description->session_width;
+    display_sizes->session_height = description->session_height;
+    g_memcpy(display_sizes->minfo,
+             description->minfo,
+             MAXIMUM_MONITOR_SIZE);
+    g_memcpy(display_sizes->minfo_wm,
+             description->minfo_wm,
+             MAXIMUM_MONITOR_SIZE);
+}
+
+/******************************************************************************/
 static int
 process_dynamic_monitor_description(struct xrdp_wm *wm,
                                     struct display_size_description *description)
@@ -1039,6 +1062,18 @@ process_dynamic_monitor_description(struct xrdp_wm *wm,
             "process_dynamic_monitor_description: Not allowing resize due to"
             " invalid dimensions (w: %d x h: %d)",
             description->session_width, description->session_height);
+        return 0;
+    }
+    if (description->session_width
+            == wm->client_info->display_sizes.session_width
+            && description->session_height
+            == wm->client_info->display_sizes.session_height)
+    {
+        LOG(LOG_LEVEL_WARNING, "process_dynamic_monitor_description:"
+            " Not resizing. Already this size. (w: %d x h: %d)",
+            description->session_width,
+            description->session_height);
+        sync_dynamic_monitor_data(wm, description);
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


metalefty:
I'll test this on Wednesday.

matt335672:
The code looks fine to me, but does this cope with situations where a monitor is moved but the overall session size doesn't change?

Example; Three monitors in a row with the left one and right one being 1280x1024 and the middle one being 1024x768.

If the user starts with an arrangement like like this:-

```
+------------+ +---+ +------------+
|            | |   | |            |
|            | +---+ |            |
|            |       |            |
+------------+       +------------+
```

Session size is (1280 + 1024 + 1280) x 1024 = 3584 x 1024

If the user then just moves the middle monitor down:-
```
+------------+       +------------+
|            |       |            |
|            | +---+ |            |
|            | |   | |            |
+------------+ +---+ +------------+
```

we'll get a message to say the layout's changed, but I think we'll just ignore it in this case, as the session size is still 3584 x 1024.

Nexarian:
@matt335672 That makes sense. The question I might have is: If the only thing that changes is the layout, and the total session size doesn't change, what is there for us to do about it? The only thing I can think of is that [this code block](https://github.com/neutrinolabs/xrdp/blob/521e5e0e42b6eaa718e61419acedffa6c07476a2/xrdp/xrdp_mm.c#L1144) should be executed essentially no matter what. We don't do the mechanics of resizing anything, but we can say "ok, we've updated the internal monitor layout."

Is there anything else you can think of here? I confess my understanding of how multiple monitors in RDP is supposed to work is still hazy.

I also noticed this: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/32525e02-fc5c-4b82-ae87-88a458444902:

- None of the specified monitors overlap.
- Each monitor is adjacent to at least one other monitor (even if only at a single point).

We're not error-checking for those use-cases as of yet, should probably add that.

Nexarian:
This ancient post seems to show what Microsoft is thinking here: https://techcommunity.microsoft.com/t5/security-compliance-and-identity/using-multiple-monitors-in-remote-desktop-session/ba-p/246840

Though it's old, I have no reason to believe the fundamentals haven't changed. What I struggle to find is more up to date documentation and/or an actual protocol example of how this is supposed to work.

matt335672 (commented):
As discussed, if session size changes, layout updates should be recorded. Once that's done, merge away!

metalefty:
It might not be related to this fix. I got another protocol error while testing. 

After 3 times move window to another display and fullscreenify, I got the following error.

image

```
[20220706-10:52:22] [DEBUG] dynamic_monitor_data: msg_type 2 msg_length 56
[20220706-10:52:22] [DEBUG] libxrdp_process_monitor_stream: The number of monitors received is: 1
[20220706-10:52:24] [DEBUG] dynamic_monitor_data: msg_type 2 msg_length 56
[20220706-10:52:24] [DEBUG] libxrdp_process_monitor_stream: The number of monitors received is: 1
[20220706-10:52:24] [WARN ] process_dynamic_monitor_description: Not resizing. Already this size. (w: 2560 x h: 1440)
[20220706-10:52:24] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[20220706-10:52:25] [INFO ] xrdp_process_offscreen_bmpcache: support level 1 cache size 7864320 MB cache entries 100
[20220706-10:52:25] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[20220706-10:52:25] [WARN ] xrdp_caps_process_codecs: unknown codec id 5
[20220706-10:52:25] [INFO ] xrdp_caps_process_codecs: RemoteFX, codec id 3, properties len 49
[20220706-10:52:25] [ERROR] SSL_write: I/O error
[20220706-10:52:26] [ERROR] xrdp_sec_send_fastpath: xrdp_fastpath_send failed
[20220706-10:52:26] [ERROR] xrdp_rdp_send_fastpath: xrdp_sec_send_fastpath failed
```

Nexarian:
@metalefty I think the problem is for some reason your session switched from RemoteFX to Fastpath during resize. Did you connect to an existing session that was previously using RemoteFX?

On one hand, switching encoding types on resize probably shouldn't be supported. On the other hand, we shouldn't crash either...

metalefty:
Yes, I connected to the existing session. I'll try with new session. Thanks!

metalefty:
Re-tried in new session, I confirmed the change fix the issue. Can you address @matt335672 's point?

metalefty:
Is it possible to backport to v0.9?

Nexarian:
@metalefty I will work on it.

Nexarian:
@metalefty I've simply incorporated this fix into https://github.com/neutrinolabs/xrdp/pull/2292 because they're similar.

metalefty:
Thanks!

PR #2301, opened 2022-07-02
Update handling of scale factor

Nexarian:
Looking at 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT (https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/ea2de591-9203-42cd-9908-be7a55237d1c),
the way we were handling it was slightly off. device_scale_factor and
desktop_scale_factor are interdependent. If either one is out of spec,
both are set to default. That wasn't how it was previously being
handled.

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -1959,30 +1959,33 @@ libxrdp_process_monitor_stream(struct stream *s,
             }
 
             in_uint32_le(s, monitor_layout->desktop_scale_factor);
-            if (monitor_layout->desktop_scale_factor < 100
-                    || monitor_layout->desktop_scale_factor > 500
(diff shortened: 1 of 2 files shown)


matt335672 (approved):
Looks OK to me

metalefty:
@Nexarian Please keep in mind to update NEWS in wiki after you merge something. It is draft release note for the next release.

matt335672:
Ditto that.  I find it a lot easier to update the NEWS when you do the merge rather than trying to pick your way through a series of changes you barely remember before a release to get a release note together. Maybe that's just me and my memory.

Nexarian:
All I needed to do here was update the wiki to add the PR: `Monitor processing logic, which was in two places, has now been unified (#1895 #2301)`

PR #2310, opened 2022-07-08
Parse more physical monitor size information

matt335672:
Draft PR to parse more physical monitor information.

This is a precursor to a scalable login screen - we need to be able to calculate the DPI for the login screen monitor, and to do this, we need access to the physical monitor size.

I need to do more testing of this before it's ready for release, but I'm going to be a bit stuck for time for a couple of weeks.

@Nexarian - can you take a look and give me any informal comments you have?

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -38,7 +38,8 @@ struct monitor_info
     int bottom;
     int flags;
 
-    /* From 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT */
+    /* From [MS-RDPEDISP] 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT, or
+     * [MS-RDPBCGR] 2.2.1.3.9.1 (TS_MONITOR_ATTRIBUTES) */
     unsigned int physical_width;
     unsigned int physical_height;
     unsigned int orientation;
@@ -205,6 +206,13 @@ struct xrdp_client_info
 
     /* xrdp.override_* values */
     struct xrdp_keyboard_overrides xrdp_keyboard_overrides;
+
+    /* These values are optionally send over as part of TS_UD_CS_CORE.
+     * They can be used as a fallback for a single monitor session
+     * if physical sizes are not available in the monitor-specific
+     * data */
+    unsigned int session_physical_width; /* in mm */
+    unsigned int session_physical_height; /* in mm */
 };
 
 /* yyyymmdd of last incompatible change to xrdp_client_info */
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -29,6 +29,7 @@
 #include "ms-rdpbcgr.h"
 
 #define MAX_BITMAP_BUF_SIZE (16 * 1024) /* 16K */
+#define TS_MONITOR_ATTRIBUTES_SIZE 20 /* [MS-RDPBCGR] 2.2.1.3.9 */
 
 /******************************************************************************/
 struct xrdp_session *EXPORT_CC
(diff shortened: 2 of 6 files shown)


Nexarian:
👀

common/xrdp_client_info.h
```diff
@@ -38,7 +38,8 @@ struct monitor_info
     int bottom;
     int flags;
 
-    /* From 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT */
+    /* From [MS-EDISP] 2.2.2.2.1 DISPLAYCONTROL_MONITOR_LAYOUT, or
```
Nexarian: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/d2954508-f487-48bc-8731-39743e0854a9

Should we use the full name here: `MS-RDPEDISP`

common/xrdp_client_info.h
```diff
@@ -205,6 +206,13 @@ struct xrdp_client_info
 
     /* xrdp.override_* values */
     struct xrdp_keyboard_overrides xrdp_keyboard_overrides;
+
+    /* These values are optionally send over as part of TS_UD_CS_CORE.
+     * They can be used as a fallback for a single monitor session
+     * if physical sizes are not available in the monitor-specific
+     * data */
+    unsigned int session_physical_width; /* in mm */
```
Nexarian, line 214: Why not put `session_physical_width` (and height) into `display_size_description`? It contains the information for all the monitors in addition to summary information such as the size of the full session.

Also, this doesn't seem to have logic that would be at parity with `session_height/width` -- Those are for computing the height/width of the ENTIRE session. Is there a similar computation for the physical width/height of the ENTIRE session when it's provided for all monitors?

I think also, how do we deal with DPI if multiple monitors are provided that each have a different scale factor? Pick the highest DPI? Lowest?

libxrdp/xrdp_sec.c
```diff
@@ -2486,8 +2546,15 @@ xrdp_sec_process_mcs_data(struct xrdp_sec *self)
...
+            case SEC_TAG_CLI_MONITOR_EX:  /* CS_MONITOR_EX     0xC008 */
+                if (xrdp_sec_process_mcs_data_monitors_ex(self, s) != 0)
+                {
+                    LOG(LOG_LEVEL_ERROR,
+                        "Processing [MS-RDPBCGR] TS_UD_CS_MONITOR_EX failed");
+                    return 1;
+                }
+                break;
             /* CS_MCS_MSGCHANNEL 0xC006
-               CS_MONITOR_EX     0xC008
```
Nexarian, line 2490: Why remove this?

libxrdp/libxrdp.c
```diff
@@ -1764,6 +1765,97 @@ libxrdp_send_session_info(struct xrdp_session *session, const char *data,
...
+   Sanitise extended monitor attributes
+
+   The extended attributes are received from either
+   [MS-RDPEDISP] 2.2.2.2.1 (DISPLAYCONTROL_MONITOR_LAYOUT), or
+   [MS-RDPBCGR] 2.2.1.3.9.1 (TS_MONITOR_ATTRIBUTES)
+
+   @param monitor_layout struct containing extended attributes
+*/
+static void
+sanitise_extended_monitor_attributes(struct monitor_info *monitor_layout)
```
Nexarian, line 1779: I like this refactor, especially because it helps clean up and clarify what extended monitor attributes are!

Nexarian:
We'll need to merge this in conjunction with: https://github.com/neutrinolabs/xorgxrdp/pull/224 -- Can you review that as well?

common/xrdp_client_info.h
matt335672: Yes we should - I'll fix it.

common/xrdp_client_info.h
matt335672: Funnily enough I started down that route. Eventually however it occurred to me that it didn't make much sense.

The physical size of the desktop is easy to understand for a single monitor, and actually has a use, which is to calculate the DPI for the monitor (in the absence of any other information). For multiple monitors however, which may not even be touching in any way, the overall physical size is not something that can be defined. For example, what would you do in a situation like this with a gap between the monitors?

```
+---------+       +---------+
|         |       |         |
|    0    |       |    1    |
|         |       |         |
+---------+       +---------+
```

So eventually I removed the overall physical size from the `display_size_description`, and just left the values from the CORE message further down the structure as a fallback, in case extended monitor information isn't available. These could be renamed maybe, if it's not clear what is going on.

Dealing with the DPI on the login screen is easy - the login screen only appears on the primary monitor, so that's the only monitor we need to worry about.

libxrdp/xrdp_sec.c
matt335672: Because we're now processing CS_MONITOR_EX under the `case SEC_TAG_CLI_MONITOR_EX:` tag, so it no longer makes sense to have this in the comment.

common/xrdp_client_info.h
matt335672: Done

Nexarian (approved):

PR #2315, opened 2022-07-14
adding large cursor support

jsorg71:


matt335672:
@jsorg71 - I'll try to get a look at this next week sometime,

jsorg71:
Take your time, I still have to add xup and xorg bits

rbreaves:
@jsorg71 Does this mean clean looking hidpi cursor support is coming online with this patch? Doing the lords work if so, I have been meaning to look into hidpi support for the cursor.

libxrdp/xrdp_caps.c
```diff
@@ -1225,6 +1242,12 @@ xrdp_caps_send_demand_active(struct xrdp_rdp *self)
         LOG_DEVEL(LOG_LEVEL_TRACE, "xrdp_caps_send_demand_active: Server Capability "
                   "CAPSSETTYPE_MULTIFRAGMENTUPDATE = %d", max_request_size);
 
+        /* large pointer 96x96 */
+        caps_count++;
+        out_uint16_le(s, CAPSETTYPE_LARGE_POINTER);
+        out_uint16_le(s, CAPSETTYPE_LARGE_POINTER_LEN);
+        out_uint16_le(s, 1); /* LARGE_POINTER_FLAG_96x96 */
```
matt335672: Can we add this to ms-rdpbcgr.h ? Something like `#define LARGE_POINTER_FLAG_96x96 1`? It just helps when searching the code.

matt335672:
One very minor comment added.

There's also the VNC module. I'm not suggesting you tackle this now, but when we merge this it might be worth adding a feature request to update that too.

pnowack:
Just FYI, there are two types of large pointer updates: The one that you implemented here (for handling up to 96x96 bitmaps) and there is another one that allows handling for bitmaps up to 384x384 pixels (`LARGE_POINTER_FLAG_384x384`).
The latter one also uses its own struct (See also `2.2.9.1.2.1.11 Fast-Path Large Pointer Update` (`[MS-RDPBCGR]`)).

jsorg71:
Thanks @pnowack , I didn't know that.  I see it here [2.2.7.2.7](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/41323437-c753-460e-8108-495a6fdd68a8).

jsorg71:
BTW, I simple was to get large cursors is to
edit ~/.Xresources and add Xcursor.size: 64

jsorg71:
@matt335672 I went to add large cursor to libvnc and seen it does not even support new cursors yet.  I think lets do a todo for this so some can enjoy large cursors with xorg backend.

matt335672:
Definitely - there's no need to hold this one up. When we merge this one, we can just add a feature request for VNC large cursors.

PR #2338, opened 2022-08-10
Adding egfx base functions.

Nexarian:


xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+
+/******************************************************************************/
+int
+xrdp_egfx_send_create_surface(struct xrdp_egfx *egfx, int surface_id,
+                              int width, int height, int pixel_format)
+{
+    int error;
+    int bytes;
+    struct stream *s;
+    char *holdp;
```
matt335672: There are some macros to do this stuff in parse.h, namely s_push_layer() and s_pop_layer(). You might want to consider using these for consistency with other code elsewhere.

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+{
+    int error;
+    int bytes;
+    int index;
+    struct stream *s;
+    char *holdp;
+
+    LOG(LOG_LEVEL_TRACE, "xrdp_egfx_send_fill_surface:");
+    make_stream(s);
+    init_stream(s, 8192);
```
matt335672: Is this guaranteed to be big enough, or should you scale it depending on `num_rects`? Somehing like `init_stream(s, 1024 + num_rects * 8);`

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+    int segment_size;
+    int segment_count;
+    struct stream *s;
+    char *hold_segment_count;
+    char *bitmap_data8;
+
+    LOG(LOG_LEVEL_TRACE, "xrdp_egfx_send_wire_to_surface1:");
+    make_stream(s);
+    bytes = bitmap_data_length + 8192;
+    bytes += 5 * (bitmap_data_length / 0xFFFF);
```
matt335672: USHRT_MAX for consistency?

matt335672 (commented):
@Nexarian - a few thoughts for you. I thought I'd comment now as next week is looking a bit hairy time-wise.

xrdp/xrdp_egfx.h
```diff
@@ -0,0 +1,173 @@
...
+
+#define XR_RDPGFX_CAPVERSION_8      0x00080004
+#define XR_RDPGFX_CAPVERSION_81     0x00080105
+#define XR_RDPGFX_CAPVERSION_10     0x000A0002
+#define XR_RDPGFX_CAPVERSION_101    0x000A0100
+#define XR_RDPGFX_CAPVERSION_102    0x000A0200
+#define XR_RDPGFX_CAPVERSION_103    0x000A0301
+#define XR_RDPGFX_CAPVERSION_104    0x000A0400
+#define XR_RDPGFX_CAPVERSION_105    0x000A0502
+#define XR_RDPGFX_CAPVERSION_106    0x000A0600 /* doc says 0x000A0601 */
```
pnowack: Update your PDF. The value was fixed some time ago (was a documentation issue). Meanwhile cap version 107 was released.

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+    init_stream(s, bytes);
+    /* RDP_SEGMENTED_DATA */
+    out_uint8(s, 0xE1); /* descriptor = MULTIPART */
+    hold_segment_count = s->p;
+    out_uint8s(s, 2); /* segmentCount set later */
```
pnowack: <s> You will run into problems here: </s>

<s>First of all, </s>`MULTIPART` is not always necessary, especially with AVC, but sometimes it is. So, you waste bandwidth here.

<s>But this is not the problem. The problem is that you assume the `segmentCount` to be 2. If you don't need more, this is fine, even for tiny frame updates.
But if you have an update `> 2 * 65535` Bytes, you have a problem, because you will need more segments.
People with large display setup will run into this, or when AVC is not available and you fall back to Calista Progressive.</s>

I suggest you to handle the segmentation outside this file and apply it in a general way.

Similar situation for `WireToSurface_2` and the `Fill` PDU above (and actually also for the blit PDUs (CacheToSurface, SurfaceToSurface)).

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+    out_uint8(s, 0xE0); /* descriptor = SINGLE */
+    /* RDP8_BULK_ENCODED_DATA */
+    out_uint8(s, 0x04); /* header = PACKET_COMPR_TYPE_RDP8 */
+    /* RDPGFX_HEADER */
+    out_uint16_le(s, XR_RDPGFX_CMDID_RESETGRAPHICS); /* cmdId */
+    out_uint16_le(s, 0); /* flags = 0 */
+    out_uint32_le(s, 340);
+    out_uint32_le(s, width);
+    out_uint32_le(s, height);
+    out_uint32_le(s, monitor_count == 0 ? 1 : monitor_count);
```
pnowack, line 694: Since I am not an xrdp member, its your choice. But personally I wouldn't apply workarounds for the own code here.

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1021 @@
...
+            " pduLength %d", cmdId, flags, pduLength);
+        if (pduLength < 8)
+        {
+            return 1;
+        }
+        if (!s_check_rem(s, pduLength - 8))
+        {
+            return 1;
+        }
+        switch (cmdId)
```
pnowack, line 876: You must be able to handle the Cache Import Offer PDU (and reply to it), even when you don't want to import anything.

xrdp/xrdp_egfx.c
pnowack: Ah, nvm. Just see that you actually skip these Bytes here and set the `segmentCount` afterwards.

xrdp/xrdp_egfx.c
jsorg71: good point, agreed.

xrdp/xrdp_egfx.c
jsorg71: yes, unlikely but yes.

xrdp/xrdp_egfx.c
jsorg71: I guess a define like MAX_PART_SIZE would be best

xrdp/xrdp_egfx.h
jsorg71: Oh nice, I sent an email to the docs team about that.  Glad they fixed it.

xrdp/xrdp_egfx.c
jsorg71: good point

xrdp/xrdp_egfx.c
Nexarian: @matt335672 I believe this is resolved.

xrdp/xrdp_egfx.c
Nexarian: @matt335672 I believe this is resolved.

xrdp/xrdp_egfx.c
Nexarian: @matt335672 I believe this is resolved.

xrdp/xrdp_egfx.h
Nexarian: @pnowack I believe this is resolved.

xrdp/xrdp_egfx.c
Nexarian: This is a workaround for Microsoft's Mac OS Client. We can't ever send a monitorcount of 0, even though internally sometimes it's set that way. So, from Microsoft's perspective, this is important to always be one.

What would you recommend I do instead?

xrdp/xrdp_egfx.c
Nexarian: Resolving.

matt335672 (commented):
I'm not at all up-to-speed on what EGFX does, so I've restricted my review to looking at possible overflows and leaks. From that perspective, I think this is fine - I haven't found any issues.

xrdp/xrdp_egfx.c
pnowack: > We can't ever send a monitorcount of 0

Well, it's RDP, it's about graphical remote desktop. You always have at least one monitor. Sending a monitor config here with 0 monitors is a bug.
If you know, that the monitor config changed, but not know yet the actual monitor layout, then defer sending this PDU until you know the new layout.
Same for the subsequent PDUs, which provide the graphics content (wire_to_surface, etc.).
There is no need to send this PDU until the screen changes or the layout changes are done.

The question would also be: If you have 0 monitors, what size for the graphics output buffer do you use in that situation?
The graphics output buffer only contains user visible surfaces, i.e. surfaces that are mapped to an output (map_surface_to_output).
You can have offscreen surfaces, that are not mapped, that are actually larger than the graphics output buffer itself, in case that is relevant for you.

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1157 @@
...
+            return 1;
+        }
+        in_uint32_le(s, version);
+        in_uint32_le(s, capsDataLength);
+        if (!s_check_rem(s, capsDataLength))
+        {
+            return 1;
+        }
+        holdp = s->p;
+        if (capsDataLength == 4)
```
pnowack, line 827: You might want to mention (for others), that caps version 101 is indirectly excluded here, but that's up to you.

xrdp/xrdp_egfx.c
Nexarian: I agree that 0 monitors is a bug in XRDP, but it's pretty baked in at this point and probably out of scope for this PR to fix it.

In the past "0 monitors with a session_width and a session_height" just meant "multi-mon is not enabled and we have one single large surface." It didn't actually matter until GFX integration, which cares about monitor numbering more than RLE or RemoteFX appeared to.

That is indeed wrong. I'll write up an issue about what's wrong here and that we should fix it (and hopefully if I have time, what needs to be fixed), then reference it in a TODO.

Sound good?

pnowack: > I agree that 0 monitors is a bug in XRDP, but it's pretty baked in at this point and probably out of scope for this PR to fix it.

Sure, `xrdp_egfx_send_reset_graphics()` is also not used yet.

Personally, I would, in case that bug is unavoidable, apply the workaround at the earliest point possible (i.e. in future commits, where `xrdp_egfx_send_reset_graphics` is actually used). This way, the area of code, where that bug exists would be reduced.

> Sound good?

Ultimately, it is up to you how you handle that, because I am not part of the xrdp project. I can only point out wrong or fishy things here.

> I'll write up an issue about what's wrong here and that we should fix it (and hopefully if I have time, what needs to be fixed), then reference it in a TODO.

Creating an issue sounds sane to avoid, that the issue is lost, while the bug still exists.

xrdp/xrdp_egfx.h
pnowack: > I believe this is resolved.

yup

xrdp/xrdp_egfx.c
```diff
@@ -0,0 +1,1157 @@
...
+    int bytes;
+    struct stream *s;
+
+    LOG(LOG_LEVEL_TRACE, "xrdp_egfx_create_surface:");
+    make_stream(s);
+    init_stream(s, 8192);
+    /* RDP_SEGMENTED_DATA */
+    out_uint8(s, 0xE0); /* descriptor = SINGLE */
+    /* RDP8_BULK_ENCODED_DATA */
+    out_uint8(s, 0x04); /* header = PACKET_COMPR_TYPE_RDP8 */
```
pnowack: Small suggestion (not an issue), you could use a `define` for the `PACKET_COMPR_TYPE_RDP8`.

Nexarian: Done!

xrdp/xrdp_egfx.c
Nexarian: https://github.com/neutrinolabs/xrdp/issues/2502

xrdp/xrdp_egfx.c
Nexarian: Done.

xrdp/xrdp_egfx.c
pnowack: > https://github.com/neutrinolabs/xrdp/issues/2502

No, that is not enough or right: Quoting the docs here (`1.5.2 Server Implementation Requirements`):

```
Furthermore, servers implementing the Remote Desktop Protocol: Graphics Pipeline Extension must
be capable of processing the following messages:
- RDPGFX_FRAME_ACKNOWLEDGE_PDU (section 2.2.2.13)
- RDPGFX_CACHE_IMPORT_OFFER_PDU (section 2.2.2.16)
- RDPGFX_CAPS_ADVERTISE_PDU (section 2.2.2.18)
```

Also, `3.2.5.16 Processing an RDPGFX_CACHE_IMPORT_OFFER_PDU message`:

```
Once the RDPGFX_CACHE_IMPORT_OFFER_PDU message has been processed, the server MUST
respond by sending the RDPGFX_CACHE_IMPORT_REPLY_PDU (section 2.2.2.17) message to the
client (section 3.2.5.17).
```

Nexarian:
It seems this has been reviewed well enough to be merged. I don't want to sit on this any further, as I definitely need to get moving on EGFX merge into devel so the benefits can be enjoyed by more standard users of XRDP.

For further issues and bug fixes, future PRs where I incorporate more of mainline_merge will be a great candidate.

xrdp/xrdp_egfx.c
Nexarian: Got it! I copied your comment into the issue so we don't lose it.

What's the best way to test this? Is there a way to force FreeRDP to send this message so we can test the negotiation? Or is there a mode on the Microsoft clients we can set to trigger it?

pnowack: mstsc sends that PDU (though not always).

PR #2341, opened 2022-08-11
Scaled login screen

matt335672:
This PR adds a scaling capability to the login screen, so that it is actually readable on high DPI (or 'retina') displays.

The DPI value for the login screen can be obtained from the client when running in a full-screen mode. If the session is
set to less than the full screen size, the DPI value cannot be calculated and behaviour is the same as current behaviour.

An additional font sans-18.fv1 is added to the distribution for higher DPI screens.

Also of note; the Windows utility to build font files has been replaced with two native UNIX utilities:-
- xrdp-dumpfv1 : Display contents of an fv1 font file
- xrdp-mkfv1 : Create new fv1 font files (only available if xrdp is built with `--with-freetype2`).

The user is able to generate their own fv1 font files and configure the system to use them, if required.

See discussion #2280 for a history of this PR, and the initial request from @HubKing.

@TOMATO-ONE has provided some excellent testing and feedback during the development of this PR.

Future work could involve using the freetype2 library to load truetype fonts directly. This would allow us to add anti-aliasing and kerning to the login screen.


metalefty:
@matt335672 Can you share some screenshots? The PR itself LGTM.

matt335672:
I'll add something to the wiki - it will help in the future I think.

matt335672:
https://github.com/neutrinolabs/xrdp/wiki/Scaleable-login-screen on the Wiki contains some shots of the login box, and answers some questions users may have.

metalefty:
Thanks!

PR #2353, opened 2022-09-01
mark count with unused attribute

kraj:
This may throw a warning when devel logs are disabled
Fixed
../../../xrdp-0.9.19/sesman/chansrv/chansrv.c:198:9: error: variable 'count' set but not used [-Werror,-Wunused-but-set-variable]                                                                                                                     int count;                                                                                                                   ^                                                                                                                1 error generated.

Upstream-Status: Pending
Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -197,7 +197,7 @@ check_timeout(void)
     struct timeout_obj *temp_tobj;
     int count;
     tui32 now;
-
+    UNUSED_VAR(count);
     LOG_DEVEL(LOG_LEVEL_DEBUG, "check_timeout:");
     count = 0;
     tobj = g_timeout_head;


matt335672:
Hi @kraj 

I can't reproduce this. Can you let me know what your compiler version and any flags are?

Also, there's a macro to do this called `UNUSED_VAR()`. It's not as pretty as the `__attribute__` but it works on a wider range of compilers.

I'll have a look at the FreeBSD CI fail. That's nothing at all to do with you.

matt335672:
@kraj - I've just noticed your branch is extremely out of date - you're 367 commits behind the development branch.

This is why the FreeBSD CI check is failing.

kraj:
> @kraj - I've just noticed your branch is extremely out of date - you're 367 commits behind the development branch.
> 
> This is why the FreeBSD CI check is failing.

its latest  clang-15 ( to be released soon)

matt335672:
@kraj - thanks for your contribution

PR #2366, opened 2022-09-13
Redesign the login screen

TOMATO-ONE:
See Discussions [#2342](https://github.com/neutrinolabs/xrdp/discussions/2342).

I have created a new login screen screen for use in the next major version.

New logo created by @metalefty.
Background colour taken from Windows 10 RDP login screen.
Login screen colours taken from Windows 10 mstsc.exe.

Add bmp logo image sized for HiDPI login screen and transparent png image for imlib2 build.

image

PR #2368, opened 2022-09-15
gitignore: add some built binaries

metalefty:
Found in #2366.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ config.sub
 configure
 depcomp
 .deps/
+fontutils/xrdp-dumpfv1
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ config.sub
 configure
 depcomp
 .deps/
+fontutils/xrdp-dumpfv1
```
matt335672, line 17: Also need fontutils/xrdp-mkfv1 which is only generated if built with `--with-freetype2`

matt335672 (commented):
LGTM apart from the one comment

PR #2391, opened 2022-10-12
xrdp.ini.5.in: Fix mixed up config options

galeksandrp:
It seems to me that descriptions for config options `require_credentials` and `enable_token_login` are mixed up.


matt335672:
Good spot - thanks.

PR #2392, opened 2022-10-13
xrdp_sec.c: add token login for autologon clients

galeksandrp:
Make `token login` available for clients with `RDP_LOGON_AUTO` flag, such as `mstsc.exe` `6.2.22000.282` (`Win11` `22621.608`).

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -1007,10 +1007,10 @@ xrdp_sec_process_logon_info(struct xrdp_sec *self, struct stream *s)
             return 1;
         }
     }
-    else if (self->rdp_layer->client_info.enable_token_login
-             && len_user > 0
-             && len_password == 0
-             && (sep = g_strchr(self->rdp_layer->client_info.username, '\x1f')) != NULL)
+    if (self->rdp_layer->client_info.enable_token_login
+            && len_user > 0
+            && len_password == 0
+            && (sep = g_strchr(self->rdp_layer->client_info.username, '\x1f')) != NULL)
     {
         LOG(LOG_LEVEL_DEBUG, "Client supplied a Logon token. Overwriting password with logon token.");
         g_strncpy(self->rdp_layer->client_info.password, sep + 1,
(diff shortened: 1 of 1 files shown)


matt335672:
Hi @galeksandrp

Thanks for this.

I've been looking at this area of the code from line 1002 to line 1035

https://github.com/galeksandrp/xrdp/blob/22e48dd91bb0e85d5f9eb7229d31b5de71bd0c49/libxrdp/xrdp_sec.c#L1002-L1035

I think it could have done with some refactoring before your change. Whereas I'm happy your change works OK, the code we've ended up with is even more complicated, and it took me a while to understand.

The relevant part of the spec is [[MS-RDPBCGR] 2.2.1.11.1.1](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/732394f5-e2b5-4ac5-8a0a-35345386b0d1). Importantly the password field is not optional, but in the resulting code, it's read in two places which seems unnecessary.

I think this would work. It's a bit simpler:-

```c
    if (unicode_utf16_in(s, len_password, self->rdp_layer->client_info.password, sizeof(self->rdp_layer->client_info.password) - 1) != 0)
    {
        LOG(LOG_LEVEL_ERROR, "ERROR reading password");
        return 1;
    }

    if (self->rdp_layer->client_info.enable_token_login
            && len_user > 0
            && len_password == 0
            && (sep = g_strchr(self->rdp_layer->client_info.username, '\x1f')) != NULL)
    {
        LOG(LOG_LEVEL_DEBUG, "Client supplied a Logon token. Overwriting password with logon token.");
        g_strncpy(self->rdp_layer->client_info.password, sep + 1,
                  sizeof(self->rdp_layer->client_info.password) - 1);
        self->rdp_layer->client_info.username[sep - self->rdp_layer->client_info.username] = '\0';
        self->rdp_layer->client_info.rdp_autologin = 1;
    }
    else if (self->rdp_layer->client_info.require_credentials &&
             !(flags & RDP_LOGON_AUTO))
    {
        LOG(LOG_LEVEL_ERROR, "Server is configured to require that the "
            "client enable auto logon with credentials, but the client did "
            "not request auto logon.");
        return 1; /* credentials on cmd line is mandatory */
    }

```

I haven't tested this yet. What do you think? Have I missed something?

PR #2394, opened 2022-10-19
Update github actions to address warnings

matt335672:
This PR fixes the following warnings issued by the CI system:-

- Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/cache
- The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

For the last of these it is necessary to remove the `-enable-fuse` and `--enable-fdkaac` switches from the 32-bit compile, as the 32-bit packages to support these options are no longer available.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,10 +134,9 @@ jobs:
       CPPCHECK_REPO: https://github.com/danmar/cppcheck.git
     steps:
       # Set steps.os.outputs.image to the specific OS (e.g. 'ubuntu20')
-      # see https://github.com/actions/cache/issues/543
       - name: Get operating system name and version.
         id: os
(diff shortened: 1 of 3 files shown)


matt335672:
Further commit added so that when `ubuntu-latest` switches to 22.04, we get no errors from the C++ CI and the OpenSSL 3.x code.

Issue #2400, opened 2022-10-28
Win 11 mstsc.exe doesn't send RemoteFX capability


metalefty:
## Win 11 

```
[2022-10-28T22:04:33.958+0900] [DEBUG] Client requested compression enabled.
[2022-10-28T22:04:33.960+0900] [DEBUG] Client supplied domain:
[2022-10-28T22:04:33.961+0900] [DEBUG] Client supplied username: meta
[2022-10-28T22:04:33.963+0900] [DEBUG] Client supplied password: <omitted from log>
[2022-10-28T22:04:33.964+0900] [DEBUG] Client supplied program:
[2022-10-28T22:04:33.966+0900] [DEBUG] Client supplied directory:
[2022-10-28T22:04:34.771+0900] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[2022-10-28T22:04:34.782+0900] [INFO ] xrdp_process_offscreen_bmpcache: support level 1 cache size 10485760 MB cache entries 100
[2022-10-28T22:04:34.798+0900] [INFO ] xrdp_caps_process_codecs: nscodec, codec id 1, properties len 3
[2022-10-28T22:04:34.814+0900] [WARN ] xrdp_caps_process_codecs: unknown codec id 5
[2022-10-28T22:04:34.832+0900] [DEBUG] xrdp_00005c68_wm_login_state_event_00000001
```

![image](https://user-images.githubusercontent.com/941609/198595900-1fcd969e-36ab-418b-abe2-b252ee902421.png)

## Win 10

```
[2022-10-28T22:02:07.907+0900] [DEBUG] Client requested auto logon.
[2022-10-28T22:02:07.908+0900] [DEBUG] Client requested compression enabled.
[2022-10-28T22:02:07.910+0900] [DEBUG] Client supplied domain:
[2022-10-28T22:02:07.911+0900] [DEBUG] Client supplied username: meta
[2022-10-28T22:02:07.913+0900] [DEBUG] Client supplied password: <omitted from log>
[2022-10-28T22:02:07.915+0900] [DEBUG] Client supplied program:
[2022-10-28T22:02:07.916+0900] [DEBUG] Client supplied directory:
[2022-10-28T22:02:08.309+0900] [INFO ] xrdp_caps_process_pointer: client supports new(color) cursor
[...]

PR #2401, opened 2022-10-28
Record codec GUID to identify unknown codec

metalefty:
Related to #2400, I was always wondering what is the unknown codec which has codec id 5 that appears in the log very often.

```
[2022-10-28T22:04:34.814+0900] [WARN ] xrdp_caps_process_codecs: unknown codec id 5
```

I added codec GUID to the log in order to identify unknown codecs.  It turned out the one is [Image RemoteFX](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/86507fed-a0ee-4242-b802-237534a8f65e).

```
[2022-10-29T02:34:55.707+0900] [INFO ] xrdp_caps_process_codecs: NSCodec(CA8D1BB9-000F-154F-589FAE2D1A87E2D6), codec id [1], properties len [3]
[2022-10-29T02:34:55.708+0900] [WARN ] xrdp_caps_process_codecs: unknown(2744CCD4-9D8A-4E74-803C0ECBEEA19C54), codec id [5], properties len [49]
[2022-10-29T02:34:55.710+0900] [INFO ] xrdp_caps_process_codecs: RemoteFX(76772F12-BD72-4463-AFB3B73C9C6F7886), codec id [3], properties len [49]      
```

@matt335672 I'd appreciate it if you review the `g_guid_to_string` function. It is the only void function in string_calls. Should it return some value?


common/string_calls.c
```diff
@@ -172,6 +172,23 @@ g_get_display_num_from_display(const char *display_text)
 
     return rv;
 }
+/*****************************************************************************/
```
matt335672: We've already got a `guid_to_str()` function in common/guid.c.

Differences are that it doesn't add the `-` signs, and it takes a `const struct guid *` 

It's used for logging from sesrun.c, and also for generating a password hash for the VNC backend.

Neither of those uses should be affected by adding '-' signs to the result, so I suggest we just replace that body of that function with yours. You'll also need to modify `GUID_STR_SIZE` in guid.h to account for the '-' signs.

common/string_calls.c
```diff
@@ -172,6 +172,23 @@ g_get_display_num_from_display(const char *display_text)
...
+void
+g_guid_to_string(char *dest, const char *src)
+{
+    unsigned char guid[16] = { 0 };
+
+    for (int i = 0; i < 16 ; i++)
+    {
+        guid[i] = (unsigned char) src[i];
+    }
+
```
matt335672: Suggest replacing the above 6 lines (not the 4 displayed) with :-

```c
const unsigned char *guid = (const unsigned char *)src;
```

common/string_calls.c
```diff
@@ -172,6 +172,23 @@ g_get_display_num_from_display(const char *display_text)
...
+    g_sprintf(dest, "%02X%02X%02X%02X-%02X%02X-%02X%02X-%02X%02X%02X%02X%02X%02X%02X%02X",
+              guid[3], guid[2], guid[1], guid[0],
+              guid[5], guid[4],
+              guid[7], guid[6],
+              guid[8], guid[9], guid[10], guid[11], guid[12], guid[13], guid[14], guid[15]);
+}
```
matt335672: Normal textual representation of a `UUID` is 8-4-4-4-12 rather than 8-4-4-16 (see [here](https://en.wikipedia.org/wiki/Universally_unique_identifier) and also [[MS-DTYP] 2.3.4.3](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/222af2d3-5c00-4899-bc87-ed4c6515e80d)). Suggest we use that instead as it's slightly more readable.

libxrdp/xrdp_caps.c
```diff
@@ -521,6 +522,7 @@ xrdp_caps_process_codecs(struct xrdp_rdp *self, struct stream *s, int len)
     int i1;
     char *codec_guid;
```
matt335672, line 523: May need `const struct guid *codec_guid` instead if we use the function in guid.c. Also the cast will need to be changed.

tests/common/test_string_calls.c
```diff
@@ -4,6 +4,7 @@
 #endif
```
matt335672, line 4: Move to `tests/common/test_guid.c` ?

matt335672 (commented):
Hi @metalefty. Looks like a worthwhile addition. There's some duplication with an existing function which I've mentioned in my comments.

common/string_calls.c
metalefty: I didn't notice that. I will rework to use guid.c.

common/string_calls.c
metalefty: Agreed.

tests/common/test_string_calls.c
metalefty: Done.

common/string_calls.c
metalefty: Replaced.

matt335672:
Hi @metalefty.

I've done some more thinking about this. There may be some mileage in updating guid.c to generate proper UUIDs as specified by [RFC4122](https://www.rfc-editor.org/rfc/rfc4122.txt) section 4.4.

I can look at that as a subsequent PR after this one is merged.

metalefty:
Moved some functions and tests to `guid.{c,h}`.

Microsoft's GUID is a little bit different from the common UUID.
See: https://devblogs.microsoft.com/oldnewthing/20220928-00/?p=107221

So I created a separate function `ms_guid_to_str()` to convert MS GUID to string.

libxrdp/xrdp_caps.c
metalefty: I will do this later in this PR. Can you have a look again at other points?

metalefty:
Sorry for the noise of frequent force-pushes. I forgot to remove some unused things in string_calls.

> I've done some more thinking about this. There may be some mileage in updating guid.c to generate proper UUIDs as specified by [RFC4122](https://www.rfc-editor.org/rfc/rfc4122.txt) section 4.4.

Indeed. Current GUID generation is actually not compliant with the spec.

libxrdp/xrdp_caps.c
matt335672: The code looks good to me now.

I've read the link you've provided, and I don't think there's any difference between Microsoft's interpretation of a GUID and RFC4122. In other words, I think you can just call get rid of the existing `guid_to_str()` and replace it with your new function completely. The existing uses of the function should be fine.

matt335672:
One comment added above. Also you're right - I need to fix the GUID generation.

libxrdp/xrdp_caps.c
metalefty: Is it fine to add hyphens for other use cases of  `guid_to_str()`?

metalefty: >  The existing uses of the function should be fine.

Ah, you've already mentioned that.

metalefty: @matt335672 I've replaced your `guid_to_str()` with my `ms_guid_to_str()`.  Can you have a look again?

libxrdp/xrdp_caps.c
```diff
@@ -521,6 +522,10 @@ xrdp_caps_process_codecs(struct xrdp_rdp *self, struct stream *s, int len)
...
     char *next_guid;
+    struct guid guid;
```
metalefty, line 525: I'm not sure `char *next_guid` can be replaced with `struct guid` so just kept.

common/guid.h
```diff
@@ -70,6 +71,7 @@ guid_is_set(const struct guid *guid);
  *            representation
  * @return str is returned for convenience
  */
-const char *guid_to_str(const struct guid *guid, char *str);
+const char *guid_to_str(const struct guid *guid, char *dest);
```
matt335672, line 74: Happy for you to change the param from `str` to `dest`, but the header comment needs updating too.

libxrdp/xrdp_caps.c
matt335672: That looks fine to me.

matt335672: That looks fine to me

matt335672 (approved):
Looks fine, with the noted change to the function header comment

PR #2408, opened 2022-11-03
Bump autoconf ver

matt335672:
The version of autoconf on Ubuntu 22.04 (2.71) produces warnings regarding obsolete macros AC_PROG_LIBTOOL, AC_FD_CC, AC_HELP_STRING and AC_TRY_COMPILE.

This PR bumps the autoconf version to 2.69 (used by CentOS 7 and Ubuntu 16.04) and replaces the obsolete macros with supported ones.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script
 
-AC_PREREQ(2.65)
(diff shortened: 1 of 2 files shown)


metalefty (approved):
I'm surprised that we were still using autoconf 2.65.

PR #2413, opened 2022-11-07
Update pamdir_suse to accommodate with TW pam.d move

YoukouTenhouin:
On newer builds of openSUSE tumbleweed the path of pam.d has moved from /usr/etc/pam.d to /usr/lib/pam.d, which prevents install script to correctly guess pam rules. Updating path in mkpamrules solves the problem.

The script should still work on older SUSE systems since there's still /etc/pam.d/common-account on those, which should allow for correct rule guess.

### DIFF
--- a/instfiles/pam.d/mkpamrules
+++ b/instfiles/pam.d/mkpamrules
@@ -9,6 +9,10 @@ outfile="$3"
 service="xrdp-sesman"
 pamdir="/etc/pam.d"
 pamdir_suse="/usr/lib/pam.d"
+if [ ! -d $pamdir_suse ]; then
+   # Older SUSE distros uses /usr/etc/pam.d
+   pamdir_suse="/usr/etc/pam.d"
+fi
 
 # Modules needed by xrdp-sesman.unix, if we get to that
 unix_modules_needed="pam_unix.so pam_env.so pam_nologin.so"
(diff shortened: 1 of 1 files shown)


instfiles/pam.d/mkpamrules
```diff
@@ -8,7 +8,7 @@ outfile="$3"
 
 service="xrdp-sesman"
 pamdir="/etc/pam.d"
-pamdir_suse="/usr/etc/pam.d"
+pamdir_suse="/usr/lib/pam.d"
```
matt335672, line 11: Since this is a shell script, how about:-

```bash
pamdir_suse="/usr/lib/pam.d"
if [ ! -d "$pamdir_suse" ]; then
    # Pre-Tumbleweed setting
    pamdir_suse="/usr/etc/pam.d"
fi
```

metalefty:
I think it's worth cherry-picking to v0.9 so I applied v0.9.21 milestone.

instfiles/pam.d/mkpamrules
YoukouTenhouin: Done, thanks for the suggestion!

matt335672:
@MosakujiHokuto - thanks for your contribution!

Issue #1317, opened 2019-03-27
xrdp-sesman ignores changes to pam configuration until full restart


dhoffend:
Version: xrdp 0.9.8-2~bpo9+1 (debian backports)

When you install xrdp with localusers only and later install/configure/integrate sssd (for example) for remote users or make any other change to the pam configuration xrdp-sesman completely ignores it. If xrdp-sesman would rediscover it's active sessions (like proposed in #819), it would be an issue and you could simple restart it, but now we've to restart xrdp-sesman (or better reboot the server) to make aware of the pam changes.

Is it possible to reread the pam configuration on HUP signal?

matt335672:
I've just had a look at this, and I'm unable to reproduce the problem as described on Ubuntu 22.04 LTS with xrdp v0.9.8.

I started xrdp-sesman and looked at the files it's reading from the PAM directory:-

```
sudo strace -p 35567 -e trace=openat 2>&1 | grep pam.d
```

I get this on an initial login:-

```
openat(AT_FDCWD, "/etc/pam.d/xrdp-sesman", O_RDONLY) = 9
openat(AT_FDCWD, "/etc/pam.d/common-auth", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-account", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-session", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-password", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/other", O_RDONLY) = 9
openat(AT_FDCWD, "/etc/pam.d/common-auth", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-account", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-password", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-session", O_RDONLY) = 10
```

On a reconnect, I get very similar:-

```
openat(AT_FDCWD, "/etc/pam.d/xrdp-sesman", O_RDONLY) = 9
openat(AT_FDCWD, "/etc/pam.d/common-auth", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-account", O_RDONLY) = 10
openat(AT_FDCWD, "/etc/pam.d/common-session", O_RDONLY) = 10
[...]

PR #2416, opened 2022-11-08
Support and document SIGHUP for xrdp-sesman

### LINKED ISSUE #2414: instfiles/xrdp-sesman.service.in has no ExecReload entry
While xrdp-sesman can reread its configuration file upon reception of a SIGHUP the service file doesn't take advantage of this and the documentation doesn't mention it.
I just kicked 50 people from one of our servers because I (stupidly) believed what "systemctl reload xrdp-sesman.service" told me.
After restarting xrdp-sesman I had to manually cleanup the fallout of the orphaned sessions.
To my surprise I just found code in sesman that will do the right thing when receiving SIGHUP.

matt335672:
Fixes #2414

@yxzzx - can you take a look at this please?

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3044,6 +3044,18 @@ g_sigterm(int pid)
 }
 
 /*****************************************************************************/
+/* does not work in win32 */
+int
+g_sighup(int pid)
+{
+#if defined(_WIN32)
+    return 0;
+#else
+    return kill(pid, SIGHUP);
+#endif
+}
+
+/*****************************************************************************/
 /* returns 0 if ok */
 /* the caller is responsible to free the buffs */
 /* does not work in win32 */
--- a/common/os_calls.h
+++ b/common/os_calls.h
@@ -192,6 +192,7 @@ char    *g_getenv(const char *name);
 int      g_exit(int exit_code);
 int      g_getpid(void);
 int      g_sigterm(int pid);
+int      g_sighup(int pid);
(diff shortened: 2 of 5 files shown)


yxzzx:
Isn't it a little bit asymmetric to do the kill with "xrdp-sesman --kill" and the reload with "kill -HUP $MAINPID". I think it would be cleaner to either remove the "--kill" code from sesman or add a "--reload" option to sesman. But that's probably just me being my usual obsessive-compulsive self...

Anyway, I think your commit will be doing the job just fine. Thanks for the fix and the documentation.

matt335672:
Hi @yxzzx 

Thanks for the swift reply.

No - that's a very good point. It absolutely should be consistent. Leave it with me.

metalefty (commented):
Changes LGTM.

BTW, is the log file reopened after reloading?

matt335672:
The code for a log restart was added for 2484928a5a47f5dba5ddd087d893c6c968b35dbb :-

https://github.com/neutrinolabs/xrdp/blob/4ff968bc980f6ff72c2e8312ce98ed67c2f6f9d8/common/log.c#L727-L782

Since the log file is shared, we never change it, as we've got no way of telling any child processes that the log file has changed. If the user tries to change it we log a warning and carry on using the old one.

That's the current behaviour anyway.

danielperna84:
Will this also affect #1317 ?

matt335672:
@danielperna84 - thanks for the heads-up.

As far as I can tell, #1317 isn't actually a problem. More information would be required to figure out exactly what was going on there.

PR #2420, opened 2022-11-10
guid_new() returns GUIDs compatible with RFC4122

matt335672:
This small PR was prompted by a [comment by @metalefty](https://github.com/neutrinolabs/xrdp/pull/2401#issuecomment-1298324018) in #2401.

Generated GUIDs for sessions are now compliant with [RFC4122 section 4.4](https://www.rfc-editor.org/rfc/rfc4122#page-14)

### DIFF
--- a/common/guid.c
+++ b/common/guid.c
@@ -30,11 +30,27 @@
 #include "os_calls.h"
 #include "string_calls.h"
 
+enum
+{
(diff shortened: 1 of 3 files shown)


metalefty (approved):
LGTM. Thanks!

PR #2364, opened 2022-09-12
Enable touchpad inertial scrolling

seflerZ:
See this [PR](https://github.com/neutrinolabs/xorgxrdp/pull/234)

@metalefty

PR #2424, opened 2022-11-12
Regain scrolling when using VNC backend

metalefty:
@seflerZ Fix for #2364.  Also added mouse buttons 6, 7, and 8 support.

This is a draft PR.  I haven't adjusted the scroll speed yet. I just bring back scroll. Comments and suggestions are welcome especially on implementation of `process_touchpad_scroll()`.

### DIFF
--- a/vnc/vnc.c
+++ b/vnc/vnc.c
@@ -680,7 +680,7 @@ process_touchpad_scroll(struct vnc *v, int axis, long param1, long param2,
      * Right -> Button7
      *
      */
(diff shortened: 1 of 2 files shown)


seflerZ:
Okay, Thanks @metalefty . I've found that we can use 'xrdp_mm->code' to force XVNC session using the old style. I've made a PR #2425  to fix this. It's pretty intutive but I don't have XVNC session configured yet. Would you please test it for me?

PR #2425, opened 2022-11-14
Disable touch events on XVNC session

seflerZ:
According to #2424 , XVNC has no touch implemented yet. Here we temporarily disable them on XVNC session and force it the old way.

### DIFF
--- a/xrdp/xrdp_types.h
+++ b/xrdp/xrdp_types.h
@@ -32,6 +32,14 @@
 #define MAX_NR_CHANNELS 16
 #define MAX_CHANNEL_NAME 16
 
+/* Code values used in 'xrdp_mm->code=' settings */
+#define XVNC_SESSION_CODE 0
+#define XRDP_SESSION_CODE 10
+#define XORG_SESSION_CODE 20
+
+/* To check whether touch events has been implemented on session type 'mm' */
+#define XRDP_MM_IMPLEMENTS_TOUCH(mm) ((mm)->code != XVNC_SESSION_CODE)
+
 struct source_info;
 
 /* lib */
--- a/xrdp/xrdp_wm.c
+++ b/xrdp/xrdp_wm.c
@@ -29,11 +29,6 @@
 #include "log.h"
 #include "string_calls.h"
 
-/* Code values used in 'xrdp_mm->code=' settings */
-#define XVNC_SESSION_CODE 0
-#define XRDP_SESSION_CODE 10
-#define XORG_SESSION_CODE 20
-
 /*****************************************************************************/
 struct xrdp_wm *
 xrdp_wm_create(struct xrdp_process *owner,
@@ -1816,9 +1811,7 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
              * The negative number is represented by complement.
              */
             delta = (device_flags & WheelRotationMask) | ~WheelRotationMask;
-
-            // XVNC session has not implemented the touch event for now
-            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
+            if (delta != 0 && XRDP_MM_IMPLEMENTS_TOUCH(self->mm))
             {
                 // Use nature scrolling, up direction is negative.
                 xrdp_wm_mouse_touch(self, TOUCH_TWO_FINGERS_UP, delta);
@@ -1831,9 +1824,7 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
         else
         {
             delta = device_flags & WheelRotationMask;
-
-            // XVNC session has not implemented the touch event for now
-            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
+            if (delta != 0 && XRDP_MM_IMPLEMENTS_TOUCH(self->mm))
             {
                 xrdp_wm_mouse_touch(self, TOUCH_TWO_FINGERS_DOWN, delta);
             }
@@ -1866,9 +1857,7 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
              * The negative number is represented by complement.
              */
             delta = (device_flags & WheelRotationMask) | ~WheelRotationMask;
-
-            // XVNC session has not implemented the touch event for now
-            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
+            if (delta != 0 && XRDP_MM_IMPLEMENTS_TOUCH(self->mm))
             {
                 // Use nature scrolling, right direction is negative.
                 xrdp_wm_mouse_touch(self, TOUCH_TWO_FINGERS_RIGHT, delta);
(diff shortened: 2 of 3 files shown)


metalefty:
Wow, that's pretty simple. I'll do some tests, thanks!

metalefty:
@seflerZ Could you rebase on current head of devel branch for easy testing?

seflerZ:
Okay发自我的手机-------- 原始邮件 --------发件人： metalefty ***@***.***>日期： 2022年11月14日周一 21:09收件人： neutrinolabs/xrdp ***@***.***>抄送： Sefler Zhou ***@***.***>, Mention ***@***.***>主    题： Re: [neutrinolabs/xrdp] Disable touch events on XVNC session (PR #2425)
@seflerZ Could you rebase on current head of devel branch for easy testing?

seflerZ:
Okay, rebase done. @metalefty

matt335672:
Thanks for this @seflerZ 

I'm happy with the approach in principle, but rather than using the code to do this, wouldn't it be better to add a query function to the module so we can just ask the module if it supports touch events?

xrdp/xrdp_wm.c
```diff
@@ -29,7 +29,10 @@
 #include "log.h"
 #include "string_calls.h"
 
-
+/* Code values used in 'xrdp_mm->code=' settings */
+#define XVNC_SESSION_CODE 0
+#define XRDP_SESSION_CODE 10
+#define XORG_SESSION_CODE 20
```
matt335672: Could you move the above definitions into xrdp_types.h and combines them with the similar ones in xrdp_mm.c?

xrdp/xrdp_wm.c
```diff
@@ -29,7 +29,10 @@
...
 
 /*****************************************************************************/
```
matt335672, line 32: Also suggest adding this define to xrdp_types.h to avoid code duplication in xrdp_wm.c:-

```c
// XVNC session has not implemented the touch event for now
#define XRDP_MM_IMPLEMENTS_TOUCH(mm) ((mm)->code != XVNC_SESSION_CODE)
```

xrdp/xrdp_wm.c
```diff
@@ -1812,7 +1815,9 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
              * The negative number is represented by complement.
              */
             delta = (device_flags & WheelRotationMask) | ~WheelRotationMask;
-            if (delta != 0)
+
+            // XVNC session has not implemented the touch event for now
+            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
```
matt335672: Can use `XRDP_MM_IMPLEMENTS_TOUCH(self->mm)` if you use the define suggested. Then the comment is not needed.

xrdp/xrdp_wm.c
```diff
@@ -1825,7 +1830,9 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
         else
         {
             delta = device_flags & WheelRotationMask;
-            if (delta != 0)
+
+            // XVNC session has not implemented the touch event for now
+            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
```
matt335672: Can use `XRDP_MM_IMPLEMENTS_TOUCH(self->mm)` if you use the define suggested. Then the comment is not needed.

xrdp/xrdp_wm.c
```diff
@@ -1858,7 +1865,9 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
              * The negative number is represented by complement.
              */
             delta = (device_flags & WheelRotationMask) | ~WheelRotationMask;
-            if (delta != 0)
+
+            // XVNC session has not implemented the touch event for now
+            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
```
matt335672: Can use `XRDP_MM_IMPLEMENTS_TOUCH(self->mm)` if you use the define suggested. Then the comment is not needed.

xrdp/xrdp_wm.c
```diff
@@ -1871,7 +1880,9 @@ xrdp_wm_process_input_mouse(struct xrdp_wm *self, int device_flags,
         else
         {
             delta = device_flags & WheelRotationMask;
-            if (delta != 0)
+
+            // XVNC session has not implemented the touch event for now
+            if (delta != 0 && self->mm->code != XVNC_SESSION_CODE)
```
matt335672: Can use `XRDP_MM_IMPLEMENTS_TOUCH(self->mm)` if you use the define suggested. Then the comment is not needed.

matt335672 (commented):
To further elaborate on my comment I've made some suggestions in a review below. Come back to me if this isn't clear.

seflerZ:
> Thanks for this @seflerZ
> 
> I'm happy with the approach in principle, but rather than using the code to do this, wouldn't it be better to add a query function to the module so we can just ask the module if it supports touch events?

Yes @matt335672 , querying the module to test if it supports touch events fully complies with the "Information Expert" desgin pattern. It's good design. But since we only have three modules and it's a temporary shutdown for the XVNC solution. If we try to use that pattern, we need to introduce new abstract method for each module. I think it's better to use macros as you suggested at present to get similar querying efforts.

seflerZ:
> To further elaborate on my comment I've made some suggestions in a review below. Come back to me if this isn't clear.

Yes, it's very intutive and clear. I'll fix them.

xrdp/xrdp_wm.c
seflerZ: Done

seflerZ: > Also suggest adding this define to xrdp_types.h to avoid code duplication in xrdp_wm.c:-
> 
> ```c
> // XVNC session has not implemented the touch event for now
> #define XRDP_MM_IMPLEMENTS_TOUCH(mm) ((mm)->code != XVNC_SESSION_CODE)
> ```

Done

seflerZ:
@metalefty @matt335672 Refactored, would you please check again?

xrdp/xrdp_types.h
```diff
@@ -32,6 +32,14 @@
 #define MAX_NR_CHANNELS 16
 #define MAX_CHANNEL_NAME 16
 
+/* Code values used in 'xrdp_mm->code=' settings */
+#define XVNC_SESSION_CODE 0
+#define XRDP_SESSION_CODE 10
```
metalefty, line 37: XRDP in this context means x11rdp. It is deprecated and replaced with xorgxrdp.
I'm not sure x11rdp supports touch events. However, that's fine for the moment.

metalefty:
One comment for the record but LGTM. Let's merge if @matt335672 approved.

metalefty (approved):

matt335672:
Agreed - thanks @seflerZ

PR #2427, opened 2022-11-14
Fix freetype2 version checking

matt335672:
The version of freetype2 returned by `pkg-config --modversion` is not the same as the freetype2 release version - more information [here](/egorodet/FreeType2/blob/master/docs/VERSIONS.TXT).

Consequently, the existing autoconfigure check for version >= 2.8.0 does not work.

This version was chosen as 2.8.x was the easiest early one to test on - Ubuntu 18.04 uses 2.8.1 and CentOS 7 uses 2.8.14.

I've just discovered the check doesn't work for Ubuntu 16.04 which only has version 2.6.1 of libfreetype6. Without this PR, the configure step works fine but the compilation fails. The failure is due to a change made in version 2.6.3 relating to removing identifiers with double underscores '__'. 

With this PR, the following error is generated on Ubuntu 16.04 if `--with-freetype2` is used:-

```
configure: error: please install version 2_8_0 or later of libfreetype6-dev or freetype-devel
```

Omitting `--with-freetype2` from the configure simply means that the new font generation utility `xrdp-mkfv1` will not be available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -281,14 +281,21 @@ if test x$use_imlib2 = xyes; then
 fi
 
 # Find freetype2
+#
+# The modversion used by pkgcheck does not correspond to the
+# freetype2 release. See docs/VERSIONS.TXT in the freetype2
+# source for a table of correspondences. If you change one
+# of the below defines, change both.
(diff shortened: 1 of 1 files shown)


metalefty (approved):
LGTM

PR #2438, opened 2022-11-24
Remove unnecesssary data from struct auth_info

### LINKED ISSUE #1599: Password Leakage
reproduce step:
1. using rdp client connect xrdp server, login
2. using `gdb` attach xrdp-sesman process
3. using `find` command to search your password if you installed `peda` plugin for gdb. You will find your password in plaintext on heap.

fix:
after authentication, xrdp-sesman should wipe out saved password on heap.

matt335672:
When using PAM authentication, a copy is made of the username and password in the auth_info structure.

The password copy is not cleared from memory when the structure is deallocated. This could mean the password is revealed to an attacker from a coredump.

One solution is to clear the password when the struct is deallocated. However, the username and password in the auth_info struct are only required for the duration of the PAM conversation function. A better solution is to remove the username and password from the auth_info struct entirely, and just use pointers for the duration of the time the callback function is used. This is the authentication and authorization phases of PAM.

### DIFF
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -37,21 +37,17 @@
 #include <stdio.h>
 #include <security/pam_appl.h>
 
-/* Defines the maximum size of a username or password. With pam there is no real limit */
-#define MAX_BUF 8192
-
+/* Allows the conversation function to find the username and password */
 struct t_user_pass
 {
-    char user[MAX_BUF];
-    char pass[MAX_BUF];
+    const char *user;
+    const char *pass;
 };
 
 struct auth_info
 {
-    struct t_user_pass user_pass;
     int session_opened;
     int did_setcred;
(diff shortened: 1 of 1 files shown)


matt335672:
I've added an extra check.

The conversation function shouldn't be called after authentication, but I can't find anything in the PAM spec that says that. So when the password pointer goes out of scope, I've removed it from the argument to the conversation function.

Tested on Ubuntu 22.04 and FreeBSD 13.

metalefty (approved):
Thanks! LGTM.

I recall someone reported this issue. It was #1599.  I think it's worth backporting this to v0.9.

matt335672:
Well remembered. This is the only place that sesman places the password on the heap. With v1.0, because we're using libipm, the password will be zeroed out when the auth call returns. I need to do a bit more work for v0.9.

PR #2460, opened 2022-12-09
Set permissions on pcsc socket dir to owner only

matt335672:
There is no reason for any user other than the current one to be able to communicate with the remote smartcard.


metalefty:
Definitey.

metalefty (approved):

Issue #2466, opened 2022-12-12
File instfiles/pam.d/xrdp-sesman.system not in release archive


gdsotirov:
File `instfiles/pam.d/xrdp-sesman.system` is not present in release archive available from Assets (e.g. [xrdp-0.9.21.tar.gz](https://github.com/neutrinolabs/xrdp/releases/download/v0.9.21/xrdp-0.9.21.tar.gz)). This leads to the following error, when [instfiles/pam.d/mkpamrules](https://github.com/neutrinolabs/xrdp/blob/devel/instfiles/pam.d/mkpamrules) script decides to use it:

```
make[3]: Entering directory '/usr/src/tmp/xrdp-0.9.21/instfiles/pam.d'
./mkpamrules auto . xrdp-sesman
Cannot find ./xrdp-sesman.system
```

I'm not sure how this archive is built, but the file should be present into it.

Related: #1558, #1560

matt335672:
Sorry @gdsotirov - this is entirely my fault.

The build works fine from a `git clone`, but the releases are built using the output of `make distcheck`. This requires all the files which are to be distributed to be referenced somehow. I should have added the file to `instfiles/pam.d/Makefile.am` when I put the patch together for #1560.

I'll put a PR together to fix this.

gdsotirov:
@matt335672 No problem and thanks for the prompt reaction!

PR #2468, opened 2022-12-12
Add xrdp-sesman.system to distributed files

matt335672:
Distributed file list also now sorted for ease of checking

References; #2466 and #2467

May be needed for any further release of the 0.9.x series.


metalefty (approved):
Let's make another release for this.

PR #2472, opened 2022-12-14
Implement uds authentication

### LINKED ISSUE #146: os_calls.c: g_sck_get_peer_cred() needs to be patched for FreeBSD
On FreeBSD, g_sck_get_peer_cred() in os_calls.c always returns `1` and it prevents xup from working. I have no pull request yet but to be patched like [this](http://lists.freedesktop.org/archives/wayland-devel/2013-February/007443.html).

### LINKED ISSUE #642: confusing message when max sessions exceeded
When sessions limit exceeded, logwindow says "login failed for display 0".  This is not user friendly. The message should be "sessions limit exceeded". 

I looked at sesman, it seems sesman doesn't return the reason of login failure to xrdp. We should think about sesman design.  The reason of login failure should be reported to the user not just written to log.

xrdp-mm

### LINKED ISSUE #909: sesman should return auth result/reason to xrdp

### LINKED ISSUE #920: Issue with pam_winbind and trusted domains.
Hi,

I've tested this with both Ubuntu 17.10 and 17.04 with the same result, everything works perfectly as long as I remember to purge xserver-xorg-legacy, I can log in with AD users via winbind etc, etc... but if a user from another (trusted) domain tries to log in it gives a blank screen for a long time and then fails to log in, showing this error in the log:-

auth.log:Oct 18 15:31:49 terminal01 xrdp-sesman[1285]: pam_unix(xrdp-sesman:auth): authentication failure; logname= uid=0 euid=0 tty=xrdp-sesman ruser= rhost=  user=CORP\user.name
auth.log:Oct 18 15:31:49 terminal01 xrdp-sesman[1285]: pam_winbind(xrdp-sesman:auth): getting password (0x00000388)
auth.log:Oct 18 15:31:49 terminal01 xrdp-sesman[1285]: pam_winbind(xrdp-sesman:auth): pam_get_item returned a password
auth.log:Oct 18 15:32:50 terminal01 xrdp-sesman[1285]: pam_winbind(xrdp-sesman:auth): internal module error (retval = PAM_AUTHINFO_UNAVAIL(9), user = 'CORP\user.name')

This user is able to log in every time with ssh, but cannot log in to xrdp, and I'm not certain why... it looks like pam_unix failed it, which is correct becuase there is no local user, and then it tried pam_winbind, but something barfed on the login name or something.

### LINKED ISSUE #1303: Bootstrap xrdp on startup for future connections
Is it possible to "bootstrap" xrdp on a headless server for accepting future connections?

Using Ubuntu Server 18.04, I install xrdp, xorgxrdp, and xfce or mate and then create an image out of the result. When booting the image on a fresh server, I'd like to have RDP connectivity as fast as possible. The first RDP connection takes up to 20 seconds to initialize (and has 5-10 seconds where the screen is completely black), but subsequent connections are much faster (they reuse the same xorg display and xfce/mate session).

Is it possible to get an xrdp session started when the machine boots up so that when I RDP into it, the xorg display and mate session are ready and initialized?

----
what I tried:

running `xrdp-sesrun 127.0.0.1 ubuntu ...` 

I get 
```
sesman config:
	ListenAddress:            127.0.0.1
	ListenPort:               3350
	EnableUserWindowManager:  1
	UserWindowManager:        startwm.sh
...
ok 1 display 10
The log reference is NULL - log not initialized properly
```

but `ps aux | grep Xorg` and `ps aux | grep mate` shows that Xorg and mate have not started running.

`xrdp-sesadmin` shows:
```
$ xrdp-sesadmin -u=ubuntu ...
[20190307-20:51:41] [DEBUG] libscp initialized
[20190307-20:51:41] [INFO ] [v1c_mng:405] connection ok
Session ID: 1704
	Session type: 2
	Screen size: 1024x768, color depth 24
	Idle time: 0 day(s) 0 hour(s) 0 minute(s)
	Connected: 2019/03/07 20:51
[20190307-20:51:41] [DEBUG] Closed socket 4 (AF_INET6 ::1 port 52586)
```

then when I connect via RDP it shows
```
$ xrdp-sesadmin -u=ubuntu ...
[20190307-20:54:10] [DEBUG] libscp initialized
[20190307-20:54:10] [INFO ] [v1c_mng:405] connection ok
Session ID: 1744
	Session type: 3
	Screen size: 1248x1209, color depth 24
	Idle time: 0 day(s) 0 hour(s) 0 minute(s)
	Connected: 2019/03/07 20:54
[20190307-20:54:10] [DEBUG] Closed socket 4 (AF_INET6 ::1 port 52600)
```

it seems the old session (id 1704) is replaced by 1744. However it clearly still has to start Xorg and mate and that takes time.

I thought running freerdp as a client against localhost could force xrdp to make a session, but unfortunately xfreerdp is an x program and requires a display.

Thanks in advance for any help.

### LINKED ISSUE #1739: MaxLoginRetry for xrdp-sesman effectively broken
#### Issue description
The MaxLoginRetry parameter in /etc/xrdp/sesman.ini is effectively broken. Attackers can make an infinite number of login attempts in a single session because they are never disconnected.

Observed on Ubuntu 20.04 with:
- xrdp 0.9.12-1 as shipped by Ubuntu
- xrdp commit 19260cc90cab7e053a661d74ee04cbde92c2615a

#### Steps to reproduce the issue
1. Install xrdp. 
2. Set /etc/xrdp/sesman.ini:MaxLoginRetry=1
3. (Re)start xrdp. Verify that xrdp-sesman has been restarted as well.
4. Connect to xrdp with any RDP client.
5. Enter random credentials in the login dialog and click [OK].
6. Repeat Step 5 several times. Note that you are never disconnected.

#### What's the expected result?
xrdp sessions should be disconnected after ${MaxLoginRetry} failed logins.

#### What's the actual result?
Sessions are never disconnected, regardless of the number of failed attempts. 

#### Additional Details
Discussion in the xrdp-questions gitter:
https://gitter.im/neutrinolabs/xrdp-questions?at=5fbce7610451324f15318c90

### LINKED ISSUE #1823: Session for winbind users is not always reconnect
When a winbind user has started a xrdp session using DOMAIN\username. Then he disconnect the session and later try’s to reconnect. But he uses only the username, the session is then not reconnected but a new session is created.

OS: Ubuntu 20.04.2 LTS
Version: 0.9.12

matt335672:
**Update : ready for review**
Fixes #642
Fixes #909
Fixes #1739
Fixes #1823 

See also #1921
This is moving some way towards the architectural changes in #1961.

This series of commits separates authentication+authorization (AA) from commands in the sesman interface. AA can be either via username/password, or via Unix Domain Socket (UDS) ownership.

Here are some notes for the commits:-
- More information is now provided to the user in the event of a session not being created.
-  sesrun can use either password or UDS logins. With some limitations, this can allow for automatic creation of sessions for local users without a password being needed.
- sesadmin now operates using UDS logins only and so a username and password are not required. To use sesadmin for another user, use `su`/`sudo`/`doas` to authenticate as the other user.
- `MaxLoginRetry` now has an effect.
- Configurations using federated naming services (e.g. AD) no longer have trouble with sessions where the same user can be referred to in more than one way.

The most significant of these is that the sesman tools are starting to become more useful.

### DIFF
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -30,15 +30,6 @@
 
 #include "scp.h"
 
-#ifdef USE_PAM
-#if defined(HAVE__PAM_TYPES_H)
-#define LINUXPAM 1
-#include <security/_pam_types.h>
-#elif defined(HAVE_PAM_CONSTANTS_H)
-#define OPENPAM 1
-#include <security/pam_constants.h>
-#endif
-#endif /* USE_PAM */
 #include <ctype.h>
 
 #include "xrdp_encoder.h"
@@ -47,10 +38,6 @@
 
 
 /* Forward declarations */
-static const char *
-getPAMError(const int pamError, char *text, int text_bytes);
-static const char *
-getPAMAdditionalErrorInfo(const int pamError, struct xrdp_mm *self);
 static int
 xrdp_mm_chansrv_connect(struct xrdp_mm *self, const char *port);
 static void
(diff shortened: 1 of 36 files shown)


matt335672:
A couple of screenshots showing the improved logging:-

## Max session limit reached
image

## Wrong password
image

Command sequence to start a session for a user from the command line, and list sessions:-

<pre>
$ <b>sudo -u testuser xrdp-sesrun -t Xvnc</b>
ok display=:11 GUID=DD94AEC9-F12D-44CE-A22C-ACFDE96F4E43
$ <b>sudo -u testuser xrdp-sesadmin -c=list</b>
Session ID: 144542
	Display: :11
	User: testuser
	Session type: Xvnc
	Screen size: 1280x1024, color depth 32
	Started: Thu Dec 15 10:19:53 2022
</pre>

I've not made major functional changes to the commands yet. I haven't implemented the admins group to be able to list all sessions, and I haven't started to tidy up the horrible interface to `xrdp-sesadmin`.

matt335672:
Just tested on FreeBSD, and found g_sck_get_peer_cred() was broken.

Additional commit fixes #146

matt335672:
This also addresses #1303 I believe, at least in some environments.

xrdp/xrdp_mm.c
```diff
@@ -233,85 +220,63 @@ xrdp_mm_get_value_int(struct xrdp_mm *self, const char *aname, int def)
...
-                type = SCP_SESSION_TYPE_XVNC;
-                break;
+        case XVNC_SESSION_CODE:
+            type = SCP_SESSION_TYPE_XVNC;
+            break;
 
-            case XRDP_SESSION_CODE:
-                type = SCP_SESSION_TYPE_XRDP;
-                break;
+        case XRDP_SESSION_CODE:
```
metalefty, line 250: I think we can remove this support for [x11rdp](https://github.com/neutrinolabs/x11rdp) in xrdp 1.0. It is replaced with xorgxrdp and been kept for compatibility for a while. It's time to remove it because xrdp 1.0 is a big jump.

However, it can be done later.

metalefty (approved):
Overall, LGTM. Let's merge and fix issues (if found) later.

xrdp/xrdp_mm.c
matt335672: Agreed.

PR #2481, opened 2022-12-21
Tiny fix in pseudocode description of startwm.sh

otaconix:
I realize there's not much value in this PR, but I noticed this, so I thought I'd fix it :slightly_smiling_face:


metalefty (approved):
Indeed, thanks!

matt335672:
For V0.9.22 see #2508

PR #2487, opened 2022-12-30
Disable autologon for empty passwords

matt335672:
See #2474 

Thanks for raising this, @igh-gr

Although there is nothing in the specification to prevent automatic logons with empty passwords, this is not a secure default.

The autologon flag INFO_AUTOLOGON ([MS-RDPBCGR] 2.2.1.11.1.1) is now ignored for empty passwords.

Tested with Remmina 1.4.29 which sets INFO_AUTOLOGON if a username is specified without a password.


metalefty (approved):

matt335672:
For V0.9.22 see #2509

PR #2492, opened 2023-01-07
Add xrdp-waitforx to wait for X to start with RandR outputs

### LINKED ISSUE #2436: Don't start window managers until there's an ouptut
Testing [fvwm3](https://github.com/fvwmorg/fvwm3/) with xorgxrdp backend on FreeBSD 13.1 it appears xrdp is running startwm.sh (UserWindowManager or DefaultWindowManager) before there's an output (ready/registered/initialized?).

9/10 when running fvwm3 this section of [code](https://github.com/fvwmorg/fvwm3/blob/master/libs/FScreen.c#L527
) will fail.   `res` is non-NULL however `res->noutput == 0`.   Resulting in `goto randr_fail` exiting fvwm3.

Putting `sleep 1` in startwm.sh allows `XRRGetScreenResources` return value struct to have `noutput > 0`.

It's possible this is just the age of the hardware I'm running on (2011 ThinkServer TS130)?   Recommending a sleep in startup.sh is the fix otherwise should xrdp wait until there's at least 1 output until it runs startwm.sh?  Or maybe this is a possible issue with FreeBSD Xorg or xorgxrdp?

derekschrock:
For some window managers (fvwm2 and fvwm3) if the X server isn't
running and has output it's possible for the window manager to fail or
reconfigure randr incorrectly.

With xrdp-waitfox:
 - Install xrdp-waitfox to the BIN dir.
 - sesman will run xrdp-waitfox as the logged in user.
 - Set an alarm to exit after 30 seconds.
 - Try to open env DISPLAY value's display (10 seconds).
 - Test for RandR extension.
 - Wait for outputs to appear (10 seconds).

Fixes #2436

### DIFF
--- a/waitforx/waitforx.c
+++ b/waitforx/waitforx.c
@@ -5,13 +5,19 @@
 #include <sys/signal.h>
 #include <unistd.h>
 
+#include "config_ac.h"
+#include "os_calls.h"
+#include "string_calls.h"
+
 #define ATTEMPTS 10
 #define ALARM_WAIT 30
 
 void
 alarm_handler(int signal_num)
 {
-    printf("Unable to find RandR outputs after %d seconds\n", ALARM_WAIT);
+    /* Avoid printf() in signal handler (see signal-safety(7)) */
+    const char msg[] = "Timed out waiting for RandR outputs\n";
+    g_file_write(1, msg, g_strlen(msg));
     exit(1);
 }
 
@@ -30,8 +36,7 @@ main(int argc, char **argv)
 
     display = getenv("DISPLAY");
 
-    signal(SIGALRM, alarm_handler);
-    alarm(ALARM_WAIT);
+    g_set_alarm(alarm_handler, ALARM_WAIT);
 
     if (!display)
     {
@@ -48,7 +53,7 @@ main(int argc, char **argv)
             printf("Opened display %s\n", display);
             break;
         }
-        sleep(1);
+        g_sleep(1000);
     }
 
     if (!dpy)
@@ -78,7 +83,7 @@ main(int argc, char **argv)
                 }
                 XRRFreeScreenResources(res);
             }
-            sleep(1);
+            g_sleep(1000);
         }
 
         if (outputs > 0)
--- a/waitforx/waitforx.c
+++ b/waitforx/waitforx.c
@@ -1,0 +1,96 @@
+#include <X11/extensions/Xrandr.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/signal.h>
+#include <unistd.h>
+
+#define ATTEMPTS 10
+#define ALARM_WAIT 30
+
+void
+alarm_handler(int signal_num)
+{
+    printf("Unable to find RandR outputs after %d seconds\n", ALARM_WAIT);
+    exit(1);
+}
+
+int
+main(int argc, char **argv)
+{
+    char *display = NULL;
+    int error_base = 0;
+    int event_base = 0;
+    int n = 0;
+    int outputs = 0;
+    int wait = ATTEMPTS;
+
+    Display *dpy = NULL;
+    XRRScreenResources *res = NULL;
+
+    display = getenv("DISPLAY");
+
+    signal(SIGALRM, alarm_handler);
+    alarm(ALARM_WAIT);
+
+    if (!display)
+    {
+        printf("DISPLAY is null");
+        exit(1);
+    }
+
+    for (n = 1; n <= wait; ++n)
+    {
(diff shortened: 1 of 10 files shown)


waitforx/waitforx.c
```diff
@@ -0,0 +1,96 @@
...
+    if (!display)
+    {
+        printf("DISPLAY is null");
+        exit(1);
+    }
+
+    for (n = 1; n <= wait; ++n)
+    {
+        dpy = XOpenDisplay(display);
+        printf("Opening display %s. Attempt %d of %d\n", display, n, wait);
```
matt335672, line 50: Using `printf` here raises an interesting question about some of the wrapper functions in `common/os_calls.h`.

Other bits of xrdp using `printf()` rather than `g_printf()`

Personally I can't see that we gain anything by using `g_printf()` over `printf()`. I think we should (over time) be removing the wrappers for standard C functions and simply use the `<stdio.h>` (etc) includes.

@metalefty - what's your take on this? Should we start a discussion?

waitforx/waitforx.c
```diff
@@ -0,0 +1,96 @@
...
+        if (dpy != NULL)
+        {
+            printf("Opened display %s\n", display);
+            break;
+        }
+        sleep(1);
```
matt335672: `sleep()` isn't a standard C function - can we use `g_sleep(1000)` here?

waitforx/waitforx.c
```diff
@@ -0,0 +1,96 @@
...
+    int outputs = 0;
+    int wait = ATTEMPTS;
+
+    Display *dpy = NULL;
+    XRRScreenResources *res = NULL;
+
+    display = getenv("DISPLAY");
+
+    signal(SIGALRM, alarm_handler);
+    alarm(ALARM_WAIT);
```
matt335672: For consistency with the other signals these should probably be wrapped.

I'd suggest writing a function which does both the timer and setting the handler, as this could be moved to Windows later (not that this is likely to happen).

How does this look?

```c
unsigned int g_set_alarm(void (*func)(int), unsigned int secs);
```

```c
/* does not work in win32 */
unsigned int
g_set_alarm(void (*func)(int), unsigned int secs)
{
#if defined(_WIN32)
    return 0;
#else
    /* Cancel any previous alarm to prevent a race */
    unsigned int rv = alarm(0);
    signal(SIGALRM, func);
    (void)alarm(secs);
    return rv;
#endif
}
```

waitforx/waitforx.c
```diff
@@ -0,0 +1,96 @@
...
+#include <sys/signal.h>
+#include <unistd.h>
+
+#define ATTEMPTS 10
+#define ALARM_WAIT 30
+
+void
+alarm_handler(int signal_num)
+{
+    printf("Unable to find RandR outputs after %d seconds\n", ALARM_WAIT);
```
matt335672: Can't use `printf()` in a signal handler. See [signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html)

Given this is unlikely to be used on Windows, here's a suggested alternative.

```c
/* Avoid printf() in signal handler (see signal-safety(7)) */
const char msg[] = "Timed out waiting for RandR outputs\n";
g_file_write(1, msg, g_strlen(msg));
```

waitforx/waitforx.c
```diff
@@ -0,0 +1,96 @@
...
+                if (res->noutput > 0)
+                {
+                    outputs = res->noutput;
+                    XRRFreeScreenResources(res);
+                    printf("Found %d output[s]\n", outputs);
+                    break;
+                }
+                XRRFreeScreenResources(res);
+            }
+            sleep(1);
```
matt335672: See above.

matt335672 (commented):
Thanks for this Derek.

waitforx/waitforx.c
metalefty: Regarding `printf()`, I personally think I would like to quit using wrapper function and use `printf()` directly.

matt335672: Thanks. I'll start a discussion about this.

waitforx/waitforx.c
derekschrock: See d7d0866

waitforx/waitforx.c
derekschrock: See 8c44ae0 and d7d0866

waitforx/waitforx.c
derekschrock: See d7d0866

matt335672:
As far as the printf/g_printf discussion goes, this is effectively resolved. See #2499. We'll stick with printf() here I think.

matt335672:
Thanks Derek.

PR #2494, opened 2023-01-09
Add file descriptor passing support for libipm

matt335672:
Add file descriptor passing to libipm.

Additional type 'h' added to libipm types to allow file descriptors to be passed between processes. This is compatible with the D-Bus specification at https://dbus.freedesktop.org/doc/dbus-specification.html

Example send code for an int and a file descriptor:-

```c
int my_int = 10;
int my_fd = g_file_open("/dev/zero");
. . .
status = libipm_msg_out_simple_send(
                 trans,
                 MSG_NO,
                 "ih", my_int, my_fd);
```

Parse with:-

```c
    int32_t my_int;
    int my_fd;
    . . .
    status = libipm_msg_in_parse(
                 trans,
                 "ih",
                 &my_int, &my_fd);
```

Wrapper functions are added to os_calls to do the hard work, and to wrap POSIX `sendmsg()` and `recvmsg()`

Unit tests also added.

This functionality will make implementing the rest of #1961 a bit easier.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1408,7 +1408,7 @@ g_sck_last_error_would_block(int sck)
 
 /*****************************************************************************/
 int
-g_sck_recv(int sck, void *ptr, int len, int flags)
+g_sck_recv(int sck, void *ptr, unsigned int len, int flags)
 {
 #if defined(_WIN32)
     return recv(sck, (char *)ptr, len, flags);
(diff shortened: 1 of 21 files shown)


matt335672:
So it doesn't get lost, this PR should enable [blacklistd](https://www.freebsd.org/cgi/man.cgi?query=blacklistd&sektion=8&format=html) support on FreeBSD/NetBSD to be added pretty simply. See [this comment](/neutrinolabs/xrdp/discussions/1961?sort=top#discussioncomment-1395073) by @metalefty

metalefty:
Sorry for the delay. I'll have some test and look.

metalefty:
After some tests, it looks good to me. Let's merge. I will restart work on blacklistd after merge.

metalefty (approved):

PR #2497, opened 2023-01-12
Use poll() instead of select()

matt335672:
poll() is specified in POSIX.1-2001 as a simpler interface for multiplexed file descriptors than select(). It also provides more
functionality than select()

This PR replaces the select() calls used in xrdp with poll() equivalents.

I'm seeing a 10-20% improvement in performance times using the file redirector with this (Windows laptop talking to Ubuntu 20.04 VM). This may be down to the increased efficiency with which poll() is able to check single file descriptors (which xrdp does a lot of). I haven't got any figures for graphical improvements.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -44,6 +44,7 @@
 #if defined(XRDP_ENABLE_VSOCK)
 #include <linux/vm_sockets.h>
 #endif
+#include <poll.h>
 #include <sys/un.h>
 #include <sys/time.h>
 #include <sys/times.h>
@@ -1594,26 +1595,24 @@ g_sck_socket_ok(int sck)
 int
 g_sck_can_send(int sck, int millis)
 {
-    fd_set wfds;
-    struct timeval time;
-    int rv;
-
-    time.tv_sec = millis / 1000;
-    time.tv_usec = (millis * 1000) % 1000000;
-    FD_ZERO(&wfds);
-
+    int rv = 0;
     if (sck > 0)
     {
-        FD_SET(((unsigned int)sck), &wfds);
-        rv = select(sck + 1, 0, &wfds, 0, &time);
+        struct pollfd pollfd;
 
-        if (rv > 0)
+        pollfd.fd = sck;
+        pollfd.events = POLLOUT;
+        pollfd.revents = 0;
+        if (poll(&pollfd, 1, millis) > 0)
         {
-            return 1;
+            if ((pollfd.revents & POLLOUT) != 0)
+            {
+                rv = 1;
+            }
         }
     }
 
-    return 0;
+    return rv;
 }
 
 /*****************************************************************************/
@@ -1622,77 +1621,64 @@ g_sck_can_send(int sck, int millis)
 int
 g_sck_can_recv(int sck, int millis)
 {
-    fd_set rfds;
-    struct timeval time;
-    int rv;
-
-    g_memset(&time, 0, sizeof(time));
-    time.tv_sec = millis / 1000;
-    time.tv_usec = (millis * 1000) % 1000000;
-    FD_ZERO(&rfds);
+    int rv = 0;
 
     if (sck > 0)
     {
-        FD_SET(((unsigned int)sck), &rfds);
-        rv = select(sck + 1, &rfds, 0, 0, &time);
+        struct pollfd pollfd;
 
-        if (rv > 0)
+        pollfd.fd = sck;
+        pollfd.events = POLLIN;
+        pollfd.revents = 0;
+        if (poll(&pollfd, 1, millis) > 0)
         {
-            return 1;
+            if ((pollfd.revents & (POLLIN | POLLHUP)) != 0)
+            {
+                rv = 1;
+            }
         }
     }
 
-    return 0;
+    return rv;
 }
 
 /*****************************************************************************/
 int
 g_sck_select(int sck1, int sck2)
 {
-    fd_set rfds;
-    struct timeval time;
-    int max;
-    int rv;
+    struct pollfd pollfd[2] = {0};
+    int rvmask[2] = {0}; /* Output masks corresponding to fds in pollfd */
 
-    g_memset(&time, 0, sizeof(struct timeval));
-    FD_ZERO(&rfds);
+    unsigned int i = 0;
+    int rv = 0;
 
     if (sck1 > 0)
     {
-        FD_SET(((unsigned int)sck1), &rfds);
+        pollfd[i].fd = sck1;
+        pollfd[i].events = POLLIN;
+        rvmask[i] = 1;
+        ++i;
     }
 
     if (sck2 > 0)
     {
-        FD_SET(((unsigned int)sck2), &rfds);
+        pollfd[i].fd = sck2;
+        pollfd[i].events = POLLIN;
+        rvmask[i] = 2;
+        ++i;
     }
 
-    max = sck1;
-
-    if (sck2 > max)
+    if (poll(pollfd, i, 0) > 0)
     {
-        max = sck2;
-    }
-
-    rv = select(max + 1, &rfds, 0, 0, &time);
-
-    if (rv > 0)
-    {
-        rv = 0;
-
-        if (FD_ISSET(((unsigned int)sck1), &rfds))
+        if ((pollfd[0].revents & (POLLIN | POLLHUP)) != 0)
         {
-            rv = rv | 1;
+            rv |= rvmask[0];
         }
 
-        if (FD_ISSET(((unsigned int)sck2), &rfds))
+        if ((pollfd[1].revents & (POLLIN | POLLHUP)) != 0)
         {
-            rv = rv | 2;
+            rv |= rvmask[1];
         }
-    }
-    else
-    {
-        rv = 0;
     }
 
     return rv;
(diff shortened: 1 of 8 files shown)


jsorg71:
Nice, this has been on the todo list for a long time.
I'm surprised poll if faster.  I kinda don't believe it.  For a small number of fds I always hear select was fastest but limited.
It will be nice to see if users notice.

matt335672:
The testing I did was limited. I just knocked up this program to check out single file descriptors:-

```c
#include <stdio.h>
#include <sys/socket.h>
#include <sys/time.h>
#include <sys/select.h>
#include <poll.h>

#define ITERATIONS (1000 * 1000)

int checkfd_using_poll(int fd)
{
    struct pollfd pollfd = {fd, POLLIN, 0};
    int rv = 0;

    if (poll(&pollfd, 1, 0) > 1)
    {
        rv = (pollfd.revents & (POLLIN | POLLHUP)) != 0;
    }

    return rv;
}

int checkfd_using_select(int fd)
{
    struct timeval timeout = {0,0};
    int rv = 0;
    fd_set set;
    FD_ZERO(&set);
    FD_SET(fd, &set);
    if (select(fd + 1, &set, NULL, NULL, &timeout) > 0)
    {
        rv = 1;
    }
    return rv;
}

int main()
{
    unsigned int i;
    struct timeval start;
    struct timeval end;
    unsigned long millisecs;

    gettimeofday(&start, NULL);

    for (i = 0 ; i < ITERATIONS ; ++i)
    {
        checkfd_using_poll(0);
    }
    gettimeofday(&end, NULL);

    millisecs = ((end.tv_sec - start.tv_sec) * 1000 +
                 (end.tv_usec - start.tv_usec) / 1000);

    printf("poll : Iterations %u ms %lu\n", ITERATIONS, millisecs);

    gettimeofday(&start, NULL);

    for (i = 0 ; i < ITERATIONS ; ++i)
    {
        checkfd_using_select(0);
    }
    gettimeofday(&end, NULL);

    millisecs = ((end.tv_sec - start.tv_sec) * 1000 +
                 (end.tv_usec - start.tv_usec) / 1000);

    printf("select : Iterations %u ms %lu\n", ITERATIONS, millisecs);
}
```

On my main development machine (elderly Xeon E3 series running Mint) I get these kind of figures:-

```
$ ./temp
poll : Iterations 1000000 ms 501
select : Iterations 1000000 ms 595
$ ./temp
poll : Iterations 1000000 ms 498
select : Iterations 1000000 ms 609
$ ./temp
poll : Iterations 1000000 ms 497
select : Iterations 1000000 ms 617
$ ./temp
poll : Iterations 1000000 ms 514
select : Iterations 1000000 ms 643
```

On the same machine in a FreeBSD 13 VM, I get much closer (and quicker) numbers:-

```
$ ./temp
poll : Iterations 1000000 ms 378
select : Iterations 1000000 ms 390
$ ./temp
poll : Iterations 1000000 ms 374
select : Iterations 1000000 ms 376
$ ./temp
poll : Iterations 1000000 ms 378
select : Iterations 1000000 ms 387
$ ./temp
poll : Iterations 1000000 ms 389
select : Iterations 1000000 ms 376
```

This is file descriptor 0. If I modify the program to dup2() file descriptor 0 on to file descriptor 500, and then run timings on fd 500 I get this on Mint:-

```
$ ./temp
poll : Iterations 1000000 ms 497
select : Iterations 1000000 ms 795
$ ./temp
poll : Iterations 1000000 ms 500
select : Iterations 1000000 ms 796
$ ./temp
poll : Iterations 1000000 ms 499
select : Iterations 1000000 ms 798
$ ./temp
poll : Iterations 1000000 ms 495
select : Iterations 1000000 ms 799
```

And this on FreeBSD 13:-

```
$ ./temp
poll : Iterations 1000000 ms 381
select : Iterations 1000000 ms 995
$ ./temp
poll : Iterations 1000000 ms 381
select : Iterations 1000000 ms 979
$ ./temp
poll : Iterations 1000000 ms 379
select : Iterations 1000000 ms 999
$ ./temp
poll : Iterations 1000000 ms 384
select : Iterations 1000000 ms 1000
```

That's a bit extreme as an example, as we aren't running 500 file descriptors, but it illustrates that `poll()` is a more scaleable solution.

common/os_calls.c
```diff
@@ -1455,26 +1456,24 @@ g_sck_socket_ok(int sck)
...
     {
-        FD_SET(((unsigned int)sck), &wfds);
-        rv = select(sck + 1, 0, &wfds, 0, &time);
+        struct pollfd pollfd;
 
-        if (rv > 0)
+        pollfd.fd = sck;
+        pollfd.events = POLLOUT;
+        pollfd.revents = 0;
+        if (poll(&pollfd, 1, millis) > 0)
```
pnowack, line 1606: You are missing a case here: When `poll()` returns `-1` and errno is set to `EINTR`, then you need to call `poll()` again, i.e. do-while loop it (applies to other poll calls in the other lines as well).

pnowack:
Very interesting those numbers. There is btw another variant, and that is `epoll`, which in some cases is supposed to perform better, but might not be available on all platforms.

jsorg71:
@matt335672 thanks for that test code.  I ran on a few systems, some old some new.  I can't find a system were select is faster.

common/os_calls.c
matt335672: Thanks @pnowack 

I can't see how this change differs from the existing `select()` code - am I missing something? In the original code the select will return -1/EINTR as will poll in the new code. The result of both calls will be to return 0 early.

The calls with a zero timeout will not return EINTR(?) so they won't need a loop.

That leaves the following according to my analysis:-

|Call                    | Analysis |
|----------------------|-------------|
|os_calls.c:1494  | Same as os_calls.c:1467 (i.e. behaviour unchanged) |
|os_calls.c:1927  | Original code also returns early on EINTR |
|xrdpapi.c:491  | Behaviour unchanged |
|xrdpapi.c:512 | Behaviour unchanged |
|xrdp-ssh-agent.c:294 | Behaviour unchanged |
|xrdp-ssh-agent.c:400 | Behaviour unchanged |
|xrdp_pcsc.c:277 | Behaviour unchanged |
|xcon.c:55 | Behaviour unchanged |
|gtcp.c:278 | Behaviour unchanged |

I've also examined what the effect of the existing behaviour would be:-

|Call                    | Behaviour on EINTR |
|----------------------|-------------|
|os_calls.c:1467  | Wait finishes early. Socket is not read until next poll |
|os_calls.c:1494  | Wait finishes early. Socket is not read until next poll |
|os_calls.c:1927  | Wait finishes early and sockets are re-polled |
|xrdpapi.c:491  | **Program can fail** |
|xrdpapi.c:512 | **Program can fail** |
|xrdp-ssh-agent.c:294 | No effect |
|xrdp-ssh-agent.c:400 | No effect |
|xrdp_pcsc.c:277 | No effect, because of unusual program logic in following poll() |
|xcon.c:55 | **Program fails** |
|gtcp.c:278 | Behaviour unchanged |

So there's a case here for fixing xrdpapi.c and xcon.c. I can certainly do that.

There's also the small problem that on an interrupt, we can't easily work out how much of the timeout has expired. A retry loop will result in the timeout possibly being exceeded, or in the pathological case, never completing.

Please come back to me if I'm missing something or I'm not being clear.

matt335672:
@pnowack - epoll() is supported on Linux and FreeBSD, so we could consider using it as an option. It's not supported on MacOS however, and there are a few people trying to get that going.

As ever, [stackoverflow](https://serverfault.com/questions/99504/why-poll-is-not-replaced-with-epoll) has an interesting take on this. I've not done a lot of my own analysis, but is seems that if we use epoll() in a way compatible with poll() we won't see any benefits. If we use it in a way which gives us benefits we cut down on our possible target systems.

There's possibly some scope for improving performance better by changing the way that `g_obj_wait()` works. At present we do the following in a typical loop:-

```
forever
{
    list rfds = <code to get fds>
    list wfds = <code to get fds>
    g_obj_wait(rfds, wfds);

   for fd in rfds; do
        if (is_readable(fd))
    {
        <do something>
    }

   for fd in wfds; do
        if (is_writeable(fd))
    {
        <do something>
    }
}
```

For each file descriptor we're making two system calls before processing it. The first poll() call in g_obj_wait() returns the readability/writeability of each fd. We then throw that away and poll each fd again! The code is, however, easy to read. Whether it's possible to remove the double fd check but keep the code readable is not something I've thought about.

jsorg71:
@matt335672, I've thought about that too.  One reason it's doing the double check is Windows.  WaitForMultipleObjects needs to be followed by WaitForSingleObject for each object.  If we remove Windows support, yea, both select and poll give us the data we need.
Note, this does change the logic as processing an fd can trigger the later fds.  Also, we would have to pass this extra fd info between the poll and the processing which we can do with a struct.  This might get more complicated when doing something like module processing in libxup or libvnc., but doable.

We are not doing 1000s of these per second so we should determine if this will even save a millisecond.  I think your test is doing a million, you can divide that out.

matt335672:
Thanks @Nexarian. Increasing the complexity to get a small performance improvement is not a good idea IMO.

Following @pnowack's comment and my own analysis, I've fixed the latent issues in xrdpapi.c and xcon.c. I've also added a note to `g_sck_can_send()` / `g_sck_can_recv()` to make it clearer that an early completion in the event of a signal should be expected.

Nexarian:
Did I say something about this? I've been watching it but didn't think I had commented...

Anyway, I might beg to differ on "increasing the complexity for a small performance gain is not a good idea." It depends on what you mean by "small." A small gain in something as critical as this system utility could have huge magnifying effects downstream, depending on how many calls we make.

And I've said this before, people who use remote desktop are ridiculously sensitive to ANY extra lag or latency. A system that "feels slightly faster" means the entire IT department ditches the "slightly slower one."

We also want to take into account scaling. I had to add a wait object for the resizing state machine. As XRDP adds features, more of these types of states will likely get added. If we build the idioms in a way that scales optimally performance wise even though it's more complex, to me, that tradeoff is worthwhile.

Now, if the performance gain is small even after we do all our calls, that's different, but I want to make sure we're taking into account the multiplicative effect. The difference between 60 and say, 65 FPS matters.

And my goal is to make XRDP ripping fast so that users have no choice but to use it as the benchmark for fast RDP :)

matt335672:
Hi @Nexarian 

Sorry - got names muddled up in my head.

Yes you're right - that was badly worded of me. I'll try to be clearer.

I don't disagree that performance improvements are important and necessary for the long-term value of this project. I also think however that writing maintainable code is also important. That makes it easier to triage and fix faults and also lowers the barriers of entry for new developers. Complexity's fine, provided it can be documented as you've admirably demonstrated with your EGFX docs. To my mind, it's all a bit of a balancing act.

PR #2504, opened 2023-01-16
Don't try to listen on the scard socket if it isn't there

### LINKED ISSUE #2501: scard_pcsc_init fails when user home directory is a network-based folder
### User Information

* Version: 0.9.21.1-1
* OS: Kali Rolling

### Error Message

```
xrdp-chansrv[####]: [ERROR] scard_pcsc_init: trans_listen failed for port /home/example/username/.pcsc11/pcscd.comm
xrdp-chansrv[####]: [ERROR] scard_pcsc_check_wait_objs: g_lis trans_check_wait_objs error
. . . <Previous entry is duplicated 20-30 times> . . .
```

### Additional Information

Network is isolated and does not use Smart Cards for authentication, therefore no immediate impact is noted at this time. However, with 10-20 users connecting or re-connecting throughout the day, the logs are being saturated with these error messages.

User home directories are mounted via [libpam_mount](https://manpages.debian.org/testing/libpam-mount/pam_mount.8.en.html) from a CIFS share on user login to "sync" files between hosts. Unix socket files do not appear to be supported on CIFS/SMB volumes.

Other applications use environment variables to determine where to create per-process or cache files, i.e. [XDG_CACHE_HOME](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which work around this issue. XRDP appears to rely on the `HOME` environment variable instead.

https://github.com/neutrinolabs/xrdp/blob/9b846b7b7e6b0b59b0b0d9f048acd9e04c65a082/sesman/chansrv/smartcard_pcsc.c#L1856-L1858

### Suggestion/Request

I understand that this is likely an edge case and will be incredibly low priority given the minimal amount of impact. That said, a long-term solution may be to use a new environment variable and default to the user's `HOME` if `NULL`. For example:

```c
home = g_getenv("XRDP_PCSC_IPC_DIR");
if (home == 0)
{
  home = g_getenv("HOME");
}
disp = g_display_num; 
g_snprintf(g_pcsclite_ipc_dir, 255, "%s/.pcsc%d", home, disp); 
```

[Link to g_getenv for your convenience.](https://docs.gtk.org/glib/func.getenv.html)

This would allow a system administrator to place `export XRDP_PCSC_IPC_DIR=/tmp/.xrdp_$UID/` in the `/etc/xrdp/startwm.sh` file to specify the environment variable per-user for each initialized RDP session.

matt335672:
Fixes #2501 

If the scard local socket can't be created, there's no point in trying to listen on it. This will just fill the chansrv log file with errors.

### DIFF
--- a/sesman/chansrv/devredir.c
+++ b/sesman/chansrv/devredir.c
@@ -803,8 +803,7 @@ devredir_proc_client_core_cap_resp(struct stream *s)
 
             case CAP_SMARTCARD_TYPE:
                 LOG_DEVEL(LOG_LEVEL_DEBUG, "got CAP_SMARTCARD_TYPE");
-                g_is_smartcard_redir_supported = 1;
-                scard_init();
+                g_is_smartcard_redir_supported = (scard_init() == 0);
                 break;
         }
         xstream_seek(s, cap_len);
--- a/sesman/chansrv/smartcard_pcsc.c
+++ b/sesman/chansrv/smartcard_pcsc.c
@@ -1853,6 +1853,8 @@ scard_pcsc_init(void)
     if (g_lis == 0)
     {
         g_lis = trans_create(2, 8192, 8192);
+        // TODO: See #2501. Use needs a way to move PCSCLITE_CSOCK_NAME
+        // to a location not under $HOME.
         home = g_getenv("HOME");
         disp = g_display_num;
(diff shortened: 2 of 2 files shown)


metalefty (approved):

matt335672:
For V0.9 see #2507

PR #2519, opened 2023-01-28
sound: fix noise between two playbacks for mp3/aac

Hiero32:
I built xrdp with --enable-mp3lame or --enable-fdkaac option, and playback on Microsoft RDP client mstsc.exe. 
Both mp3 and aac works.
However, an issue was found for both mp3 and aac.
Reproduction step is as follows:
1. playback an audio file using VLC.
2. stop playback after a couple of seconds.
3. playback the same audio file again. 

I hear "short sound" first then the audio file's sound at step 3. 
The "short sound" seems the chunk just after stop at step 2.  
I guess:
1. The chunk was sent the mstsc.exe, but not decoded at step 2.  
2. The chunk remains in the buffer of mstsc.exe.
3. Then it was decoded at step 3.

I thought, it might fix the issue to send silence data to mstsc.exe just before sound_send_close() to push out remaining chunk. 
And I changed g_bbuf_size for mp3.
1152*4 is samples per frame (1152 bytes) x 16bit(2 Bytes) x 2ch.  
This works fine.
I think the cause might be an issue of mstsc.exe.
This workaround is helpful to avoid this issue.

### DIFF
--- a/sesman/chansrv/sound.c
+++ b/sesman/chansrv/sound.c
@@ -1090,11 +1090,17 @@ static int
 process_pcm_message(int id, int size, struct stream *s)
 {
     static int sending_silence = 0;
+    static int silence_start_time = 0;
     switch (id)
     {
         case 0:
             if ((g_client_does_fdk_aac || g_client_does_mp3lame) && sending_silence)
             {
+                if ((g_time3() - silence_start_time) < 1000)
+                {
+                    /* do not send data within 1000mS after SNDC_CLOSE is sent. to avoid stutter */
+                    return 0;
+                }
                 sending_silence = 0;
             }
             return sound_send_wave_data(s->p, size);
@@ -1107,6 +1113,7 @@ process_pcm_message(int id, int size, struct stream *s)
                 char *buf = (char *) g_malloc(g_bbuf_size, 0);
                 if (buf != NULL)
                 {
+                    silence_start_time = g_time3();
                     sending_silence = 1;
                     for (int i = 0; i < send_silence_times; i++)
                     {
--- a/sesman/chansrv/sound.c
+++ b/sesman/chansrv/sound.c
@@ -1090,31 +1090,12 @@ static int
 process_pcm_message(int id, int size, struct stream *s)
 {
     static int sending_silence = 0;
-    static int discard_size = 0;
-    static int silence_start_time = 0;
     switch (id)
     {
         case 0:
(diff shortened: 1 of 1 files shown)


metalefty:
Thank you for your contribution. Could you format your code and squash commits? Run `scripts/run_astyle.sh`.

metalefty:
Also, could you address failing CI tests?

BTW, I think this is worth backporting to v0.9. Will cherry-pick to v0.9 after merge.

Hiero32:
I found an issue that sound tends to be stuttered for sound with video.
It might be caused by changing g_bbuf_size value.
I need more investigation,

metalefty:
I confirmed your changeset fixes the issue. Thank you very much for your contribution!

metalefty:
@Hiero32 Do you have any further changes for this?

Hiero32:
Sorry for waiting.
I fixed a couple of issues which I found during test.
- sometimes sound stuttered during playback
- sound stuttered on skip operation of vlc

There still exists sound stuttered a bit on skip operation of vlc.
However, it occurs without this patch.
I do not found other issue so far.
If necessary, I will squash.

metalefty:
Thanks, I'll test your additional patch soon. I'd appreciate if you squash commits then we cherry-pick just one commit when backporting your fix to older version.

metalefty:
Hmm, after some tests on my side, the first patch (with additional fix) occurs less stutter for me.

Hiero32:
Hmm,
I committed new patch.
This is close to first one.
Try this one.
Sorry to bother you.

Hiero32:
I think I found a condition which causes stutter.
If sound data is sent within 1000mS after SNDC_CLOSE is sent, stutter occurs.
This seems the characteristic of mstsc.exe..
Committed patch handles it.
One side effects is it takes 1000mS that sound comes out after skip operation.
I feel it is better than stuttered sound.
I hope this is the last patch.

metalefty:
I still have a doubt if 1000ms is a valid value (can be shortened?) but your fix gains significant improvement. Let's fix later if there's another glitch.

Hiero32:
I tested 600mS through 1000mS.
The result was at least 1000mS is necessary.
Anyway, this patch is a workaround for an issue of mstsc.exe.
Following text is described in 3.2.5.2.1.7 Processing a Close PDU of [MS-RDPEA].
- The client finishes any audio that arrived before this PDU and that remains to be rendered.

If mstsc.exe conformed to this spec, this patch was not necessary.

Issue #1723, opened 2020-11-13
Connected to GNOME desktop environment instead of ubuntu-gnome custom version


giuliotripi:
I installed xrdp on a fresh Ubuntu 20.04.1 install.
When I connect from xrdp I expect to land in the Ubuntu version of Gnome, like I do when I login form the physical PC.
I tried to add a `.xsession` file in the home of the user I was logging in to, with content 
```
gnome-session --session=ubuntu
```
but it didn't change anything.

That's what the desktop looks like when I log in from a remote PC
![image](https://user-images.githubusercontent.com/7399875/99128409-011b6380-260b-11eb-80aa-7b50a2a28197.png)
That's what the desktop looks like when I log in from the local PC
![image](https://user-images.githubusercontent.com/7399875/99128626-94ed2f80-260b-11eb-91b8-a52e01c9cb62.png)

drewer:
apply to ~/.xsessionrc

cho export GNOME_SHELL_SESSION_MODE=ubuntu
export XDG_CURRENT_DESKTOP=ubuntu:GNOME
export XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg

matt335672:
Thanks to @drewer for an excellent answer. I was looking for a bit more explanation as to what is happening here, and the above post put me very much on the right track.

I can't add much to @drewer's answer, except for the following two things.

1. The standard login scripts `/etc/profile.d/xdg_dirs_desktop_session.sh` and `/etc/X11/Xsession.d/60x11-common_xdg_path` use the `DESKTOP_SESSION` variable (if defined) to set both `XDG_CONFIG_DIRS` and `XDG_DATA_DIRS`. So rather than the set above, you should probably use:-

   ```
   export DESKTOP_SESSION=ubuntu
   export GNOME_SHELL_SESSION_MODE=ubuntu
   export XDG_CURRENT_DESKTOP=ubuntu:GNOME
   ```
2. if you want to do this on a global level, add the three lines above to the `pre_start()` function in `/etc/xrdp/startwm.sh`.

The reason that the xrdp default screen is different is simply that the GDM display manager works somewhat differently from the `/etc/X11/Xsession` script used by xrdp `startwm.sh`.

[...]

PR #2522, opened 2023-01-29
use startup command from /usr/share/xsession if DISPLAY_SESSION is set

akarl10:
gnome shell has a way to start different desktop profiles, but all use the same gnome-session binary.

also XDG somehow expects that the DESKTOP_SESSION environment variable is set. This value is derived from the .desktop file name in /usr/share/xsessions

There is also a XDG_CURRENT_DESKTOP that comes from that file

on debian/ubuntu setting STARTUP should work, not sure about other distros.

### DIFF
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -55,6 +55,20 @@ post_start()
   return 0
 }
 
+get_xdg_session_startupcmd()
+{
+  # DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
+  # If set and valid then the STARTUP command will be taken from there
+  if [ -n "$1" ] && [ -d /usr/share/xsessions ] \
+    && [ -f "/usr/share/xsessions/$1.desktop" ]; then
+    STARTUP=$(grep ^Exec= "/usr/share/xsessions/$1.desktop")
+    STARTUP=${STARTUP#Exec=*}
+    XDG_CURRENT_DESKTOP=$(grep ^DesktopNames= "/usr/share/xsessions/$1.desktop")
+    XDG_CURRENT_DESKTOP=${XDG_CURRENT_DESKTOP#DesktopNames=*}
+    echo "env XDG_CURRENT_DESKTOP=\"$XDG_CURRENT_DESKTOP\" $STARTUP"
+  fi
+}
+
 #start the window manager
 wm_start()
 {
@@ -66,6 +80,15 @@ wm_start()
   # debian
   if [ -r /etc/X11/Xsession ]; then
     pre_start
+    
+    # STARTUP is the default startup command.
+    # if $1 is empty and STARTUP was not set
+    # /etc/X11/Xsession.d/50x11-common_determine-startup will fallback to
+    # x-session-manager
+    if [ -z "$STARTUP" ] && [ -n "$DESKTOP_SESSION" ]; then
+      STARTUP="$(get_xdg_session_startupcmd "$DESKTOP_SESSION")"
+    fi
+    
     . /etc/X11/Xsession
     post_start
     exit 0


matt335672:
The intention of this (if I'm reading it right) is to give a system manager the chance to force user of a particular desktop. Is that right?

akarl10:
kind of right.
nowadays "login screen" managers like gdm or sddm take the launch command from from /usr/share/xsessions and dont use x-session-manager

if you look for example what gnome defines:
```
[Desktop Entry]
Name=Ubuntu on Xorg
Comment=This session logs you into Ubuntu
Exec=env GNOME_SHELL_SESSION_MODE=ubuntu /usr/bin/gnome-session --session=ubuntu
TryExec=/usr/bin/gnome-shell
Type=Application
DesktopNames=ubuntu:GNOME
X-GDM-SessionRegisters=true
X-Ubuntu-Gettext-Domain=gnome-session-43
```

you see that the launch command also contains parameters and in that case environment variables.
if you look at https://github.com/neutrinolabs/xrdp/issues/1723 you see that XDG_CURRENT_DESKTOP is initialized by DesktopNames.

sesman/startwm.sh
```diff
@@ -62,6 +62,17 @@ wm_start()
     . /etc/default/locale
     export LANG LANGUAGE
   fi
+  
+  # DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
+  # If set and valid then the STARTUP command will be taken from there
```
matt335672: Add a note that not all distros support this.

I can't find in in the CentOS 8 startup scripts for example.

matt335672 (commented):
One suggestion:-

sesman/startwm.sh
akarl10: true.. What is a bit strange to me is that on debian does a exec in Xsession so post_start will never be called.

do you know if thats also the case in other distros?

if all do a exec then it might be better to call the Xsession script with the fisrt parameter being $STARTUP

matt335672: I don't think there's any real commonality between the distros. They all seem to do their own thing.

Another thing you could do is wrap this up in a function and call it from the debian start stanza. We can then add it to any others that will benefit from it.

akarl10: ok, i did a rewrite. now the function gives the startup command without setting the environment just by calling the function.

This way a user could also set the preferred desktop by putting DESKTOP_SESSION in .profile. In my previous version that did not work

sesman/startwm.sh
```diff
@@ -55,17 +55,38 @@ post_start()
...
+{
+  # DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
+  # If set and valid then the STARTUP command will be taken from there
+  if [ -n "$1" ] && [ -d /usr/share/xsessions ] \
+    && [ -f "/usr/share/xsessions/$1.desktop" ]; then
+    STARTUP=$(grep ^Exec= "/usr/share/xsessions/$1.desktop")
+    STARTUP=${STARTUP#Exec=*}
+    XDG_CURRENT_DESKTOP=$(grep ^DesktopNames= "/usr/share/xsessions/$1.desktop")
+    XDG_CURRENT_DESKTOP=${XDG_CURRENT_DESKTOP#DesktopNames=*}
+    echo "env XDG_CURRENT_DESKTOP=$XDG_CURRENT_DESKTOP $STARTUP"
```
matt335672: Suggest `echo "env XDG_CURRENT_DESKTOP=\"$XDG_CURRENT_DESKTOP \" $STARTUP"`, to add "" to the XDG_CURRENT_DESKTOP

sesman/startwm.sh
```diff
@@ -55,17 +55,38 @@ post_start()
...
+  
   # debian
   if [ -r /etc/X11/Xsession ]; then
     pre_start
+    
+    # STARTUP is the default startup command.
+    # if $1 is empty and STARTUP was not set
+    # /etc/X11/Xsession.d/50x11-common_determine-startup will fallback to
+    # x-session-manager
+    STARTUP="$(get_xdg_session_startupcmd "$DESKTOP_SESSION")"
```
matt335672: Small point, but you can avoid a sub-process invocation for the common case (i.e. DESKTOP_SESSION not set) by testing DESKTOP_SESSION here. Also, it's possible (??) that STARTUP may be set by the user. How about:-

```sh
if [ -z "$STARTUP" ] && [ -n "$DESKTOP_SESSION" ]; then
    STARTUP="$(get_xdg_session_startupcmd "$DESKTOP_SESSION")"
fi
```

I'm making this up as I go along - please come back at me if you disagree.

matt335672 (commented):
Two minor comments

sesman/startwm.sh
akarl10: true

sesman/startwm.sh
akarl10: sounds reasonable. if we don't do that we may create a regression if one used the STARTUP "trick" in .profile

matt335672: Good point about the regression. That hadn't occurred to me.

matt335672:
LGTM.

Do you want to squash it into a single commit? Then I'll merge it.

PR #2523, opened 2023-01-30
Adding define guards to xrdp.h

Nexarian:
xrdp.h didn't have the standard define guards to prevent duplicate includes of a header. This broke the egfx prototype, so make sure we add this as it's important hygiene for the future.

### DIFF
--- a/xrdp/xrdp.h
+++ b/xrdp/xrdp.h
@@ -18,6 +18,9 @@
  * main include file
  */
 
+#ifndef _XRDP_XRDP_H_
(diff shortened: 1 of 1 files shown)


metalefty (approved):

matt335672:
Good catch @Nexarian.

I had a quick look through all the others, and I found these too:-

- mc/mc.h
- neutrinordp/xrdp-neutrinordp.h
- sesman/chansrv/fifo.h
- xup/xup.h

I'll put a PR together for them.

PR #2524, opened 2023-01-30
Adding utility functions for xrdp regions.

Nexarian:
These are utility functions that are necessary for some ancillary functions of EGFX. Given they're very straightforward and closed ended, thought I would add them as a standalone PR.

But I've taken this a step further now. I took the liberty of adding unit tests for these functions, which now gives an opportunity to demonstrate the use of cmocka mocking library!

This makes the changes more extensive, but I think it's worth the effort. Feedback on the approach I've taken with cmocka is appreciated!

### DIFF
--- a/tests/xrdp/test_xrdp_region.c
+++ b/tests/xrdp/test_xrdp_region.c
@@ -1,0 +1,164 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Jay Sorg, Christopher Pitstick 2004-2023
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Test driver for XRDP routines
+ *
+ * If you want to run this driver under valgrind to check for memory leaks,
+ * use the following command line:-
+ *
+ * CK_FORK=no valgrind --leak-check=full --show-leak-kinds=all \
+ *     .libs/test_xrdp
+ *
+ * without the 'CK_FORK=no', memory still allocated by the test driver will
+ * be logged
+ */
+
+#if defined(HAVE_CONFIG_H)
+#include "config_ac.h"
+#endif
+
+#include "log.h"
+#include "os_calls.h"
+#include <stdlib.h>
+
+#if defined(XRDP_PIXMAN)
+#include <pixman.h>
+#else
+#include "pixman-region.h"
+#endif
+
+#include "test_xrdp.h"
+#include "xrdp.h"
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <setjmp.h>
+#ifdef __cplusplus
+#error "cmocka is not supported with C++"
+#else
+#include <cmocka.h>
+#endif
+
+#define UNUSED(x) (void)(x)
+
+// Mock functions
+static pixman_box16_t test_box =
+{
+    0, 0, 100, 100
+};
+
+pixman_box16_t *__wrap_pixman_region_extents(pixman_region16_t *region)
+{
+    check_expected_ptr(region);
+    return mock_ptr_type(pixman_box16_t *);
+}
+
+static void test_xrdp_region_get_bounds__negligent_path(void **state)
+{
+    struct xrdp_region *region = g_new0(struct xrdp_region, 1);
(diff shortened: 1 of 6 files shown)


common/xrdp_sockets.h
```diff
@@ -26,8 +26,8 @@
 #define CHANSRV_PORT_OUT_BASE_STR  "xrdp_chansrv_audio_out_socket_%d"
 #define CHANSRV_PORT_IN_BASE_STR   "xrdp_chansrv_audio_in_socket_%d"
 #define CHANSRV_API_BASE_STR       "xrdpapi_%d"
-#define XRDP_X11RDP_BASE_STR       "xrdp_display_%d"
-#define XRDP_DISCONNECT_BASE_STR   "xrdp_disconnect_display_%d"
+#define XRDP_X11RDP_BASE_STR       "xrdp_disp_%d"
+#define XRDP_DISCONNECT_BASE_STR   "xrdp_disconnect_disp_%d"
 #define SCP_LISTEN_PORT_BASE_STR   "sesman.socket"
```
matt335672: Is there a good reason to change these?

If so fine, but bear in mind you'll also need to update https://github.com/neutrinolabs/xrdp/wiki/The-socketdir-directory where these are documented.

configure.ac
```diff
@@ -519,14 +519,22 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
...
+perform_unit_tests=yes; # Assume packages will be found
 if test "x$ensure_tests_deps" == "xyes"; then
     PKG_CHECK_MODULES([CHECK], [check >= 0.10.0],
-        [perform_unit_tests=yes],
+        [],
         [AC_MSG_ERROR([please install check, the unit test framework])])
+    # Check if cmocka is available - needed for unit testing
+    PKG_CHECK_MODULES([CMOCKA], [cmocka],
+        []
+        [AC_MSG_WARN([please install cmocka, the mocking framework])])
```
matt335672: Does this need to be an error, or am I misunderstanding? The user has specified they want test dependencies checked for.

tests/xrdp/test_xrdp_region.c
```diff
@@ -0,0 +1,171 @@
...
+#ifdef __cplusplus
+// cmocka.h lacks c++ guards for CI
+extern "C" {
+#include <cmocka.h>
+}
+#define C_LINKAGE extern "C"
+#else
+#include <cmocka.h>
+#define C_LINKAGE
+#endif
```
matt335672, line 54: See also our gitter conversation.

matt335672 (commented):
Hi @Nexarian

Well done on getting this together, I can see how it's become a bit more of a job than you intended!

A few comments below. See also our gitter chat.

configure.ac
Nexarian: Fixed in the other PR.

common/xrdp_sockets.h
Nexarian: I don't know how I got this, but I specifically remember messing with `make distcheck` and getting an error that there was a risk that this specific string could exceed the length of sa.sun_path on [this line](https://github.com/neutrinolabs/xrdp/blob/dd03a1ad10f4d4dd65c82d545aefa3d9d68005a6/sesman/tools/dis.c#L65), but I can no longer reproduce it, nor do I understand why I got this error in the first place. Ignore this, I won't check it in.

matt335672: Interesting - thanks for the reference.

According to [unix(7)](https://www.man7.org/linux/man-pages/man7/unix.7.html), `sa.sun_path` will be at least 92 bytes. POSIX isn't as definitive and says it is [typically between 92 and 108 bytes](https://pubs.opengroup.org/onlinepubs/009696699/basedefs/sys/un.h.html). Either should be fine for `XRDP_DISCONNECT_STR`

We should really be using `g_snprintf()` to bound this write, but given where it is I don't think it's a big concern.

matt335672:
Are you still going to stick with C++ unit testing? I've got some ideas on how we can remove it relatively simply.

matt335672:
Sorry - just read your gitter comment.

common/xrdp_sockets.h
Nexarian: Agreed. Resolving.

Nexarian:
More fixes need to be merged before I can merge this:
- https://github.com/neutrinolabs/xrdp/pull/2544
- https://github.com/neutrinolabs/xrdp/pull/2545

tests/xrdp/test_xrdp_region.c
```diff
@@ -0,0 +1,171 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <setjmp.h>
+#ifdef __cplusplus
+// cmocka.h lacks c++ guards for CI
+extern "C" {
+#include <cmocka.h>
+}
+#define C_LINKAGE extern "C"
+#else
```
matt335672, line 52: I'm pretty sure we should be removing all the C+ stuff. How about having an `#error "cmocka is not supported with C++"` if `__cplusplus` is defined? Then all the C_LINKAGE stuff can go too.

matt335672 (commented):
One comment - looking good.

tests/xrdp/test_xrdp_region.c
Nexarian: Updated.

PR #2526, opened 2023-01-31
Add missing include guards

matt335672:
Prompted by #2523 

I noted while I was doing this, that we have two fifo.h files:-
- [common/fifo.h](https://github.com/neutrinolabs/xrdp/blob/devel/common/fifo.h)
- [sesman/chansrv/fifo.h](https://github.com/neutrinolabs/xrdp/blob/devel/sesman/chansrv/fifo.h)

I have no idea why this is, but it may be related to performance. It looks like the first one will be consistent, and the second one will be very fast with occasional hiccups if the fifo needs to be resized.

Anyone get any comments on this?

### DIFF
--- a/mc/mc.h
+++ b/mc/mc.h
@@ -18,6 +18,9 @@
  * media center
  */
 
+#ifndef MC_H
+#define MC_H
+
 /* include other h files */
 #include "arch.h"
 #include "parse.h"
@@ -102,3 +105,5 @@ struct mod
     int height;
     int bpp;
 };
+
+#endif // MC_H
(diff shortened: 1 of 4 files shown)


matt335672:
Unrelated to this PR we also have two rail.h files!

Nexarian (approved):

Nexarian:
No idea on the fifos, but I also believe we should standardize. We should do the Big-O complexity calculations to determine which one has the better runtime complexity.

If I remember algorithms and data structures, we should go with the one that is fast in the average case and slow in the resize case because:
1. Resizing isn't actually as slow as we think it is (https://www.youtube.com/watch?v=YQs6IC-vgmo)
2. The amortized average case gets you more speed.

And you know I'm all about that speed :)

Nexarian:
I'm going to guess the two rails are due to the same reason we have a lot of duplication: Windows versus versus not-Windows version.

metalefty (approved):

PR #2528, opened 2023-01-31
Chansrv fuse display variable

akarl10:
as discussed briefly here https://github.com/neutrinolabs/xrdp/discussions/2525

there is a draft for that feature

### DIFF
--- a/sesman/chansrv/chansrv_fuse.c
+++ b/sesman/chansrv/chansrv_fuse.c
@@ -516,10 +516,25 @@ xfuse_init(void)
     {
         /* mount_name is relative to $HOME, e.g. ~/xrdp_client,
          * or ~/thinclient_drives */
-        g_snprintf(g_fuse_root_path, sizeof(g_fuse_root_path), "%s/%s",
-                   g_getenv("HOME"), g_cfg->fuse_mount_name);
-    }
+        int home_path_length = 0;
+        g_strncpy(g_fuse_root_path, g_getenv("HOME"), sizeof(g_fuse_root_path));
+        home_path_length = g_strlen(g_fuse_root_path);
 
+        if (home_path_length + 1 > sizeof(g_fuse_root_path))
+        {
+            LOG(LOG_LEVEL_ERROR,
+                "Fuse root path variable is not big enough to hold the full path");
+            return -1;
+        }
+        if (g_fuse_root_path[home_path_length] != '/')
+        {
+            g_fuse_root_path[home_path_length] = '/';
+            home_path_length ++;
+            g_fuse_root_path[home_path_length] = '\0';
+        }
+        format_user_info(g_fuse_root_path + home_path_length, sizeof(g_fuse_root_path) - home_path_length, g_cfg->fuse_mount_name);
+
+    }
     /* Remove all trailing '/' from the root path */
     p = g_fuse_root_path + g_strlen(g_fuse_root_path);
     while ( p > g_fuse_root_path && *(p - 1) == '/')
--- a/sesman/chansrv/chansrv_fuse.c
+++ b/sesman/chansrv/chansrv_fuse.c
@@ -516,25 +516,14 @@ xfuse_init(void)
     {
         /* mount_name is relative to $HOME, e.g. ~/xrdp_client,
          * or ~/thinclient_drives */
-        int home_path_length = 0;
-        g_strncpy(g_fuse_root_path, g_getenv("HOME"), sizeof(g_fuse_root_path));
-        home_path_length = g_strlen(g_fuse_root_path);
-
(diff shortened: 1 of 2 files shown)


sesman/chansrv/chansrv_fuse.c
```diff
@@ -516,8 +516,28 @@ xfuse_init(void)
     {
         /* mount_name is relative to $HOME, e.g. ~/xrdp_client,
          * or ~/thinclient_drives */
-        g_snprintf(g_fuse_root_path, sizeof(g_fuse_root_path), "%s/%s",
-                   g_getenv("HOME"), g_cfg->fuse_mount_name);
+        unsigned int home_path_length = 0;
+        unsigned int remaining_strlen = sizeof(g_fuse_root_path);
+
+        g_strncpy(g_fuse_root_path, g_getenv("HOME"), remaining_strlen);
```
matt335672: `g_strncpy()` always adds a terminating zero at dst[len] - it's a real gotcha. So this call will write one byte past the end of g_fuse_root_path! Personally I don't like this function. You probably need `remaining_strlen - 1` here.

Alternatively:-

```c
unsigned int len = g_snprintf(g_fuse_root_path, sizeof(g_fuse_root_path), "%s/", g_getenv("HOME"));
if (len < sizeof(g_fuse_root_path))
{
    format_user_info(g_fuse_root_path + len, sizeof(g_fuse_root_path) - len, g_cfg->fuse_mount_name);
}
```

sesman/chansrv/chansrv_fuse.c
```diff
@@ -2810,6 +2834,41 @@ static unsigned int format_user_info(char *dest, unsigned int len,
         g_strncpy(username, uidstr, sizeof(username) - 1);
     }
 
+    if (getenv("DISPLAY") == NULL)
+    {
+        /* if environment variable is not set, set it to empty string */
+        g_strncpy(display, "", 1);
+        g_strncpy(displaynum, "-1", 3);
+    }
```
matt335672, line 2827: just use :-
```c
display[0] = '\0';
displaynum[0] = '\0';
```

sesman/chansrv/chansrv_fuse.c
```diff
@@ -2810,6 +2834,41 @@ static unsigned int format_user_info(char *dest, unsigned int len,
...
+    else
+    {
+        const char *start = display;
+        const char *end = NULL;
+        unsigned int length = 0;
+        g_strncpy(display, getenv("DISPLAY"), sizeof(display) - 1);
+        while (*start != '\0' && *start != ':')
```
matt335672: Can also use `g_get_display_num_from_display()` in string_calls.h:-

https://github.com/neutrinolabs/xrdp/blob/b75a5ce76359ab2fbbd1608e1551d2a0733c4d1d/common/string_calls.h#L163-L171

None of this code needs to be highly optimised. It's only called once when chansrv starts up. Prefer readability over efficiency for this.

matt335672 (commented):
Thanks for this @akarl10. Comments below.

sesman/chansrv/chansrv_fuse.c
akarl10: ahh, totally forgot that \*printf returns the length written (again without \0 as you said)

that way it is way readable. What we miss here is if HOME has a terminating / we have //, but I think that was also the case before.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -2810,6 +2818,18 @@ static unsigned int format_user_info(char *dest, unsigned int len,
...
+        display[0] = '\0';
+        displaynum[0] = '\0';
+    }
+    else
+    {
+        g_strncpy(display, getenv("DISPLAY"), sizeof(display) - 1);
+        g_snprintf(displaynum, sizeof(displaynum), "%d", g_get_display_num_from_display(display));
+    }
+
     return g_format_info_string(dest, len, format, map);
```
matt335672, line 2835: That should be OK, but `g_get_display_num_from_display()` can return -1, in which case '%d' will also map to -1. This shouldn't happen, as to get here the call has to have succeeded in the past. I don't think there's a better thing to do here. Thoughts?

Also, we've got an 80-char limit in the [style guide](https://github.com/neutrinolabs/xrdp/wiki/Coding-Style). If you could wrap this line that would be great.

akarl10: don't know any solution to the -1 issue. But I think mapping to -1 should not do any harm since the constructed path would still be valid.

I also shortened the comment a few lines above

matt335672:
LGTM - I'll merge this.

Thanks @akarl10

PR #2534, opened 2023-02-04
Add core cmocka framework

Nexarian:
Initial integration with cmocka.

The intent is to first apply this to the xrdp unit tests, but until something uses it directly we'll hold off on updating the Makefile.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -519,14 +519,22 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
 
 CFLAGS="$save_CFLAGS"
 
-# perform unit tests if libcheck found
+# perform unit tests if libcheck and libmocka found
+perform_unit_tests=yes; # Assume packages will be found
 if test "x$ensure_tests_deps" == "xyes"; then
     PKG_CHECK_MODULES([CHECK], [check >= 0.10.0],
-        [perform_unit_tests=yes],
+        [],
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -519,14 +519,22 @@ AC_CHECK_HEADER([X11/extensions/Xrandr.h], [],
...
+perform_unit_tests=yes; # Assume packages will be found
 if test "x$ensure_tests_deps" == "xyes"; then
     PKG_CHECK_MODULES([CHECK], [check >= 0.10.0],
-        [perform_unit_tests=yes],
+        [],
         [AC_MSG_ERROR([please install check, the unit test framework])])
+    # Check if cmocka is available - needed for unit testing
+    PKG_CHECK_MODULES([CMOCKA], [cmocka],
+        []
+        [AC_MSG_WARN([please install cmocka, the mocking framework])])
```
matt335672: (Duplicate comment) Does this need to be an error, or am I misunderstanding? The user has specified they want test dependencies checked for.

Nexarian: I'm not sure I understand your concern. If cmocka isn't installed, and it was requested, it's asking that you install it?

Similar to the error message for check, right?

Nexarian: I figured out what you meant. I didn't realize it was originally a warning. I've updated it to be an error. Hopefully we're good to go :)

matt335672 (approved):
LGTM!

PR #2535, opened 2023-02-05
Acceptable use policy optional addition.

Nexarian:
WIP, don't review this yet!

### DIFF
--- a/xrdp/xrdp_login_wnd.c
+++ b/xrdp/xrdp_login_wnd.c
@@ -26,6 +26,7 @@
 #include "xrdp.h"
 #include "log.h"
 #include "string_calls.h"
+#include <string.h>
 
 #define ASK "ask"
 #define ASK_LEN g_strlen(ASK)
@@ -124,7 +125,7 @@ xrdp_wm_delete_all_children(struct xrdp_wm *self)
     struct xrdp_bitmap *b;
     struct xrdp_rect rect;
 
-    for (index = self->screen->child_list->count - 1; index >= 0; index--)
+    for (index = self->screen->child_list->count - 1; index >= 0; --index)
     {
         b = (struct xrdp_bitmap *)list_get_item(self->screen->child_list, index);
         MAKERECT(rect, b->left, b->top, b->width, b->height);
@@ -506,6 +507,8 @@ xrdp_wm_show_edits(struct xrdp_wm *self, struct xrdp_bitmap *combo)
                 if ((g_strncasecmp(name, "password", 255) == 0) ||
                         (g_strncasecmp(name, "pampassword", 255) == 0))
                 {
+                    g_strncpy(b->caption1, self->session->client_info->password, 255);
+                    b->edit_pos = g_mbstowcs(0, b->caption1, 0);
                     b->password_char = '*';
 
                     if (username_set)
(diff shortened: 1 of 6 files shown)


Nexarian:
I recognize the changes to the xrdp.ini.in file can't be checked in (probably have to be checked in as an example template and/or with the documentation), but otherwise, this is ready to review.

The most important workflow change here is that it will now populate the password box with the password that's sent over the wire from the client, and the user could change it.

What do you all think?
Example AUP login screen

xrdp/xrdp_login_wnd.c
```diff
@@ -742,6 +745,17 @@ xrdp_login_wnd_get_monitor_dpi(struct xrdp_wm *self)
...
+split_string_by_newline(const char *str, struct list *lines)
+{
+    char *copy = strdup(str);
+    char *pch = strtok(copy, "\n");
+    while (pch != NULL)
+    {
+        list_add_item(lines, (tintptr) pch);
+        pch = strtok(NULL, "\n");
+    }
+}
```
matt335672, line 759: Memory problems here I think. You're allocating memory with a single call to `strdup()`, and then expecting `auto_free` to fix it for you on `list_delete()`. However, you'll be calling free on addresses mid-way through the allocated block.

How about (in string_calls.c or list.c) :-

```c
static const char *
append_fragment(const char *start, const char *end, struct list *list)
{
    unsigned int len = end - start;
    char *copy = malloc(len + 1);
    g_memcpy(copy, start, len);
    copy[len] = '\0';
    list_add_item(list, (tintptr)copy);
    return end + 1;
}

struct list *
split_string_by_newline(const char *str)
{
    struct list *result = list_create();
    result->auto_free = 1;

    if (str != NULL)
    {
        const char *p;

        while ((p = g_strchr(str, '\n')) != NULL)
        {
            str = append_fragment(str, p, result);
        }

        if (*str != '\0')
        {
            append_fragment(str, str + g_strlen(str), result);
        }
    }
    
    return result;
}
```

It's got a slightly different signature admittedly, but all memory is freed on `list_delete()`

matt335672: Also, it behaves differently from `strtok()` for contibuous delimiters.

xrdp/xrdp_login_wnd.c
```diff
@@ -951,6 +964,51 @@ xrdp_login_wnd_create(struct xrdp_wm *self)
...
+    {
+        char *file_path = globals->ls_acceptable_use_policy_filename;
+        int file_size = g_file_get_size(file_path);
+        if (file_size < 1)
+        {
+            LOG(LOG_LEVEL_ERROR, "xrdp_login_wnd_create: error reading data from file [%s]",
+                file_path);
+            return 0;
+        }
+        char *str = (char *)g_malloc(sizeof(char) * (file_size + 1024), 1);
```
matt335672, line 978: Check malloc result?

matt335672 (commented):
A few comments for you

xrdp/xrdp_login_wnd.c
Nexarian: This is what happens when you've spent too much of your career in languages with garbage collectors and then go code in an unmanaged memory language when you're tired. You're right of course!

I think I'll split this part up into smaller PRs as always to make them easier to review.

PR #2536, opened 2023-02-08
Add memory allocation checking to the list module

matt335672:
The list module lacks memory allocation checking, and consequently can coredump on `list_create()` or `list_add_item()`.

This PR adds checks to these calls to prevent this.  `list_add_item()` now returns a boolean which can be used to check the
addition actually happened. This is compatible with existing calls which currently just assume the list has to succeed.

An additional call `list_create_sized()` allows a list of at least a certain size to be created. This can avoid memory allocations while the list is being added to, and is also more efficient for larger lists.

### DIFF
--- a/common/list.c
+++ b/common/list.c
@@ -22,23 +22,53 @@
 #include <config_ac.h>
 #endif
 
+#include <stdlib.h>
+
 #include "arch.h"
 #include "os_calls.h"
 #include "string_calls.h"
 #include "list.h"
 #include "log.h"
 
+enum
+{
+    DEFAULT_LIST_SIZE = 10,
+    DEFAULT_GROW_BY_SIZE = 10
+};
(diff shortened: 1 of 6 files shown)


metalefty:
That's nice. I think this should be backported to v0.9.

matt335672:
Added some really simple tests for the list module to the test suite. Good enough for now?

If so, I'll merge and backport to 0.9

metalefty (approved):
LGTM

matt335672:
Just found a couple more possible memory leaks in `list_insert_item()` and `list_append_list_strdup()`. These have been fixed in the same way, and test-cases have been added.

I've also made use of standard-library `realloc()` rather than `malloc()` followed by `memcpy()` as this is potentially more efficient.

PR #2544, opened 2023-02-11
Add unit test parameter for more coverage.

Nexarian:
This adds a unit test parameter. The rules are as follows:
- Max features are distchecked
- All jobs run unit tests
- Unless they are g++ or min, then both are disabled.

Needed to update the Ubuntu OS for 32 bit builds and add more dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,34 +21,43 @@ jobs:
             feature_set: min
             arch: amd64
             os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


matt335672 (approved):
Looks fine to me, but we should probably standardise the case between inittests and DISTCHECK. I don't really mind which!

Nexarian:
I'll check this in now to get the ball rolling, then we can iterate.

PR #2545, opened 2023-02-11
Add free to egfx unit tests

Nexarian:
Missed a free on this unit test. Doesn't directly affect performance or memory leakage, but it's good hygiene to clean this up.


matt335672 (approved):
Definitely agree with this. You can run the test exes with valgrind using something like:-

- `CK_FORK=no valgrind tests/common/.libs/test_common` for common
- `CK_FORK=no valgrind tests/xrdp/.libs/test_xrdp` for XRDP.

The `CK_FORK=no` stops the check library from forking which makes it easier to track the memory.

I note that I've got some clearing up of my own to do here for the bitmap stuff! If you try the above command on the xrdp tests you'll see what I mean.

PR #2549, opened 2023-02-13
Add support for RFB protocol version 3.7 and 3.8

Jesse-Bakker:


vnc/vnc.c
```diff
@@ -1710,8 +1715,27 @@ lib_mod_connect(struct vnc *v)
         error = trans_force_read_s(v->trans, s, 12);
         if (error == 0)
         {
-            s->p = s->data;
-            out_uint8a(s, "RFB 003.003\n", 12);
+            in_uint8p(s, version_str, 12);
+            if (g_strncmp(version_str, "RFB 003.00", 10) != 0)
```
matt335672, line 1720: Need to check `\n` in `version_str[11]` as well.

https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#protocolversion

I appreciate this makes the logging a bit trickier.

vnc/vnc.c
```diff
@@ -1720,20 +1744,78 @@ lib_mod_connect(struct vnc *v)
...
+                        error = trans_force_read_s(v->trans, s, n_sec_lvls);
+                    }
+                    else
+                    {
+                        // Read size of reason
+                        error = trans_force_read_s(v->trans, s, 4);
+                        if (error == 0)
+                        {
+                            in_uint32_be(s, i);
+                            error = trans_force_read_s(v->trans, s, i);
```
matt335672, line 1796: If the server supplies a large value here, bad things can happen as the buffer in 's' is only sized for 8192 bytes. `struct trans` is used in a slightly odd way by the VNC module, so may be best to check the input is less than a large value like 1024, and fail if it's too big.

matt335672:
@Jesse-Bakker - thanks for this.

It's been a busy week here, I'll get a proper review to you probably early next week now.

vnc/vnc.c
```diff
@@ -1720,20 +1744,78 @@ lib_mod_connect(struct vnc *v)
...
+                        in_uint8(s, j);
+                        // Choose the highest security level that we support
+                        if (j > sec_lvl && j <= 2)
+                        {
+                            sec_lvl = j;
+                        }
+                    }
+                    init_stream(s, 8192);
+                    out_uint8(s, sec_lvl);
+                    s_mark_end(s);
```
matt335672: The [spec](/rfbproto/rfbproto/blob/master/rfbproto.rst#security) is a little unclear on this:-

> If the server listed at least one valid security type supported by the client, the client sends back a single byte indicating which security type is to be used on the connection:

I read that as we shouldn't send back a zero here if we don't get a value we like.

vnc/vnc.c
```diff
@@ -1720,20 +1744,78 @@ lib_mod_connect(struct vnc *v)
...
+            g_sprintf(text, "VNC security level is %d (1 = none, 2 = standard)", sec_lvl);
             v->server_msg(v, text, 0);
 
-            if (i == 1) /* none */
+            if (sec_lvl == 1) /* none */
             {
-                check_sec_result = 0;
+                if (version >= 8)
+                {
+                    check_sec_result = 1;
```
matt335672: Note that in this case (version 3.8), if security checking fails, the server should also send a reason string. At the moment you aren't logging this (round about line 1881).

vnc/vnc.c
```diff
@@ -1802,8 +1884,7 @@ lib_mod_connect(struct vnc *v)
     {
         v->server_msg(v, "VNC sending share flag", 0);
         init_stream(s, 8192);
-        s->data[0] = 1;
-        s->p++;
+        out_uint8(s, 1);
         s_mark_end(s);
```
matt335672, line 1944: Good spot - thanks.

matt335672 (commented):
Thanks @Jesse-Bakker

There's a few comments below. Get back to me on anything I haven't explained well, or you disagree with.

common/parse.h
```diff
@@ -440,6 +440,15 @@ parser_stream_overflow_check(const struct stream *s, int n, int is_out,
...
+/******************************************************************************/
+#define in_str(s, v, n) do \
+    { \
+        S_CHECK_REM((s), (n + 1)); \
+        (v) = (s)->p; \
+        (v[n]) = '\0'; \
+        (s)->p++; \
+    } while (0)
+
 /******************************************************************************/
```
matt335672, line 452: This isn't copying the string from the input buffer.

when the macro runs:-

- `(v) = (s)->p;` sets `err_reason` to be `(s)->p` (which points into `(s)->data`)
- `(v[n]) = '\0';` writes a null into the input buffer
- `(s)->p++` advances one character along the buffer. This isn't useful, but you're not seeing any bad effects as the macro is only used at the end of the input.

I don't think it's good form to overwrite the input buffer in this way - you should be copying the string. Here's a possibility (not tested):-

```c
#define in_str(s, n, v, vlen) do \
    { \
        int copylen = MIN((n), (vlen - 1)); \
        in_uint8a((s), (v), copylen); \
        (v)[copylen] = '\0'; \
        if ((n) > copylen) \
        { \
            in_uint8s((s), (n) - copylen); \
        } \
    } while (0)
```

Use with :-

```c
char err_reason[128];
. . .
in_str(s, i, err_reason, sizeof(err_reason));
```

It's a pretty complex macro however.

vnc/vnc.c
```diff
@@ -1715,13 +1716,18 @@ lib_mod_connect(struct vnc *v)
...
         {
-            in_uint8p(s, version_str, 12);
+            in_uint8a(s, version_str, 12);
             if (g_strncmp(version_str, "RFB 003.00", 10) != 0)
             {
                 version_str[11] = '\0';
                 LOG(LOG_LEVEL_ERROR, "Invalid server version string %s", version_str);
                 error = 1;
             }
+            if (version_str[11] != '\n')
```
matt335672, line 1726: Need an `else if` here.

If the previous test succeeds (invalid version string), the newline will be overwritten with a NUL for error reporting, and then this test will trigger too!

vnc/vnc.c
```diff
@@ -1771,10 +1777,28 @@ lib_mod_connect(struct vnc *v)
                         if (error == 0)
                         {
                             in_uint32_be(s, i);
+                            if (i <= 1024)
+                            {
+                                error = trans_force_read_s(v->trans, s, i);
+                            }
```
matt335672, line 1783: Logic problem here. If this line returns 0, the string is read again at line 1796 below.

Also, `i` is `int`rather than `unsigned int`, so the test against 1024 is not sufficient.

Suggest the following, if `in_str()` is redefined:-

```c
char err_str[128];  // replaces char *err_str;
. . .
                // The client chooses the security type
                error = trans_force_read_s(v->trans, s, 1);
                if (error == 0)
                {
                    in_uint8(s, n_sec_lvls);
                    if (n_sec_lvls > 0)
                    {
                        error = trans_force_read_s(v->trans, s, n_sec_lvls);
                    }
                    else
                    {
                        // Server has closed the connection
                        error = 1;
                        g_snprintf(err_reason, sizeof(err_reason),
                                  "<reason unavailable>");
                        // Read size of reason
                        if (trans_force_read_s(v->trans, s, 4) == 0)
                        {
                            in_uint32_be(s, i);
                            if (i > 0 && i < sizeof(err_reason) &&
                                trans_force_read_s(v->trans, s, i) == 0)
                            {
                                in_str(s, i, err_reason, sizeof(err_reason));
                            }
                        }
                        LOG(LOG_LEVEL_ERROR, "Connection closed by server : %s",
                            err_reason);
                    }
                }
```

NB Lines need reformatting to keep width to 80 chars (see https://github.com/neutrinolabs/xrdp/wiki/Coding-Style)

vnc/vnc.c
```diff
@@ -1870,8 +1897,37 @@ lib_mod_connect(struct vnc *v)
...
+                    }
+                    else
+                    {
+                        i = 0;
+                    }
+                    if (error == 0)
+                    {
+                        in_str(s, err_reason, i);
+                        v->server_msg(v, err_reason, 0);
+                    }
```
matt335672, line 1924: If we get to this point, the server has disconnected, and yet error is still 0 (it should probably be 2).

Also, `i` can be negative again.

Suggest something like:-

```c
            if (i != 0)
            {
                if (version < 8)
                {
                    g_snprintf(text, sizeof(text), "VNC password failed");
                }
                else
                {
                    g_snprintf(err_reason, sizeof(err_reason),
                               "<reason unavailable>");
                    if (trans_force_read_s(v->trans, s, 4) == 0)
                    {
                        in_uint32_be(s, i);
                        if (i > 0 && i < sizeof(err_reason) &&
                                trans_force_read_s(v->trans, s, i) == 0)
                        {
                            in_str(s, i, err_reason, sizeof(err_reason));
                        }
                    }
                    g_snprintf(text, sizeof(text), "VNC password failed : %s",
                               err_reason);
                }

                v->server_msg(v, text, 0);
                error = 2;
            }
            else
            {
                v->server_msg(v, "VNC password ok", 0);
            }
```

matt335672 (changes requested):
Thanks for that @Jesse-Bakker

A few comments below.

matt335672:
PS : Don't worry about the code check fails - we're looking at it. See #2567

matt335672:
Superceded by #3460

PR #2551, opened 2023-02-14
Make pam.d directory configurable

yifanjiang:
@matt335672 This is some my thought according to the discussion of https://github.com/neutrinolabs/xrdp/pull/2547 . And the change worked in my environment, does it make sense to you? Thanks.


yifanjiang:
Sorry for the build error, I am going to double check.

PR #2552, opened 2023-02-14
Make pam.d directory configurable

yifanjiang:
Change according to the discussion in https://github.com/neutrinolabs/xrdp/pull/2547 .

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -558,6 +558,12 @@ if test "x$enable_strict_locations" != "xyes"; then
   localstatedir="/var";
 fi
 
+AC_ARG_WITH([pamconfdir],
+  [AS_HELP_STRING([--with-pamconfdir=DIR],
+                  [Use directory for pam.d config (default: /etc/pam.d)])],
+                  [], [with_pamconfdir="$sysconfdir/pam.d"])
+AC_SUBST([pamconfdir], [$with_pamconfdir])
+
 PKG_INSTALLDIR
 
 AC_CHECK_HEADERS([sys/prctl.h])
@@ -638,6 +644,7 @@ echo "  exec_prefix             $exec_prefix"
 echo "  libdir                  $libdir"
 echo "  bindir                  $bindir"
 echo "  sysconfdir              $sysconfdir"
+echo "  pamconfdir              $pamconfdir"
 echo "  localstatedir           $localstatedir"
 echo "  runstatedir             $runstatedir"
 echo "  socketdir               $socketdir"
--- a/sesman/verify_user_pam.c
+++ b/sesman/verify_user_pam.c
@@ -204,7 +204,15 @@ get_service_name(char *service_name)
     service_name[0] = 0;
 
     if (g_file_exist("/etc/pam.d/xrdp-sesman") ||
-            g_file_exist(XRDP_SYSCONF_PATH "/pam.d/xrdp-sesman"))
+#ifdef __LINUX_PAM__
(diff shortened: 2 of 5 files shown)


configure.ac
```diff
@@ -558,6 +558,12 @@ if test "x$enable_strict_locations" != "xyes"; then
   localstatedir="/var";
 fi
 
+AC_ARG_WITH([pamconfdir],
+  [AS_HELP_STRING([--with-pamconfdir=DIR],
+                  [Use directory for pam.d config (default: /etc)])],
```
metalefty: I think the default should location be `$sysconfdir/etc` rather than `/etc` when pamconfdir is not specified. Otherwise, this change will change current behaviour.

metalefty: Example: https://cgit.freedesktop.org/systemd/systemd/commit/?id=5c390a4ae0d383b2003074ed011d47876c7e630c

matt335672:
Hi @yifanjiang 

I'm out of time this week to have a good look at this, but after a quick glance this indeed looks like I was thinking of. This will definitely be useful to SuSE and any other distros that don't use `/etc/pam.d`.

A quick question- is there a good reason why you didn't add `/pam.d` to the pam config dir?

Speak soon.

yifanjiang:
> A quick question- is there a good reason why you didn't add `/pam.d` to the pam config dir?

Hi @matt335672 , thanks for looking into this. The reason was to make it a little more flexible for downstream packaging. While I took a second look, keeping a consistent lexical style with socketdir and sesmanruntimedir might be a better choice. So I update through the changes.

metalefty (approved):
I'm fine with this now. Waiting for @matt335672 's approval.

sesman/verify_user_pam.c
```diff
@@ -204,7 +204,7 @@ get_service_name(char *service_name)
     service_name[0] = 0;
 
     if (g_file_exist("/etc/pam.d/xrdp-sesman") ||
-            g_file_exist(XRDP_SYSCONF_PATH "/pam.d/xrdp-sesman"))
```
matt335672, line 207: Before line 207, can you add:--

```c
#ifdef __LINUX_PAM__
            /* /usr/lib/pam.d is hardcoded into Linux-PAM */
            g_file_exist("/usr/lib/pam.d/xrdp-sesman")) ||
#endif
#ifdef OPENPAM_VERSION
            /* /usr/local/etc/pam.d is hardcoded into OpenPAM */
            g_file_exist("/usr/local/etc/pam.d/xrdp-sesman")) ||
#endif
```

matt335672 (commented):
I was all ready to approve this, then I read this [useful comment](https://github.com/neutrinolabs/xrdp/pull/2547#issuecomment-1430897118) from @thkukuk

I suggest the following amendment.

@yifanjiang - I apologise for messing you about like this. However, what we've now got something that should be able to support the use-cases mentioned by @thkuk

sesman/verify_user_pam.c
yifanjiang: @matt335672 That's completely fine, I am happy there could be a better connection between xrdp and pam. Change already committed.

matt335672:
Thanks for the quick edits @yifanjiang 

Again, I apologise for the confusion.

matt335672:
@metalefty - I think this should be backported to v0.9. Do you agree?

metalefty:
@matt335672 Yes, it is definitely useful.

PR #2553, opened 2023-02-15
Add function to split string into list.

Nexarian:


common/list.c
```diff
@@ -298,3 +297,40 @@ list_dump_items(struct list *self)
...
+    {
+        str = append_fragment(str, p, result);
+    }
+
+    if (*str != '\0')
+    {
+        append_fragment(str, str + g_strlen(str), result);
+    }
+    return result;
+}
```
matt335672, line 367: Sorry to be a bit of a pain, but after #2536 was merged last week we can check for allocation failures, so we probably ought to for new functions like this.

Here's a modified version. Feel free to hack it about as appropriate:-

```c
/**
 * Appends a string fragment to a list
 * @param[in,out] start Pointer to start of fragment (by reference)
 * @param end Pointer to one past end of fragment
 * @param list List to append to
 * @result 1 for success
 *
 * In the event of a memory failure, 0 is returned and the list is deleted
 */
static int
append_fragment(const char **start, const char *end, struct list *list)
{
    const unsigned int len = end - *start;
    char *copy = (char *)malloc(len + 1);
    if (copy == NULL)
    {
        list_delete(list);
        return 0;
    }
    g_memcpy(copy, *start, len);
    copy[len] = '\0';
    if (!list_add_item(list, (tintptr)copy))
    {
        g_free(copy);
        list_delete(list);
        return 0;
    }
    *start = end + 1;
    return 1;
}

struct list *
split_string_into_list(const char *str, char character)
{
    struct list *result = list_create();
    if (result == NULL)
    {
        return result;
    }
    result->auto_free = 1;

    if (str == NULL)
    {
        return result;
    }

    const char *p;
    while ((p = g_strchr(str, character)) != NULL)
    {
        if (!append_fragment(&str, p, result))
        {
            return NULL;
        }
    }

    if (*str != '\0')
    {
        if (!append_fragment(&str, str + g_strlen(str), result))
        {
            return NULL;
        }
    }
    return result;
}
```

common/list.h
```diff
@@ -94,5 +94,7 @@ int
 list_append_list_strdup(struct list *self, struct list *dest, int start_index);
 void
 list_dump_items(struct list *self);
+struct list *
+split_string_into_list(const char *str, char character);
```
matt335672, line 115: Can do with a header comment saying NULL is returned on allocation error, like some of the other functions here (if you implement the above).

matt335672 (commented):
Couple of comments.

common/list.c
Nexarian: Not a pain at all! Working on updating

common/list.h
Nexarian: Updated!

.github/workflows/build.yml
```diff
@@ -133,6 +133,7 @@ jobs:
           echo "CFLAGS=$CFLAGS_${{ matrix.arch }}" >> $GITHUB_ENV
           echo "LDFLAGS=$LDFLAGS_${{ matrix.arch }}" >> $GITHUB_ENV
       - uses: actions/checkout@v3
+      - uses: ros-tooling/setup-ros@v0.6
```
matt335672: This looks interesting, but I'm not clear on what exactly it's doing(?)

common/list.c
```diff
@@ -298,3 +297,70 @@ list_dump_items(struct list *self)
...
+ * Appends a string fragment to a list
+ * @param[in,out] start Pointer to start of fragment (by reference)
+ * @param end Pointer to one past end of fragment
+ * @param list List to append to
+ * @result 1 for success
+ *
+ * In the event of a memory failure, 0 is returned and the list is deleted.
+ */
+int
+append_fragment(const char **start, const char *end, struct list *list)
```
matt335672: Great idea to make this non-static so we can test it, as you have below. In this case however it's still a private function, it might be worth changing the name to something a bit less generic, like `split_string_append_fragment()`. What do you think?

tests/common/test_list_calls.c
```diff
@@ -112,8 +112,46 @@ START_TEST(test_list__simple_append_list)
...
+    ck_assert_str_eq((const char *)l->items[3], "sat");
+    ck_assert_str_eq((const char *)l->items[4], "on");
+    ck_assert_str_eq((const char *)l->items[5], "my");
+    ck_assert_str_eq((const char *)l->items[6], "hat.");
+
+    list_delete(l);
+}
+END_TEST
+
+/******************************************************************************/
```
matt335672, line 156: Nice example!

.github/workflows/build.yml
matt335672: Scrub the above. I see you're looking at the build issues. I'm still haven't fixed my email notifications on gitter.

matt335672 (approved):

Issue #2491, opened 2023-01-06
[RFE] enforce single graphical session on systemd platforms


akarl10:
since https://github.com/systemd/systemd/issues/24932 got rejected by @poettering himself I think the only longterm solution is that every display session manager implements a safeguard that prevents "launching" multiple graphical sessions and display a message to the user that he must log off the offending session (maybe also providing a option to force terminate the offending session)

Multiple graphical sessions per user are simply not possible with systemd. This is by design
if this is not intentional then the broken by design (`graphical-session.target` being a user unit, in addition the assumption that the DBUS_SESSION_BUS_ADDRESS is session exclusive while systemd make this the same for every user session), but still it is not something we can fix

If someone has a better idea how to treat this issue, just propose that, but I think something has to be done given the amount of bug reports that pop up that the display keeps black

some findings:
```
loginctl show-user $UID
```
displays the active sessions. Locally created graphical sessions fill the `DISPLAY` property (xrdp does not do that, but maybe that is only possible if the session is local)

A similar thing should be done in sddm/gdm/... (out of scope for this report)

matt335672:
I agree. I think your systemd RFE illustrates that this isn't going to change for systemd-based systems.

The difficulty is where to put something like this. We can easily add something to `startwm.sh`which implements what you suggest above. That solves the console login followed by xrdp login, but not the other way round.

A script in `/etc/X11/xinit/xinitrc.d` or `/usr/etc/X11/xinit/xinitrc.d` might be adequate for many systems. I'm welcome to suggestions, personally.

akarl10:
here a script that should work.

I put it in /etc/X11/Xsession.d/10enforce-single-graphical-session

but of course that is debian/ubuntu specific. Also this only works for X11 session. Xrdp should detect pre-existing wayland sessions, but the other way around probably does not work.

Good news: it seems all you need to do is stop `graphical-session.target` if it is running

```bash
if systemctl --quiet --user is-active graphical-session.target; then
  xmessage -button Logout:1,Terminate\ running\ graphical\ session:2 "Only a single graphical session is allowed.
If you terminate the current graphical session all unsaved data in that session will be lost"
  action=$?
  if [ $action -eq 2 ]; then
    systemctl --user stop graphical-session.target
    sleep 1
  else
    exit 0
  fi
fi
```

matt335672:
That's a great contribution - thanks. I like the use of `graphical-session.target`.

[...]

PR #2557, opened 2023-02-18
improve "#2522 use startup command from /usr/share/xsession ..."

Hiero32:
It is good way to set STARTUP in startwm.sh.
Many desktop environments do not start wthout it on debian/ubuntu.
Users need to write lines in .xsessionrc/.xsession., however it is difficult to set environment variables correctly.

This fixes an issue in #2522 and adds some improvements.
- Escape of double quote in following line does not work as expected.
```
    echo "env XDG_CURRENT_DESKTOP=\"$XDG_CURRENT_DESKTOP\" $STARTUP"
```
- replace ';' to ':',  if ';' exists in XDG_CURRENT_DESKTOP. 
This is a workaround for some desktop environments. e.g. gnome-flashback-metacity.
- add to source the file "/etc/xrdp/export_desktop_session", if exists.
#2522 expects DESKTOP_SESSION is set in sesman.ini in the SessionVariables section.
I think it is a bit inconvenient for users to modify sesman.ini.
I added to source the file /etc/xrdp/export_desktop_session, if exists.
Simplest content in /etc/xrdp/export_desktop_session is to just one line to export DESKTOP_SESSION environment variable as follows.
```
export DESKTOP_SESSION=ubuntu
```
Because /etc/xrdp/export_desktop_session is shell script, more functionality can be added.
For example, if multiple users use the server and only userA wants to use gnome-flashback-metacity, and others want to use ubuntu, following /etc/xrdp/export_desktop_session works.
```
if [ "$USER" = "userA" ]; then
  export DESKTOP_SESSION=gnome-flashback-metacity
else
  export DESKTOP_SESSION=ubuntu
fi
```

### DIFF
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -85,13 +85,14 @@ wm_start()
   if [ -r /etc/X11/Xsession ]; then
     pre_start
 
-    # if /etc/xrdp/export_desktop_session file exists, souurce it.
-    # /etc/xrdp/export_desktop_session script exports DESKTOP_SESSION environment varible
-    # The value shall be one of "ls -1 /usr/share/xsessions|cut -d. -f1".
-    # e.g. export DESKTOP_SESSION=ubuntu
-    if [ -r /etc/xrdp/export_desktop_session ]; then
-      . /etc/xrdp/export_desktop_session
-    fi
+    # if you want to start preferred desktop environment,
+    # add following line,
+    #  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=<your preferred desktop>
+    # in either of following file.
+    # 1. ~/.profile
+    # 2. create a file (any filename is OK) in /etc/profile.d
+    # <your preferred desktop> shall be one of "ls -1 /usr/share/xsessions/"
+    # e.g.  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION==ubuntu
 
     # STARTUP is the default startup command.
     # if $1 is empty and STARTUP was not set
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -57,8 +57,7 @@ post_start()
 
 get_xdg_session_startupcmd()
 {
-  # DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
-  # If set and valid then the STARTUP command will be taken from there
+  # If DESKTOP_SESSION is set and valid then the STARTUP command will be taken from there
   # GDM exports environment variables XDG_CURRENT_DESKTOP and XDG_SESSION_DESKTOP.
   # This follows it.
   if [ -n "$1" ] && [ -d /usr/share/xsessions ] \
@@ -90,7 +89,7 @@ wm_start()
     #  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=<your preferred desktop>
     # in either of following file.
     # 1. ~/.profile
-    # 2. create a file (any filename is OK) in /etc/profile.d
+    # 2. create a file (any_filename.sh is OK) in /etc/profile.d
     # <your preferred desktop> shall be one of "ls -1 /usr/share/xsessions/|cut -d. -f1"
     # e.g.  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=ubuntu
 
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -91,8 +91,8 @@ wm_start()
     # in either of following file.
     # 1. ~/.profile
     # 2. create a file (any filename is OK) in /etc/profile.d
-    # <your preferred desktop> shall be one of "ls -1 /usr/share/xsessions/"
-    # e.g.  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION==ubuntu
+    # <your preferred desktop> shall be one of "ls -1 /usr/share/xsessions/|cut -d. -f1"
+    # e.g.  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=ubuntu
 
     # STARTUP is the default startup command.
     # if $1 is empty and STARTUP was not set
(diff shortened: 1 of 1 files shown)


matt335672:
Personally I'm not keen on the /etc/xrdp/export_desktop_session. This can easily be added into `/etc/profile.d` by the system manager, or even `.profile` for the user (which works around your user test above).

The rest of it looks good to me though.

What do you think @akarl10 ?

akarl10:
looks good.
agree about `export_desktop_session`.
What I recommended in my ppa instruction is putting something like
```bash
[ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=<your preferred desktop>
```
in ~/.profile, but of course /etc/profile.d is a good place too

sesman/startwm.sh
```diff
@@ -80,15 +84,23 @@ wm_start()
...
-    
+
+    # if /etc/xrdp/export_desktop_session file exists, souurce it.
+    # /etc/xrdp/export_desktop_session script exports DESKTOP_SESSION environment varible
+    # The value shall be one of "ls -1 /usr/share/xsessions|cut -d. -f1".
+    # e.g. export DESKTOP_SESSION=ubuntu
+    if [ -r /etc/xrdp/export_desktop_session ]; then
+      . /etc/xrdp/export_desktop_session
+    fi
+
```
matt335672, line 95: As discussed in the PR, please remove this bit.

By all means add a comment to point the user to `~/.profile` or `/etc/profile.d/`

sesman/startwm.sh
```diff
@@ -85,13 +85,14 @@ wm_start()
...
-    if [ -r /etc/xrdp/export_desktop_session ]; then
-      . /etc/xrdp/export_desktop_session
-    fi
+    # if you want to start preferred desktop environment,
+    # add following line,
+    #  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION=<your preferred desktop>
+    # in either of following file.
+    # 1. ~/.profile
+    # 2. create a file (any filename is OK) in /etc/profile.d
+    # <your preferred desktop> shall be one of "ls -1 /usr/share/xsessions/"
```
akarl10: I would suggest `"ls -1 /usr/share/xsessions/" without .desktop suffix`

sesman/startwm.sh
```diff
@@ -85,13 +85,14 @@ wm_start()
...
+    # e.g.  [ -n "$XRDP_SESSION" ] && export DESKTOP_SESSION==ubuntu
```
akarl10: should probably be DESKTOP_SESSION=ubuntu

akarl10: I like the documentation. This way we give some kind of "best practice" and prevent too many custom solutions/hacks that might break if upstream changes something

Hiero32:
I tried to put
```
[ -n "XRDP_SESSION" ] && export DESKTOP_SESSION=ubuntu
```
into ~/.profile.
It works. However, I found an small issue in the following steps.
1. local login (gnome-flashback-metacity) (from GDM) -> XRDP_SESSION is not set.  DESKTOP_SESSION=gnome-flashback-metacity is set
2. logout
3. xrdp  login -> XRDP_SESSION is set. DESKTOP_SESSION=ubuntu is set
4. logout
5. local login (gnome-flashback-metacity)  if interval between 3. and 5. is less than around 15sec XRDP_SESSION, XRDP_PULSE_SINK_SOCKET, XRDP_PULSE_SOURCE_SOCKET, XRDP_SOCKET_PATH and PULSE_SCRIPT are still set.

akarl10:
> I tried to put
> 
> ```
> [ -n "XRDP_SESSION" ] && export DESKTOP_SESSION=ubuntu
> ```
> 
> into ~/.profile. It works. However, I found an small issue in the following steps.
> 
>     1. local login (gnome-flashback-metacity) (from GDM) -> XRDP_SESSION is not set.  DESKTOP_SESSION=gnome-flashback-metacity is set
> 
>     2. logout
> 
>     3. xrdp  login -> XRDP_SESSION is set. DESKTOP_SESSION=ubuntu is set
> 
>     4. logout
> 
>     5. local login (gnome-flashback-metacity)  if interval between 3. and 5. is less than around 15sec XRDP_SESSION, XRDP_PULSE_SINK_SOCKET, XRDP_PULSE_SOURCE_SOCKET, XRDP_SOCKET_PATH and PULSE_SCRIPT are still set.

say Hooray to user@.service
nowadays some desktop environments don't start applications directly with "exec" but using a transient systemd user unit. the environment used by applications executed this way is `systemctl --user show-environment`

the 3-5 sec timeout is the time it takes to stop that unit after the LAST session of the user (if you would have a ssh session active this environment will not clear)

Hiero32:
Understood. Thanks.

matt335672:
@Hiero32 - there's probably not much you can do in the scope of this change to fix this problem. See #2491 for more background on this. @akarl10 has spent quite a bit of effort looking into this.

Personally, once you've fixed up @akarl10's documentation sections I'd be happy to merge this.

akarl10:
I think in https://github.com/neutrinolabs/xrdp/pull/2557/commits/a2a8a0ba89a4708db074a800073541a5d9ed533d the documentation is already fixed. At least I am happy with this inline doc

If we stick to the .profile recommendation we could remove the comment about SessionVariables (note that sesman.ini never contained DESKTOP_SESSION so this would be the only place that was mentioned)

matt335672:
Agreed.

@Hiero32 - do you want to remove this comment at lines 60 and 61?

```
 # DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
 # If set and valid then the STARTUP command will be taken from there
```

then we'll merge this.

Hiero32:
There are a couple of reasons I prefer .profile and /etc/profile.d.
- .profile and /etc/profile.d support both per-user and system-wide.
- log out then log in is enough to reflect modification of .profile or /etc/profile.d.
- sesman.ini can be over written when apt upgrade is executed.
- reboot is required to reflect modification of sesman.ini.

I replaced
```
# DESKTOP_SESSION should be set in sesman.ini in the SessionVariables section.
# If set and valid then the STARTUP command will be taken from there
```
with
```
# If DESKTOP_SESSION is set and valid then the STARTUP command will be taken from there
```

Also I confirmed /etc/profile.
I found .sh suffix is necessary for filename in /etc/profile.d.
changed to
```
 # 2. create a file (any_filename.sh is OK) in /etc/profile.d

matt335672:
Thanks @Hiero32 for sticking with this.

PR #2567, opened 2023-02-27
Fix CI build break

matt335672:
This is based on [this comment](/actions/runner-images/issues/7192#issuecomment-1445902492) which was posted in the middle on the night Eastern time. So you wont have seen this @Nexarian 

My own personal preference would be to remove the `apt-get upgrade` line as this is quicker, but this should do the job too.


Nexarian (approved):

Nexarian:
I actually prefer this :)

PR #2586, opened 2023-03-09
waitforx logging improvements

matt335672:
As mentioned on Gitter, this makes a few changes to the waitforx executable:-
1) Executable is now in `$libexecdir/xrdp/waitforx` as it isn't user facing and doesn't need to be on the PATH.
2) The logging and error status mechanisms are significantly improved to allow for better logging of the problem(s) encountered with the X server.

You might find you also need neutrinolabs/xorgxrdp#254 in some situations.


sesman/Makefile.am
```diff
@@ -5,7 +5,7 @@ AM_CPPFLAGS = \
   -DXRDP_SYSCONF_PATH=\"${sysconfdir}\" \
   -DXRDP_CFG_PATH=\"${sysconfdir}/xrdp\" \
   -DXRDP_SBIN_PATH=\"${sbindir}\" \
-  -DXRDP_BIN_PATH=\"${bindir}\" \
```
metalefty, line 8: Good idea.

matt335672:
I'm going to merge this. Further comments on waitforx can be added to #2592

PR #2587, opened 2023-03-09
[refactoring] xrdp: enable loading any keymap files

metalefty:
by separating functions. This makes it easier to create unit tests.

-------

## Background

I'm in the initial implementation stage of TOML config parser. Firstly, I started to TOML parser for keymap files.

I would like to confirm TOML parser is valid by comparing results of current INI parser and TOML parser are the same in unit tests.  For this purpose, I need functions that can load any keymap file.

Pseudo test code will be like this:

```C

struct xrdp_keymap *keymap_ini;
struct xrdp_keymap *keymap_toml;

get_keymaps_by_file("./km-00000409.ini", keymap_ini);
get_keymaps_by_file_toml("./km-00000409.toml", keymap_toml);

compare_xrdp_keymap(keymap_ini, keymap_toml);
```

### DIFF
--- a/xrdp/lang.c
+++ b/xrdp/lang.c
@@ -287,8 +287,8 @@ int km_load_file(const char *filename, struct xrdp_keymap *keymap)
 {
     int fd;
 
-    fd = g_file_open(filename);
     LOG(LOG_LEVEL_INFO, "Loading keymap file %s", filename);
+    fd = g_file_open(filename);
 
     if (fd != -1)
     {
@@ -306,6 +306,7 @@ int km_load_file(const char *filename, struct xrdp_keymap *keymap)
     }
     else
     {
+        LOG(LOG_LEVEL_ERROR, "Error loading keymap file %s (%s)", filename, g_get_strerror());
         return 1;
     }
 
--- a/xrdp/lang.c
+++ b/xrdp/lang.c
@@ -231,7 +231,6 @@ km_read_section(int fd, const char *section_name, struct xrdp_key_info *keymap)
 int
 get_keymaps(int keylayout, struct xrdp_keymap *keymap)
 {
-    int fd;
     int basic_key_layout = keylayout & 0x0000ffff;
     char *filename;
     struct xrdp_keymap *lkeymap;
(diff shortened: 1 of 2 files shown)


metalefty:
Changed function name a bit.

xrdp/lang.c
```diff
@@ -258,36 +257,21 @@ get_keymaps(int keylayout, struct xrdp_keymap *keymap)
...
-            km_read_section(fd, "shiftcapslock", keymap->keys_shiftcapslock);
-            km_read_section(fd, "shiftcapslockaltgr", keymap->keys_shiftcapslockaltgr);
-
-            if (g_memcmp(lkeymap, keymap, sizeof(struct xrdp_keymap)) != 0)
-            {
-                LOG(LOG_LEVEL_WARNING,
-                    "local keymap file for 0x%08x found and doesn't match "
-                    "built in keymap, using local keymap file", keylayout);
-            }
+        lkeymap = (struct xrdp_keymap *)g_malloc(sizeof(struct xrdp_keymap), 0);
```
matt335672, line 261: Should probably add a check for malloc failure here.

xrdp/lang.c
```diff
@@ -297,3 +281,33 @@ get_keymaps(int keylayout, struct xrdp_keymap *keymap)
...
+
+/*****************************************************************************/
+int km_load_file(const char *filename, struct xrdp_keymap *keymap)
+{
+    int fd;
+
+    fd = g_file_open(filename);
+    LOG(LOG_LEVEL_INFO, "Loading keymap file %s", filename);
+
+    if (fd != -1)
```
matt335672, line 293: Need to log error if keymap couldn't be opened

xrdp/xrdp.h
```diff
@@ -411,6 +411,9 @@ get_char_from_scan_code(int device_flags, int scan_code, int *keys,
 int
 get_keymaps(int keylayout, struct xrdp_keymap *keymap);
 
+int
+km_load_file(const char *filename, struct xrdp_keymap *keymap);
+
```
matt335672, line 416: I notice this is a global. Is the intention to use it elsewhere later on?

matt335672 (commented):
A few comments for you.

xrdp/lang.c
metalefty: Agreed but this is the same as before. I would like to add check later.

xrdp/lang.c
metalefty: Indeed.

xrdp/xrdp.h
metalefty: Yes, this is intentional and I'm going to add some unit tests later.

matt335672 (approved):

PR #2595, opened 2023-03-16
Add planar compression

Nexarian:
Used for the bitmaps transmitted for the login screen over the egfx channels.

This leverages standard bitmap planar compression. See these references:
- https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/9b422f69-8e05-4c6d-b6fb-fa02ef75a8f2
- https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/0d2e1a03-e123-46b2-b2b8-ed730a794ae4
- https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/749fa4d4-0d88-4fe5-bf4c-914cd05776cb

Of course I should add that I didn't write this. The credit for that goes to @jsorg71 -- I'm packaging it up and helping merge it.

I probably should write unit tests for this, but I'm not actually sure how...

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -2094,7 +2094,9 @@ libxrdp_process_monitor_stream(struct stream *s,
         }
     }
 
-    /* set wm geometry if the encompassing area is well formed. Otherwise, log and return an error.*/
+    /* set wm geometry if the encompassing area is well formed.
+       Otherwise, log and return an error.
+    */
     if (all_monitors_encompassing_bounds.right
             > all_monitors_encompassing_bounds.left
             && all_monitors_encompassing_bounds.bottom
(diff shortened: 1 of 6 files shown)


matt335672 (commented):
Looks fine to me. There are no obvious leaks. I'm not really qualified to comment on the methods.

matt335672:
Is this sufficiently important to add to NEWS?

Nexarian:
No, because this is "sleeper code" -- It's not hooked up yet. In the next week, due to the RDP IO conference, I expect Jay and I will enable support for CPU accelerated H264 GFX in devel. If there is a release after that, this can be bundled in that news.

If for some reason that DOESN'T happen, I would expect a minor note that "gfx code is being merged into this release, but it is not yet active. Stay turned for this to be turned on in future releases."

As discussed, I've opted to do it this way to make it so that adding GFX to XRDP is done in digestible pieces, rather than all at once, because a gigantic monolithic PR is both hard to review, and hard to learn from later when doing code archaeology.

matt335672:
Fine - thanks for the update.

I agree with you about the digestible pieces comment. I'm struggling with getting the sesman changes done for the same reason.

matt335672:
PS : Have a great week!

PR #2601, opened 2023-03-20
Add sesmanlib

matt335672:
**Edit** - this is now ready for review. I've also got a few questions below which I'd like some comments on.

This is the next step in separating sesexec out from sesman (#1961).

Both sesexec and sesman will need to share common code (identified below). In addition, there are already some couplings between sesman and related executables:-
- sesrun uses the sesman config module.
- authtest uses the sesman auth module
- Both sesman and chansrv have duplicated code to implement "clipboard restrict" parameters, introduced with #2082

This PR creates a new shared library for all code which is shared between sesman and these other executables.

The shared modules are:-
1) The access module (will be needed by sesexec too)
2) The auth module(s) (needed by sesexec and used by authtest)
3) A new 'clip restrict' module for the code currently duplicated in sesman and chansrv
4) The config module (needed by sesexec and used by sesrun)

At this stage I'd really appreciate comments on this approach. I think a library is the right approach, but I'm really not sure about the following:-
- Is there a better name for it?
- Is it in the right place in the directory tree?
- Have I left anything out?
- I've renamed modules with common names, so config.c becomes sesman_config.c. Is that a reasonable thing to do?

Thanks.


sesman/libsesman/sesman_config.c
```diff
@@ -614,39 +638,16 @@ config_dump(struct config_sesman *config)
 
     /* Security configuration */
     g_writeln("Security configuration:");
-    g_writeln("    AllowRootLogin:            %d", sc->allow_root);
+    g_writeln("    AllowRootLogin:  esm         %d", sc->allow_root);
```
metalefty: What does esm mean？

matt335672: That looks like finger trouble to me - I'll remove it. Thanks!

metalefty:
Overall, LGTM.

metalefty (approved):

PR #2609, opened 2023-03-25
common: change g_malloc, g_free, g_memset, g_memcpy, and g_memmove to…

jsorg71:
… macros.  This should allow compilers to optimize these calls.

### DIFF
--- a/common/arch.h
+++ b/common/arch.h
@@ -20,6 +20,7 @@
 #define ARCH_H
 
 #include <stdlib.h>
+#include <string.h>
 
 #if defined(HAVE_STDINT_H)
 #include <stdint.h>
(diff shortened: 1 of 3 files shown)


matt335672:
@metalefty - do you want this in teh v0.9 branch as well?

metalefty:
I think it would be nice if we can have benefits from compiler optimization. Let's take this to v0.9 branch.

Nexarian:
Curious: How much of a performance gain do we think we'll get from this compiler optimization?

matt335672:
Not a vast amount, but we may well get some code shrinkage (which will have caching benefits). The compiler can optimise copies of small fixed memory sizes into one or more move instructions, avoiding a function call.

This is also a step to moving to #2499 - using library functions makes it (hopefully) easier to attract other contributions, although we've got a long way to go on this!

PR #2639, opened 2023-04-19
[devel] sesman: Prevent the use of 'alternate shell'

### LINKED ISSUE #850: prevent the use of "alternate shell"
Hi, is there a way to prevent Users from executing a single application instead of running the intended Desktop?  I guess it would make sense to have an option in sesman.ini to set the intended behavior.

matt335672:
By setting the new config value 'AllowAlternateShell' to 'no' it is now possible to prevent the use of an alternate shell, which can be set by the connecting user.
The default remains unchanged and any shell is allowed if the config value is not specified. It can also be set explicitly to 'yes' to achieve the same outcome.

Fixes: #850


matt335672:
Port of #2634 from the v0.9 stream

PR #2649, opened 2023-04-28
remove unnecessary include `check.h`

jat001:
same problem as #2124

`check.h` removes `HAVE_STDINT_H`, `log.h` includes `arch.h` which requires `HAVE_STDINT_H` to be defined.

`test_libxrdp.h` already includes `check.h`. so just remove it.

### DIFF
--- a/tests/libxrdp/test_libxrdp_main.c
+++ b/tests/libxrdp/test_libxrdp_main.c
@@ -4,7 +4,6 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <check.h>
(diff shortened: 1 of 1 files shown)


matt335672:
@jat001 - thanks for your contribution.

Which platform did you discover this on by the way?

jat001:
> @jat001 - thanks for your contribution.
> 
> Which platform did you discover this on by the way?

Arch Linux

matt335672:
Thanks.

@metalefty - I'll see if a backport to v0.9 is also necessary.

PR #2667, opened 2023-05-12
Use config_ac.h consistently and correctly

iskunk:
This is a cleanup in the vein of #647.

Not all `.c` files were #including `config_ac.h` directly, and due to the way the #include was handled, some files were seeing the config definitions twice:

* `common/pixman-region16.c`
* `libipm/libipm.c`
* `libipm/libipm_send.c`
* `libipm/libipm_recv.c`
* `libipm/eicp.c`
* `libipm/ercp.c`
* `libipm/scp.c`

This PR makes everything consistent.

(It might be desirable to give `common/pixman-region.c` a different filename extension, perhaps `.inc`, to make clear that it is not meant to be compiled directly.)

### DIFF
--- a/common/pixman-region.c
+++ b/common/pixman-region.c
@@ -63,8 +63,11 @@
  * PERFORMANCE OF THIS SOFTWARE.
  */
 
-#if defined(HAVE_CONFIG_H)
-#include <config_ac.h>
+#ifndef PIXMAN_REGION_MAX
+#error "This file should be #included from pixman-region16.c, not compiled directly"
+#endif
+#ifndef CONFIG_AC_H
+#error "config_ac.h not visible in pixman-region.c"
 #endif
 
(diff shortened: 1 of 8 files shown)


fontutils/windows/fontdump.c
```diff
@@ -16,6 +16,10 @@
  * limitations under the License.
  */
 
+#if defined(HAVE_CONFIG_H)
+#include <config_ac.h>
+#endif
+
 #include <windows.h>
```
matt335672: This utility is intended for Windows only (where autoconf isn't used), so should not be modified.

matt335672 (commented):
Thanks @iskunk  - one comment below.

fontutils/windows/fontdump.c
iskunk: All right, I'll remove it from that file.

matt335672:
Thanks for your contribution @iskunk

PR #2670, opened 2023-05-13
add large cursor support, posix shm

jsorg71:


matt335672 (approved):
I like this - it's a pretty clean and it shows how we can migrate away from SysV shared memory.

I'm happy to merge this as a first step to getting rid of SysV shared memory in devel.

matt335672:
Jay - 32-bit CI is fixed (#2674).

matt335672:
Apologies - I've just merged a conflicting change to os_calls.c

PR #2672, opened 2023-05-15
Simplify interaction with systemd

iskunk:
When xrdp is being managed with systemd...

* Forking is not necessary, as systemd does not need the program to exit;
* PID files are not necessary, when the program does not fork;
* The `--kill` invocation is not needed, as systemd can send the SIGTERM itself;
* The `--reload` invocation is not needed, as systemd can send the SIGHUP itself.

This PR simplifies the systemd units accordingly. This will not only make managing the files in `/run` easier, it will also reduce the operating footprint that needs to be accounted for in AppArmor profiles and other security frameworks.

### DIFF
--- a/instfiles/xrdp-sesman.service.in
+++ b/instfiles/xrdp-sesman.service.in
@@ -6,13 +6,11 @@ StopWhenUnneeded=true
 BindsTo=xrdp.service
 
 [Service]
-Type=forking
-PIDFile=@localstatedir@/run/xrdp-sesman.pid
+Type=exec
 EnvironmentFile=-@sysconfdir@/sysconfig/xrdp
 EnvironmentFile=-@sysconfdir@/default/xrdp
-ExecStart=@sbindir@/xrdp-sesman $SESMAN_OPTIONS
-ExecStop=@sbindir@/xrdp-sesman $SESMAN_OPTIONS --kill
-ExecReload=@sbindir@/xrdp-sesman $SESMAN_OPTIONS --reload
+ExecStart=@sbindir@/xrdp-sesman $SESMAN_OPTIONS --nodaemon
+ExecReload=kill -HUP $MAINPID
 
 [Install]
 WantedBy=multi-user.target
--- a/instfiles/xrdp.service.in
+++ b/instfiles/xrdp.service.in
@@ -5,12 +5,10 @@ Requires=xrdp-sesman.service
 After=network.target xrdp-sesman.service
 
 [Service]
-Type=forking
-PIDFile=@localstatedir@/run/xrdp.pid
+Type=exec
 EnvironmentFile=-@sysconfdir@/sysconfig/xrdp
 EnvironmentFile=-@sysconfdir@/default/xrdp
-ExecStart=@sbindir@/xrdp $XRDP_OPTIONS
-ExecStop=@sbindir@/xrdp $XRDP_OPTIONS --kill
+ExecStart=@sbindir@/xrdp $XRDP_OPTIONS --nodaemon
 
 [Install]
 WantedBy=multi-user.target


matt335672:
Hi @iskunk 

Thanks for this.

Please rebase on latest devel to fix 32-bit CI problems (#2674).

I'm personally happy with this. I should point out however that we're changing a user-visible interface in that `xrdp-sesman --reload` will no longer work, but this is a major version upgrade and the same functionality is exposed through systemctl anyway. We just shouldn't back-port this to 0.9

@metalefty - what do you think?

iskunk:
Rebased as requested.

It might be good to consider whether `--kill` and `--reload` are worth keeping, given that literally all these do is send SIGTERM or SIGHUP to the appropriate process (which then has to be allowed for in the AppArmor profile). Unix daemons typically just document the signal, and let the user determine which process to send it to with `killall`/`pkill`/etc.

matt335672:
I'd rather not remove the existing commands, although I largely agree with where you're coming from.

Here are a couple of quotes from [systemd.service(5)](https://www.man7.org/linux/man-pages/man5/systemd.service.5.html) [www.man7.org]:-

>      ExecReload=
>          ...
>          Note however that reloading a daemon by sending a signal (as with
>           the example line above) is usually not a good choice, because this
>           is an asynchronous operation and hence not suitable to order
>           reloads of multiple services against each other. It is strongly
>           recommended to set ExecReload= to a command that not only triggers
>           a configuration reload of the daemon, but also synchronously waits
>           for it to complete.

and
>      ExecStop=
>          ...
>           Note that it is usually not sufficient to specify a command for
>           this setting that only asks the service to terminate (for example,
>           by sending some form of termination signal to it), but does not
>           wait for it to do so. Since the remaining processes of the services
>           are killed according to KillMode= and KillSignal= or
>           RestartKillSignal= as described above immediately after the command
>           exited, this may not result in a clean stop. The specified command
>           should hence be a synchronous operation, not an asynchronous one.

Systemd recommended usage is at odds with the standard UN*X approach. At the moment systemd appears to be losing this particular battle, but if we ever need to implement synchronous reload or stop, we'll need to do it via the existing command mechanism, or something similar.

iskunk:
Systemd offers `Type=notify` (described in the same man page), in which the service uses `sd_notify(3)` to inform the system that it is usable. But that is systemd-specific, of course.

I don't believe there's a general solution to fully synchronous Unix service control. But at the same time, I don't foresee that xrdp will have many services depending on it, to bring about the need for this in the first place.

metalefty:
Overall, I'm fine with this. Fedora already applies a patch similar to this PR.
https://src.fedoraproject.org/rpms/xrdp/blob/f38/f/xrdp-0.9.4-service.patch

iskunk:
In my first draft of this change, I similarly got rid of the `Type=` line altogether, going with the systemd default of `Type=simple`. But then I noticed in the documentation that in this mode, systemd considers the service started immediately after the `fork()` call prior to executing the service. Whereas in `Type=exec`, systemd considers the service started after the `exec()` call succeeds. And thus, as the man page states outright, the systemd default of `Type=simple` won't tell you right off the bat if there was a problem executing the service.

(I don't understand why the systemd doc makes such a big deal of services starting up synchronously, when it defaults to a service mode that can't even report the basic fact of whether execution of the service binary was successful or not. I can see all this being of interest if you're gunning for three-second boot times, but for real-world usage, it's an odd design decision.)

matt335672:
For anyone else following this conversation, the text in question is in [systemd.service(5)](https://www.man7.org/linux/man-pages/man5/systemd.service.5.html) [man7.org]

I agree that this seems a bit complex, but in systemd's defence (that's the first time I've ever written that!) it's trying to solve a problem that's been glossed over for decades in the UN*X world. Sun's SMF in Solaris 10 came close to solving it, but from memory. that was a nightmare to work with.

Digressing slightly, I want to get to a place where sesman can be restarted. We're not far off that, but at that point I can see a use for adding `sd_notify()` to the code, as sesman needs to rediscover existing services before it announces it's ready. We already detect systemd for makefiles, so adding a C pre-processor symbol is trivial.

What I suggest for now is to stick with `Type-exec` as you have here, and later we can consider moving to `Type-notify`.

To respond to your other previous point, the reload and terminate can be made synchronous by adding them to the SCP protocol we already have. Authentication would be done via Unix Domain Socket ownership. This gives us a finer-grain of access control (if we want it) than a signal.

Is that OK? Can I merge this now?

iskunk:
> Digressing slightly, I want to get to a place where sesman can be restarted. We're not far off that, but at that point I can see a use for adding `sd_notify()` to the code, as sesman needs to rediscover existing services before it announces it's ready. We already detect systemd for makefiles, so adding a C pre-processor symbol is trivial.

But wouldn't adding an `sd_notify()` call be for the benefit of services that depend on sesman (presumably sesman would use it to inform systemd that it is ready to accept connections), rather than sesman finding its own dependencies?

> What I suggest for now is to stick with `Type-exec` as you have here, and later we can consider moving to `Type-notify`.

Sounds good!

> To respond to your other previous point, the reload and terminate can be made synchronous by adding them to the SCP protocol we already have. Authentication would be done via Unix Domain Socket ownership. This gives us a finer-grain of access control (if we want it) than a signal.

If/when you go this route, I would suggest using a separate control program (an `xrdpctl` or `xrdp-control` or the like) for this purpose, instead of options to the main binaries. With the existing `--kill`/`--reload` options, the permissions required to send the signals to the running process need to be added to the AppArmor profiles for the main binaries---meaning that the normal server processes are then able to send those same signals at will. (The target of those signals can be restricted, so an attacker won't be able to send SIGHUP/SIGTERM to arbitrary processes, but it's still a permission that could be useful to them and would not be necessary if some other program were sending the signals.)

Caveat: An alternative that I've considered is, rather than having the AppArmor confinement apply to any invocation of the binary (as is standard practice), it would be possible for the systemd unit to apply the confinement itself (see `AppArmorProfile=` in `systemd.exec(5)`). Thus, any other invocations of the xrdp programs, such as from the command line, would be unconfined. So a program-options approach would then require no additional permissions in the profile. But I'm unsure if using this systemd feature is desirable, not least since having different confinements on the differing invocations may be confusing.

> Is that OK? Can I merge this now?

Please do! However the init system interface is elaborated later on, I think for now this is a good move.

matt335672:
Thanks @iskunk 

The `sd_notify()` would indeed tell systemd that sesman is accepting connections. This might be useful (for example) if a later script is going to use `xrdp-sesrun` to set up a session for a user before they log in.

Thanks for your other detail - that's useful. I can see the point of a separate binary for this. We can maybe use `xrdp-sesadmin` as that's a command-line utility we already have. I'm planning on greatly increasing the scope of this program over time.

iskunk:
Thank you likewise!

Though service startup synchronicity is a known issue in theory, I'd be curious to see what problems xrdp users run into when it is not addressed. The Unix signal mechanism is simple and well-understood, and there's value in that, too... I'd hope there is a good ROI for implementing a more complex mechanism, or one that is systemd-specific.

I wasn't aware of `xrdp-sesadmin`! Yes, that looks like a fine place for a one-stop shop xrdp run-time control interface.

Overall, I'm happy to see all the progress happening here under your name, and am looking forward to the result :-)

metalefty:
@iskunk thanks for your work.

PR #2686, opened 2023-05-17
Re-work FIFO code

### LINKED ISSUE #2529: Investigate combining fifo.c files
Not a huge priority.

See #2526 for the background on this.

Unit tests can also be added for the new fifo library module.

matt335672:
Fixes #2529 

I took the best features of the existing fifo implementations and merged them. The result is faster than both and has no awkward edge cases.

The only significant other change is I've allowed an item destructor to be specified when the fifo is created. This allows the fifo to be cleared or deleted without the code having to explicitly empty it first.

At the moment fifos are used for :-
1) sending data between the xrdp encoder thread and the main xrdp thread
2) Queueing incoming audio data

I've tested both of these, and also added a fifo test module to the test suite.


jsorg71:
This looks good to me, nice work.

jsorg71 (approved):
It minor but for assigned functions we normally don't use the (*self->func_name)() we just use self->func_name() but I really don't have a strong preference.

matt335672:
Thanks for that. I've updated the coding style [on the wiki](https://github.com/neutrinolabs/xrdp/wiki/Coding-Style) to correspond.

PR #2703, opened 2023-05-26
Add script to make release tarball

metalefty:
to avoid mistake in release procedure.

Sorry for the v0.9.22 release. I made a script to generate release tarball not to make such mistake again.


matt335672 (approved):
Great idea.

This LGTM. It's nice and simple, and using `set -e` prevents the need to check lots of error returns.

PR #2709, opened 2023-05-30
move to posix shm

jsorg71:


matt335672 (commented):
Code all looks fine.

It strikes me that you could remove the older `process_server_paint_rect_shmem()` and `process_server_paint_rect_shmem_ex()` calls at this point as we're in devel and these are no longer being called by xorgxrdp. Or is this a bad idea?

Nexarian:
@matt335672 They're still used by the GFX mainline_merge branch. Maybe a comment to merge the two is in order, but I don't want to break mainline_merge just yet, it'll take a bit to integrate with this going forward.

matt335672:
Fine - that sounds very much like it's a 'bad idea' to remove them then!

matt335672:
@jsorg71 - Can we merge this now?

Thanks

Nexarian:
Just a note that I'm good with this. It will break `mainline_merge` but I'm happy to rebase and refactor on top of this. We need this change, and my AVC444 work should not be considered blocking any more.

Nexarian (approved):

jsorg71:
Yes this will break mainline_merge, I'll help fix that too.

Nexarian:
This also needs to go in: https://github.com/jsorg71/xorgxrdp/tree/posix_shm

jsorg71:
I think that I got the xorgxrdp part done.  A good reason to leave the sysv share memory part in for a bit is that newer xrdp can work with older xorgxrdp.

Nexarian:
This is great, I was working on adding it to `mainline_merge` today!

jsorg71:
I had to do another PR to get everything working. #2786 That plus the PR from xorgxrdp should work.
To get xorgxrdp_helper working, I think we'll need to watch for message 64 and accept the fd and resend it.  First we should get it working in mainline_merge without xorgxrdp_helper.  I think we'll need that for message 63 too.

Nexarian:
Cool, I'm happy to try my hand at getting this working. Not quite sure I understand how to do the handoff with xorgxrdp_helper but as you said, that comes later.

Should probably have one of our Friday chats soon. I have a lot of updates about AVC444.

PR #2719, opened 2023-06-08
Log xrdp termination signals

matt335672:
I've made some changes to xrdp so that when a sub-process terminates on an unexpected signal, a message is logged.

This is primarily to support #2697. Without this, there is no log message generated if xrdp encounters a SECCOMP violation unless the user looks in the system or audit log.

With this change, messages like the following appear in xrdp.log:-

```
[2023-06-08T16:27:59.780+0100] [ERROR] [process_pending_sigchld_events(xrdp_listen.c:866)] Child 11738 terminated unexpectedly with signal "Bad system call"
```

and in xrdp-sesman.log, for a SIGKILL:-

```
[2023-06-08T16:39:35.869+0100] [WARN ] [pid:14702 tid:140681994113152] [session_process_child_exit(session.c:869)] Window manager (pid 14911, display 12) exited with signal "Killed". This could indicate a window manager config problem
```

I have a question though. I've used g_strsignal() which returns strings like "Bad system call" or "Killed". I can't help feeling that the symbolic names SIGSYS, SIGKILL, etc might be more informative for these errors.

Thoughts please? @iskunk - do you have any opinions on this?

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -21,6 +21,7 @@
 #if defined(HAVE_CONFIG_H)
 #include "config_ac.h"
 #endif
+#include <signal.h>
 #include <string.h>
 #include <strings.h>
 #include <stdlib.h>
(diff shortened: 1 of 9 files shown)


iskunk:
I agree that proper signal names are preferable. The "friendly" names don't really help anyone... experts will have to recall which signal is being described, and novices won't know what signals are anyway.

I was going to say (from the other issue) that the message shouldn't presume a SIGSYS is due to a seccomp filter tripping, but that's no longer the case here.

Overall, a  nice bit of polish for an otherwise unforgiving security mechanism!

matt335672:
That was more challenging that I expected.

Messages like the following are now logged:-

```
[2023-06-12T15:43:17.556+0100] [ERROR] [process_pending_sigchld_events(xrdp_listen.c:867)] Child 17516 terminated unexpectedly with signal SIGSYS
```

iskunk:
Just to be clear, whose feedback are you waiting for?

matt335672:
Anyone in the team who has a bit of free time to look at this.

I've had a private chat with @Nexarian about merging over the short term. I'm now expecting him to merge this when he's ready.

metalefty (approved):

PR #2734, opened 2023-06-19
Xrdp pcsc1

last1284:


matt335672:
Thanks for this @last1284 

I agree that the PCSC situation isn't ideal at the moment. However, this area still needs a lot of work before it's production ready.

I think what I'll do before we drop the next major release is pull in @zorgluf's changes, but place all the PCSC code under an `--enable` flag. That should let people at least us it, and we can start looking into ironing out some of the issues.

Can I ask you to separate out your keyboard change into a separate PR? That's not at all contentious, and we can merge it quickly.

PR #1077, opened 2018-03-23
Wtmp

moobyfr:
Add the utmp/wtmp support.

PR #890, opened 2017-09-27
Feature wtmp

moobyfr:
WIP utmp/wtmp
- should work under linux
- on freebsd wtmp is ok, but updwtmpx() isn't known

PR #2745, opened 2023-06-30
Add utmp/wtmp support

matt335672:
I thought it might be easiest to discuss this now by way of another PR.

For the history, see #38, #890, #1077 and more recently discussion #2744. There are probably others!

The points I took away from the discussion were:-
1) PAM on at least Linux and FreeBSD is capable of updating wtmp and lastlog, and should be used where possible. This is because these files have a Y2038 problem on glibc systems. See the links in #2744 for info on this.
2) utmp is a harder problem to solve, and in the short term can still be updated with traditional interfaces.
3) btmp is not used enough to worry about.

I've taken @moobyfr's wtmp branch, rebased it and added a few extra commits. The rebasing was a little challenging as all the code has now moved to sesexec rather than being in sesman. The commits do the following:-
1) Fix a name collision between the `struct exit_status` I added to os_calls and the one in (Linux) utmpx.h
2) Removed wtmp updating code and moved it to the PAM files.
3) Added `--enable-utmp` to configure.ac to allow this feature to be used only where required.
4) Added support for some Linux-specific utmpx fields.
5) Fixed a testing problem I found with g_strncpy().

On a Ubuntu 22.04 system with this branch, I now have the following when logging on:-

```
$ who -a | grep testuser
testuser ? xrdp:11      2023-06-30 11:29   ?        137441 (::ffff:<IP4-addr>)
w testuser
 11:30:35 up  1:56,  3 users,  load average: 0.48, 0.35, 0.17
USER     TTY      FROM             LOGIN@   IDLE   JCPU   PCPU WHAT
testuser xrdp:11  <snipped>        11:29    0.00s  0.00s  0.08s /usr/libexec/gnome-session-binary
$ utmpdump /run/utmp | grep testuser
Utmp dump of /run/utmp
[7] [137441] [:b  ] [testuser] [xrdp:11     ] [<snipped> ] [0.0.0.0        ] [2023-06-30T10:29:28,096368+00:00]
$ who /var/log/wtmp | tail -1
testuser :11          2023-06-30 11:29 (<snipped>)
$ lastlog -u testuser
Username         Port     From             Latest
testuser         :11      <snipped>        Fri Jun 30 11:29:26 +0100 2023
```

After logging out normally:-

```
$ who -a | grep testuser
$ w testuser
 11:37:03 up  2:03,  2 users,  load average: 0.08, 0.32, 0.24
USER     TTY      FROM             LOGIN@   IDLE   JCPU   PCPU WHAT
$ utmpdump /run/utmp | grep testuser
Utmp dump of /run/utmp
$ who -a /var/log/wtmp | tail -2
testuser ? :11          2023-06-30 11:29   ?        137236 (<snipped>)
           :11          2023-06-30 11:34            137236 id=      term=0 exit=0
$ lastlog -u testuser
Username         Port     From             Latest
testuser         :11      <snipped>        Fri Jun 30 11:29:26 +0100 2023
```

A couple of other points:-
1. Arch linux doesn't seem to be shipping with either pam_lastlog.so or pam_lastlog2.so at the moment.
2. Debian/Fedora are still on pam_lastlog.so
3. OpenSUSE Tumbleweed has moved to pam_lastlog2.so

On another note, I'm going to be unavailable for a lot of July, but I've got a bit of time next Monday to work on this.


moobyfr:
LGTM, thx for getting this back and synced

matt335672:
I'm waiting on @Nexarian before I merge this - I don't want to disrupt some of the stuff he's looking at.

In the meantime I'll fix the conflicts.

matt335672:
A lot of time has passed, so I'm rebasing it. I'll retest soon and hopefully we can get this merged.

Nexarian:
Yeah, at this point I don't think anything I've worked on will affect this in light of the recent merges!

matt335672:
@metalefty - I've retested this and it still works!

Are you happy for me to merge it?

metalefty:
Let me test.

matt335672:
A couple of testing notes - these will be added to the release notes under the 'packagers and maintainers' section.

# Linux
1) You'll need to compile with `--enable-utmp` to get utmp support.
2) wtmp/lastlog support is provided via PAM. I've updated the PAM files in instfiles, so if you're using `make install` you should be fine. Otherwise you may need to modify `/etc/pam.d/xrdp-sesman` yourself.

# FreeBSD
FreeBSD has its UTX database which replaces all of the above files. It is supported via the FreeBSD [pam_lastlog module](https://github.com/freebsd/freebsd-src/tree/main/lib/libpam/modules/pam_lastlog) which I think is called normally anyway. In other words, this should already work on FreeBSD without this PR.

metalefty:
I did a quick test only on Linux. LGTM overall.

BTW, the from entry is not updated after reconnecting another client from a different IP address.  Isn't it implemented yet?
```
[almalinux@a9xrdp ~]$ lastlog -u almalinux
Username         Port     From                                       Latest
almalinux        :10      ::ffff:192.168.96.7                       Wed Feb 21 12:20:08 +0000 2024
```

metalefty (approved):
LGTM, let's merge.

matt335672:
That's a really good point.

I've had a look at the [pam_lastlog source code](/linux-pam/linux-pam/blob/master/modules/pam_lastlog/pam_lastlog.c). The `lastlog` file is only updated when a session is opened or closed.

The PAM session lifetime is that of the user session - i.e. it starts when the user logs a session in and ends when that session finishes.

For a reconnect we don't start a new session - we only go through the `auth` and `account` stages of PAM. So I think this is sort-of expected if we use PAM to implement this functionality.

matt335672:
I've added your observation to the `known issues` on the NEWS page. I've also deleted the old `wtmp` branch from the repo.

PR #2766, opened 2023-08-07
Clipboard updates

### LINKED ISSUE #2763: When the client clipboard contains an image, ConvertSelection(target="STRING") requests hang forever
### xrdp version

0.9.21

~~~
xrdp 0.9.21

  Configure options:
      --build=x86_64-redhat-linux-gnu
      --host=x86_64-redhat-linux-gnu
      --program-prefix=
      --disable-dependency-tracking
      --prefix=/usr
      --exec-prefix=/usr
      --bindir=/usr/bin
      --sbindir=/usr/sbin
      --sysconfdir=/etc
      --datadir=/usr/share
      --includedir=/usr/include
      --libdir=/usr/lib64
      --libexecdir=/usr/libexec
      --localstatedir=/var
      --sharedstatedir=/var/lib
      --mandir=/usr/share/man
      --infodir=/usr/share/info
      --enable-pixman
      --enable-painter
      --enable-vsock
      --with-socketdir=/run/xrdp
      --with-imlib2
      build_alias=x86_64-redhat-linux-gnu
      host_alias=x86_64-redhat-linux-gnu
      CFLAGS=-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection 
      LDFLAGS=-Wl,-z,relro  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld
      PKG_CONFIG_PATH=:/usr/lib64/pkgconfig:/usr/share/pkgconfig

  Compiled with OpenSSL 1.1.1k  FIPS 25 Mar 2021
~~~

Rocky Linux 8.8 (Green Obsidian)

### Installation method

### Which backend do you use?

xorgxrdp

happens even in an empty X session

VM

### What's your client?

happens with both xfreerdp on Linux and Microsoft client on Windows

### Area(s) with issue?

Clipboard

### Steps to reproduce

1. Run an RDP client connecting to xrdp on some server.
2. Copy an image on the client-side.
3. Running the following command on the server-side (you may have to run it a few times, xrdp seems to do some caching):
xclip -out -selection clipboard

### ✔️ Expected Behavior

xsel should return immediately with the output "Error: target STRING not available", as on normal X.

### ❌ Actual Behavior

xsel hangs indefinitely.

### Anything else?

_No response_

matt335672:
Fixes #2763 

This PR contains three commits
1) The first of these adds more checking to a [Format Data Response PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/21cda270-0923-4302-898a-1b89f4c4904d). In particular, if the client responds with CB_RESPONSE_FAIL for any requested format, the X11 client is informed that the requested selection is not available, rather than the message being ignored. Additionally, pasted text is now given its own function, and more error checking is added.
2) The second of these only advertises text to the X11 clients when a text format (or a file list) is advertised by the Windows client.
3) The third of these is a tidy-up. There's a lot more that can be done in this area, but I've stuck with the minimum so that a backport to v0.9.x is feasible.

With these commits, when a picture is selected on the Windows client, the following can be seen using `xclip`:-

```
$ xclip -o -selection clipboard -t TARGETS
TARGETS
TIMESTAMP
MULTIPLE
image/bmp
$ xclip -o -selection clipboard -t STRING
Error: target STRING not available
```

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -179,6 +179,7 @@ x-special/gnome-copied-files
 #include "xcommon.h"
 #include "chansrv_fuse.h"
 #include "ms-rdpeclip.h"
+#include "xrdp_constants.h"
 
 static char g_bmp_image_header[] =
 {
@@ -245,9 +246,21 @@ static int g_cliprdr_flags = CB_USE_LONG_FORMAT_NAMES |
 static int g_formatIds[16];
 static int g_num_formatIds = 0;
 
-static int g_file_format_id = -1;
+/* Format ID assigned to "FileGroupDescriptorW" by the client */
+static int g_file_group_descriptor_format_id = -1;
 
 static char g_last_atom_name[256] = "";
+
+/*
+ * Values for the named formats we send to the client in
+ * a Format List PDU
+ */
+
+enum
+{
+    CB_FORMAT_FILE_GROUP_DESCRIPTOR = 0xc0bc
+};
+
 
 /*****************************************************************************/
 static char *
@@ -646,7 +659,7 @@ clipboard_send_format_announce(int xrdp_clip_type)
             case XRDP_CB_FILE:
                 LOG_DEVEL(LOG_LEVEL_DEBUG, "clipboard_send_format_announce: XRDP_CB_FILE");
                 /* canned response for "file" */
-                out_uint32_le(s, 0x0000c0bc);
+                out_uint32_le(s, CB_FORMAT_FILE_GROUP_DESCRIPTOR);
                 clipboard_out_unicode(s, "FileGroupDescriptorW", 21);
                 out_uint32_le(s, 0x0000c0ba);
                 clipboard_out_unicode(s, "FileContents", 13);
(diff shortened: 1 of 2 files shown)


metalefty:
Thanks for the fix, I'll test. I think backport to v0.9 is necessary.

matt335672:
The patches I've added cherry-pick nicely on to v0.9, which is one reason I didn't make more changes in this area! I'll do some more tidy-ups on devel only.

Since the cherry-pick was so trivial, I've added #2767 against v0.9. Test whichever one works best for you and I'll apply any fixes to both.

metalefty:
LGTM.

```
% xclip -o -selection clipboard -t TARGETS
TARGETS
TIMESTAMP
MULTIPLE
image/bmp
```
```
% xclip -o -selection clipboard -t STRING
xclip: Error: window id 0x1c00001 cannot convert CLIPBOARD selection to target 'STRING'
```

PR #2779, opened 2023-08-30
Release v0.9.23

metalefty:


matt335672:
One comment - I think the link you've used to the CVE in NEWS.md will die when the temporary fork is deleted. Would it be better to use this (which isn't working yet)?

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-40184

I've added the above to NEWS on the wiki for the development version.

metalefty:
It is not a private fork URL but a GitHub Security Advisory. It won't disappear even after the deletion of a private fork. But I'll use cve.mitre.org as we used to or cve.org link instead.

matt335672:
OK - thanks for the info.

PR #2782, opened 2023-09-04
Cppcheck 2.8 removed the dependency on z3

firewave:
Cppcheck removed the z3 dependency in https://github.com/danmar/cppcheck/commit/8f386e15fdedff37486c683d933ccc9a1e307388 so only perform the related steps if necessary.


matt335672:
Nice spot @firewave - thanks.

PR #2783, opened 2023-09-04
waitforx: fix build

metalefty:
```
--- waitforx ---
CCLD     waitforx
ld: error: unable to find library -lX11
ld: error: unable to find library -lXrandr
cc: error: linker command failed with exit code 1 (use -v to see invocation)
*** [waitforx] Error code 1
```


matt335672:
Quite right - thanks @metalefty

PR #2786, opened 2023-09-04
posix shm, can not unmap shmem_ptr until encoder is done with it

jsorg71:
Fix issue where main thread unmap shared memory before worker thread is done with it.

### DIFF
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -1386,11 +1386,14 @@ process_server_paint_rect_shmfd(struct mod *amod, struct stream *s)
             {
                 bmpdata = (char *)shmem_ptr;
                 bmpdata += shmem_offset;
+                /* we give up ownership of shmem_ptr
+                   will get cleaned up in server_paint_rects_ex or
+                   xrdp_mm_process_enc_done(rfx, gfx) */
                 rv = amod->server_paint_rects_ex(amod, num_drects, ldrects,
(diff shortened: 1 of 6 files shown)


xup/xup.c
```diff
@@ -1386,11 +1386,14 @@ process_server_paint_rect_shmfd(struct mod *amod, struct stream *s)
             {
                 bmpdata = (char *)shmem_ptr;
                 bmpdata += shmem_offset;
+                /* we give up ownership of shmem_ptr
+                   will get cleaned up in server_paint_rects_ex or
+                   xrdp_mm_process_enc_done(rfx, gfx) */
```
matt335672, line 1391: Nice comment - makes it clear what's going on.

matt335672:
LGTM

@Nexarian - this is more your area than mine. Amy comment?

Nexarian:
I actually just verified that this works on mainline_merge with x264! :)

Now to get it working with the helper.

jsorg71:
I think I'll merge the xorgxrdp part now.  It's good we still keep the sysV shm messages in xrdp(xup) so a newer xrdp will work with older xorgxrdp.  The other way around won't work but we could bump the header version so xorgxrdp won't compile.  We also have a version message in the API for run time check.  xup sends a version number to xorgxrdp, so far it's always been 0.0.0.1.

PR #2787, opened 2023-09-05
fixed Cppcheck `unusedVariable` warnings

firewave:
This is part of the `style` checks in Cppcheck which are not enabled in the CI. Those are currently rather produce a lot of findings which might not be of interest but these are easy and safe cleanups.


metalefty (approved):
Approved, thanks for your contribution!

PR #2794, opened 2023-09-18
Improve Unicode support

### LINKED ISSUE #942: Clipboard question
I’m seeing a weird behavior when trying to copy paste from xrdp session to windows machine.

If I copy some text in pycharm, then try to paste it into Skype for Business, I get an error like its invalid.  I think same thing with paste into outlook.  I can paste into Notepad just fine though.

I’m wondering if anyone has an idea what is happening here.  I know there are multiple clipboards but it’s not clear to me how to debug.

Maybe there’s an issue with RichText or something?

### LINKED ISSUE #2603: Clipboard can't handle emojis pasted from Windows
This was reported by @howudodat on Gitter

Pasting this string from Windows into an RDP session results in the clipboard temporarily locking up:-

Simple Test.😥

Trying to view the string with `xclip -o -selection clipboard -t STRING` or `xclip -o -selection clipboard -t UTF8_STRING` results in `xclip` hanging.

An analysis shows that the string from windows breaks down as the following UTF-16 characters in the CB_FORMAT_DATA_RESPONSE:-

|Char|Value|
|:---|:---|
|S| 0x53 |
|i| 0x69 |
|m| 0x6d |
|p| 0x70 |
|l| 0x6c |
|e| 0x65 |
| | 0x20 |
|T| 0x54 |
|e| 0x65 |
|s| 0x73 |
|t| 0x74 | 
|.| 02e |
|😥| 0xd83d 0xde25 |

This string is passed to `g_wcstombs()` :-

https://github.com/neutrinolabs/xrdp/blob/42be5762870a5720545cf79e09130358cbf91b20/sesman/chansrv/clipboard.c#L1351

which returns -1

If the string is pasted without the emoji, the paste succeeds.

Problem seems to be that the encoding of U+1F625 as the surrogate pair "0xd83d 0xde25" is a windows extension :-

https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF

A fix to this will avoid use of the `wcstombs()` and `mbstowcs()` calls

**Edit: 2023-10-20.** Surrogate pairs are NOT a Windows extension - they're part of the UTF-16 spec.

matt335672:
**Updated 2023-10-20: Ready for review**

Fixes #2603 #942 

I've been looking into the way we handle Unicode in general.

We've been using `g_mbstowcs()` and `g_wcstombs()` to handle conversions between UTF-8 and both UTF-16 and UTF-32.

These calls suffer from a number of problems when used for this purpose:-
1. The type `wchar_t` is not portable between platforms.
2. The Locale needs to be set for these functions to work correctly.
3. On the UN*X platforms I've looked at, UTF-32 is supported by these functions and we assume UTF-16 is just the bottom bits of a UTF-32 character. That works fine for characters in the [Unicode BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) but fails for other characters such as emojis which require surrogate pairs (see [#2603](/neutrinolabs/xrdp/issues/2603), possibly [#942](/neutrinolabs/xrdp/issues/942))
4. The functions are brittle - if a badly-formed or unrecognised character is found the functions just return -1 with no chance of recovery (possibly [#942](/neutrinolabs/xrdp/issues/942))

UTF-16 is used for Windows communication, and UTF-32 is used for other reasons, mostly related to font handling in the login window code.

This PR moves UTF-16 support into the Windows marshalling and unmarshalling code (common/parse.[hc]), and changes everything else to use UTF-8 mostly with UTF-32 conversions applied where needed.

New routines are provided to handle the conversions which are robust when presented with incorrect data. These routines are locale-independent. Extensive unit tests have been added for these.

The types `char32_t` and `char16_t`  are used for UTF-32 and UTF-16 characters respectively. These are back-ported from C11.

The calls `g_mbstowcs()` and `g_wcstombs()` and the type `twchar` are no longer required and have been removed. There is one remaining use of a bare `mbstowcs()` call in the [genkeymap tool](/neutrinolabs/xrdp/blob/da8fa059cdee6603fd0affe9582896eb6f6065e3/genkeymap/genkeymap.c#L152) which I haven't touched as it seems to be the best way to achieve what is required.


metalefty (approved):
Overall, LGTM. Let's fit it later if something is not working well.

matt335672:
Thanks @metalefty 

@firewave is working on #2829 at the moment which is getting there but will be quite a disruptive merge. It involves small mods to lots of files.. Let's get that one in first so I don't mess up his workflow.

firewave:
As this is ready to merge - put it in. I have no issues with rebasing/conflicts and the other PR still needs work. No need to postpone other (more substantial) changes because of it.

matt335672:
Thanks @firewave - I'll do that then. It might simplify the work I'm doing with smartcards.

PR #2803, opened 2023-09-25
Support NetBSD, OpenIndiana, Gentoo Linux, and MacOS.

mstone2001:
I use XRDP extensively on many different machines I manage, but had a few platforms where I couldn't get it to work.  So I went though the effort of getting these platforms to work, and wanted to offer the changes.  Almost all of the changes were focused on the os_calls file, and also added some additional startup configs for each platform.  I made these changes against the development branch, but I also have a working version against the last official release.  I originally had problems with shared memory and xorgxrdp, but looks like the posix changes in the current development branch fixed all of that, great work!  (Although OpenIndiana requires some build flags, will open a pull request for that ).

NetBSD:

-    g_sck_get_peer_cred:  Added NetBSDs flavor of this.
-    g_sleep:  NetBSD has a quirk where you can't usleep for more than 1,000,000 microseconds( 1 second ), so created a hybrid sleep, usleep method for netbsd.
-    service scripts are the same as for freebsd, added checks for this to install.
-    Note I used pkgsrc to install all required libraries.  Also note that building against native BSD pam libraries causes undefined symbols, so I built against openpam.  

OpenIndiana:

-    g_sck_get_peer_cred:  Added OpenIndiana's flavor of this.
-    Have to re-register for SIGCHLD after signal is handled.
-    Getting strange EINTRs, standard way of handling this is to retry on EINTR.  Modified the g_waitpid function to do the automatic retry.  This resolved a number of strange errors.  This should be a standard way to handle EINTR, so I did not #ifdef it.  I tried it on Linux platforms and did not see any issue with it.  Also modified a few calls doing a direct waitpid to call the g_waitpid instead.
-   Modified alarm registration to also handle SIGINT.  Again, believe this should not need to be #ifdef, but could be.
-   While testing, added an extra parameter to the new xwait program to allow the wait time to be configurable.
-   Build flags required -D_XOPEN_SOURCE=600 so posix calls are allowed.  Also need this for xorgxrdp.
-   Added manifest and method files for SMF for services.  
-   Note this code should work on Solaris, but I do not have access to any Solaris environments so can't test.  
-   Commented out line specifying config file when running Xorg, doesn't work on OpenIndiana.
-   One issue.  I could not get PAM to work on OpenIndiana, I had numerous problems.  It looks like enable-pam defaults to yes, and I couldn't find a way to turn it off.  So I changed enable-pam to default to no.  A second option would be to change this to --disable-pam if you want PAM to be the default, or add an explicit --enable-builtin option but I was getting tired of fighting with autoconf to get it right.  I can look at this change some more if it is an issue. 

GentooLinux:

-   Added support for OpenRC and startup files to the build.  Added --enable-openrc to install them.  

MacOS:

-    Added plist files to run xrdp and xrdp-sesman.  Note these rely on a tool available with the macports install, daemondo, which I find is really useful to get traditional services to play nicely with launchd.  But of course, user can just modify the installed files to run however they want.  All dependent libraries were installed via macports including build tools ( autoconf, automake, etc. )
-   Added install hook to comment out Xorg and Xvnc blocks automatically, since I believe the only way possible to run on MacOS is through vnc-any.  
-   Made minor enhancement to facilitate how I run it.  You have a nice gateway method where you can configure the sesman to act as a gateway for login. However, when I use this with the any-vnc configuration, I could not find a way to configure it because I need the gateway password to be the password the user entered, and the password to be the password for the VNC server.  So added the "ask" option for the gateway username and gateway password so they can use the rdp username and password.  Running this way, I don't have to rely on VNC with its low security password, and instead can do real authentication.  I can also lock down the VNC server to localhost only.  My current VNC server of choice is currently OSXvnc-server( now vine server ).  I find the performance is quite good.
-   I was getting a crash in the PAM code.  I found one issue where an object was being freed outside of a null check on it.  For some reason this only manifested on MacOS.

I believe I have summarized all of the changes.  I also added some additional logging various places while I was identifying the issues.  I can also offer to help support these in the future, since I plan to actively use xrdp on these platforms.  If you have any questions let me know.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3318,17 +3318,24 @@ g_waitpid(int pid, int *stat_loc, int options)
 #else
     int rv = 0;
 
-    again:
-        rv = waitpid(pid, stat_loc, options);
-	//retry EINTR.
-	if( rv == -1 && errno == EINTR )
-	{
-	    goto again;
-        }
+again:
+    rv = waitpid(pid, stat_loc, options);
+    //retry EINTR.
+    if( rv == -1 && errno == EINTR )
+    {
+        goto again;
+    }
 
     if( rv == -1 ) 
     {
-	LOG(LOG_LEVEL_ERROR, "waitpid returned %s", g_get_strerror());
+	if( errno == ECHILD )
+	{
+	    LOG(LOG_LEVEL_INFO, "waitpid returned %s", g_get_strerror());
+	}
+	else
+        {
+	    LOG(LOG_LEVEL_ERROR, "waitpid returned %s", g_get_strerror());
+	}
     }
 
     return rv;
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3004,6 +3004,7 @@ g_set_alarm(void (*func)(int), unsigned int secs)
     /* Cancel any previous alarm to prevent a race */
     unsigned int rv = alarm(0);
     signal(SIGALRM, func);
+    signal(SIGINT, func);
     (void)alarm(secs);
     return rv;
 #endif
@@ -3356,7 +3357,11 @@ g_waitpid_status(int pid)
         }
         else
         {
-            LOG(LOG_LEVEL_WARNING, "wait for pid %d returned unknown result", pid);
+            LOG(LOG_LEVEL_WARNING, "wait for pid %d returned unknown result %s", pid, g_get_strerror());
+	    if( errno == EINTR )
+	    {
+		return g_waitpid_status(pid);
+            }
         }
     }
 
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3278,7 +3278,7 @@ g_waitchild(struct exit_status *e)
     e->reason = E_XR_UNEXPECTED;
     e->val = 0;
 
-    rv = waitpid(-1, &wstat, WNOHANG);
+    rv = g_waitpid(-1, &wstat, WNOHANG);
 
     if (rv == -1)
     {
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -100,8 +100,11 @@ extern char **environ;
 
 /* sys/ucred.h needs to be included to use struct xucred
  * in FreeBSD and OS X. No need for other BSDs except GNU/kFreeBSD */
+/* Solaris uses __sun and needs .h */
 #if defined(__FreeBSD__) || defined(__APPLE__) || defined(__FreeBSD_kernel__)
 #include <sys/ucred.h>
+#elif defined(__sun)
+#include <ucred.h>
 #endif
 
 /* for solaris */
@@ -149,11 +152,14 @@ g_mk_socket_path(void)
                 LOG(LOG_LEVEL_ERROR,
                     "g_mk_socket_path: g_create_path(%s) failed",
                     XRDP_SOCKET_PATH);
+
+		        LOG(LOG_LEVEL_TRACE, "g_mk_socket_path() returned 1");
                 return 1;
             }
         }
         g_chmod_hex(XRDP_SOCKET_PATH, 0x1777);
     }
+
     return 0;
 }
 
@@ -591,6 +597,7 @@ g_sck_vsock_socket(void)
 int
 g_sck_get_peer_cred(int sck, int *pid, int *uid, int *gid)
 {
+    LOG_DEVEL(LOG_LEVEL_TRACE, "g_sck_get_peer_cred(%d)", sck);
 #if defined(SO_PEERCRED)
     socklen_t ucred_length;
     struct myucred
@@ -603,6 +610,7 @@ g_sck_get_peer_cred(int sck, int *pid, int *uid, int *gid)
     ucred_length = sizeof(credentials);
     if (getsockopt(sck, SOL_SOCKET, SO_PEERCRED, &credentials, &ucred_length))
     {
+        LOG_DEVEL(LOG_LEVEL_TRACE, "g_sck_get_peer_cred() returned 1");
         return 1;
     }
     if (pid != 0)
@@ -626,6 +634,8 @@ g_sck_get_peer_cred(int sck, int *pid, int *uid, int *gid)
 
     if (getsockopt(sck, SOL_LOCAL, LOCAL_PEERCRED, &xucred, &xucred_length))
     {
+        LOG(LOG_LEVEL_ERROR, "getsockopt() failed: %s", strerror(errno));
+        LOG_DEVEL(LOG_LEVEL_TRACE, "g_sck_get_peer_cred() returned 1");
         return 1;
     }
     if (pid != 0)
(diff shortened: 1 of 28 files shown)


metalefty:
This pull request is large a bit. Could you separate into per operating system?

common/os_calls.c
```diff
@@ -149,11 +149,14 @@ g_mk_socket_path(void)
                 LOG(LOG_LEVEL_ERROR,
                     "g_mk_socket_path: g_create_path(%s) failed",
                     XRDP_SOCKET_PATH);
+
+		        LOG(LOG_LEVEL_TRACE, "g_mk_socket_path() returned 1");
```
ziggythehamster, line 153: This seems to be indented inconsistently with the preceding statement

configure.ac
```diff
@@ -86,8 +90,8 @@ AC_ARG_ENABLE(tests,
               [Ensure dependencies for the tests are installed]),
               [ensure_tests_deps=yes], [])
 AC_ARG_ENABLE(pam, AS_HELP_STRING([--enable-pam],
-              [Build PAM support (default: yes)]),
-              [], [enable_pam=yes])
+              [Build PAM support (default: no)]),
+              [], [enable_pam=no])
```
ziggythehamster, line 94: I think that on Linux, this should default to yes

instfiles/launchdaemons/org.xrdp.xrdp-sesman.plist
```diff
@@ -0,0 +1,20 @@
...
+	<string>--start-cmd</string>
+	<string>/usr/local/sbin/xrdp-sesman</string>
+	<string>-n</string>
+	<string>;</string>
+	<string>--pid=exec</string>
+</array>
+<key>Disabled</key><true/>
+<key>KeepAlive</key><true/>
+</dict>
+</plist>
```
ziggythehamster, line 20: This may be a dumb question, but why use this other binary instead of just writing a plist that works without the extra dependency? That would allow xrdp to work whether installed by MacPorts or Homebrew

waitforx/waitforx.c
```diff
@@ -21,20 +21,23 @@ alarm_handler(int signal_num)
      *
      * Prefix the message with a newline in case another message
      * has been partly output */
+
+    if( signal_num == SIGALRM ) 
+    {
     const char msg[] = "\n<E>Timed out waiting for RandR outputs\n";
     g_file_write(1, msg, g_strlen(msg));
     exit(XW_STATUS_TIMED_OUT);
+    }
```
ziggythehamster, line 30:
```suggestion
    {
        const char msg[] = "\n<E>Timed out waiting for RandR outputs\n";
        g_file_write(1, msg, g_strlen(msg));
        exit(XW_STATUS_TIMED_OUT);
    }
```

xrdp/Makefile.am
```diff
@@ -110,3 +110,9 @@ dist_xrdppkgdata_DATA = \
   sans-18.fv1 \
   cursor0.cur \
   cursor1.cur
+
+if MACOS
+# must be tab below. Common out Xorg and Xvnc since they don't work on MacOs. 
+install-data-hook:
+	sed -i '' '235,249s/^/;/g' $(DESTDIR)$(sysconfdir)/xrdp/xrdp.ini
```
ziggythehamster, line 117: this breaks if the file changes in potentially minor ways

xrdp/xrdp_mm.c
```diff
@@ -2605,16 +2605,35 @@ xrdp_mm_connect_sm(struct xrdp_mm *self)
...
-                    if (!g_strcmp(gw_username, "same"))
-                    {
-                        gw_username = xrdp_mm_get_value(self, "username");
-                    }
-
-                    if (gw_password == NULL ||
-                            !g_strcmp(gw_password, "same"))
-                    {
+		            if( gw_username != NULL )
+		            {
```
ziggythehamster, line 2609: inconsistent indentation

instfiles/openrc/xrdp-sesman
```diff
@@ -0,0 +1,4 @@
+#!/sbin/openrc-run
+
+command="/usr/local/sbin/xrdp-sesman"
```
metalefty, line 3: This shouldn't be hardcoded to '/usr/local' prefix.

metalefty (changes requested):
Anyway, could you separate your PR into per operating system? Gentoo changes seems easy so it can be merged soon if you separate PR into per operating system.

matt335672:
@mstone2001 - thanks very much for this. You've clearly put a lot of effort into it.

I agree with @metalefty that a PR per operating system is necessary. This will make the reviewing a lot easier:-
- A lot of the current work on MacOS is being done by @unstabler, and he will certainly want to look at the Mac portions.
- Many years ago I was adminning a Solaris 10 network. A lot of that consisted of back-porting stuff to build on Solaris 10, so I have some expertise in this area although it is a little dated. I'll be very happy to review and even test the OpenIndiana changes, but having reviewed the [Oracle OTN license](https://www.oracle.com/downloads/licenses/solaris-cluster-express-license.html) I don't think I'll be downloading Solaris 11 as it isn't absolutely clear that this is suitable for developing open-source software.

Regarding the C source file formatting, I think all you need to do to fix this is run the `./scripts/run_astyle.sh` script from the top level.

mstone2001:
> This pull request is large a bit. Could you separate into per operating system?

Sure I can do that.  Should I just fork the repo once for each operating system and submit a pull request for each?  Or is there a better way?

configure.ac
mstone2001: I'll probably change the option to --enable-native then, which will effectively disable pam.  Since pam is the default I don't see a reason to have an --enable-pam option, correct me if I'm wrong.  Pam works everywhere except OpenIndiana, it may be possible to get it working I can take a second look after the changes are merged.

instfiles/launchdaemons/org.xrdp.xrdp-sesman.plist
mstone2001: I can test this without the extra dependency, the macports daemondo handles launchctl strangeness, but I have seen many configs that don't use it.

waitforx/waitforx.c
mstone2001: Yep, I'll have to review the indentation.

xrdp/Makefile.am
mstone2001: I can bullet proof it by using grep to look for the [Xorg] and [Xvnc] blocks to get the line numbers.

xrdp/xrdp_mm.c
mstone2001: Yep, apologies, I really messed up indentation.

instfiles/openrc/xrdp-sesman
mstone2001: Ahh I see.  Yes, I believe there are some prefix replacements available in the other scripts, I'll look at this.

mstone2001:
> Anyway, could you separate your PR into per operating system? Gentoo changes seems easy so it can be merged soon if you separate PR into per operating system.

Sure, hopefully will have some time this week to break this up.  I'll fork 4 versions and submit a pull request for each, unless you can suggest a better way.

mstone2001:
> @mstone2001 - thanks very much for this. You've clearly put a lot of effort into it.
> 
> I agree with @metalefty that a PR per operating system is necessary. This will make the reviewing a lot easier:-
> 
> * A lot of the current work on MacOS is being done by @unstabler, and he will certainly want to look at the Mac portions.
> * Many years ago I was adminning a Solaris 10 network. A lot of that consisted of back-porting stuff to build on Solaris 10, so I have some expertise in this area although it is a little dated. I'll be very happy to review and even test the OpenIndiana changes, but having reviewed the [Oracle OTN license](https://www.oracle.com/downloads/licenses/solaris-cluster-express-license.html) I don't think I'll be downloading Solaris 11 as it isn't absolutely clear that this is suitable for developing open-source software.
> 
> Regarding the C source file formatting, I think all you need to do to fix this is run the `./scripts/run_astyle.sh` script from the top level.

Thanks for that, wasn't aware you had a script.  Yes, Oracle is not the most open source friendly company, so I just wanted to note that it "should work", since OpenIndiana is just a fork of Solaris 10, and I did my changes looking at the Solaris 10 documentation.  I targeted this for OpenIndiana since that is what I use.

metalefty:
> Sure, hopefully will have some time this week to break this up. I'll fork 4 versions and submit a pull request for each, unless you can suggest a better way.

You don't need to use 4 forks.  You can create branches as many as you want in one fork.

mstone2001:
> > Sure, hopefully will have some time this week to break this up. I'll fork 4 versions and submit a pull request for each, unless you can suggest a better way.
> 
> You don't need to use 4 forks. You can create branches as many as you want in one fork.

That makes sense, I'll do that then, a branch for each OS, and separate out the changes.

metalefty:
Let me close this. Changes will be separated and raised as new PRs per operating system.

PR #2807, opened 2023-09-26
Support Gentoo Linux

mstone2001:
As requested, breaking up my push into one per platform.  This is 1 of 4.

GentooLinux:

Added support for OpenRC and startup files to the build. Added --enable-openrc to install them.
As per comments, replace path in open rc files with actual path from make.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -188,6 +188,11 @@ AC_ARG_ENABLE(rdpsndaudin, AS_HELP_STRING([--enable-rdpsndaudin],
               [], [enable_rdpsndaudin=no])
 AM_CONDITIONAL(XRDP_RDPSNDAUDIN, [test x$enable_rdpsndaudin = xyes])
 
+AC_ARG_ENABLE(openrc, AS_HELP_STRING([--enable-openrc],
+              [Use openrc in (default: no)]),
+              [], [enable_openrc=no])
+AM_CONDITIONAL(XRDP_OPENRC, [test x$enable_openrc = xyes])
+
 AC_ARG_WITH(imlib2, AS_HELP_STRING([--with-imlib2=ARG], [imlib2 library to use for non-BMP backgrounds (ARG=yes/no/<abs-path>)]),,)
 
 AC_ARG_WITH(freetype2, AS_HELP_STRING([--with-freetype2=ARG], [freetype2 library to use for rendering fonts (ARG=yes/no/<abs-path>)]),,)
@@ -574,6 +579,7 @@ AC_CONFIG_FILES([
   instfiles/pam.d/Makefile
   instfiles/pulse/Makefile
   instfiles/rc.d/Makefile
+  instfiles/openrc/Makefile
   keygen/Makefile
   waitforx/Makefile
   libipm/Makefile
@@ -628,6 +634,7 @@ echo "  ipv6only                $enable_ipv6only"
 echo "  vsock                   $enable_vsock"
 echo "  auth mechanism          $auth_mech"
 echo "  rdpsndaudin             $enable_rdpsndaudin"
+echo "  openrc                  $enable_openrc"
 echo
 echo "  with imlib2             $use_imlib2"
 echo "  with freetype2          $use_freetype2"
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -188,6 +188,11 @@ AC_ARG_ENABLE(rdpsndaudin, AS_HELP_STRING([--enable-rdpsndaudin],
               [], [enable_rdpsndaudin=no])
 AM_CONDITIONAL(XRDP_RDPSNDAUDIN, [test x$enable_rdpsndaudin = xyes])
 
+AC_ARG_ENABLE(openrc, AS_HELP_STRING([--enable-openrc],
+              [Use openrc in (default: no)]),
+              [], [enable_openrc=no])
+AM_CONDITIONAL(XRDP_OPENRC, [test x$enable_openrc = xyes])
```
metalefty, line 194: Is it possible to autodetect OpenRC like HAVE_SYSTEMD?

mstone2001: Unfortunately, they also use etc/init.d to store their configuration files, so that rules out the easiest way.   I checked my running system, and it doesn't have any systemd binaries installed, so I suppose a valid check might be just to check for systemd.  The gentoo wiki warns that systemd can show up even if you don't use it due to package dependencies.  So not 100% sure that is reliable.  I know for their packages they don't autodetect, you specify in your "use flags" whether you want systemd or not.

instfiles/openrc/xrdp
```diff
@@ -0,0 +1,4 @@
+#!/sbin/openrc-run
+
+command="__BASE__/sbin/xrdp"
+pidfile="/run/${RC_SVCNAME}.pid"
```
metalefty, line 4: PID directory might be configured by XRDP_PID_PATH. I don't recommend hardcoding this.

configure.ac
matt335672: How about checking for the presence of the `rc-status` command? Would that be sufficient?

instfiles/openrc/xrdp
matt335672: `XRDP_PID_PATH` appears to map to `${localstatedir}/run` in many Makefile.am files, so this could be achieved by expanding the `install-data-hook:`

Personally I'm not that keen on using `install-data-hook` in this way - the way the systemd files are done is cleaner to my mind. That is however outside the scope of this change.

matt335672 (commented):
No additional comments to @metalefty

instfiles/openrc/xrdp
metalefty: > Personally I'm not that keen on using install-data-hook in this way - the way the systemd files are done is cleaner to my mind. That is however outside the scope of this change.

I thought the same thing. I didn't mentioned that because that is out of scope so we're totally on the same page.

configure.ac
mstone2001: After poking around the gentoo forums, seems like trying to autodetect this is discouraged.  For gentoo, you specify when you install a package whether you want to use systemd or not, so they state you should have an option to tell it whether to use systemd or not.

instfiles/openrc/xrdp
mstone2001: After reading about these comments and reviewing the code, it seems like there are a couple of different ways the code is currently configuring itself for installs. I was following the install-data-hook pattern.  I'd volunteer to open a separate pull request to look at unifying all of the install logic to whatever you think is the best practice way to do it. Since I'm already adding some new platforms with new install files I could look at making them all common.  In any case, for the current pull request, do you want me to make the PID_PATH changes?

mstone2001: I resolved the other conversation since we were talking about similar things.  I fixed some whitespace errors in the current commit.  Let me know what you'd like me to do regarding this pull request, is it acceptable as is or do you want me to make some additional changes?

matt335672:
Hi @mstone2001,

This PR has suffered slightly in that merging it will cause some regressions in these files:-
- sesman/chansrc/smartcard.c
- sesman/chansrc/smartcard_pcsc.c
- xrdp/xrdp_font.c

Can you rebase to the latest wavefront in devel?

You may not know what I'm talking about - I didn't understand any of this when I made my first PR to this project. The concept is explained pretty well [here](https://git-scm.com/book/en/v2/Git-Branching-Rebasing), but it's a bit of a learning curve if you're used (as I was) to something like Subversion. Please feel free to come back to me with any questions.

PR #2810, opened 2023-09-29
Remove client-server shortcut paste code

### LINKED ISSUE #1839: Large clipboard pastes fail
When copying a large amount of text containing non-ascii characters locally, and then trying to paste it on the remote linux session, the paste operation will output a single character instead of the clipboard content. Since reproducing this issue requires a large amount of text, I won't include it here, but essentially copying the beginning section of [Mary Shelley's Frankenstein ](https://www.gutenberg.org/files/84/84-h/84-h.htm#letter1) (beginning of Letter 1 until the end of Letter 4) reproduces it consistently for me.

I found a commit from 2012 ([60322a3](https://github.com/neutrinolabs/xrdp/commit/60322a3a42217954aa066728fcbbea8db90539af)) that introduced code meant to split clipboard data passing between client and server in small chunks when data size was larger than 32KB. This caused some long clipboard data text blocks to be interpreted as "COMPOUND_TEXT" instead of "UTF8_STRING" by the X server. I tried handling "COMPOUNT_TEXT" correctly, but some multi-byte Unicode characters were being mangled.

I eventually just disabled the splitting code in clipboard_data_in() (the g_clip_c2s.doing_response_ss and g_clip_c2s.in_request ifs) and it worked. I don't know if there's been any change to the RDP protocol since 2012, but every client I tried (native Microsoft RDP client and FreeRDP client) already seems to split large bodies of text into chunks, so handling it on the server-side no longer seems necessary.

Let me know if you want me to provide more information or run some more tests.

matt335672:
Fixes #1839 

@jsorg71 - I'd appreciate your feedback on this PR. Are I missing something obvious?

When significant amounts of data is coming from the client in a fragmented CLIPRDR_DATA_RESPONSE PDU, this code provides a way to start copying it to a requesting client before it is all read.

The only advantage of this code is to provide a slight speedup before a paste is visible on the server.

There are significant problems with this code. Notably, it is very difficult to parse Unicode text coming through this route. Each UTF-16 character can occupy up to 4 bytes, and a fragmentation boundary could occur at any point within a UTF-16 character.

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -1489,161 +1489,6 @@ clipboard_process_clip_caps(struct stream *s, int clip_msg_status,
 }
 
 /*****************************************************************************/
-static int
-ss_part(char *data, int data_bytes)
-{
-    int index;
-    char *text;
-
-    LOG_DEVEL(LOG_LEVEL_DEBUG, "ss_part: data_bytes %d read_bytes_done %d "
-              "incr_bytes_done %d", data_bytes,
-              g_clip_c2s.read_bytes_done,
-              g_clip_c2s.incr_bytes_done);
-    /* copy to buffer */
-    if (g_clip_c2s.type == g_utf8_atom)
-    {
-        /* todo unicode */
-        text = (char *)g_malloc(data_bytes, 0);
-        index = 0;
-        data_bytes /= 2;
-        while (index < data_bytes)
-        {
-            text[index] = data[index * 2];
-            index++;
-        }
-        text[index] = 0;
-        g_memcpy(g_clip_c2s.data + g_clip_c2s.read_bytes_done, text, data_bytes);
-        g_clip_c2s.read_bytes_done += data_bytes;
-        g_free(text);
-    }
-    else
-    {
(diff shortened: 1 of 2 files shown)


matt335672:
In the absence of any feedback I'm going to merge this change as it fixes a customer problem. Happy to revert if it proves necessary.

jsorg71:
Hi Matt,  Sorry for delay,
I think you can see what this ss code was doing.  The clipboard channel chunks out clipboard data to transmit, the X11 clipboard chunks out data to transfer between X11 clients.  Hook them up so the app starts getting data right away.  Seems great except you need to know the size in the beginning and converting UTF 16 to UTF 8 you can not know the size.  I did notice you can guess the size too small but guessing too large hangs X.  We could get around that by dumping zeros at the end maybe.
I think it's ok to just remove it.  If we need something like this back, we can improve on it.

matt335672:
I got the gist of the code. The main problem I had was more related to #2603 rather than the size. Addressing #2603 means coping with UTF-16 characters being either 2 or 4 bytes (e.g. 😥 maps to 0x3d 0xd8 0x25 0xde in the byte stream). Converting that to UTF-8 on the fly in the ss code means caching up to 3 bytes potentially when we hit a fragmentation point. It's doable, but messy.

PR #2812, opened 2023-10-01
Support MacOS.

mstone2001:
This is 4 of 4 of my OS changes.  As per comments I'm going to look at removing the daemondo program and see if it will work with manilla plist files.  

MacOS:

Added plist files to run xrdp and xrdp-sesman. Note these rely on a tool available with the macports install, daemondo, which I find is really useful to get traditional services to play nicely with launchd. But of course, user can just modify the installed files to run however they want. All dependent libraries were installed via macports including build tools ( autoconf, automake, etc. )
Added install hook to comment out Xorg and Xvnc blocks automatically, since I believe the only way possible to run on MacOS is through vnc-any.
Made minor enhancement to facilitate how I run it. You have a nice gateway method where you can configure the sesman to act as a gateway for login. However, when I use this with the any-vnc configuration, I could not find a way to configure it because I need the gateway password to be the password the user entered, and the password to be the password for the VNC server. So added the "ask" option for the gateway username and gateway password so they can use the rdp username and password. Running this way, I don't have to rely on VNC with its low security password, and instead can do real authentication. I can also lock down the VNC server to localhost only. My current VNC server of choice is currently OSXvnc-server( now vine server ). I find the performance is quite good.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3365,14 +3365,16 @@ g_setpgid(int pid, int pgid)
 void
 g_clearenv(void)
 {
+    LOG_DEVEL(LOG_LEVEL_TRACE, "g_clearenv()");
 #if defined(_WIN32)
 #else
-#if defined(BSD)
+#if defined(BSD) || defined(__sun) || defined(__APPLE__)
     environ[0] = 0;
 #else
     environ = 0;
 #endif
 #endif
+    LOG_DEVEL(LOG_LEVEL_TRACE, "--g_clearenv()");
 }
 
 /*****************************************************************************/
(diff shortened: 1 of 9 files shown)


mstone2001:
So, based on the previous comments, I investigated the need for the external daemondo for the launchd plists.  And it looks like it is not necessary, so I have simplified the plist files to directly launch the xrdp and xrdp-sesman services.
I also cleaned up trailing whitespace and ran the astyle program to fix the indentation issues.

alexmyczko:
is this vnc only or will rdp also work?

mstone2001:
> is this vnc only or will rdp also work?

This is using xrdp via the vnc method.  So you run a vnc server on our Mac( I use Osxvnc, formally vine server ), and configure xrdp to use the local vnc connection.  This way, you can use Remote Desktop to connect and authenticate, and lock down the VNC server to localhost only so it's more secure.

PR #2813, opened 2023-10-02
Remove dependency on signal() function

matt335672:
Replaces uses of signal() with sigaction() which should be far more portable.

Tested on Linux and FreeBSD. See also #2811 

sigaction() has been in POSIX since 2001, so I've not felt it necessary to add any feature tests for this.

@mstone2001 - this should solve a lot of your System-V issues I hope. Please review if you have the time.

**Edit:** I've added unit tests for the signal functions now, as these will be very useful in tracking down any odd behaviour in #2811 and maybe others

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -2987,10 +2987,28 @@ g_set_alarm(void (*func)(int), unsigned int secs)
 #if defined(_WIN32)
     return 0;
 #else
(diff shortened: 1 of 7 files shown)


matt335672:
I'm merging this now, as the tests pass on Linux, FreeBSD and even OpenIndiana

PR #2815, opened 2023-10-05
Improve tsusers/tsadmins group support

### LINKED ISSUE #2806: Parameter TerminalServerUsers in sesman.ini does not work in combination with winbind AD integration
### xrdp version

0.9.12

~~~
xrdp 0.9.12
  Copyright (C) 2004-2018 Jay Sorg, Neutrino Labs, and all contributors.

  Configure options:
      --enable-ipv6
      --enable-jpeg
      --enable-fuse
      --enable-rfxcodec
      --enable-opus
      --enable-painter
      --enable-vsock
      --build=x86_64-linux-gnu
      --prefix=/usr
      --includedir=${prefix}/include
      --mandir=${prefix}/share/man
      --infodir=${prefix}/share/info
      --sysconfdir=/etc
      --localstatedir=/var
      --disable-silent-rules
      --libdir=${prefix}/lib/x86_64-linux-gnu
      --libexecdir=${prefix}/lib/x86_64-linux-gnu
      --disable-maintainer-mode
      --disable-dependency-tracking
      --with-socketdir=/run/xrdp/sockdir
      build_alias=x86_64-linux-gnu
      CFLAGS=-g -O2 -fdebug-prefix-map=/build/xrdp-GJgww4/xrdp-0.9.12=. -fstack-protector-strong -Wformat -Werror=format-security 
      LDFLAGS=-Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now -Wl,--as-needed
      CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2 
      PKG_CONFIG_PATH=/build/xrdp-GJgww4/xrdp-0.9.12/pkgconfig

  Compiled with OpenSSL 1.1.1f  31 Mar 2020
~~~

Ubuntu 20.04

### Installation method

### Which backend do you use?

xorgxrdp

Unity

Both WMs and physical machines

### What's your client?

Microsoft RDP client

### Area(s) with issue?

Authentication

### Steps to reproduce

We use winbind under Ubuntu 20 to integrate our servers into out Activate Directory. 
Our user/group names look like "DOMAIN+someName"

If I use the following parameter values in sesman.ini: ...
TerminalServerUsers=Domain+someGroup
TerminalServerAdmins=tsadmins
AlwaysGroupCheck=true

... then nonody can log in anymore.

### ✔️ Expected Behavior

It would be fine if winbind Active Directory groups would be supported by parameter TerminalServerUsers as well.

### ❌ Actual Behavior

What goes wrong:
In function g_check_user_in_group (os_calls.c), line 3535 the list groups->gr_mem is always empty in our case. Probably, winbind is not filling this list. Consequently, function g_check_user_in_group returns always ok=0.

### Anything else?

Would it be possible to use the OS function getgrouplist() which seems to work in combination with winbind as well. Here a possible implementation:

~~~C
int g_check_user_in_group(const char *username, int gid, int *ok)
{
#if defined(_WIN32)
    return 1;
#else
    struct passwd *pw;
    int numberOfGroups = 10000;
    gid_t *groups = new gid_t[numberOfGroups];
    int i;
    int exitCode = 1;

    pw = getpwnam(username);
    if (pw != 0)
    {    
      if (getgrouplist(pw->pw_name, pw->pw_gid, groups, &numberOfGroups) != -1) 
      {
        exitCode = 0;
        
        for (i = 0; i < numberOfGroups; ++i)
        {
          if (groups[i] == gid)
          {
            *ok = 1;
            break;
          }
        }
      }
    }
    
    delete [] groups;

    return exitCode;
#endif
}
~~~

matt335672:
Fixes #2806 

The first couple of commits in this PR use [getgrouplist](https://man7.org/linux/man-pages/man3/getgrouplist.3.html) (if available) to check whether a user is in a group in `g_check_user_in_group()`.

This approach is more likely to succeed where a directory service is being used for the following reasons:-
- Some directory services allow group member enumerations to be disabled for speed or security reasons.
- When using directory services a user may be referred to by more than one name. Using a name comparison to check the group member list is risky in these situations.

Although very widely available, `getgrouplist()` isn't defined by POSIX. If `getgrouplist()` is not detected, os_calls falls back to using the existing code.

`g_check_user_in_group()` now returns success if the passed-in group is the primary group of the user. That allows us to remove a couple of other checks in sesman (3rd commit).

The 4th commit reworks the group checking in sesman/sesexec, to allow primarily for much better logging. This is intended primarily for the devel branch. Commits 1-3 can be easily back-ported to V0.9 if required.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3573,42 +3573,95 @@ g_getgroup_info(const char *groupname, int *gid)
 }
 
 /*****************************************************************************/
-/* returns error */
-/* if zero is returned, then ok is set */
-/* does not work in win32 */
+#ifdef HAVE_GETGROUPLIST
+int
+g_check_user_in_group(const char *username, int gid, int *ok)
+{
+    int rv = 1;
+    struct passwd *pwd_1 = getpwnam(username);
+    if (pwd_1 != NULL)
+    {
+        // Get number of groups for user
+        //
+        // Some implementations of getgrouplist() (i.e. muslc) don't
+        // allow ngroups to be <1 on entry
+        int ngroups = 1;
+        GETGROUPS_T dummy;
+        getgrouplist(username, pwd_1->pw_gid, &dummy, &ngroups);
+
+        if (ngroups > 0) // Should always be true
+        {
+            GETGROUPS_T *grouplist;
+            grouplist = (GETGROUPS_T *)malloc(ngroups * sizeof(grouplist[0]));
+            if (grouplist != NULL)
+            {
+                // Now get the actual groups. The number of groups returned
+                // by this call is not necessarily the same as the number
+                // returned by the first call.
+                int allocgroups = ngroups;
+                getgrouplist(username, pwd_1->pw_gid, grouplist, &ngroups);
+                ngroups = MIN(ngroups, allocgroups);
+
+                rv = 0;
+                *ok = 0;
+
+                int i;
+                for (i = 0 ; i < ngroups; ++i)
+                {
+                    if (grouplist[i] == (GETGROUPS_T)gid)
+                    {
+                        *ok = 1;
+                        break;
+                    }
+                }
+                free(grouplist);
+            }
+        }
+    }
+    return rv;
+}
+/*****************************************************************************/
+#else // HAVE_GETGROUPLIST
 int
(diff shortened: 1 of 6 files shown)


metalefty:
+1 to backport since the reporter is using v0.9.

PR #2832, opened 2023-10-23
Allow init_stream() to take unsigned size arguments

matt335672:
Support using an unsigned argument for the size argument to init_stream(). This argument should not need a cast if passed an unsigned int (e.g.) from the sizeof operator.

The correct fix is probably to move the stream struct to using `unsigned int` or `size_t` rather than `int`. This is somewhat less intrusive for now.


matt335672:
On second thoughts, this is a bad idea because of possible overflows - closing.

PR #2872, opened 2023-12-11
common/arch.h: added e2k arch definition

r-a-sattarov:
Ref: https://en.wikipedia.org/wiki/Elbrus_(computer)


matt335672:
Thanks.

I found this link too which makes things a little clearer:-

https://lvee.org/en/abstracts/303

PR #2879, opened 2023-12-16
Initial rfx progressive integration

Nexarian:
- Mostly base functions and utilities necessary to enable RFX Progressive
- Add more EGFX work.
- Update encoder.
- Does not yet include caps determination to enable RFX progressive (yet).

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -71,6 +71,6 @@ libcommon_la_SOURCES = \
   $(PIXMAN_SOURCES)
 
 libcommon_la_LIBADD = \
-  -lpthread \
+  -lpthread -lrt \
   $(OPENSSL_LIBS) \
   $(DLOPEN_LIBS)
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -2114,10 +2114,14 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec *self, struct stream *s)
     in_uint16_le(s, supportedColorDepths);
     LOG_DEVEL(LOG_LEVEL_TRACE, "Received [MS-RDPBCGR] TS_UD_CS_CORE "
               "<Optional Field> supportedColorDepths %s",
-              supportedColorDepths == 0x0001 ? "RNS_UD_24BPP_SUPPORT" :
-              supportedColorDepths == 0x0002 ? "RNS_UD_16BPP_SUPPORT" :
-              supportedColorDepths == 0x0004 ? "RNS_UD_15BPP_SUPPORT" :
-              supportedColorDepths == 0x0008 ? "RNS_UD_32BPP_SUPPORT" :
+              supportedColorDepths == RNS_UD_24BPP_SUPPORT
+              ? "RNS_UD_24BPP_SUPPORT" :
+              supportedColorDepths == RNS_UD_16BPP_SUPPORT
+              ? "RNS_UD_16BPP_SUPPORT" :
+              supportedColorDepths == RNS_UD_15BPP_SUPPORT
+              ? "RNS_UD_15BPP_SUPPORT" :
+              supportedColorDepths == RNS_UD_32BPP_SUPPORT
+              ? "RNS_UD_32BPP_SUPPORT" :
               "unknown");
 
     if (!s_check_rem(s, 2))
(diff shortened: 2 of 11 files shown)


xrdp/xrdp_encoder.c
```diff
@@ -77,16 +84,24 @@ xrdp_encoder_create(struct xrdp_mm *mm)
 
     client_info = mm->wm->client_info;
 
+    /* RemoteFX 7.1 requires LAN but GFX does not */
     if (client_info->mcs_connection_type != CONNECTION_TYPE_LAN)
     {
-        return 0;
+        if ((mm->egfx_flags & 3) == 0)
```
metalefty: Where is `3` come from?

Nexarian: You know I was wondering that myself (Most of this is code's lineage comes from Jay's original egfx branch that was started many years ago).

I think it's this: In this simplified version of GFX, this flag can only ever be 1 (For H264) or 2 (For Progressive RFX). Anding with 3 means "It is neither 1 nor 2":

```
0b00 & 3 = 0
0b01 & 3 = 1
0b10 & 3 = 2
0b11 & 3 = 3
```

So basically, if it's not a LAN connection, and it's not EGFX, drop the connection because the only other use for the encoder is "non-progressive RFX" which requires LAN per the spec.

There's probably a clearer way to write this. I think egfx_flags needs to become an enum, too. Thoughts?

metalefty: I agree. Either enum or define as macro would be clearer.

Nexarian: Updated!

libxrdp/xrdp_sec.c
```diff
@@ -2133,7 +2133,15 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec *self, struct stream *s)
     {
         client_info->bpp = 32;
     }
-
+    if (earlyCapabilityFlags & 0x100) /* RNS_UD_CS_SUPPORT_DYNVC_GFX_PROTOCOL */
```
metalefty: Let's add this to ms-rdpbcgr.h.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/00f1da4a-ee9c-421a-852f-c19f92343d73

metalefty:
Logic looks good to me. Some trivial thoughts.

libxrdp/xrdp_sec.c
```diff
@@ -2133,7 +2133,15 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec *self, struct stream *s)
...
+    {
+        LOG_DEVEL(LOG_LEVEL_INFO, "client supports gfx");
```
metalefty: I think this should be LOG instead of LOG_DEVEL.

libxrdp/xrdp_sec.c
Nexarian: Updated.

libxrdp/xrdp_sec.c
Nexarian: Fixed. Do you think the "client DOES NOT support gfx" message should also be a LOG?

metalefty: Either is fine. Let's leave it as-is now.

Nexarian: So... Should I merge this then?

metalefty:
@Nexarian I have created `gfx_mainline_merge_work` branch. Let's merge into the branch now. After everything's done, let's merge `gfx_mainline_merge_work` into `devel`.

metalefty (approved):

common/Makefile.am
```diff
@@ -71,6 +71,6 @@ libcommon_la_SOURCES = \
   $(PIXMAN_SOURCES)
 
 libcommon_la_LIBADD = \
-  -lpthread \
+  -lpthread -lrt \
```
matt335672, line 74: I don't think this is needed - it seems to build OK without it. Am I missing a compile option?

xrdp/xrdp_wm.c
```diff
@@ -767,10 +767,19 @@ xrdp_wm_init(struct xrdp_wm *self)
...
 
-                /*
-                 * Skip the login box and go straight to the connection phase
-                 */
-                xrdp_wm_set_login_state(self, WMLS_START_CONNECT);
+                if (self->session->client_info->gfx && !self->mm->egfx_up)
+                {
+                    /* gfx session but have not recieved caps advertise yet,
+                       set flag so we will connect to backend later */
+                    self->mm->gfx_delay_autologin = 1;
```
matt335672, line 774: This flag appears to be set but not used.

matt335672 (commented):
Definitely worth getting this into its own branch

xrdp/xrdp_wm.c
Nexarian: Correct. It's intended for future use, just like the xrdp_egfx.c and xrdp_egfx.h files. It WILL be used once GFX is turned on.

common/Makefile.am
Nexarian: No, again, future use. Trying to get in base/fundamental changes that I know we will need in the future.

Nexarian:
> @Nexarian I have created `gfx_mainline_merge_work` branch. Let's merge into the branch now. After everything's done, let's merge `gfx_mainline_merge_work` into `devel`.

Hmm, I had a slightly different approach. I was thinking that we slowly merge the pieces that we're sure we need into `devel` in smaller chunks like this. In this way it gives us an opportunity to tune/review smaller pieces over time. We can put the actual enabling of RFX Progressive/GFX under a build flag and call it experimental until we decide to officially release it. It also avoids the entirety of the `mainline_merge` problem where it became something of a second-devel/release that many were unofficially using.

However, I'm open to doing it this way as well. Given that there are unfinished "for future use" portions in this PR, I'll merge into `gfx_mainline_merge_work`

metalefty:
> It also avoids the entirety of the mainline_merge problem where it became something of a second-devel/release that many were unofficially using.

I think it is not a big problem now because `gfx_mainline_merge_work` will be merged into `devel` in a month and we will focus on GFX for a while. So other developments are suspended and not so many other pull requests will be merged into `devel`.

common/Makefile.am
matt335672: I'm fine with that. Thanks for the clarification.

The reason I raised it was there's probably a better way to add this with autotools. I'll investigate further when we've got more dependencies to play with.

xrdp/xrdp_wm.c
matt335672: Thanks.

matt335672:
I'm happy with pausing other developments for now. #2745 is probably the biggest outstanding change I've got and that can wait.

PR #2891, opened 2023-12-28
Gfx mainline merge work

Nexarian:


xrdp/xrdp_encoder.c
```diff
@@ -70,41 +78,70 @@ xrdp_enc_data_done_destructor(void *item, void *closure)
...
+    if (self == NULL)
+    {
+        return NULL;
+    }
     self->mm = mm;
 
     if (client_info->jpeg_codec_id != 0)
     {
-        LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_encoder_create: starting jpeg codec session");
+        LOG_DEVEL(LOG_LEVEL_INFO, "xrdp_encoder_create: "
```
matt335672: General thought: would it be useful to make these LOG() rather than LOG_DEVEL()? I'm thinking ahead to triaging user problem reports regarding which codec is being loaded.

xrdp/xrdp_mm.c
```diff
@@ -79,7 +67,13 @@ xrdp_mm_create(struct xrdp_wm *owner)
...
               self->wm->client_info->h264_codec_id);
 
-    self->encoder = xrdp_encoder_create(self);
+    if ((self->wm->client_info->gfx == 0) &&
+            ((self->wm->client_info->h264_codec_id != 0) ||
+             (self->wm->client_info->jpeg_codec_id != 0) ||
+             (self->wm->client_info->rfx_codec_id != 0)))
+    {
+        self->encoder = xrdp_encoder_create(self);
+    }
```
matt335672, line 77: Do you need these tests before this call here? It looks like they're already in `xrdp_encoder_create()`

xrdp/xrdp_mm.c
```diff
@@ -2789,7 +3154,10 @@ xrdp_mm_connect_sm(struct xrdp_mm *self)
     }
 }
 
-
+#define MIN_MS_BETWEEN_FRAMES 40
+/* can not change this to zero yet, g_obj_wait in os_calls.c treats
+   everything less then 1 to mean wait forever */
+#define MIN_MS_TO_WAIT_FOR_MORE_UPDATES 1
 /*****************************************************************************/
```
matt335672, line 3282: Do you want me to put a PR in to fix this for you? Looking through the code, the only places I've found where 0 is used for an infinite wait are chansrv/chansrv.c and sesexec/sesexec.c

matt335672 (commented):
Looks OK so far to me although I certainly can't say I understand all that is going on.

The GFX stuff appears pretty spread through the code. Is that just the way it is, or once we've got something going is there some scope for modularising it a bit more?

xrdp/xrdp_mm.c
Nexarian: Sure!

xrdp/xrdp_encoder.c
Nexarian: Yeah this would be good to change.

Nexarian:
> The GFX stuff appears pretty spread through the code. Is that just the way it is, or once we've got something going is there some scope for modularizing it a bit more?

I don't think we've thought much about how to modularize it or make it better. This code is largely a lift-and-shift from Jay's original prototypes that go back to 2017/2018. So if you have suggestions they'd be welcome. I think now is the time to think about how to re-flow the code. The good news with this code is it has been pretty thoroughly tested as people have been using variations of this in `mainline_merge` for years (though that's mostly been H264 a lot of the code such as the egfx_caps code is shared)

xrdp/xrdp_mm.c
Nexarian: It looks like the only check that doesn't overlap here is the `self->wm->client_info->gfx == 0` check, which I'm not sure is actually needed... Or rather, I'm not entirely certain what it does.

xrdp/xrdp_encoder.c
Nexarian: Updated.

Nexarian:
To respond further to why GFX is spread out throughout the code, there are several main things this change targets:
- It adds the sec code to check for a GFX capable client. This seems fine to me.
- It modifies the resize-on-the-fly code to be GFX compatible. This is awkward, and is the subject of that gigantic wiki post I wrote a few years ago. It could be improved, but I need another pair of eyes on it. All I know is it's stable and it works and has for years, but the fact that you need to delete the GFX session before resizing and reconstitute it is awkward and against what the MS specification says you need to do.
- It makes some changes to the way painting is done because GFX modes require bitmap sizes of either 16x16 or 64x64 pixels, which means our dirty region detection logic can't do "only the pixels that changed." Instead, it has to do it block-wise.
- It adds a few things, like changes to order processing, that will matter for H264 but we might as well get in now.
- Due to a quirk of the way GFX is initialized, it changes the way login works with autologin. This is a necessary evil until we do a more aggressive refactoring of how [auto]login works.
- Add RFX Progressive to the encoder facilities that already exist. I added Jay's changes to make this work to librfxcodec years ago and made sure they worked with mainline_merge, so tweaking is all that's needed.
- Changes to how data is transmitted. In addition to fastpath, we now have xrdp_egfx_send_wire_to_surface1 and 2. Since those are fundamentally different, we have to break out conditional logic.
- The most awkward part is probably the changes to xrdp_mm.c -- It's not clear that the functions with the prefix `xrdp_mm_egfx` belong there. They could go in a separate file.

After reading through the code I feel it's less "random" and more "invasive." GFX changes a lot of things from how capabilities are handled to how dirty regions are detected to how encoding works to how resizing works. It just touches a lot of stuff. There probably is room to simplify it. The FreeRDP codebase did a great job of this, but it's clear they've put a lot of effort into refactoring and code cleanup beyond what's probably worthwhile to do here.

Resizing and xrdp_mm would be targets for simplification, if any.

xrdp/xrdp_mm.c
Nexarian: I read through your changes in https://github.com/neutrinolabs/xrdp/pull/2904 and merged in what looked like the relevant bits. I don't understand why GitHub is diffing and adding other changes that are already there.

matt335672: What you've done is fine. I think github is confused because of the timing of the merge of the FreeBSD CI update. I'll delete my PR.

matt335672:
@Nexarian - Thanks for taking another look at it. I think we can look at re-simplifying later on. I'll happily pick up the xrdp_mm stuff.

matt335672:
Hi @Nexarian 

I'm running your current merges as of 8th Jan. Configuring with `./configure --enable-devel-all --enable-ipv6`

I'm using an xfreerdp client to control modes:-

```
xfreerdp /gfx /u:testuser /size:1280x1024 /v:xrdp-test.test.lan
```

I'm not having any session connection problems with a first connection, but I'm seeing the green banding you've mentioned elsewhere.

When I log out, I'm getting memory access errors. I've run with valgrind, and I'm attaching logs for xrdp and valgrind for the connection.
[xrdp.log](https://github.com/neutrinolabs/xrdp/files/13860172/xrdp.log)
[valgrind.log](https://github.com/neutrinolabs/xrdp/files/13860175/valgrind.log)

Before logging out there are no reported memory access violations, which is good.

matt335672:
FWIW, the valgrind issues are caused by the main xrdp process exiting while the encoder thread is still active.

This patch stops the valgrind errors:-

```diff
--- a/xrdp/xrdp_process.c
+++ b/xrdp/xrdp_process.c
@@ -306,6 +306,8 @@ xrdp_process_main_loop(struct xrdp_process *self)
     }
     /* Run end in module */
     xrdp_process_mod_end(self);
+    xrdp_wm_delete(self->wm);
+    self->wm = NULL;
     libxrdp_exit(self->session);
     self->session = 0;
     self->status = -1;
```

Nexarian:
Thanks, I incorporated your PR! I've also noticed that resize-on-the-fly appears to be broken, so I'll dig into that tonight.

Nexarian:
As you can see from the commit, fixed it. I missed when lifting-and-shifting the code from `mainline_merge_shm` that we need to process the "memory allocation complete" event regardless of whether the helper is enabled. This means that the changes I added to xorgxrdp are also important and this PR does not stand alone.

metalefty:
I'll test it.

metalefty:
@Nexarian Does this PS includes 20 commits included in `gfx_mainline_merge_work` branch?

Nexarian:
@metalefty I'm confused. There are only 7 commits beyond devel?

metalefty:
Ah, sorry. My head was messed up due to a lack of sleep. Just ignore it. 

BTW, did you talk something about multimon with Jay?

Nexarian:
Jay's multimon work is here:
- [XRDP](https://github.com/jsorg71/xrdp/tree/workspot_multimon)
- [XORGXRDP](https://github.com/jsorg71/xorgxrdp/tree/workspot_multimon)

What I can do is work on integrating these into this branch. The only bug I'm aware of in the `gfx_mainline_merge_work` is the graphical corruption with RFX PRO that we need to hunt down, but I agree with you the next priority should be pulling in the multimon work if you and Matt can investigate the corruption.

metalefty:
Then, could you go on multimon work? I will look into the graphical corruption.

matt335672:
The corruption is going to be a tough one to find. I spent a couple of hours playing with it on Monday and got nowhere - it's certainly not related to out-of-bounds memory accesses. It might be necessary to put a test program together which illustrates the fault and look at the resulting PDUs. Anyone got any other ideas?

Nexarian:
The other idea I have is to draw the damage rects on the screen before
encoding (maybe in a red rectangle). That way we can see if it's an
encoding issue or simply a problem of sending the wrong buffer. This need
not be out of bounds but rather might just be sending the wrong part of the
screen at the wrong time.

What is telling is the corruption nearly always has "radioactive green"
which from experience I've learned is the xorg way to say "unallocated
memory" meaning it's sending data it shouldn't be.

jsorg71:
The GFX decoder does not like rects in the list that contain tiles that didn't change.  I older RemoteFX decoder was ok with that.  I don't see my patch for rdpCapture2 that fixes that in gfx_mainline_merge_work.  Also, the frame ack is used to 'lock' the shared memory.  I think some of the dynamic monitor changes violate this assumption.  For debugging, we can change xorgxrdp to always allocate a new shared memory so they can't both be using it at the same time.

Nexarian:
This patch?
https://github.com/jsorg71/xorgxrdp/commit/3d05ddd307e8b5fd9f2cd47e0596e10cb24374f4
I'll add it tonight and try it out!

jsorg71:
No I think it's this one.

```
diff --git a/module/rdpCapture.c b/module/rdpCapture.c
index 94f16aa..8ebfa66 100644
--- a/module/rdpCapture.c
+++ b/module/rdpCapture.c
@@ -1170,7 +1170,14 @@ rdpCapture2(rdpClientCon *clientCon, RegionPtr in_reg, BoxPtr *out_rects,
             rcode = rdpRegionContainsRect(in_reg, &rect);
             LLOGLN(10, ("rdpCapture2: rcode %d", rcode));

-            if (rcode != rgnOUT)
+            if (rcode == rgnOUT)
+            {
+                LLOGLN(0, ("rdpCapture2: rgnOUT"));
+                rdpRegionInit(&tile_reg, &rect, 0);
+                rdpRegionSubtract(in_reg, in_reg, &tile_reg);
+                rdpRegionUninit(&tile_reg);
+            }
+            else
             {
                 crc = crc_start();
                 if (rcode == rgnPART)```

**Comment by Nexarian** [comment]
That didn't seem to help unfortunately :/

**Comment by Nexarian** [comment]
I also tried changing the rdpCapture2 to the "wyhash" variation that I have in mainline_merge_shm, no luck either.

**Comment by matt335672** [comment]
This may be subjective, but after merging @Nexarian's last two commits, I'm finding it harder to generate green areas on the display. That's good in a way, but I've yet to come up with some way of reproducing the problem effectively which allows for testing. I've knocked up some X11 drawing programs, but these all seem to be working perfectly.

How are the rest of you reproducing the screen corruption?

**Comment by Nexarian** [comment]
Here's what I do:
- Open a terminal, drag it to the middle of the screen.
- Run `glxgears`
- Click between activating the glx window and the terminal window

It will eventually create spots within the terminal window that, if you only move the window slightly, will not disappear. This is, presumably, because of the CRC check algorithm that doesn't update tiles that aren't changing. However, on screen they're corrupt so it's very interesting. If the edges of the terminal go over those tiles, they will be corrected.

Not a very thorough set of instructions to reproduce, but it works consistently for me.

**Comment by matt335672** [comment]
With the latest gfx_mainline_merge_work commits 06e83036a5f87d9c5c5e762b2bdcf2034efb7309 and neutrinolabs/xorgxrdp@8519407d7fbcb71fd046ce42e5674c308e9ea52b I'm still getting the occasional green banding.

A couple of observations:-

1) It seems that xup doesn't need write access to the shared memory area. It would be reasonable to disable this to prevent a stray pointer in xrdp causing issues with the display:-

```patch
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -1275,7 +1275,7 @@ process_server_set_pointer_shmfd(struct mod *amod, struct stream *s)
         {
             Bpp = (bpp == 0) ? 3 : (bpp + 7) / 8;
             shmembytes = width * height * Bpp + width * height / 8;
-            if (g_file_map(fd, 1, 1, shmembytes, &shmemptr) == 0)
+            if (g_file_map(fd, 1, 0, shmembytes, &shmemptr) == 0)
             {
                 cur_data = (char *)shmemptr;
                 cur_mask = cur_data + width * height * Bpp;
@@ -1368,7 +1368,7 @@ process_server_paint_rect_shmfd(struct mod *amod, struct stream *s)
     {
         if (num_fds == 1)
         {
-            if (g_file_map(fd, 1, 1, shmem_bytes, &shmem_ptr) == 0)
+            if (g_file_map(fd, 1, 0, shmem_bytes, &shmem_ptr) == 0)
             {
                 bmpdata = (char *)shmem_ptr;
                 bmpdata += shmem_offset;
```

2) I'm wondering if this is down to syncing issues between xorgxrdp and xup. I'm running a VM which is logging to a hard drive array, and as the file is opened with `O_SYNC` xup is a bit laggy. Enabling more logging seems to make the problem worse. In fact if I add this delay, I see it a lot more:-
```patch
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -1715,6 +1715,7 @@ lib_mod_process_message(struct mod *mod, struct stream *s)
     {
         LOG_DEVEL(LOG_LEVEL_INFO,
                   "lib_mod_process_message: type 3 len %d", len);
+        g_sleep(100);
         for (index = 0; index < num_orders; index++)
         {
             phold = s->p;
```

How is the syncing supposed to work? If a `crect` (say) is sent over the link, how does `xup` know how when the client has read the `crect` data so the source can be overwritten in the server? I'm sure I'm misunderstanding something here, but I can't see what.

jsorg71:
@matt335672 Here is what I did on higher latency connection to get rid of xorgxrdp overwriting the shared memory before the encoder is done.
This breaks something in dynamic resize though.
```
diff --git a/module/rdpClientCon.c b/module/rdpClientCon.c
index 2cfebf4..ca9a285 100644
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -2753,12 +2753,12 @@ rdpScheduleDeferredUpdate(rdpClientCon *clientCon)
     uint32_t msToWait;
     uint32_t minNextUpdateTime;
 
-    if (clientCon->updateRetries > UPDATE_RETRY_TIMEOUT) {
-        LLOGLN(10, ("rdpScheduleDeferredUpdate: clientCon->updateRetries is %d"
-                    " and has exceeded the timeout of %d retries."
-                    " Overriding rect_id_ack to INT_MAX.", clientCon->updateRetries, UPDATE_RETRY_TIMEOUT));
-        clientCon->rect_id_ack = INT_MAX;
-    }
+    //if (clientCon->updateRetries > UPDATE_RETRY_TIMEOUT) {
+    //    LLOGLN(10, ("rdpScheduleDeferredUpdate: clientCon->updateRetries is %d"
+    //                " and has exceeded the timeout of %d retries."
+    //                " Overriding rect_id_ack to INT_MAX.", clientCon->updateRetries, UPDATE_RETRY_TIMEOUT));
+    //    clientCon->rect_id_ack = INT_MAX;
+    //}
 
     curTime = (uint32_t) GetTimeInMillis();
     /* use two separate delays in order to limit the update rate and wait a bit
```

Nexarian:
This conditional has been the most bug-prone part of dynamic resize and part of the reason why `mainline_merge` was still a "prototype." What can happen is that when a resize happens, it may occur in a situation where an ACK is never received. This means that the counter for frame_ack on the xorgxrdp size will be suck in a waiting forever state, an `infinite loop` of another sort. This was also the first "crash" I found when testing resize long ago. this doesn't happen in the current devel branch because there is no progressive or incremental encoding; every frame is transmitted fully so there's no drift between a key frame and an incremental one, RFX or H264 or otherwise.

So my hack was to say "if we've waited far too long, reset the ack counter and assume that a resize happened." However, how long do we wait? I set it to 200 (not 200 ms, just 200 increments), but on extremely high latency connections, that may be too short. I set the value based on empirical testing with my systems, but it apparently is only stable for low latency connections. Trying to find a sweet spot value here is a losing battle. If this reset happens at the wrong time, it breaks memory chain-of-custody and now corrupted memory is served to xrdp. This happens with H264 as well. Jay tried commenting it out, but the crashes return if we do.

My prototype solution to work around this was to say "OK, well if we have to hit this timeout, return a 'key frame'." Dynamic resize is supposed to send a few key frames after the session is restored, but that doesn't appear to be enough, you also have to send it when this condition is hit -- In theory this would work for H264, but I don't know what the equivalent of a "Key Frame" for RFX PRO is or how to tell the system to force one?

Other potential solutions include:
- Don't rely on frame_ack to maintain memory continuity. I think Jay was thinking about this.
- When a resize occurs, have an additional signal be sent to xorgxrdp to reset the frame counter. I *thought* I had tried this and it wasn't working, but I probably missed something.

Anyway, I really need another pair of eyes on this and/or thoughts.

Nexarian:
@jsorg71 Is this the way to force an RFX PRO key frame? https://github.com/neutrinolabs/librfxcodec/commit/34d051363112b16b7df511f08524f69a925939a5#diff-c3dcdcd6d2256de1b92089f68faff20bbb9e370b86680668dce9ef5eeefadd30R345

matt335672:
I can confirm that Jay's patch above fixes my remaining green screen issues.

@Nexarian :-
1) I don't think we can rely on a UNIX signal being transportable between the encoder and xorgxrdp (if that's what you meant).
2) I don't fully understand the problem here, but I can see that with the encoder running in a separate thread, there are going to be a lot of difficulties getting this right. I thought the first stage of the resize was to stop the encoder anyway. Have I got that wrong?

I need to read some code now so I understand why Jay's patch does what it does.

jsorg71:
@matt335672 I think the read only thing above is probably a good idea.
@Nexarian I don't think we need to make a big change here.
Thinking about this, we are now capturing a frame and sending it out, then starting a timer.  I think we should not start a timer at that time.  We should start a timer or do another capture when the frame ack comes in from xup.
I think that would be more efficient.

Nexarian:
Added the read-only change.

The timer analogy is interesting. The main concern I have with this is performance. If we wait until the ack to send another frame, do we slow things down substantially? From what I know of the RDP protocol, this is exactly how acks are supposed to work but I believe it's optional in the spec?

Still, worth a try. I will also try my idea about sending an RFX PRO keyframe (hopefully tomorrow).

jsorg71:
@Nexarian I looked at librfxcodec.  I have to add some flags to allow reset.  The only way you can reset it now is to delete and recreate it.
I'm with you on the timer and performance.  We make sure latency is same as now.

jsorg71:
@Nexarian Another thing we can do is have 2 shared memory buffers.   We can be encoding one when capture happens on another.
Another note about the frame acks.  Yes they are optional in the client but the xup xorg acks are always happening.  It is the way the shared memory is 'locked'

Nexarian:
Why leave it at 2? Why not simply have a running queue of captures that have happened to encode and captures that have been encoded that are being sent? Maybe we limit the size to a certain value, but that seems like a pretty good way to handle it. On resize both the encoding queue and the capture queues are cleared. In that way, we generate frames literally as fast as the system can push them. No locking needs to happen as they're already in order?

Nexarian:
Now that I think more about it after writing that, that would require an additional copy of the memory which slows us down, so maybe NOT such a good idea...

jsorg71:
There is definitely something we can do in that area.  Lets work on that after the first GFX release.
I'll create a PR for the timer changes and I think Matt was talking about the configure.ac patch, I'll include that too.
Then I'll work on adding a flag to the rfxcodec lib for keyframe.

Nexarian:
I added a KEY_FRAME_REQUESTED parameter.

Based on a novice reading of what it might take to reset the rfxcodec library I created this: https://github.com/neutrinolabs/librfxcodec/pull/60, I'm guessing there's more to it though.

jsorg71:
Mostly right but I don't think you need to clear those work buffers but you do need to clear the history tiles.  If you look at rfxcodec_encode_destroy, you need to free all those enc->rbs[x][y] and set them to nil.

matt335672:
I've been doing a bit of testing around the GFX functionality while @jsorg71 and @Nexarian are getting the kinks out of the low-level stuff.

Whichever encoder is selected, only the backends which call  `server_paint_rects()` or `server_paint_rects_ex()` will use it. The VNC encoder and NeutrinoRDP just use xrdp orders to move stuff about on the screen. However, the logging doesn't make this clear, and suggests that a GFX encoder is being used in these situations when clearly it isn't.

I'm keen on getting this made a bit clearer, as I'm generally triaging the log files submitted by users!

Is there any reason why the call to `xrdp_encoder_create()` can't be moved later when the module is loaded? That way we can call a module-specific function to see whether we need to start the encoder or not.

If you're happy with that I'll look at doing that and improving the logging in this area.

jsorg71:
@matt335672 That sounds great.  I think most of the bits left for Nex and I are in dynamic resize.

matt335672:
I've been playing around with moving the EGFX surface creation and mapping (along with the encoder) until after the login screen, when we've worked out if it's actually needed or not. The capability exchange over the virtual channel is still happening early on. This seems like it should be OK according to the spec.

It makes the logging a lot clearer for the non-GFX codepaths. I'm clearly doing something wrong however, as the GFX option isn't working - I'm just getting a blue screen with the odd bit of disturbance on it now and then.

I'll carry on looking after the weekend.

Nexarian:
`EGFX surface creation and mapping (along with the encoder) until after the login screen, when we've worked out if it's actually needed or not`

This is definitely ideal and would make a lot of things easier. I'm really glad you're thinking about this. I will say I don't think this will be particularly easy, as my understanding is that until now the login screen uses all of the pre-negotiated session parameters just like a normal session, and initializing EGFX post login is almost like changing those parameters mid-stream.

Nexarian:
@jsorg71 @matt335672 The mechanism I cooked up to request a key frame from xorgxrdp works on my systems in the case that you've shown where you introduce an artificial g_sleep in conjunction with https://github.com/neutrinolabs/librfxcodec/pull/60.

```
diff --git a/.gitmodules b/.gitmodules
index 66ec2efd..6060c53b 100644
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,8 @@
 [submodule "librfxcodec"]
        path = librfxcodec
        url = https://github.com/neutrinolabs/librfxcodec.git
-       branch = .
+       branch = gfx_mainline_merge_work
        ignore = untracked
-
 [submodule "libpainter"]
        path = libpainter
        url = https://github.com/neutrinolabs/libpainter.git
diff --git a/librfxcodec b/librfxcodec
index 25f61280..135a920a 160000
--- a/librfxcodec
+++ b/librfxcodec
@@ -1 +1 @@
-Subproject commit 25f6128026bddf7d3588f5f5466603c87ac6c0b3
+Subproject commit 135a920a0f662294af430017e3dac6a09053ff9e
diff --git a/xrdp/xrdp_encoder.c b/xrdp/xrdp_encoder.c
index c2892b76..ecf961cd 100644
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -363,6 +363,8 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
     struct rfx_tile *tiles;
     struct rfx_rect *rfxrects;
     int alloc_bytes;
+    int encode_flags;
+    int encode_passes = 0;
 
     LOG_DEVEL(LOG_LEVEL_DEBUG, "process_enc_rfx:");
     LOG_DEVEL(LOG_LEVEL_DEBUG, "process_enc_rfx: num_crects %d num_drects %d",
@@ -423,15 +425,23 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
                 }
 
                 out_data_bytes = self->max_compressed_bytes;
-                tiles_written = rfxcodec_encode(self->codec_handle,
+
+                encode_flags = 0;
@@ -363,6 +363,8 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
     struct rfx_tile *tiles;
     struct rfx_rect *rfxrects;
+++ b/librfxcodec
@@ -1 +1 @@
-Subproject commit 25f6128026bddf7d3588f5f5466603c87ac6c0b3
+Subproject commit 135a920a0f662294af430017e3dac6a09053ff9e
diff --git a/xrdp/xrdp_encoder.c b/xrdp/xrdp_encoder.c
index c2892b76..ecf961cd 100644
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -363,6 +363,8 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
     struct rfx_tile *tiles;
     struct rfx_rect *rfxrects;
     int alloc_bytes;
+    int encode_flags;
+    int encode_passes = 0;
 
     LOG_DEVEL(LOG_LEVEL_DEBUG, "process_enc_rfx:");
     LOG_DEVEL(LOG_LEVEL_DEBUG, "process_enc_rfx: num_crects %d num_drects %d",
@@ -423,15 +425,23 @@ process_enc_rfx(struct xrdp_encoder *self, XRDP_ENC_DATA *enc)
                 }
 
                 out_data_bytes = self->max_compressed_bytes;
-                tiles_written = rfxcodec_encode(self->codec_handle,
+
+                encode_flags = 0;
+                if (((int)enc->flags & KEY_FRAME_REQUESTED) && encode_passes == 0)
+                {
+                    encode_flags = RFX_FLAGS_PRO_KEY;
+                }
+                tiles_written = rfxcodec_encode_ex(self->codec_handle,
                                                 out_data + XRDP_SURCMD_PREFIX_BYTES,
                                                 &out_data_bytes, enc->data,
                                                 enc->width, enc->height,
                                                 enc->width * 4,
                                                 rfxrects, enc->num_drects,
                                                 tiles, enc->num_crects,
-                                                self->quants, self->num_quants);
+                                                self->quants, self->num_quants,
+                                                encode_flags);
             }
+            ++encode_passes;
         }
 
         LOG_DEVEL(LOG_LEVEL_DEBUG,
index 8720b005..a01b2b01 100644
--- a/xup/xup.c
+++ b/xup/xup.c
@@ -1715,6 +1715,7 @@ lib_mod_process_message(struct mod *mod, struct stream *s)
     {
         LOG_DEVEL(LOG_LEVEL_INFO,
                   "lib_mod_process_message: type 3 len %d", len);
+        g_sleep(100);
         for (index = 0; index < num_orders; index++)
         {
             phold = s->p;
```

In this way we don't have to mess with that "timer" code for stability of the session. I'm loathe to mess with the xorgxrdp workflow RIGHT NOW b/c the entire point of that code is to prevent a deadlock that is tricky to reproduce in dynamic resize. I don't want to risk re-introducing it again.

However, I am open to both approaches. We simply have two options that appear to work now.

Nexarian:
I did find another bug that is unrelated to the above. If you minimize the session with GLX gears running, the session does have green artifacts. But this happens with or without the "timer" code commented.

matt335672:
I've hit a bit of a problem with the approach I'm using for only initialising EGFX when we use it.

With FreeRDP / Remmina, I can do the caps exchange on the virtual channel when the channel comes up, and then only create and map an EGFX surface when I need it. That's all working fine now.

However with Windows 10/11, as soon as I send the caps confirm message I can't use standard drawing commands at all. It's possible I'm missing something else related to the painter - I'll carry on digging.

matt335672:
Couple of observations on b494c04508b147f465d8bba570857deaf0dd6675.
1) On a Windows 10, or Windows 11 client, setting the colour depth of the remote session to anything other than 32 results in this dialogue:-
image
   Using a debugger, I can see that the error is generated immediately the EGFX capsconfirm message is sent back. 
2) The above error results in the channel closing. This isn't a fatal error though, so we end up with an xrdp process which has to be killed manually - the error checking isn't sufficient to pick up on this at the moment.

Here's a patch for the first of these issues:-

```patch
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -2140,8 +2140,16 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec *self, struct stream *s)
     }
     if (earlyCapabilityFlags & RNS_UD_CS_SUPPORT_DYNVC_GFX_PROTOCOL)
     {
-        LOG(LOG_LEVEL_INFO, "client supports gfx protocol");
-        self->rdp_layer->client_info.gfx = 1;
+        if (client_info->bpp < 32)
+        {
+            LOG(LOG_LEVEL_WARNING,
+                "client requested gfx protocol with insufficient color depth");
+        }
+        else
+        {
+            LOG(LOG_LEVEL_INFO, "client supports gfx protocol");
+            self->rdp_layer->client_info.gfx = 1;
+        }
     }
     else
     {
diff --git a/xrdp/xrdp_mm.c b/xrdp/xrdp_mm.c
index 7802df41..d2a76b56 100644
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -1422,8 +1422,7 @@ int
 egfx_initialize(struct xrdp_mm *self)
 {
     LOG_DEVEL(LOG_LEVEL_TRACE, "egfx_initialize");
-    if (!(self->wm->client_info->mcs_early_capability_flags
-            & RNS_UD_CS_SUPPORT_DYNVC_GFX_PROTOCOL))
+    if (!(self->wm->client_info->gfx))
     {
         return 0;
     }
```

metalefty:
Good catch!

matt335672:
There's a bit more to do in this area I think. Leave it with me and I'll get a more comprehensive PR put together.

Nexarian:
@jsorg71 Here is my attempt to fix the "RFX gets corrupted if you minimize" issue. It's somewhat hacky, but I didn't see the corruption issue after adding this.

Basically: If suppress goes from 1 -> 0, request a keyframe.
https://github.com/neutrinolabs/xorgxrdp/pull/277

metalefty:
I'm testing GFX multimon at the following versions but I got "protocol error" on mstsc. Is something missed to cherry-pick?

* https://github.com/neutrinolabs/xrdp/commit/54acca43cfe85d60f8a8963fe7ac84cd293d5be2
* https://github.com/neutrinolabs/xorgxrdp/commit/b93ad93881d9cbfaca1214c1fdb46208e12a4c1e

Nexarian:
I haven't merged Jay's changes as I haven't had a chance to test it. I suppose now is the time. I'll just send it, I'm realistically not going to have much time in the next week.

Nexarian:
I merged in Jay's work and did a quick test on my M1 MacBook with 3x 4k monitors, it works! 👍

Nexarian:
Hmm, now dynamic resizing for a single monitor appears to be broken :/. We can investigate.

matt335672:
I can look at the resizing early next week.

Nexarian:
Stack trace.  Seems the mechanism to use multi monitor doesn't like when something changes via the virtual channel.

```
  #0  0x000055b99bc9887d server_egfx_cmd (xrdp + 0x1287d)
  #1  0x00007f069043d363 process_server_egfx_shmfd (libxup.so + 0x2363)
  #2  0x00007f069043d7c8 lib_mod_process_orders (libxup.so + 0x27c8)
  #3  0x00007f069043e996 lib_mod_process_message (libxup.so + 0x3996)
  #4  0x00007f0691438bc4 trans_check_wait_objs (libcommon.so.0 + 0x15bc4)
  #5  0x000055b99bc9c8ed xrdp_mm_check_wait_objs (xrdp + 0x168ed)
  #6  0x000055b99bca05ce xrdp_process_main_loop (xrdp + 0x1a5ce)
  #7  0x000055b99bc92f1b xrdp_process_run (xrdp + 0xcf1b)
  #8  0x000055b99bc9421a xrdp_listen_fork (xrdp + 0xe21a)
  #9  0x000055b99bc8b106 main (xrdp + 0x5106)
  #10 0x00007f069114e083 __libc_start_main (libc.so.6 + 0x24083)
  #11 0x000055b99bc8b90e _start (xrdp + 0x590e)
```

matt335672:
I've had a quick look at the single monitor resizing. What seems to be happening is when the encoder and the EGFX surface have been removed, xorgxrdp is sending an orders command even though output is suppressed. That's causing a SEGV due to the missing encoder. I'll look into it in more detail tomorrow - not sure where the best place to fix this is.

matt335672:
To add more detail to the stack trace:-

```
hread 1 "xrdp" received signal SIGSEGV, Segmentation fault.
server_egfx_cmd (mod=0x55a087f290a0, cmd=0x55a087f2c574 "\v", cmd_bytes=85, data=0x7f256bc5b000 "", data_bytes=3244032) at xrdp_mm.c:4144
4144	    tc_mutex_lock(mm->encoder->mutex);
(gdb) where
#0  server_egfx_cmd (mod=0x55a087f290a0, cmd=0x55a087f2c574 "\v", cmd_bytes=85, 
    data=0x7f256bc5b000 "", data_bytes=3244032) at xrdp_mm.c:4144
#1  0x00007f256da6f70b in process_server_egfx_shmfd (
    amod=amod@entry=0x55a087f290a0, s=s@entry=0x55a087f24a90) at xup.c:1279
#2  0x00007f256da710e2 in lib_mod_process_orders (mod=mod@entry=0x55a087f290a0, 
    type=type@entry=62, s=s@entry=0x55a087f24a90) at xup.c:1670
#3  0x00007f256da7179e in lib_mod_process_message (mod=mod@entry=0x55a087f290a0, 
    s=s@entry=0x55a087f24a90) at xup.c:1778
#4  0x00007f256da6980b in lib_data_in (trans=0x55a087f249d0) at xup.c:132
#5  0x00007f256db13c10 in trans_check_wait_objs (self=0x55a087f249d0)
    at trans.c:442
#6  0x00007f256da71c75 in lib_mod_check_wait_objs (mod=0x55a087f290a0)
    at xup.c:1899
#7  0x000055a087507289 in xrdp_mm_check_wait_objs (self=0x55a087e6bed0)
    at xrdp_mm.c:3668
#8  0x000055a087519f53 in xrdp_wm_check_wait_objs (self=0x55a087b983e0)
    at xrdp_wm.c:2351
#9  0x000055a08750de85 in xrdp_process_main_loop (self=self@entry=0x55a087b20480)
    at xrdp_process.c:287
#10 0x000055a0874f6067 in xrdp_process_run (in_val=in_val@entry=0x0)
    at xrdp_listen.c:152
#11 0x000055a0874f76cc in xrdp_listen_fork (self=self@entry=0x55a087b1f180, 
    server_trans=server_trans@entry=0x55a087b20850) at xrdp_listen.c:802
#12 0x000055a0874f7c81 in xrdp_listen_main_loop (self=0x55a087b1f180)
    at xrdp_listen.c:987
```

jsorg71:
Oh, sorry guys, I didn't test resize much.  I'll check why suppress is not working.

matt335672:
I can see what's happening on the resize.

The encoder is now responsible for GFX surface management - it is instructed what to do by xorgxrdp.

The resize logic isn't expecting this. The first thing is does is to stop the encoder and manually delete surfaces. When we [send the mod_server_version message](https://github.com/neutrinolabs/xrdp/blob/33539c1d8bc7a2c290cf43350314b0c24ebec778/xrdp/xrdp_mm.c#L1746), xorgxrdp sends the GFX monitor layout back. We then try to pass this to the encoder which isn't there and get the SEGV.

I think the solution is going to be to not stop the encoder as part of the GFX resize. We'll then end up with the version message being processed normally. All the surface management for the resize can be driven from xorgxrdp.

We'll probably need to solve the issue with the MacOS client as well. I've managed to get a VM running Monterey to check this out.

I'm out of time today to try things today, but I've got all day Monday to spend on this.

Nexarian:
> I think the solution is going to be to not stop the encoder as part of the GFX resize.

This might be hard. In order to change the size of the encoders, you have to delete and re-create them. This is true from what I understand for RFX, but also for the upcoming X264, NVENC, and YAMI encoders (the width and height of all of them are read-only properties set as an argument to the constructor). However, we might be able to get away with doing it in one step of the state machine, or eliminate the state machine entirely, which leads to...

> We'll probably need to solve the issue with the MacOS client as well

As I wrote [here](https://github.com/neutrinolabs/xrdp/wiki/On-EGFX-architecture-and-Dynamic-Monitors), the entire reason the state machine exists is because of the fact that I couldn't figure out how to not shut down the GFX channel and re-create it in order to have a stable resize with the Mac OS client. If that requirement goes away, then we can collapse a lot of this into a single atomic step that doesn't require a lot of "waiting until something is done" including re-creating the encoder at a different size.

Nexarian:
Another potential way to solve this that is probably too much work, is to queue up any surface commands from xorgxrdp until the state machine finishes what it needs to, then replay those commands back to the client after the encoder and the channel are back...

jsorg71:
rdpSendGfxMonitors does not look at suppress.  Maybe it should.  Anyway, just comment out the call to rdpSendGfxMonitors seems to fix it.  If surface 0 is first monitor and 1 is second one etc we should be ok.

Nexarian:
That *almost* fixes this. But the following test case is still broken with that command commented:
1. Connect with one monitor
2. [Optional] Resize the session (this works)
3. Disconnect
4. Connect with > 1 monitors
Result: The other monitors are black, the first monitor is the resolution that you left it when you disconnected.

Nexarian:
This change https://github.com/neutrinolabs/xrdp/pull/2891/commits/75b41afa04aa387acd569a691ca92e33107b78bd has the same net effect, but it simply blocks a crash if the encoder isn't there. It doesn't fix the bug I outlined earlier.

matt335672:
I've had a look at the latest xorgxrdp merges, and I'm getting an idea of what's going on. I can see that for GFX and H.264 we're sending the GFX graphics messages directly from `rdpClientConSendPaintRectShmFd()` to the encoder, which makes a certain amount of sense and is fairly readable. We only do this when the client advertises itself as GFX capable. This happens when the version message is received, so we should be able to handle connecting to the same session from both GFX and non-GFX clients.

The bit I'm confused about is why we need to create and map the surfaces from within xorgxrdp. It's going to be harder to support [[MS-RDPEGFX] 3.3.5.19](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/a720af02-0a47-44e2-a934-dc8534062b44) behaviour. This is probably going to have to be done in xrdp_mm.c, so we'll end up duplicating code between these two locations.

Can anyone help me with why this approach has been taken?

jsorg71:
Yea, I was talking with Nex.  Xorgxrdp odes not need to reset.  As long as the surface ID can be predicable for each monitor.  I'll do a PR to remove that.

metalefty:
Let's merge into devel.

PR #2904, opened 2024-01-04
Allow a zero timeout for g_obj_wait()

matt335672:
At the moment, timeouts of -1 or 0 for g_obj_wait() both mean 'no timeout', i.e. wait forever. This commit changes a timeout of 0 to mean 'return immediately'. This is more consistent with poll() and WaitForMultipleObjects(), and can be used by the new encoder logic.

There are a very few places on xrdp where a timeout of 0 is currently specified to mean 'wait forever'. These are replaced with values of -1.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,7 +4,7 @@ FreeBSD_task:
       SSL: libressl
   matrix:
     freebsd_instance:
-      image_family: freebsd-12-4
+      image_family: freebsd-13-2
   prepare_script:
     - pkg install -y $SSL git autoconf automake libtool pkgconf opus jpeg-turbo fdk-aac pixman libX11 libXfixes libXrandr nasm fusefs-libs check imlib2 freetype2 cmocka
     - git submodule update --init --recursive
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -71,6 +71,6 @@ libcommon_la_SOURCES = \
   $(PIXMAN_SOURCES)
 
 libcommon_la_LIBADD = \
-  -lpthread \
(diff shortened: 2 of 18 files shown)


matt335672:
While doing this I've discovered the timer queue processing in chansrv can be significantly improved. I'll fix that in devel at some stage

matt335672:
Issue #2906 added about chansrv timer queue mentioned above.

BTW the FreeBSD CI failures above can be fixed by using a later FreeBSD (either by rebasing `gfx_mainline_merge_work` or cherry-picking 05c78bb284764d06bf7750e37cea19c773f439ff)

metalefty:
Yes, FreeBSD 12 is EoL'd

metalefty:
@matt335672 Could you cherry-pick 05c78bb284764d06bf7750e37cea19c773f439ff to `gfx_mainline_merge_work`?

matt335672:
I've updated `gfx_mainline_merge_work` in the main repository to address the FreeBSD issue.

@Nexarian, @jsorg71 - can you update your local repositories?

Nexarian:
Will update

Nexarian:
This PR looks good to me, but for some reason I don't understand why the diff has changes in it that it shouldn't. This should just apply to the g_obj_wait changes? Right?

Nexarian:
This is what you intended to change, correct? https://github.com/neutrinolabs/xrdp/pull/2891/commits/353d76f7543f2992aec326d036fcbbaff89cb1d1

matt335672:
I'll delete this now - @Nexarian has already merged it.

PR #2909, opened 2024-01-09
Improve portability

### LINKED ISSUE #2908: xrdp compiling on Solaris, two changes required
### xrdp version

0.9.80

~~~
xrdp 0.9.80

  Configure options:
      

  Compiled with OpenSSL 1.0.2zh  30 May 2023
~~~

Solaris 11.4.62

### Installation method

git clone & make install

### Which backend do you use?

tbd

gnove

Solaris zones

### What's your client?

tbd

### Area(s) with issue?

Crashes such as segfault, Compile error

### Steps to reproduce

The first issue is the requirements for <limits.h> in os_calls.c in order to compile.
In the same file, the second issue is the method used to clear the environment is "harsh" (environ=0) and cause a core on Solaris, it is better to use clearenv which would work everywhere.

```
index 765ad472..624e0621 100644
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -35,6 +35,7 @@
 /* fix for solaris 10 with gcc 3.3.2 problem */
 #if defined(sun) || defined(__sun)
 #define ctid_t id_t
+#include <limits.h>
 #endif
 #include <unistd.h>
 #include <errno.h>
@@ -3450,10 +3451,14 @@ g_clearenv(void)
 #else
 #if defined(BSD)
     environ[0] = 0;
+#else
+#if defined(sun) || defined(__sun)
+    clearenv();
 #else
     environ = 0;
 #endif
 #endif
+#endif
 }
 
```

### ✔️ Expected Behavior

attempt to start the Xorg server

### ❌ Actual Behavior

core because of "environ=0" in g_clearenv()
[2024-01-09T09:27:46.976+0100] [ERROR] waitforx failed with unexpected signal SIGSEGV

### Anything else?

_No response_

matt335672:
Fixes #2908 

Improves portability
- Use clearenv() if it exists
- Don't rely on <limits.h> being pulled in by <sys/param.h>

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -95,15 +95,10 @@ struct sockaddr_hvs
 #endif
 
 #include "os_calls.h"
+#include "limits.h"
 #include "string_calls.h"
 #include "log.h"
 #include "xrdp_constants.h"
-
-/* for clearenv() */
-#if defined(_WIN32)
-#else
-extern char **environ;
-#endif
 
 #if defined(__linux__)
 #include <linux/unistd.h>
@@ -3446,13 +3441,15 @@ g_setpgid(int pid, int pgid)
 void
 g_clearenv(void)
 {
-#if defined(_WIN32)
-#else
-#if defined(BSD)
+#if defined(HAVE_CLEARENV)
+    clearenv();
+#elif defined(_WIN32)
+#elif defined(BSD)
+    extern char **environ;
     environ[0] = 0;
 #else
+    extern char **environ;
     environ = 0;
-#endif
 #endif
 }
 
--- a/configure.ac
+++ b/configure.ac
@@ -211,7 +211,7 @@ AM_COND_IF([DEVEL_DEBUG],
 AC_SEARCH_LIBS([setusercontext], [util])
 
 # Define HAVE_XXXXX macros for some system functions
-AC_CHECK_FUNCS([setusercontext getgrouplist])
+AC_CHECK_FUNCS([setusercontext getgrouplist clearenv])
 
 # The type used by getgrouplist() is the same type used by getgroups()
 AC_TYPE_GETGROUPS
--- a/sesman/tools/sesrun.c
+++ b/sesman/tools/sesrun.c
@@ -68,8 +68,8 @@
 #   define DEFAULT_BPP 32
 #endif
 
-#ifndef DEFAULT_TYPE
-#   define DEFAULT_TYPE "Xorg"
+#ifndef DEFAULT_SESSION_TYPE
+#   define DEFAULT_SESSION_TYPE "Xorg"
 #endif
 
 /**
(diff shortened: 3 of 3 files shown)


abrossard:
Hi, I confirm that it compiles and runs fine on Solaris 11. Thanks for the very quick response.
There are compilation warnings, but not related, so this fix is fine from my point of view.
For references:

```
ssl_calls.c: In function ‘ssl_rc4_set_key’:
ssl_calls.c:224:5: warning: ‘RC4_set_key’ is deprecated [-Wdeprecated-declarations]
  224 |     RC4_set_key((RC4_KEY *)rc4_info, len, (tui8 *)key);
...
ssl_calls.c: In function ‘ssl_rc4_crypt’:
ssl_calls.c:251:5: warning: ‘RC4’ is deprecated [-Wdeprecated-declarations]
  251 |     RC4((RC4_KEY *)rc4_info, len, (tui8 *)data, (tui8 *)data);
...
and so on for MD5_init, MD5_Update, MD5_Final, 
...
verify_user_pam.c: In function ‘common_pam_login’:
verify_user_pam.c:257:15: warning: assignment to ‘int (*)(int,  struct pam_message **, struct pam_response **, void *)’ from incompatible pointer type ‘int (*)(int,  const struct pam_message **, struct pam_response **, void *)’ [-Wincompatible-pointer-types]
  257 |     pamc.conv = verify_pam_conv;

sesrun.c: In function ‘usage’:
sesrun.c:179:50: warning: format ‘%s’ expects argument of type ‘char *’, but argument 2 has type ‘int’ [-Wformat=]
  179 |     g_printf("    -t <type>             Default:%s\n", DEFAULT_TYPE);
      |                                                 ~^
      |                                                  |
      |                                                  char *
      |                                                 %d
sesrun.c:179:49: warning: ‘%s’ directive argument is null [-Wformat-overflow=]
  179 |     g_printf("    -t <type>             Default:%s\n", DEFAULT_TYPE);

```

matt335672:
Thanks for the swift reply!

Before I merge this, let's look at your other issues.

- The openssl warning about `RC4_set_key` is interesting. We only try to call this function for OpenSSL versions under 3.0. What version of openSSL are you running on Solaris?
- The PAM one may be a concern. Are you using PAM, and does it seem to be working? There's a program called `authtest` which is built as part of the build but not distributed. See if that works. If not, I may have a fix for you.
- The `sesrun` error may cause `xrdp-sesrun --help` to segfault or display something odd for the `-t` default. This is happening as the macro `DEFAULT_TYPE` is being defined elsewhere in the Solaris includes. I've pushed another commit to rename this macro to `DEFAULT_SESSION_TYPE`.

abrossard:
For openssl, it turns out that both versions are installed, 1.02.2 and 3.09. The latter is stored under /usr/openssl/3/bin and that wasn't in my path.
The login works and so does PAM, so I wouldn't worry about that one.

matt335672:
Thanks.

I'll squash this and merge it.

matt335672:
I'm not proposing to backport this to 0.9 as there seems to be little interest in Solaris on that branch.

PR #2921, opened 2024-01-24
GFX tidy-ups

matt335672:
A number of tidy-ups to the GFX code:-

**Edit 2024-1-25 : NeutrinoRDP is now working with GFX. Comment updated**

## Fixes
- Don't enable GFX if the client doesn't support 32 bpp.

  mstsc.exe sets 'GFX available' in the early capability flags for bpp < 32, but when you try to use it, mstsc.exe fails.
- Add error checking to the `xrdp_mm_egfx_send_planar_bitmap()` codepaths

  Without this, you can end up with xrdp processes which don't exit when the client exits.
- Improved codec logging when creating an encoder thread.
- Fixed behaviour if a GFX codec cannot be chosen.

## Changes
- If using GFX, always wait for the egfx virtual channel to come up before starting the Window Manager state machine.

  This keeps the codepath for autologin the same as the non-autologin case, and makes sure the channel is up before we draw anything.
- Only start the encoder thread for xorgxrdp as other session types don't need it.

  This is a fairly significant change and encompasses 4 commits in this PR.

- Update NeutrinoRDP to not pass the "drdynvc" virtual channel through to the target.

   Dynamic virtual channels for neutrinordp have not been working since v0.9.9. This doesn't fix that, but it does prevent a target which attempts to use drdynvc from blowing up GFX.

## What's working
- GFX/RFX/no-codec session types tested with mstsc.exe and xorg and xvnc backends
- neutrinordp

## What's not working
- neutrinordp mouse buttons. This appears to be a regression in devel, so not related to GFX.

### DIFF
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -113,26 +113,26 @@ set_keyboard_overrides(struct mod *mod)
 
     if (mod->allow_client_kbd_settings)
     {
-        settings->kbd_type = mod->client_info.keyboard_type;
-        settings->kbd_subtype = mod->client_info.keyboard_subtype;
+        settings->kbd_type = mod->client_info->keyboard_type;
+        settings->kbd_subtype = mod->client_info->keyboard_subtype;
         /* Define the most common number of function keys, 12.
            because we can't get it from client. */
         settings->kbd_fn_keys = 12;
-        settings->kbd_layout = mod->client_info.keylayout;
+        settings->kbd_layout = mod->client_info->keylayout;
 
         /* Exception processing for each RDP Keyboard type */
-        if (mod->client_info.keyboard_type == 0x00)
+        if (mod->client_info->keyboard_type == 0x00)
         {
             /* 0x00000000 : Set on Server */
             LOG(LOG_LEVEL_WARNING, "keyboard_type:[0x%02x] ,Set on Server",
-                mod->client_info.keyboard_type);
+                mod->client_info->keyboard_type);
         }
-        else if (mod->client_info.keyboard_type == 0x04)
+        else if (mod->client_info->keyboard_type == 0x04)
         {
             /* 0x00000004 : IBM enhanced (101- or 102-key) keyboard */
             /* Nothing to do. */
         }
-        else if (mod->client_info.keyboard_type == 0x07)
+        else if (mod->client_info->keyboard_type == 0x07)
         {
             /* 0x00000007 : Japanese keyboard */
             /* Nothing to do. */
@@ -281,7 +281,7 @@ lxrdp_connect(struct mod *mod)
         LOG(LOG_LEVEL_ERROR, "NeutrinoRDP proxy connection: status [Failed],"
             " RDP client [%s], RDP server [%s:%d], RDP server username [%s],"
             " xrdp pamusername [%s], xrdp process id [%d]",
-            mod->client_info.client_description,
+            mod->client_info->client_description,
             mod->inst->settings->hostname,
             mod->inst->settings->port,
             mod->inst->settings->username,
@@ -294,7 +294,7 @@ lxrdp_connect(struct mod *mod)
         LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy connection: status [Success],"
             " RDP client [%s], RDP server [%s:%d], RDP server username [%s],"
             " xrdp pamusername [%s], xrdp process id [%d]",
-            mod->client_info.client_description,
+            mod->client_info->client_description,
             mod->inst->settings->hostname,
             mod->inst->settings->port,
             mod->inst->settings->username,
@@ -574,7 +574,7 @@ lxrdp_end(struct mod *mod)
     LOG(LOG_LEVEL_INFO, "NeutrinoRDP proxy connection: status [Disconnect],"
         " RDP client [%s], RDP server [%s:%d], RDP server username [%s],"
         " xrdp pamusername [%s], xrdp process id [%d]",
-        mod->client_info.client_description,
+        mod->client_info->client_description,
         mod->inst->settings->hostname,
         mod->inst->settings->port,
         mod->inst->settings->username,
@@ -636,7 +636,7 @@ lxrdp_set_param(struct mod *mod, const char *name, const char *value)
     }
     else if (g_strcmp(name, "client_info") == 0)
     {
-        g_memcpy(&(mod->client_info), value, sizeof(mod->client_info));
+        mod->client_info = (struct xrdp_client_info *)value;
         /* This is a Struct and cannot be printed in next else*/
         LOG_DEVEL(LOG_LEVEL_DEBUG, "Client_info struct ignored");
     }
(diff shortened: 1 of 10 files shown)


metalefty:
Thanks for the tidy-up! All the changes you made are significant for stability.

matt335672:
I've worked out what's happening with NeutrinoRDP.

The neutrinordp target is sending a [DVC Capabilities Request PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedyc/c07b15ae-304e-46b8-befe-39c6d95c25e0) [MS-RDPEDYC] through xrdp to the other end. This is responding with a [DVC Capabilities Response PDU](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedyc/d45cb2a6-e7bd-453e-8603-9c57600e24ce) which we are unpacking and then not forwarding to the target (via neutrinordp) but processing it ourselves. This results in a second call to `xrdp_mm_drdynvc_up()` and hence `egfx_initialize()`.

As fas as I can tell, this won't have worked properly since #1224 was merged for v0.9.9 in Oct 2018. So the simplest fix for now is probably:-
1) to not pass DRDYNVC through to the target as it won't get a capabilities response PDU since v0.9.9
2) Raise an issue on this. No users have yet done so.

Nexarian:
I'll take a look at this today. There's A LOT of code here that make what look to be significant changes to the workflow. That being said, all of this looks super important for stability and thanks for doing so!

Nexarian:
Also update the comment on the PR about "what's not working" :)

matt335672:
Header comment updated!

The only significant change to the workflow is the encoder I think.
- If GFX is available we always start a GFX surface and use that. If xup isn't loaded, `xrdp_mm_egfx_send_planar_bitmap()` is used for all updates (login screen, vnc, neutrinordp). This doesn't need an encoder.
- When xup is loaded, it starts the encoder as soon as the connection is confirmed. It can then make the module calls which use it.

Nexarian:
Something in this breaks resize-on-the-fly with RFX Progressive with the Mac OS client. There are multiple really good fixes here. I will cherry-pick and merge the ones that I've verified don't break anything.

To test resize-on-the-fly if you don't have a MacBook you can use FreeRDP. Unfortunately Microsoft put DRM protection on the .RDP files for the Azure Windows RDP client (yes, DRM for an RDP client...) so there is still no easy-to-access Windows client (other than FreeRDP) that supports resize-on-the-fly for testing.

matt335672:
That sounds good - thanks. I'll wait for you.

Once you've done the cherry-picking I'll slim these commits down, rebase them and figure out what the problem is (if it happens on FreeRDP).

Nexarian:
https://github.com/neutrinolabs/xrdp/pull/2921/commits/cf7bcf5879ae9f915b7ebbaf6719108c42662192 works but if you switch between 16 bit and 32 bit the session will still crash. Still, it's a good fix and I'll merge it in. We need to fix that bug though (it might also be in devel/release as of today though)

Nexarian:
https://github.com/neutrinolabs/xrdp/pull/2921/commits/bab04ed4124b3a3f1a5bcf0a2717ac0e21080489 seems to cause issues with resize, skipping that one.

Nexarian:
https://github.com/neutrinolabs/xrdp/pull/2921/commits/c17fb9c1b45e3135e52d7cdadeb0150ac9fe64ce worked great. Merged.

Nexarian:
https://github.com/neutrinolabs/xrdp/pull/2921/commits/3ddf5261a3b4ab196847796adf41d086c04a2976 broke resizing. Skipping.

Nexarian:
I couldn't get https://github.com/neutrinolabs/xrdp/pull/2921/commits/9725c79227d18aa0a09004f2719a353ec77a9b57 and https://github.com/neutrinolabs/xrdp/pull/2921/commits/f2466580a55be86f9dfaadfdbeeaf6a047220666 to work, it simply created a dead session.

Nexarian:
https://github.com/neutrinolabs/xrdp/pull/2921/commits/45e741257ff3b1d4ff35c436f645cafb76f6e8c4 also breaks resizing.

At this point I think it's probably appropriate to rebase and get a copy of FreeRDP to validate the rest of these changes work with resizing.

That's all I'll do for now.

matt335672:
You comment bout switching between 16 and 32 bit is interesting - this means switching between GFX and non-GFX sessions. Possibly there's some state info hanging around in xorgxrdp which causes it.

I'll rebase and reproduce the problems you mention above.

matt335672:
I'm having problems getting dynamic resizing working at all with FreeRDP.

xrdp: 7277860f2111942a867828e317a409b7705fe242
xorgxrdp : neutrinolabs/xorgxrdp@b93ad93881d9cbfaca1214c1fdb46208e12a4c1e

Command line I'm using is:-

```
xfreerdp /gfx /v:xrdp-test.test.lan /u:testuser /d: /p: /dynamic-resolution
```

With FreeRDP 3.2, when I try to resize the client I get a blank screen and this in the log:-

```
[10:44:21:341] [23017:000059ea] [WARN][com.freerdp.core.rdp] - [rdp_handle_sc_flags][0x55560861a230]: [CONNECTION_STATE_FINALIZATION_CLIENT_SYNC] unexpected server message, expected flag FINALIZE_SC_SYNCHRONIZE_PDU [0x00000001] [have NO_FLAG_SET [0x00000000]]
[10:44:21:342] [23017:000059ea] [WARN][com.freerdp.core.rdp] - [rdp_handle_sc_flags][0x55560861a230]: [CONNECTION_STATE_FINALIZATION_CLIENT_SYNC] unexpected server message, expected flag FINALIZE_SC_SYNCHRONIZE_PDU [0x00000001] [have NO_FLAG_SET [0x00000000]]
[10:44:21:442] [23017:000059ea] [WARN][com.freerdp.core.rdp] - [rdp_handle_sc_flags][0x55560861a230]: [CONNECTION_STATE_FINALIZATION_CLIENT_SYNC] unexpected server message, expected flag FINALIZE_SC_SYNCHRONIZE_PDU [0x00000001] [have NO_FLAG_SET [0x00000000]]
```

with FreeRDP 2.90 I don't get the messages, but I get a blank screen.

@Nexarian - I'll carry on investigating, but if you could post what version of FreeRDP works for you (and what the command line is) that might be useful.

matt335672:
See #2929 - I'll come back to the above when I've got a better handle on how resize is supposed to work - at the moment I'm not convinced I'm on top of the spec.

matt335672:
Rebased, but please ignore for now. I'm going to look into whether I can optimise the GFX path through the resize state machine first.

matt335672:
I'm going to close this now. Given the way resize is going to work, I think we're going to need the encoder for all GFX codepaths.

We're working a lot on client-initiated dynamic resize at the moment. The server-initiated resizes (i.e. via xrandr) also need some looking at, possibly after our first GFX drop.

If we keep the GFX instructions in xorgxrdp rather than moving them into xrdp we'll need to add corresponding code to the vnc module. This doesn't feel quite right to me, but we can look into that later.

Nexarian:
> we'll need to add corresponding code to the vnc module. This doesn't feel quite right to me, but we can look into that later.

Agreed.

PR #2925, opened 2024-01-25
Set up alternative definitions for some window messages

matt335672:
This fixes a regression introduced by #2864

FreeRDP silently redefines WM_LBUTTONUP, WM_LBUTTONDOWN, WM_RBUTTONUP and WM_RBUTTONDOWN. Consequently on my neutrinordp build at least, the right and left buttons don't work.

This is the least disruptive change. Renaming the defines totally will break xorgxrdp.

This probably needs backporting to v0.9 - I'll test that tomorrow.

### DIFF
--- a/neutrinordp/xrdp-color.c
+++ b/neutrinordp/xrdp-color.c
@@ -22,6 +22,10 @@
 #endif
 
 #include "xrdp-neutrinordp.h"
+#include "defines.h"
+#include "log.h"
+#include "os_calls.h"
+#include "string_calls.h"
 
 char *
 convert_bitmap(int in_bpp, int out_bpp, char *bmpdata,
--- a/neutrinordp/xrdp-neutrinordp.c
+++ b/neutrinordp/xrdp-neutrinordp.c
@@ -17,6 +17,17 @@
  * limitations under the License.
  */
 
+#include <freerdp/freerdp.h>
+#include <freerdp/codec/bitmap.h>
+
+// FreeRDP defines some macros that we have different values for.
+// To catch this we need to include the freerdp includes before our
+// local ones (see gcc bug #16358)
+#undef WM_LBUTTONUP
+#undef WM_LBUTTONDOWN
+#undef WM_RBUTTONUP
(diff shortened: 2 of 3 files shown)


Nexarian (approved):

metalefty (approved):

matt335672:
Thanks both.

This doesn't affect V0.9 directly as I didn't implement the user of the named constants for neutrinordp in that branch.

matt335672:
I'm going to add a note to this, as I don't understand why the definition of `WM_LBUTTONDOWN` in `/usr/local/include/freerdp/rail.h` is not triggering a macro replacement error.

Consider these files:-

```
// temp.h
#define WM_LBUTTONDOWN 0x202
```

```
// temp.c
#define WM_LBUTTONDOWN 102
#include <temp.h>

int main()
{
    return 0;
}
```

Then :-

```
$ gcc -I. -g temp.c
In file included from temp.c:4:
./temp.h:1: warning: "WM_LBUTTONDOWN" redefined
    1 | #define WM_LBUTTONDOWN 0x202
      | 
temp.c:1: note: this is the location of the previous definition
    1 | #define WM_LBUTTONDOWN 102
      | 
```

However, if I move temp.h to the compiler's default include path:-

```
sudo cp temp.h /usr/local/include
gcc -g temp.c
```

I don't get any errors, despite the macro definition still being accepted. The replacement is happening silently:-

```
gcc -E -dM temp.c | grep WM_LBUTTONDOWN
#define WM_LBUTTONDOWN 0x202
```

I assume this is some gcc specialisation - clang does the same thing so I'm assuming it's not a bug. Anyone got any thoughts?

matt335672:
Outstanding gcc bug request [#16358](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=16358) from 2004

That suggests moving the freerdp includes before the local includes will catch this. I'll rework this PR so the compiler does more of the work for us.

matt335672:
I've reworked this so the compiler now catches the macro redefinitions. Consequently it's necessary to add some `#undef` directives for the re-used macros.

PR #2933, opened 2024-02-01
Gfx mainline merge multimon1

jsorg71:
This adds the mutimon changes that include the xup message 62 for GFX message passthrough from the backend(Xorg)
Alos, the shared memory area is use a bit differently. Not you do not need a huge shared memory to cover the entire desktop. It only need to cover the monitor. This allows larger desktop with / height with RemoteFX sessions and GFX sessions.


Nexarian:
@jsorg71 While I merged this in, we need to fix dynamic monitor resizing. More detail in the main PR.

PR #2939, opened 2024-02-08
GFX: Relegate some logs to LOG_DEVEL

metalefty:
I think these logs are too much for regular use with LOG_LEVEL=DEBUG.


matt335672:
Agreed. These all seem reasonable changes to me.

PR #2940, opened 2024-02-08
Update astyle to 3.4.12

metalefty:
as astyle 3.1 is no longer available in the latest FreeBSD.

Also, astyle upstream repository has been moved to GitLab. Switched git to obtain astyle source. 
https://gitlab.com/saalen/astyle

### DIFF
--- a/scripts/install_astyle.sh
+++ b/scripts/install_astyle.sh
@@ -80,12 +80,8 @@ fi
     # Put everything in this directory
     FILESDIR=$INSTALL_ROOT/$ASTYLE_VER
 
-    git clone --sparse ${REPO_URL} "$workdir"
-    cd "$workdir"
-    git sparse-checkout set AStyle
-    git checkout tags/${ASTYLE_VER}
-
-    cd "AStyle"
+    git clone -b "${ASTYLE_VER}" --depth 1 ${REPO_URL} "$workdir"
+    cd "$workdir"/AStyle
 
     make_args="DESTDIR=$FILESDIR"
 
--- a/scripts/install_astyle.sh
+++ b/scripts/install_astyle.sh
@@ -15,8 +15,8 @@ INSTALL_ROOT=~/astyle.local
 # ----------------------------------------------------------------------------
 usage()
 {
-    echo "** Usage: $0 <svn-tags-url> <tag-name>"
-    echo "   e.g. $0 https://svn.code.sf.net/p/astyle/code/tags 3.1"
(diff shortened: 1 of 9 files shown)


metalefty:
There is a little bit difference in astyle result between 3.1 and 3.4. Commit 47d4ab1339b74089f33ae4a581b5f54d316a0a1f shows the difference. I'm not sure how to configure astyle to get the same result.

scripts/run_astyle.sh
```diff
@@ -7,7 +7,7 @@
 # Note: the script must be run from the root directory of the xrdp repository
 
 INSTALL_ROOT=~/astyle.local
-ASTYLE_FROM_XRDP=$INSTALL_ROOT/3.1/usr/bin/astyle
+ASTYLE_FROM_XRDP=$INSTALL_ROOT/3.4.12/usr/bin/astyle
 MIN_ASTYLE_VER="3.1"
```
matt335672, line 11: We should probably allow the version to be passed in to the script (like cppcheck) so we only have to set it in one place, but this is fine for now.

scripts/install_astyle.sh
```diff
@@ -80,9 +80,12 @@ fi
...
     FILESDIR=$INSTALL_ROOT/$ASTYLE_VER
 
-    svn checkout ${REPO_URL}/${ASTYLE_VER}/AStyle "$workdir"
-
+    git clone --sparse ${REPO_URL} "$workdir"
     cd "$workdir"
+    git sparse-checkout set AStyle
+    git checkout tags/${ASTYLE_VER}
+
+    cd "AStyle"
```
matt335672, line 85: Happy with this, but install_cppcheck.sh does it a different way (from https://stackoverflow.com/questions/791959) :-

```
git clone -b "$ASTYLE_VER" --depth 1 "$REPO_URL" "$workdir"
```

matt335672 (commented):
A couple of minor comments.

matt335672:
Great news that astyle is being continued. It looks like the code changes are just fixing a few strange things that astyle did anyway, so I'm happy with that.

scripts/install_astyle.sh
metalefty: It looks simpler! I like it.

scripts/run_astyle.sh
metalefty: Let's address it later.

metalefty:
Here's the astyle check log with cache deleted.
https://github.com/metalefty/xrdp/actions/runs/7829461430/job/21361981667

PR #2943, opened 2024-02-12
Remove hard-coded version from scripts/run_astyle.sh

matt335672:
This script now works the same way as cppcheck. The version to be used is specified once in the github CI action.

One subtlety I found is that `echo -e` isn't a recognised option in dash. I've replaced this line:-

```sh
LOWEST_VERSION=`echo -e "$MIN_ASTYLE_VER\n$ASTYLE_FROM_OS_VER_OUTPUT" | sort -V | head -n1`
```

with this one:-

```sh
LOWEST_VERSION=`{ echo "$MIN_ASTYLE_VER" ; echo "$ASTYLE_VER_OUTPUT"; } | sort -V | head -n1`
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -213,6 +213,6 @@ jobs:
       - run: sudo scripts/install_astyle_dependencies_with_apt.sh
       - run: scripts/install_astyle.sh $ASTYLE_REPO $ASTYLE_VER
       - name: Format code with astyle
-        run: scripts/run_astyle.sh
+        run: scripts/run_astyle.sh -v $ASTYLE_VER
       - name: Check code formatting
         run: git diff --exit-code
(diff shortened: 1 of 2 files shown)


rowlap:
FWIW this is flagged by shellcheck; I'm not up to speed on github automation these days, but it should be possible to catch with https://github.com/marketplace/actions/shellcheck#run-shellcheck-only-in-a-single-directory

```
In run_astyle.sh line 37:
    LOWEST_VERSION=`echo -e "$MIN_ASTYLE_VER\n$ASTYLE_FROM_OS_VER_OUTPUT" | sort -V | head -n1`
                         ^-- SC3037 (warning): In POSIX sh, echo flags are undefined.
```

matt335672:
@rowlap - thanks. That' useful to know.

Issue #2423, opened 2022-11-12
Scroll is not working with VNC backend after #2364


metalefty:
I don't see any scroll events when using VNC after #2364.  I'll raise a PR for this soon. 

@seflerZ I'd appreciate it if you review my PR later.

## VNC
https://user-images.githubusercontent.com/941609/201479177-8b3b2289-0243-4e7d-a6dd-930b0769f0d8.mp4

## xorgxrdp

https://user-images.githubusercontent.com/941609/201479768-4ea8ba88-478e-4d6a-b01d-7759a8d222f0.mp4

matt335672:
I can confirm this.

I've checked out these commits:-
- d80d096e before I merged #2364. I get events for buttons 4 and 5 in xev using VNC.
- e1697879 - the merge commit for #2364. I don't get the same events for buttons 4 and 5.

Nice find @metalefty

metalefty:
Raised #2424 for this.

metalefty:
Solved by #2425.

PR #2948, opened 2024-02-16
[v0.9] Backport touchpad intertial scrolling

metalefty:
Resolves: https://github.com/neutrinolabs/xorgxrdp/issues/265

v0.9 had only a workaround for too fast scroll issue by environment variable `XRDP_XORG_TOUCHPAD_SCROLL_HACK=yes` however the hack is not working on some environments. I think we can backport inertial scrolling as a fundamental solution now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ xrdp/xrdp
 xrdp/xrdp.ini
 xrdp_configure_options.h
 xrdpapi/xrdp-xrdpapi-simple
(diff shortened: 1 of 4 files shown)


matt335672:
For the record, the original PR with commentary is at #2364

Do we also need the relevant bits of #2423 / #2424 / #2425 ?

metalefty:
Definitely, thank you!

PR #2954, opened 2024-02-21
GFX: Relegate some logs to LOG_LEVEL_DEBUG

metalefty:
Related to: #2939
Backport to: v0.10

Per-frame logs are too much.

PR #2957, opened 2024-02-21
[v0.10] cherry picks

metalefty:
Backport #2954, #2955, #2956, #2962, #2967, #2970, #2972.

Other additional pull requests might be cherry-picked later. I'll keep this open for a while.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,6 +2,7 @@
 
 AC_PREREQ([2.69])
 AC_INIT([xrdp], [0.9.90], [xrdp-devel@googlegroups.com])
+AC_DEFINE([VERSION_YEAR], 2024, [Copyright year])
 AC_CONFIG_HEADERS(config_ac.h:config_ac-h.in)
 AM_INIT_AUTOMAKE([1.7.2 foreign])
 AC_CONFIG_MACRO_DIR([m4])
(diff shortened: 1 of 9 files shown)


metalefty:
It might be better to use `git cherry-pick -x` for cherry-picking. 

```
       -x
           When recording the commit, append a line that says "(cherry picked
           from commit ...)" to the original commit message in order to
           indicate which commit this change was cherry-picked from. This is
           done only for cherry picks without conflicts. Do not use this
           option if you are cherry-picking from your private branch because
           the information is useless to the recipient. If on the other hand
           you are cherry-picking between two publicly visible branches (e.g.
           backporting a fix to a maintenance branch for an older release from
           a development branch), adding this information can be useful.
```

matt335672:
Can you add #2955 ? It will be needed for any CI when Node 16 gets turned off in a couple of months?

I was just about to add individual PRs for these but I can't see any advantage to that.

I didn't know about `cheey-pick -x`. Thanks for commenting on that.

metalefty:
Added!

metalefty:
I learned `cherry-pick -x` from FreeBSD comitter's guide. It is used in MFC (Merge from current), very similar workflow to our backporting.

https://docs.freebsd.org/en/articles/committers-guide/#mfc-with-git

configure.ac
```diff
@@ -1,7 +1,9 @@
 # Process this file with autoconf to produce a configure script
 
 AC_PREREQ([2.69])
-AC_INIT([xrdp], [0.9.90], [xrdp-devel@googlegroups.com])
+=======
+AC_INIT([xrdp], [0.10.80], [xrdp-devel@googlegroups.com])
```
metalefty: Oops, this should be 0.9.90.

metalefty:
Merging this.

PR #2962, opened 2024-02-22
GFX: Fix disconnect on resize of busy windows

matt335672:
When a resize is underway on a busy X server, it is possible for a queued EGFX cmd (order #62) to be processed after the decoder has been deleted. This causes a client disconnect with no useful error message.

### DIFF
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -4150,11 +4150,13 @@ server_egfx_cmd(struct xrdp_mod *mod,
     mm = wm->mm;
     if (mm->encoder == NULL)
     {
+        // This can happen when we are in the resize state machine, if
+        // there are messages queued up by the X server
         if (data != NULL)
         {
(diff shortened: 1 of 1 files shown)


Nexarian (approved):

matt335672:
@metalefty - can you cherry-pick this one into #2957 please?

metalefty:
Sure, the user might have encountered this issue during the acceptance test.  I have not yet determined that the problem they encountered is the same as this problem because I couldn't reproduce the issue on my side but the symptom sounds very similar to this.

Nexarian:
I'm glad you've learned to use GLXGears to test resizing/monitor changes. Because it's such a simple app, it "spams" frames and is very good at finding timing problems.

My other stress test I do is resize while playing a YouTube video (pick your favorite, I usually do music videos such as Epic Rap Battles of History), and if the video (after a few glitchy frames) recovers it's good. However, if it breaks, usually you resize into a black-screen-of-death

PR #2969, opened 2024-02-27
Increase GFX output buffer size

### LINKED ISSUE #2968: GFX connection reliably crashes with SIGABRT
This issue is likely to be identical to the third issue I reported in https://github.com/neutrinolabs/xrdp/pull/2942#issuecomment-1948904733.

Testing the latest v0.10 branch c3cb8554b61dd5f3fb03b47c694ab03ab65b2e78 on AlmaLinux 9.3.

Here's a simple program to reproduce: https://gist.github.com/metalefty/99b688ec67e5debb4d662fb6a12fc069

1. Connect to xrdp session from 4k monitor (I tested 3840x2160)
2. Start terminal emulator and maximize the window
3. Run the program to reproduce 
    * `./a.out 200 300`
    * adjust cols arguments to fit your window, and set rows to some big value for example 200.
5. xrdp process crashes with SIGABRT when the output of the program drawn

```
[2024-02-26T04:01:33.731+0000] [ERROR] Child 194657 terminated unexpectedly with signal SIGABRT
```

When trying to reconnect, xrdp process crashes SIGABRT as soon as reconnected because the same screen will be drawn again. If resize happens when reconnecting (reconnect from a different size), it gets back working. As far as I tested, this issue occurs when the session size is 4k. I can't reproduce this issue with 2560x1600 geometry. It is a smaller size than 3840x2160. It doesn't crash when the cols are less than 1/2 of the total screen width.

paint

On FreeBSD, xrdp process doesn't crash but I get a black blank screen instead. As same as AlmaLinux, it gets back to working when I make resizing happen (moving mstsc to other monitors). It is something related to the GFX encoder.

I also tried v0.10 + cherry-pick #2962 but it doesn't affect this issue.

@Nexarian I remember you have a 4k monitor. Can you confirm if you can reproduce this issue?

matt335672:
Fixes #2968 

Size the GFX output buffer pessimistically based on the largest monitor in the configuration.

A couple of comments:-
1) The sizing is really pessimistic. It assumes that no compression happens to the incoming data at all.
2) This may not be needed - see [this comment](https://github.com/neutrinolabs/xrdp/issues/2968#issuecomment-1966194501) from @jsorg71 

Because of 2) above, this is still in draft.

### DIFF
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -86,6 +86,40 @@ xrdp_enc_data_done_destructor(void *item, void *closure)
 }
 
 /*****************************************************************************/
+static unsigned int
+get_largest_monitor_pixels(struct xrdp_mm *mm)
+{
+    unsigned int max_pixels;
+
+    struct xrdp_client_info *client_info = mm->wm->client_info;
+    struct display_size_description *display_sizes;
+    display_sizes = &client_info->display_sizes;
+
+    if (display_sizes->monitorCount < 1)
+    {
+        max_pixels = display_sizes->session_width *
+                     display_sizes->session_height;
+    }
+    else
+    {
+        max_pixels = 0;
+        struct monitor_info *minfo = display_sizes->minfo;
+        unsigned int i;
+        for (i = 0 ; i < display_sizes->monitorCount; ++i)
+        {
+            unsigned int pixels = (minfo[i].right + 1) - minfo[i].left;
+            pixels *= (minfo[i].bottom + 1) - minfo[i].top;
+            if (pixels > max_pixels)
+            {
+                max_pixels = pixels;
+            }
+        }
+    }
+
+    return max_pixels;
+}
+
+/*****************************************************************************/
 struct xrdp_encoder *
 xrdp_encoder_create(struct xrdp_mm *mm)
 {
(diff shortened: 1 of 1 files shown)


metalefty:
The patch 1) works great for me.

metalefty:
Can this be merged?  This is needed for v0.10 to continue the user acceptance test. 
Let's merge another PR to devel if this is not needed (case 2).

matt335672:
If it's needed for user acceptance testing, I think we should merge it as you suggest.

Nexarian:
I think @jsorg71 was investigating an anomaly that this shouldn't be necessary; but if it fixes a user facing issue then we should still merge it. No objections here.

jsorg71:
@matt335672 I think we still need this fix.  I just want to figure out why it crashed instead of just returning a few less tiles.
I did try to reproduce this crash and I can not.  I don't have a 4k screen but I set the max buffer to 1000 and I got one tile back which is expected.  Anyway lets not hold this up.

metalefty:
Thanks, merging and will cherry-pick to v0.10.

matt335672:
Agreed that this needs more investigation. I've re-opened the original fault report which was closed automatically.

@jsorg71 - I don't have a 4k monitor either, but I managed to frig something with the console on a windows VM which was bigger than my virtual monitor. Also, I think that you can probably just edit a mstsc ini file which isn't full sceen, and just put the larger dimensions in. I don't have access to an environment to check this today though.

jsorg71:
Thanks Matt, I can repro and I found the issue.  I'll test a couple of days, then do a PR.

matt335672:
That's great news @jsorg71.

Can you undo 90e4aca26a7f256b93a31168da3a2682d0ca22df as part of the PR? I was never really happy with it as anything other than a short-term fix.

PR #2972, opened 2024-02-29
Suppress some logs when no errors

metalefty:
This outputs per-frame logs when using VNC backend and on login screen.

Backport to v0.10.


matt335672:
LGTM

PR #2973, opened 2024-02-29
GFX: selectable lossy compression levels

metalefty:
Reboot of #2539. This feature is also requested by @unstabler and another enterprise user.

Classic RemoteFX [requires connectionType==LAN](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdprfx/516f1a1c-20ec-4ea1-b92a-7995d8a322cf) so we can't use the parameter to adjust compression level. In contrast, GFX can be used whatever value connectionType is set to. Therefore, we can use this parameter to tell the encoder compression level.

This is still a proof of concept so this PR is a draft.

### Ultra Low Quality
image

### Low Quality
image

### Standard Quality
image

### DIFF
--- a/xrdp/xrdp_encoder.c
+++ b/xrdp/xrdp_encoder.c
@@ -37,11 +37,30 @@
 #define OUT_DATA_BYTES_DEFAULT_SIZE (16 * 1024 * 1024)
 
 #ifdef XRDP_RFXCODEC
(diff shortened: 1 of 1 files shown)


metalefty:
There is room to adjust quantization values but this one can be merged.

jsorg71:
This looks good to me, maybe a rebase first.
As an expansion later maybe we can read these from xrdp.ini.

PR #2976, opened 2024-03-05
Remove duplicate DEBUG output

rowlap:
allow_multimon and new_cursors are printed twice.

I left the lines which make most sense with their surroundings, but feel free to switch to the other option.


metalefty:
Good catch!

metalefty:
Backport to v0.9 and v0.10.

PR #2990, opened 2024-03-11
Detect missing drdynvc

### LINKED ISSUE #2984: GFX hangs without channel drdynvc=true
This issue is reported by an enterprise user during the user acceptance test. Not a blocking issue to release v0.10.0. It is fine tofix this in the next minor release such as v0.10.1, v0.10.2.

`drdynvc` channel is necessary to get GFX working however the channel can be disabled in `xrdp.ini`. The problem is that without `drdynvc=true`, xrdp shows a blank screen instead of the login prompt, and spins at 100% CPU calling poll().

We need to define behaviour when `drdynvc` turned off.

1. Output log at LOG_LEVEL_WARNING and fall back to former codec
2. Output log at LOG_LEVEL_ERROR and exit xrdp daemon
3. Other

@matt335672 What do you think?

matt335672:
Fixes #2984 

This was slightly more complicated than I'd hoped.

Biggest problem was that the `[Channels]` section in `xrdp.ini` wasn't scanned until `xrdp_wm_init()`, by which time we'd tried to start GFX.

Solution is to scan the `[Channels]` earlier, and disable GFX if for any reason the static channel `drdynvc` can't be started. 'Any reason' here includes the channel being disabled.

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -1408,6 +1408,23 @@ libxrdp_disable_channel(struct xrdp_session *session, int channel_id,
 
 /*****************************************************************************/
 int
+libxrdp_drdynvc_start(struct xrdp_session *session)
+{
+    struct xrdp_rdp *rdp;
+    struct xrdp_sec *sec;
+    struct xrdp_channel *chan;
+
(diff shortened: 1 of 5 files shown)


Nexarian (approved):
LGTM. Thanks for the refactoring.

metalefty (approved):
LGTM, too. Thanks!

PR #3013, opened 2024-03-22
gfx send multiple wire to surface messages when compressed data is la…

jsorg71:
…rger than max_compressed_bytes


matt335672 (approved):
I'm happy with this.

I've tested with @metalefty's test program in #2968 in a single (emulated) 4K monitor. I was unable to crash it.

I also re-tested running under valgrind, and got no memory violations.

matt335672:
I'm going to merge this into `devel`.

@metalefty - it's up to you whether you want to back-port this to v0.10.x, or leave that one with the version that's undergone user-acceptance testing.

Issue #3010, opened 2024-03-21
Triage before v0.10.0 release


metalefty:
There are several ongoing PRs. Let's triage which to include in v0.10.0 official release. 

* #3009
    * #2968
* #2910 
* #2959
* #2943

matt335672:
Personally I'm happy with all of them except #2968. @jsorg71 is looking at fixing this properly and if this is done we don't need that one.

There are a couple of other things I'd like to add for consideration:-
1) Commit 44710129317744d32d90199c2ab890b9265da4b2 which is currently part of #2974. I'm happy to leave the rest of that PR for now, but this commit will make it easier for Debian to support their own non-privileged user code for v0.10. I can factor this into a separate PR.
2) I've made some comments on commit ccead296e615b92afc17daf67447876fb3c122e0 regarding a tidy-up. Again, I can put a PR together for this.

Thoughts?

metalefty:
Sounds good. Could you raise a separate PR?
Then merge these PRs into devel, cherry-pick to v0.10 next.

matt335672:
Progress report before the weekend.

I've created #3011 relating to the unprivileged user and #3012 relating to the SECCOMP filter.

The following are merged into devel:-
- #2910
- #2959
- #2943
- #3009 
- #3011

I'm waiting for Nex to comment on #3012 as this is a relatively high-risk area and I want to be sure I've got it right.

When I've got approval from Nex I'll backport all of the above to v0.10

We've merged a fix for #2968 on #2969. This uses a lot of memory for big screens, but works. Jay is working on improving this, so when this lands (in devel?) we can consider a back-port to v0.10.

metalefty:
Thanks! Let's wait for Nex.

PR #3017, opened 2024-03-27
Cherry picks to v0 10

matt335672:
See #3010


metalefty:
@matt335672 Could you add #3013?

matt335672:
I've updated NEWS 0.10 for this, so I'll merge it now in preparation for the next beta.

PR #3031, opened 2024-04-18
Merge updates from librfxcodec

matt335672:
1) add SSE2 simd for dwt_shift_rem and diff_count

   make dwt_shift_rem easier to read
   move common rlgr defines to common file
   move common dwt defines to common file

2) Fix 'make distcheck'


matt335672:
Note: As I submit this, it looks like ppa.launchpad.net is down so the CI is failing.

PR #3034, opened 2024-04-19
Release v0.10.0-beta.3

metalefty:


matt335672 (approved):

metalefty:
Note: The difference in update_news.sh script from xorgxrdp is xrdp has submodule. The script works properly even when the current directory is submodule dir.

PR #3055, opened 2024-05-01
Add ibus unicode input support

### LINKED ISSUE #1990: Support Unicode Keyboard events in xrdp
xrdp: A Remote Desktop Protocol server.
Copyright (C) Jay Sorg 2004-2014
See http://www.xrdp.org for more information.
Version 0.9.5

In the same environment of Chinese and related dependencies, vnc is a device that can bring external Chinese into the remote ubuntu 18.04 system

I currently use the input method in ubuntu, but I hope that the Chinese from the external client can be brought into the ubuntu system

-----------------------------------------------------
apt install -y language-pack-zh-hans   fonts-droid-fallback ttf-wqy-zenhei ttf-wqy-microhei fonts-arphic-ukai fonts-arphic-ukai fonts-arphic-uming

seflerZ:
Close #1990 . This PR was made by me and @matt335672. It introduced serveral unicode input interfaces and has the ibus interface implemented (That means you should have ibus configured in the server side first).

The unicode input is a kind of method of sending unicode codes directly from the client to the server. With it, you can input any characters or glyphs (like Emoji 😁) even if they are not present in the keymap of the keyboard. It is typically used by mobile devices like Android\iOS phones, especially in Asia languages.

Another benifit of the unicode input is you can use your device's voice input to generate a bunch of texts quickly. Although we only have the ibus implemented now, it is easy to do for other input services like fcitx.


matt335672:
Hi @seflerZ 

The CI  checks are failing because the project merged #2829 recently which adds the `-Wmissing-prototypes` to the compiler. Your branch is currently based in an earlier part of `devel` which doesn't have this check.

There's a patch attached which fixes this. It does the following:-
1) [Typo] Replace `xrdp_input_unicode_destory()` with `xrdp_input_unicode_destroy()`
2) Uses '(void)' in declarations/definitions of functions with no arguments. '()' is not recommended for modern C code, although it's fine in C++.
3) Marks functions which do not need external linkage with 'static'

Suggest you rebase on recent `devel`. This will bring #2829 into your environment. You can then add the attached patch or make your own. I've got a few extra comments to add, but those can wait

[patch.txt](https://github.com/neutrinolabs/xrdp/files/15185756/patch.txt)

matt335672:
Thanks @seflerZ 

Can you rebase this on the latest devel? It will make it easier to read and review, and make the Git history read a little better when we merge this.

Let me know if you need a hand.

seflerZ:
@matt335672 Sorry, I made the merge from devel. I've rebased it.

PR #3073, opened 2024-05-14
add GFX h264 software via --enable-x264

jsorg71:


xrdp/xrdp_encoder_x264.c
```diff
@@ -0,0 +1,242 @@
...
+xrdp_encoder_x264_create(void)
+{
+    struct x264_global *xg;
+
+    LOG_DEVEL(LOG_LEVEL_TRACE, "xrdp_encoder_x264_create:");
+    xg = g_new0(struct x264_global, 1);
+    if (xg == NULL)
+    {
+        return NULL;
+    }
```
matt335672: Above 4 lines (and indeed, the xg variable) not needed.

common/xrdp_client_info.h
```diff
@@ -170,7 +180,7 @@ struct xrdp_client_info
     int mcs_early_capability_flags;
 
     int max_fastpath_frag_bytes;
-    int capture_code;
+    int old_capture_code;
```
matt335672: Is this now just here for offset purposes? If so, suggest we call it something like `to_be_reused` so that it doesn't look like it still has a function. There's probably a better name than that.

matt335672 (commented):
Couple of minor observations.

common/xrdp_client_info.h
jsorg71: yes, I think enums are the same size as int but I think compiler can decide.

Nexarian:
Fix the merge conflict, then I have no further objections.

jsorg71:
Sure, I'll rebase and fix that up.
I'm also thinking about adding a xrdp.ini setting for prefer h264 over progressive.  Currently, if the client supports h264, it's used.

Nexarian:
We do need a setting to allow users to select which protocol they want to use. Strictly speaking, they should also be able to select RLE (non-GFX) as well...

Issue #1707, opened 2020-10-21
Allow IP_FREEBIND to be specified as a socket option for xrdp


matt335672:
From the Linux manpage `ip(7)`:-

>       IP_FREEBIND (since Linux 2.4)
>              If enabled, this boolean option allows binding to an IP address
>              that  is nonlocal or does not (yet) exist.  This permits listen‐
>              ing on a socket, without requiring the underlying network inter‐
>              face  or  the  specified dynamic IP address to be up at the time
>              that the application is trying to bind to it. 

On systemd-enabled systems, not all network interfaces may not be up when xrdp starts. This poses problems for a system administrator who wishes to bind to the specific IP addresses of those interfaces rather than 'all interfaces'.

See #1610 for an instance where this has been encountered by a user.

This socket option seems at the moment to only be available for Linux.

metalefty:
Maybe this snippet helps: https://github.com/microsoft/WSL/issues/460

```C
// FreeBSD uses IP_BINDANY or IP_NONLOCALOK instead of IP_FREEBIND
// Use requires root rights in FreeBSD. Linux allows this without capabilities
#if (__FreeBSD__ || __FreeBSD_kernel__) && !defined(IP_FREEBIND)
#if defined(IP_BINDANY)
#define IP_FREEBIND IP_BINDANY
#elif defined(IP_NONLOCALOK)
#define IP_FREEBIND IP_NONLOCALOK
#endif // IP_BINDANY || IP_NONLOCALOK
#endif // __FREEBSD__ || __FreeBSD_kernel__
```

matt335672:
Yes that's useful - thanks.

PR #3086, opened 2024-05-22
Fix a systemd dependency ("network-online.target")

schen0x:
Hi, this PR fixes a problem that the "xrdp.service" fail to auto-start when instructed to listen on a specific interface

- By changing the "network.target" systemd dependency to "network-online.target"
- The "network-online.target", in short, means at least one network interface has finished IP level setup.
- The previously used "network.target" is vague and does not provide such guarantee (ref: man systemd.special(7)).
- Which often cause "xrdp.service" fail to auto-start when the service is configured to listen on a specific interface (e.g. in xrdp.ini, "port=tcp://192.168.0.1:3389"). Because the interface may have not finish setting up its IP, when "xrdp.service" starts.

### DIFF
--- a/instfiles/xrdp.service.in
+++ b/instfiles/xrdp.service.in
@@ -2,7 +2,7 @@
 Description=xrdp daemon
 Documentation=man:xrdp(8) man:xrdp.ini(5)
 Requires=xrdp-sesman.service
-After=network.target xrdp-sesman.service
+After=network-online.target xrdp-sesman.service
 
 [Service]
 Type=exec


matt335672:
I can see where you're coming from @schen0x 

I can see there are no definite guarantees even with this, but it's better than the current state.

For further guarantees, see #1707. I'm waiting on #2974 to be approved before I implement that as there are implications for FreeBSD.

@metalefty - what do you think? LGTM.

metalefty:
LGTM, too. @bsmojver added a very similar patch to Fedora package.

PR #3090, opened 2024-05-23
Remove the wait for the ibus daemon to start

### LINKED ISSUE #3080: Need to fix the delay, that is not needed if ibus is already started
https://github.com/neutrinolabs/xrdp/blob/98d0003715be1e0693817f49925d48b558df69bb/sesman/chansrv/input_ibus.c#L224

matt335672:
Fixes #3080 and tested by @seflerZ 

The initial implementation of Uinicode input via IBus used a startup delay of 3 seconds to wait for the daemon to be ready before connecting to it.

This commit introduces a poll-wait loop which can remove the delay entirely if the daemon is up when chansrv starts the interface.

### DIFF
--- a/sesman/chansrv/input_ibus.c
+++ b/sesman/chansrv/input_ibus.c
@@ -220,8 +220,21 @@ xrdp_input_unicode_init(void)
         return 0;
     }
 
-    /* Wait becasue ibus daemon may not be ready in first login. Do we have a flag to avoid busy waiting? */
-    sleep(3);
+    /* Wait because the ibus daemon may not be ready on first login */
+    const char *addr = ibus_get_address();
+    unsigned int cnt = 0;
+    while (!addr && cnt < 10)
+    {
+        usleep(500 * 1000); // half a second
+        addr = ibus_get_address();
+        ++cnt;
+    }
+    if (!addr)
+    {
+        LOG(LOG_LEVEL_ERROR,
+            "xrdp_ibus_init: Timed out waiting for iBus daemon");
+        return 1;
+    }
 
     LOG(LOG_LEVEL_INFO, "xrdp_ibus_init: Initializing the iBus engine");
     ibus_init();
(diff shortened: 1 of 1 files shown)


sesman/chansrv/input_ibus.c
```diff
@@ -220,8 +220,21 @@ xrdp_input_unicode_init(void)
...
     }
 
-    /* Wait becasue ibus daemon may not be ready in first login. Do we have a flag to avoid busy waiting? */
-    sleep(3);
+    /* Wait because the ibus daemon may not be ready on first login */
+    const char *addr = ibus_get_address();
+    unsigned int cnt = 0;
+    while (!addr && cnt < 10)
+    {
+        usleep(500 * 1000); // half a second
```
seflerZ, line 226: I think we can increase the timeout a little longer, such as 3-5 seconds. Longer timeout has no side effects because it will end as soon as the iBus is ready. We can cover 99.99%+ computers for that.

matt335672: @seflerZ - I think the timeout is 5 seconds with the above (10 * 0.5 secs). Is that enough?

seflerZ:
Hi @matt335672 , according to @JulyIghor 's feedback. We have a duplicate registration in input_ibus.c. Would you apply this patch to fix it up? I've tested it works.
```patch
From 29c8bfbd6d7227956c6baa944aac4b2e5649d7ba Mon Sep 17 00:00:00 2001
From: seflerZ <sefler@126.com>
Date: Thu, 30 May 2024 14:42:22 +0800
Subject: [PATCH] remove duplicate enable&disable func registration

---
 sesman/chansrv/input_ibus.c | 2 --
 1 file changed, 2 deletions(-)

diff --git a/sesman/chansrv/input_ibus.c b/sesman/chansrv/input_ibus.c
index 7b803e3a..b7ff2992 100644
--- a/sesman/chansrv/input_ibus.c
+++ b/sesman/chansrv/input_ibus.c
@@ -155,8 +155,6 @@ xrdp_input_main_loop(void *in_val)
     factory = ibus_factory_new(ibus_bus_get_connection(bus));
     g_object_ref_sink(factory);
     g_signal_connect(factory, "create-engine", G_CALLBACK(xrdp_input_ibus_create_engine), NULL);
-    g_signal_connect(factory, "enable", G_CALLBACK(xrdp_input_ibus_engine_enable), NULL);
-    g_signal_connect(factory, "disable", G_CALLBACK(xrdp_input_ibus_engine_disable), NULL);

     ibus_factory_add_engine(factory, "XrdpIme", IBUS_TYPE_ENGINE);

--
2.34.1

```

PR #3106, opened 2024-06-03
Fix build with devel logging, but without --enable-pixman

matt335672:
Mentioned in #3105 as an observation.

If `./configure` is used with devel logging, but without `--enable-pixman`, the stub pixman development files are used.

However, in this configuration, the `pixman_region_selfcheck()` function is declared, but not defined.

Fix is to add a function definition in this situation.

This is a regression introduced in 7e58209b195052205831d53a9f4cd0b4604a293d

### DIFF
--- a/common/pixman-region.c
+++ b/common/pixman-region.c
@@ -1631,3 +1631,78 @@ PREFIX (_extents) (region_type_t *region)
     return (&region->extents);
 }
 
+#ifdef USE_DEVEL_LOGGING
+/*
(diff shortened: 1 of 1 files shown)


matt335672:
My mistake on the regression - both v0.9 and V0.10 are fine.

The regression was introduced in 5eed76f32b9f73d1dd00aa1a4d3bc6878eb9da07 (2023-10-20)

PR #3107, opened 2024-06-03
Fix resize crash

matt335672:
Fixes a couple of problems with GFX resizing.

See #3105 for the background.

There's also detail on the commit comments about the changes.


xrdp/xrdp_encoder.c
```diff
@@ -1457,6 +1470,7 @@ proc_enc_msg(void *arg)
         }
 
     } /* end while (cont) */
+    g_set_wait_obj(self->xrdp_encoder_term_done);
```
Nexarian, line 1473: I'm honestly surprised this wasn't already here. Good fix!

PR #3111, opened 2024-06-10
Fix min/max monitor placing issue

### LINKED ISSUE #3075: xrdp v0.10 incorrect behavior when minimizing and restoring a mstsc.exe connection in fullscreen with multiple monitors (v0.9 works fine)
### xrdp version

0.10.0

0.10.0

various 

### Installation method

git clone & make install

### Which backend do you use?

xorgxrdp

i3 and GNOME

kvm virtual machine

### What's your client?

Win11 mstsc.exe

### Area(s) with issue?

Graphic glitches, Other

### Steps to reproduce

In Summary, when you open a multi-monitor full-screen remote connection to xrdp 0.10, it's not possible to minimize and then maximize correctly the remote desktop connection: only one monitor is displayed when you maximize. The same workflow works correctly with xrdp 0.9.

Equipment utilized:
- laptop running Windows 11, screen resolution: 1920x1080 (probably irrelevant)
- external monitor, screen resolution: 1920x1080 (probably irrelevant)
- external monitor is configured to extend (not mirror) the laptop's display
- the remote Linux desktop is configured with `i3` because `i3` makes it very easy to identify monitors and desktops by number, but the issue also affects GNOME

Reproducer:
- open Remote Desktop Connection on the Windows 11 laptop
- select "Use all my monitors for the remote session" in the Display tab
- also select "Display the connection bar when I use the full screen"
- connect to a Linux machine where xrdp is installed
  - you should now see a full-screen xrdp connection on two monitors which works as expected
- in the Remote Desktop Connection bar, click the "Restore Down Button"
  - the resulting window correctly encapsulates the multi-monitor configuration, i.e. using the scrollbars it's possible to see both monitors contained in the Remote  Desktop Connection Window
- click the Maximize button in the Remote Desktop Connection Window

### ✔️ Expected Behavior

- xrdp 0.9 has the expected behavior:
  - the window is restored full-screen to both monitors
  - you can see the remote desktop connection again on two monitors in full screen

### ❌ Actual Behavior

- **v0.10 behaves incorrecly:**
  - the windows is restored full-screen to one-monitor only
  - the monitor you see depends on your desktop environment, but it's only one
    - for instance, with `i3`, if you had Desktop 1 and Desktop 2, now you can only see an empty Desktop 3 which didn't exist before, and the status bar does not show indicators for Desktop 1 and Desktop 2 anymore
    - GNOME's behavior is more difficult to understand/predict because it's doesn't have the same way of numbering and identifying desktops uniquely 
  - however the output of `xrandr` still shows two connected monitors, which seems correct

### Anything else?

The workaround is to disconnect and reconnect, and the monitors are once again displayed correctly.

matt335672:
Fixes #3075 

When a multi-monitor session has the top-left vertex of the primary monitor at a desktop location other than (0,0), minimising and maximising the session results in the (0,0) co-ordinate of the entire desktop being placed at the top-left of the primary monitor.

The implementation seems to be at odds with [MS-RDPEGFX] 2.2.2.14 which suggests the monitorDefArray of the RDPGFX_RESET_GRAPHICS_PDU should be the same as that in the Monitor Layout PDU ([MS-RDPBCGR] 2.2.12.1)

### DIFF
--- a/xrdp/xrdp_egfx.h
+++ b/xrdp/xrdp_egfx.h
@@ -222,6 +222,10 @@ xrdp_egfx_send_wire_to_surface2(struct xrdp_egfx *egfx, int surface_id,
                                 int codec_id, int codec_context_id,
                                 int pixel_format,
                                 void *bitmap_data, int bitmap_data_length);
+/*
+ * NB: mi below must be such that the (top,left) co-ordinate of
+ * the primary monitor is (0.0)
+ */
 struct stream *
 xrdp_egfx_reset_graphics(struct xrdp_egfx_bulk *bulk, int width, int height,
                          int monitor_count, struct monitor_info *mi);
--- a/xrdp/xrdp_mm.c
+++ b/xrdp/xrdp_mm.c
@@ -1444,7 +1444,7 @@ xrdp_mm_egfx_caps_advertise(void *user, int caps_count,
         error = xrdp_egfx_send_reset_graphics(self->egfx,
                                               screen->width, screen->height,
                                               self->wm->client_info->display_sizes.monitorCount,
-                                              self->wm->client_info->display_sizes.minfo_wm);
+                                              self->wm->client_info->display_sizes.minfo);
         LOG(LOG_LEVEL_INFO, "xrdp_mm_egfx_caps_advertise: xrdp_egfx_send_reset_graphics "
             "error %d monitorCount %d",
             error, self->wm->client_info->display_sizes.monitorCount);


matt335672:
This is a tiny change, but I suspect it will need a lot of testing, particularly as we've got at least one enterprise user who has deployed GFX.

I've tested this with two screens and monitor hot-plug/unplug on Windows 10 so far. The primary screen was to the right of the secondary.

metalefty:
LGTM so far.

matt335672:
I've done as much testing as I can with my equipment.

I think I'll merge this into devel and set up a cherry-pick for v0.10. If we get no more reported problems with devel we can update v0.10 too.

PR #3117, opened 2024-06-14
Add missing scancode map for ABNT2 KPPT

matt335672:
The Brazilian ABNT2 Keyboard layout contains a keypad decimal key which doesn't exist on other keypads:-

https://www.kbdlayout.info/kbdbr/virtualkeys

This key is curently mapped in xorgxrdp to keycode 134 (basic mapping), but isn't present in the scancode map. It needs to be added so that it is available to VNC sessions and will be mapped for xorgxrdp when we move to evdev keycode mappings.


matt335672:
Closing this - I've just found the pause key needs some work too (special keycode), so I'll get the two in together.

Issue #3102, opened 2024-05-31
Clipboard not working for images


StrilGit:
### xrdp version

0.10.0

### Detailed xrdp version, build options

xrdp 0.10.0
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-fuse
      --enable-jpeg
      --enable-rfxcodec
      --enable-mp3lame
      --enable-vsock
      --enable-painter

  Compiled with OpenSSL 3.0.2 15 Mar 2022

### Operating system & version

Ubuntu 22.04

### Installation method

dnf / apt / zypper / pkg / etc

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

XFCE

### Environment xrdp running on

VM

### What's your client?

Microsoft MSTSC

### Area(s) with issue?

Clipboard

### Steps to reproduce

Copy Image (e.g. screenshot) on Windows. Paste in LibreOffice

### ✔️ Expected Behavior

Image should be pasted

### ❌ Actual Behavior

"The contents of the clipboard could not be pasted" --> Nothing inserted

### Anything else?

When I am using the repository-version 0.9.17, the image is pasted without problems

griffondemo:
Hello, 

I would like to share some findings/issues we are facing with clipboard support and xrdp software.  Please assess if this is an issue, a bug or if a possible workaround exists.    Thank you for your support

We are facing exactly the same issue as StrilGit.   

We have tested this feature on Ubuntu 22.04 & Ubuntu 24.04.  

- On Ubuntu 22.04, we have compiled from source xrdp latest version (i.e. 0.10.x) and clipboard support is not working as expected 

- On Ubuntu 24.04, we have used the package from Ubuntu Repository (0.9.24.x or 0.9.25.x)  and clipboard support is not working as expected 

**The Clipboard feature seems broken starting with version xrdp 0.9.23.x.**  

If you compile from source xrdp package version 0.10.x, same behavior happens on Ubuntu 22.04 & Ubuntu 24.04, ie clipboard not working 

Operating system & version
Ubuntu 22.04  and later 

Installation method
pkg/Compile from source 

Which backend do you use?
xorgxrdp

What desktop environment do you use?
Gnome Desktop

Environment xrdp running on
VM/ Physical machine 

What's your client?
Microsoft MSTSC

[...]

PR #3088, opened 2024-05-23
Further cherry-picks for v0.10

matt335672:
Backports the following PRs from the devel branch:-
- #3082
- #3086
- #3103
- #3104
- #3107
- #3112

PR #3120, opened 2024-06-17
clipboard: fix a bug when pasting image to LibreOffice

metalefty:
While here, embed correct file size in BMP file header.

Fixes:          #3102
Sponsored by:   Krämer Pferdesport GmbH & Co KG

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -181,22 +181,8 @@ x-special/gnome-copied-files
 #include "ms-rdpeclip.h"
 #include "xrdp_constants.h"
 
-static char g_bmp_image_header[] =
-{
-    /*
-     * Bitmap file header
-     * https://en.wikipedia.org/wiki/BMP_file_format#Bitmap_file_header
-     *
-     * NOTE: All of the integer values are stored in little-endian format
-     */
-    0x42, 0x4d,                 /* signature */
-    0x00, 0x00, 0x00, 0x00,     /* file size, to be filled later */
-    0x00, 0x00,                 /* reserved 1 */
-    0x00, 0x00,                 /* reserved 2 */
-    0x36, 0x00, 0x00, 0x00      /* offset to pixel array (14 + 40 bytes) */
-};
-
-#define BMPFILEHEADER_LEN sizeof(g_bmp_image_header)
+#define BMPFILEHEADER_LEN       14
+#define BMPINFOHEADER_LEN       40
 
 extern int g_cliprdr_chan_id;   /* in chansrv.c */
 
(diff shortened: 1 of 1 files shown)


metalefty:
Rebased onto #3088 to fix FreeBSD CI.

sesman/chansrv/clipboard.c
```diff
@@ -1148,20 +1158,35 @@ clipboard_process_data_response_for_image(struct stream *s,
...
+    /* assemble bmp header */
+    g_memcpy(bmp_file_header, g_bmp_image_header, BMPFILEHEADER_LEN);
+    bmp_size = (uint32_t) g_clip_c2s.total_bytes;
+#if defined(L_ENDIAN)
+    for (int i = 2, j = 0; i < 6; i++, j += 8)
+#else
+    for (int i = 5, j = 0; i >= 2; i--, j += 8)
+#endif
+    {
+        bmp_file_header[i] = ((bmp_size >> j) & 0xff);
```
metalefty: @matt335672 Can you have a look at this area especially? I'm not sure if this is a good way to embed file size in BMP file header.

matt335672: Yes, that could do with tidying up.

Personally, I'd use the stream functions we already have. They've been well-tested, an xrdp developer should be familiar with them, and a lot of magic numbers can be avoided. Something like this maybe (compiles, but not tested):-

```c
    /* Assemble bmp header */
    /* See https://en.wikipedia.org/wiki/BMP_file_format#Bitmap_file_header */
    make_stream(hdr_s);
    if (hdr_s == 0)
    {
        g_free(g_clip_c2s.data);
        g_clip_c2s.total_bytes = 0;
        return 0;
    }
    init_stream(hdr_s, BMPFILEHEADER_LEN);
    out_uint8(hdr_s, 'B');
    out_uint8(hdr_s, 'M');
    out_uint32_le(hdr_s, g_clip_c2s.total_bytes);
    out_uint16_le(hdr_s, 0);
    out_uint16_le(hdr_s, 0);
    out_uint32_le(hdr_s, BMPFILEHEADER_LEN + 40);

    /* Copy header and data to output stream */
    g_memcpy(g_clip_c2s.data, hdr_s->data, BMPFILEHEADER_LEN);
    in_uint8a(s, g_clip_c2s.data + BMPFILEHEADER_LEN, len);
    free_stream(hdr_s);
```
You can remove `g_bmp_image_header`, and simply define `BMPFILEHEADER_LEN` as 14.

It's not as efficient, but that probably doesn't matter. It's not being called that often to need to optimise it.

metalefty: Thanks, it worked and applied tidy-up. I also replaced the magic number `40` with `BMPINFOHEADER_LEN`.

matt335672 (approved):
LGTM

Also needs to go into `devel`

PR #3127, opened 2024-06-19
[v0.10] Fix min/max monitor placing issue

matt335672:
when a multi-monitor session has the top-left vertex of the primary monitor at a desktop location other than (0,0), minimising and maximising the session results in the (0,0) co-ordinate of the entire desktop being placed at the top-left of the primary monitor.

The implementation seems to be at odds with [MS-RDPEGFX] 2.2.2.14 which suggests the monitorDefArray of the RDPGFX_RESET_GRAPHICS_PDU should be the same as that in the Monitor Layout PDU ([MS-RDPBCGR] 2.2.12.1)

(cherry picked from commit 095f0d0e4cd2589bcd6ab995a1f4c29c7fbc3d33)


metalefty (approved):
Let's merge.

PR #3136, opened 2024-06-21
Fix wrong version in configure.ac

### LINKED ISSUE #3135: Wrong version in configure.ac
[commit 4c2a7e8](https://github.com/neutrinolabs/xrdp/commit/4c2a7e893b023892b9b3e6ecb21902cf6c326608)
pushed the version from **0.9.80** to **0.10.80** 
i might be mistaken but that seems to be a typo and it should be **0.10.0**

Caldwell-74:
closes #3135

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 # Process this file with autoconf to produce a configure script
 
 AC_PREREQ([2.69])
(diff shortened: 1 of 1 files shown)


metalefty:
Not wrong.

PR #3138, opened 2024-06-24
Update scancode processing

### LINKED ISSUE #355: keyboard issue with xorgxrdp
We are using an old Java applicaton (Oracle Virtual Desktop Client for SunRay) that can be made to work with xrdp/VNC with xmodmap, but we cannot get it to work with xorgxrdp. Several keys map completely incorrectly, and xmodmap changes appear to be ignored. Running xev on the console vs in xorgxrdp shows, for example, that the console uses keycode 116 for the down arrow, while xorgxrdp is using keycode 104 for the down arrow; however, keycode 104 at the console is KP_Enter, and that's how the Java application is also treating keycode 104 (ie. in the Java application, using the down arrow results in the application thinking Enter has been pressed). Running xorg-genkeymap at the console and then using the resulting keymap for the default keyboard layout (km-0409.ini in our case) has no effect on this either.

### LINKED ISSUE #621: xdotool does not type keys with AltGr modifier
Hi guys,

I'm experimenting since some weeks with xrdp to get the best remote desktop experience. One of the problem I currently can't solve is this:
I use KeeWeb as a keepass password manager in Linux/XFCE/xrdp within a KVM. Keeweb can autotype usernames and passwords when triggered via shortcut. For this to work it uses xdotool.

Problem:
when I use Keeweb in a XFCE4 session on xrdp the '@' (at) just gets a 'q' because it seems the AltGr modifier doesn't work. When doing the exact same within a "real" X session started via lightdm everything works fine.

My current investigations shows the following difference in the input method of the started Xorg:
setxkbmap -query -v on a Xorg started via xrdp-sesman:
```
keycodes:   xfree86+aliases(qwertz)
types:      complete
compat:     complete
symbols:    pc+de(nodeadkeys)+inet(pc105)
geometry:   pc(pc105)
rules:      base
model:      pc105
layout:     de
variant:    nodeadkeys
```

When starting the same session via lightdm:
```
keycodes:   evdev+aliases(qwertz)
types:      complete
compat:     complete
symbols:    pc+de(nodeadkeys)+inet(evdev)
geometry:   pc(pc105)
rules:      evdev
model:      pc105
layout:     de
variant:    nodeadkeys

```
As you can see the "real" Xorg session uses evdev as keyboard rule while xrdp uses base. When I set 
setxkbmap -rules evdev in terminal in xrdp session then xdotool types the '@' correctly but the keyboard is not working correctly anymore, e.g. the cursor keys don't work.

I already tried to generate a new mapping file once within the "real" Xorg session and once within the xrdp session. But nothing helps.

I'm not sure if this is a problem of xdotool or Keeweb or xrdp... I can only see that it's working in a normal Xorg session started by lightdm it's not working with xrdp.

I have a feeling that this might on of several issues that arise because xrdp-sessman seems not to create a real X session like lightdm does...

Any suggestions what I could try to fix this? I guess it's a configuration issue but I don't know what else to check.

Thanks and cheers

matt335672:
Fixes #621
Fixes #355 

This is a significant re-write of xrdp keyboard handling.

See also xorgxrdp PR neutrinolabs/xorgxrdp#316

This builds on the work in #3039 / #3113 with significant changes:-
1) The ad-hoc ways of indexing scancodes have been brought together in a 'scancode index' which maps an RDP scancode into an 8-bit value. This index can then be used to access an array containing state information about a particular key.
2) RDP scancode defines are provided in a new development file xrdp_scancode_defs.h
3) (related to the above) A lot of magic numbers have been removed.
4) Some keys which only worked with xorgxrdp (pause and ABNT2 keypad point) now work with VNC as well.
5) The confusingly named function `xrdp_load_keyboard_layout()` which is only needed for xorgxrdp has been renamed `xrdp_init_xkb_layout()` and is now only called when needed.
6) The keycode set (`base` or `evdev`) can be specified for the X server at run-time.The corresponding XKB rules are passed to `xorgxrdp`.
7) A minor unreported problem with xorgxrdp has been resolved.

I've got a bit more testing to do before I can merge this, but I'd welcome any comments.


matt335672:
Unless anyone's got any great objections, I'll merge this and neutrinolabs/xorgxrdp#316 into devel after we've shipped v0.10.1 (see #3155). It seems pretty solid to me, but I've not got a huge array of keyboards here for testing. Getting it out into the community should flush out any last few problems before we ship it in a major release.

Issue #3149, opened 2024-07-08
xrdp-0.10.0.tar.gz does not include 'docs/man/xrdp-mkfv1.8.in'.


tanakakz-alpha:
### xrdp version

0.10.0

### Detailed xrdp version, build options

~~~
./configure --enable-jpeg --enable-tjpeg --enable-pixman --enable-opus --enable-mp3lame --enable-rdpsndaudin --enable-pamuserpas --with-imlib2=yes --enable-vsock --with-freetype2=yes
~~~

### Operating system & version

Ubuntu 22.04

### Installation method

other

### Area(s) with issue?

Compile error

### Steps to reproduce

~~~
wget https://github.com/neutrinolabs/xrdp/releases/download/v0.10.0/xrdp-0.10.0.tar.gz
tar xzvf xrdp-0.10.0.tar.gz
cd xrdp-0.10.0/
./bootstrap
./configure --enable-jpeg --enable-tjpeg --enable-pixman --enable-opus --enable-mp3lame --enable-rdpsndaudin --enable-pamuserpas --with-imlib2=yes --enable-vsock --with-freetype2=yes
make
~~~

### ✔️ Expected Behavior

make successfully.

### ❌ Actual Behavior

~~~
make[3]: *** No rule to make target 'xrdp-mkfv1.8.in', needed by 'xrdp-mkfv1.8'.  Stop.
~~~

### Anything else?

xrdp-0.10.0.tar.gz does not contain 'docs/man/xrdp-mkfv1.8.in'.
v0.10.0.tar.gz contains "docs/man/xrdp-mkfv1.8.in".

xrdp-0.10.0.tar.gz is missing the required files.

metalefty:
Good catch! A mistake in Makefile at #2341 caused this issue.  I submitted #3150.

@matt335672 Could you have a look?

metalefty:
It will be fixed in the next release (v0.10.1 is expected in July or August), thanks!

PR #3150, opened 2024-07-08
docs: always include docs/man/xrdp-mkfv1.8.in to dist tarball

metalefty:
Files included in distribution tarball must always be enumerated, not be enumerated conditionally.

Resolves:   #3149


matt335672 (approved):
Good point - that looks good to me.

Issue #3083, opened 2024-05-21
GFX: graphic glitches when a monitor is more than 4096 pixels wide or high


metalefty:
### xrdp version

0.10.0-beta.3

### Detailed xrdp version, build options

~~~
xrdp 0.10.0-beta.3
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-imlib2
      --enable-jpeg
      --enable-painter
      --enable-pam-config=freebsd
      --enable-pixman
      --enable-rfxcodec
      --enable-strict-locations
      --enable-vsock
      --localstatedir=/var
      --with-pkgconfigdir=/usr/local/libdata/pkgconfig
      --disable-devel-debug
      --disable-devel-logging
      --enable-fdkaac
      --disable-fuse
      --enable-ipv6
      --disable-mp3lame
      --enable-opus
      --x-libraries=/usr/local/lib
      --x-includes=/usr/local/include
      --prefix=/usr/local
      --mandir=/usr/local/share/man
      --disable-silent-rules
      --infodir=/usr/local/share/info/
      --build=amd64-portbld-freebsd14.0
      build_alias=amd64-portbld-freebsd14.0
      CC=cc
      CFLAGS=-O2 -pipe  -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing
      LDFLAGS= -lssl  -fstack-protector-strong
      LIBS=-L/usr/local/lib
      CPPFLAGS=-isystem /usr/local/include
      CXX=c++
      CXXFLAGS=-O2 -pipe -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -isystem /usr/local/include
      PKG_CONFIG=pkgconf
      PKG_CONFIG_LIBDIR=/ssd/ports-wrkdir/home/meta/FreeBSD/ports/git/net/xrdp-devel/work/.pkgconfig:/usr/local/libdata/pkgconfig:/usr/local/share/pkgconfig:/usr/libdata/pkgconfig
      OPENSSL_CFLAGS=-I/usr/include
      OPENSSL_LIBS=-L/usr/lib -lcrypto -lssl
      CPP=cpp

  Compiled with OpenSSL 3.0.13 30 Jan 2024
~~~

### Operating system & version

FreeBSD 14 but actually not OS specific

### Installation method

[...]

PR #3152, opened 2024-07-08
[v0.10] Support screens larger than 4096 pixels (update submodule)

metalefty:
Fixes:  #3083

---

Jay is working on NEON (accelerations for ARM64) however bug fixes should be shipped ASAP to the stable version v0.10.


rowlap:
Sorry to nitpick, but can we use "pixels" spelling in the title, in aid of searching?

metalefty:
Thanks, it was my mistake.

metalefty:
I will correct the commit message later.

matt335672:
LGTM

PR #3158, opened 2024-07-11
[v0.9] Fix xrdp.ini man page

metalefty:
- require_credentials and enable_login_token was swapped
- enable_login_token was unintendedly overwritten with domain_separator
- Set bold and italic properly

---
@matt335672 There was an issue in your recent cherry-pick.


matt335672 (approved):
Looks like I made a bit of a mess of that!

Thanks for the fix.

PR #3159, opened 2024-07-12
chkpriv: respect $sysconfdir for config files directory

metalefty:
While here, ignore build artifacts of chkpriv tools.

Follow-up to:   #2974


matt335672 (approved):
Definitely worth doing - thanks.

Issue #3165, opened 2024-07-17
Specific filenames cause File transfer / drive redirection error


tsz8899:
### xrdp version

0.10.80

### Detailed xrdp version, build options

~~~
xrdp 0.10.80
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-fuse
      --enable-ipv6
      --enable-rdpsndaudin
      --enable-mp3lame
      LDFLAGS=-ldl
~~~

### Operating system & version

debian deepin

### Installation method

git clone & make install

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

xfce dde

### What's your client?

mstsc xfreerdp wfreerdp

### Area(s) with issue?

Clipboard, File transfer / drive redirection

### Steps to reproduce

Specific filenames cause File transfer / drive redirection error or crash

The anomalies are not due to specific single characters but are caused by combinations of multiple normal characters utf-8

1.When the filename contains specific characters, copying via the file clipboard fails. 
For example: "日照香炉生紫烟遥看瀑布挂前川飞流直下三千尺疑是银河落九天.txt" 
Pasting from local to xrdp session works normally. 
Copying within the xrdp session to thinclient_drives also works. 
But copying from the xrdp session and pasting to the local disk does not work.

2.thinclient_drives cannot see files with specific character names.
For example: "暮从碧山下，山月随人归。却顾所来径，苍苍横翠微。相携及田家，童稚开荆扉。绿竹入幽径，青萝拂行衣。欢言得所憩，美酒聊共挥。长歌吟松风，曲尽河星稀。我醉君复乐，陶然共忘机。.txt"
In the thinclient_drives directory of the xrdp session, other files can be seen, but this file is not visible. 
This phenomenon also occurs in file managers like thunar and dde-file-manager on Debian Deepin.

3.Copying files with specific character names causes thinclient_drives to crash. 
For example: "暮从碧山下山月随人归却顾所来径苍苍横翠微相携及田家童稚开荆扉绿竹入幽径青萝拂行衣欢言得所憩美酒聊共挥长歌吟松风曲尽河星稀我醉君复乐陶然共忘机.txt" 
Pasting from local to xrdp session works normally. 
But pressing CTRL+C on this file within the xrdp session causes thinclient_drives to crash instantly. 
[...]

PR #3172, opened 2024-07-22
Fix potential name buffer overflows in redirector

matt335672:
Found while investigating #3165 

The state buffers used by the following structs in chansrv_fuse.c are one byte too small for filenames of length XFS_MAXFILENAMELEN:-
- struct state_lookup
- struct state_create
- struct state_rename

I don't think this needs a CVE - in practice, there is no runtime danger, as XFS_MAXFILENAMELEN is 255, and the arrays sized to hold these strings will be followed by non-byte aligned data. Nevertheless this should be fixed to prevent problems if the value is changed.


metalefty (approved):
LGTM. Ship to v0.10?

matt335672:
I'll do that too.

PR #3173, opened 2024-07-22
Chansrv clip fixes

matt335672:
Fixes much of #3165

- Much longer filenames are now supported. This is useful when dealing with some character encodings which use 4-byte UTF-8 encoded characters.
- Some hard-coded buffer sizes used for filenames have been removed.
- An issue related to pasting file lists from the clipboard when using the FreeRDP client has been fixed

Special thanks to @tsz8899 for raising the original issue and working with the team on resolving it.

Update 2024-8-2 : No longer draft


matt335672:
I'm merging this into `devel `as both @tsz8899 and I have tested it.

I'll put a cherry-pick together for v0.10.x

PR #3194, opened 2024-08-05
[V0.10] chansrv clip fixes

matt335672:
Backport of #3165 / #3173 to V0.10

- Much longer filenames are now supported by the redirector. This is useful when dealing with some character encodings which use 4-byte UTF-8 encoded characters.
- Some hard-coded buffer sizes used for filenames have been removed.
- An issue related to pasting file lists from the clipboard when using the FreeRDP client has been fixed

Special thanks to @tsz8899 for raising the original issue and working with the team on resolving it.


metalefty (approved):
LGTM. Merge at your convenience.

PR #3213, opened 2024-08-19
Fix github runner version for v0.10 at 22.04

matt335672:
At the time of writing, the github runner for ubuntu-latest is 22.04, but this will change to 24.04 later in the year.

This change has not yet been announced, but the change from ubuntu 20.04 to 22.04 happened [around November '22](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/) so it's reasonable to assume the change to 24.04 will happen round about November '24.

We've already made sure that `devel` will run on 24.04 in #3206. This necessitated a small change to both the libpainter (neutrinolabs/libpainter#23) and librfxcodec (neutrinolabs/librfxcodec#72) submodules. The change to `devel` means that `devel` can carry on using `ubuntu-latest`, and we (hopefully) won't notice.

For the V0.10 branch, I believe we have three choices to keep CI running after November '24:-
1) Make a v0.10.x branch for each submodule, based at their current v0.10 location and make the necessary changes to the branches.
2) Update the v0.10.x branch to use the latest versions of the submodules
3) Fix v0.10.x at Ubuntu CI version 22.04.

My personal preference is for 1) or 3). @akarl10 has suggested 3), and in the absence of any other suggestions I'm happy to go with that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,21 +20,21 @@ jobs:
           - CC: gcc
             feature_set: min
             arch: amd64
-            os: ubuntu-latest
+            os: ubuntu-22.04
             unittests: false
             DISTCHECK: false
 
           - CC: g++
             feature_set: min
             arch: amd64
-            os: ubuntu-latest
+            os: ubuntu-22.04
             unittests: false
             DISTCHECK: false
 
           - CC: clang
             feature_set: min
             arch: amd64
-            os: ubuntu-latest
+            os: ubuntu-22.04
             unittests: false
             DISTCHECK: false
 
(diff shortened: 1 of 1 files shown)


metalefty (approved):
Approved. We might make changes to make v0.10 CI 24.04 capable by November but there's no reason not to do this.

matt335672:
Agreed.

I note that the 20.04 runner is still available now, so it's reasonable to assume that the 22.04 runner will be available in two years. This is possibly longer than the lifetime of xrdp v0.10.x.

PR #3214, opened 2024-08-19
GFX: configurable x264 parameters

metalefty:
This PR makes x264 parameters such as bitrate configurable.

- Add gfx.toml for configurations for GFX
- newconfig is a new config parser for toml
- x264 parameters are stored in an array of structures (idx is connectyionType)
- Each parameter will be inherited from the default unless redefined for each connection type
  - index 0 is default since connectyionType == 0 is not used in MS-RDPBCGR

Design is not very sophisticated, comments are welcomed.
- newconfig should be moved to common?

TODO:
- Make order of preferred codecs (H264 or RFX) configurable

### DIFF
--- a/xrdp/xrdp_tconfig.h
+++ b/xrdp/xrdp_tconfig.h
@@ -1,0 +1,66 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Koichiro Iwao
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ *
+ * @file xrdp_tconfig.c
+ * @brief TOML config loader and structures
+ * @author Koichiro Iwao
+ *
+ */
+#ifndef _XRDP_TCONFIG_H_
+#define _XRDP_TCONFIG_H_
+
+/* The number of connection types in MS-RDPBCGR 2.2.1.3.2 */
+#define NUM_CONNECTION_TYPES 7
+#define GFX_CONF XRDP_CFG_PATH "/gfx.toml"
+
+/* nc stands for new config */
+struct xrdp_tconfig_gfx_x264_param
+{
+    char preset[16];
+    char tune[16];
+    char profile[16];
+    int vbv_max_bitrate;
+    int vbv_buffer_size;
+    int fps_num;
+    int fps_den;
+};
+
+struct xrdp_tconfig_gfx
+{
+    /* store x264 parameters for each connection type */
+    struct xrdp_tconfig_gfx_x264_param x264_param[NUM_CONNECTION_TYPES];
+};
+
+static const char *const rdpbcgr_connection_type_names[] =
+{
+    "default", /* for xrdp internal use */
+    "modem",
+    "broadband_low",
+    "satellite",
+    "broadband_high",
+    "wan",
+    "lan",
+    "autodetect",
+    0
+};
+
+int tconfig_load_gfx(const char *filename, struct xrdp_tconfig_gfx *config);
+
+#endif
--- a/tests/xrdp/Makefile.am
+++ b/tests/xrdp/Makefile.am
@@ -1,5 +1,6 @@
 AM_CPPFLAGS = \
   -DXRDP_TOP_SRCDIR=\"$(top_srcdir)\" \
+  -DXRDP_CFG_PATH=\"${sysconfdir}/xrdp\" \
   -I$(top_builddir) \
   -I$(top_srcdir)/xrdp \
   -I$(top_srcdir)/libxrdp \
@@ -31,6 +32,7 @@ test_xrdp_SOURCES = \
     test_xrdp_egfx.c \
     test_xrdp_keymap.c \
     test_xrdp_region.c \
+    test_tconfig.c \
     test_bitmap_load.c
 
 test_xrdp_CFLAGS = \
@@ -59,6 +61,7 @@ test_xrdp_LDADD = \
     $(top_builddir)/xrdp/xrdp_encoder.o \
     $(top_builddir)/xrdp/xrdp_process.o \
     $(top_builddir)/xrdp/xrdp_login_wnd.o \
+    $(top_builddir)/xrdp/xrdp_tconfig.o \
     $(top_builddir)/xrdp/xrdp_main_utils.o \
     $(top_builddir)/libpainter/src/libpainter.la \
     $(top_builddir)/librfxcodec/src/librfxencode.la \
@@ -69,3 +72,10 @@ test_xrdp_LDADD = \
     $(IMLIB2_LIBS) \
     @CHECK_LIBS@ \
     @CMOCKA_LIBS@
+
+if XRDP_X264
+AM_CPPFLAGS += -DXRDP_X264 $(XRDP_X264_CFLAGS)
+test_xrdp_LDADD += \
+    $(top_builddir)/xrdp/xrdp_encoder_x264.o \
+    $(XRDP_X264_LIBS)
+endif
--- a/tests/xrdp/test_tconfig.c
+++ b/tests/xrdp/test_tconfig.c
@@ -1,0 +1,50 @@
+#if defined(HAVE_CONFIG_H)
+#include "config_ac.h"
+#endif
+
+#include "xrdp_tconfig.h"
+#include "test_xrdp.h"
+#include "xrdp.h"
+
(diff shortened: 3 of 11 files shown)


rowlap:
Is there scope for fps to match the client, without needing to be hardcoded?
```
#define X264_DEFAULT_FPS_NUM 24
#define X264_DEFAULT_FPS_DEN 1
```

We see the artificial capping of refresh rate to <= 30fps to be one of the significant causes of user-perceived latency.

metalefty:
These macros are not for hardcoding the fps value,  actually, the last resort value if fps values are not found anywhere in the config file.

Multiple factors affect the whole refresh rate.  There is also a cap on the capture rate on xorgxrdp side. The `fps_num` and `fps_den` values can configure the frame rate of x264 encoding but for debugging purposes at the moment.  It is not intended for users to simply change fps_num and try to get more than 30fps. It doesn't make significant changes because screen capture rate is around 25fps at the moment.

Network latency is also a big factor in user-perceiving latency. Total latency matters but network latency is not easily controllable. So it makes sense that fps is the only factor we can control. If network latency is 40ms, the total latency will be 80ms (25fps == 40ms/frame). That's huge!

I also would like to tackle the latency issue but not in the scope of this PR right now.

matt335672:
Just about to start a review.

One think I must say up front is I'm not keen on the 'newconfig' name. Time has a habit of passing more quickly that any of us like, and all of a sudden, that 'new' can seem quite 'old'.

Just up the road from where we live, there's a national park called the [New Forest](https://en.wikipedia.org/wiki/New_Forest). I guess it was new in 1086, but here we are over 1000 years later and no-one's quite got round to renaming it yet!

xrdp/xrdp_newconfig.h
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+
+#ifndef _XRDP_NEWCONFIG_H_
+#define _XRDP_NEWCONFIG_H_
+
+#include "toml.h"
+#include "string_calls.h"
+#include "ms-rdpbcgr.h"
+
```
matt335672: Suggest moving the above includes into xrdp_newconfig.c and replacing with an incomplete type declaration for toml_table_t:-

```c
struct toml_table_t;
typedef struct toml_table_t toml_table_t;
```

This stops these defines being pulled in to files which just use this include to access the module.

xrdp/xrdp_newconfig.h
```diff
@@ -0,0 +1,70 @@
...
+#if !defined(NUM_CONNECTION_TYPES)
+#define NUM_CONNECTION_TYPES CONNECTION_TYPE_AUTODETECT
+#endif
```
matt335672: Suggest replacing with `#define NUM_CONNECTION_TYPES 7` and a comment. The `#if` could result in NUM_CONNECTION_TYPES being defined elsewhere, and silently overwriting the value in this module */

xrdp/xrdp_newconfig.h
```diff
@@ -0,0 +1,70 @@
...
+
+#define GFX_CONF XRDP_CFG_PATH "/gfx.toml"
+
+#define NCLOG(log_level, args...) LOG(log_level, "NewConfig: " args)
```
matt335672: Above two can also go into the C file.

xrdp/xrdp_newconfig.h
```diff
@@ -0,0 +1,70 @@
...
+    "broadband_high",
+    "wan",
+    "lan",
+    "autodetect",
+    0
+};
+
+
+int nc_load_gfx_x264_ct(toml_table_t *tfile, const int connection_type,
+                        struct xrdp_nc_gfx_x264_param *param);
```
matt335672: Does `nc_load_gfx_x264_ct` need to be here? It's only used in the C file I think. Could be removed from here and made static.

xrdp/xrdp_newconfig.c
```diff
@@ -0,0 +1,232 @@
...
+nc_load_gfx_x264_ct(toml_table_t *tfile, const int connection_type,
+                    struct xrdp_nc_gfx_x264_param *param)
+{
+    if (connection_type > NUM_CONNECTION_TYPES)
+    {
+        NCLOG(LOG_LEVEL_ERROR, "Invalid connection type is given");
+        return 1;
+    }
+
+    toml_table_t *x264 = toml_table_in(tfile, "x264");
```
matt335672: Could return NULL here?

xrdp/xrdp_newconfig.c
```diff
@@ -0,0 +1,232 @@
...
+        free(datum.u.s);
+    }
+    else if (connection_type == 0)
+    {
+        NCLOG(LOG_LEVEL_WARNING,
+              "x264 param preset is not set for connection type [%s], "
+              "adopting the default value \"" X264_DEFAULT_PRESET "\"",
+              rdpbcgr_connection_type_names[connection_type]);
+        g_strncpy(param[connection_type].preset,
+                  X264_DEFAULT_PRESET, sizeof(X264_DEFAULT_PRESET) - 1);
```
matt335672: Should be `sizeof(param[connection_type].preset) - 1`

xrdp/xrdp_newconfig.c
```diff
@@ -0,0 +1,232 @@
...
+        free(datum.u.s);
+    }
+    else if (connection_type == 0)
+    {
+        NCLOG(LOG_LEVEL_WARNING,
+              "x264 param tune is not set for connection type [%s], "
+              "adopting the default value \"" X264_DEFAULT_TUNE "\"",
+              rdpbcgr_connection_type_names[connection_type]);
+        g_strncpy(param[connection_type].tune,
+                  X264_DEFAULT_TUNE, sizeof(X264_DEFAULT_TUNE));
```
matt335672: Should be `sizeof(param[connection_type].tune) - 1`

xrdp/xrdp_newconfig.c
```diff
@@ -0,0 +1,232 @@
...
+        free(datum.u.s);
+    }
+    else if (connection_type == 0)
+    {
+        NCLOG(LOG_LEVEL_WARNING,
+              "x264 param profile is not set for connection type [%s], "
+              "adopting the default value \"" X264_DEFAULT_PROFILE "\"",
+              rdpbcgr_connection_type_names[connection_type]);
+        g_strncpy(param[connection_type].profile,
+                  X264_DEFAULT_PROFILE, sizeof(X264_DEFAULT_PROFILE));
```
matt335672: Should be `sizeof(param[connection_type].profile) - 1`

xrdp/xrdp_newconfig.h
metalefty: Not actually needed.

xrdp/xrdp_newconfig.c
metalefty: Indeed, need to check it.

xrdp/xrdp_newconfig.h
metalefty: `GFX_CONF` is referred from `xrdp_encoder_x264.c` so it should be in header file.

metalefty:
> One think I must say up front is I'm not keen on the 'newconfig' name. Time has a habit of passing more quickly that any of us like, and all of a sudden, that 'new' can seem quite 'old'.
> 
> Just up the road from where we live, there's a national park called the [New Forest](https://en.wikipedia.org/wiki/New_Forest). I guess it was new in 1086, but here we are over 1000 years later and no-one's quite got round to renaming it yet!

I recalled New York about that. New York is quite younger than New Forest though. 

About the "newconfig" name, I don't think it is a good name either and want to change it. Whatever is fine if it recalls TOML or a new config system and doesn't conflict with other names. I have some candidates:

- stomp (Simple, Stupid, or S-whatever-word TOML parser)
- tomling (TOML+ing, [a male kitten](https://en.wiktionary.org/wiki/tomling) - I didn't know)

I like tomling. It's a cute name.

matt335672:
I'm happy with tomling myself, but I'd also be happy with sticking with 'config' if we can make it work with existing files. Naming things is one of the hardest problems in computing I think!

metalefty:
I also thought about 'config' but there is already `xrdp_config` structure for storing current ini configurations.  So I avoided sticking `config` or `xrdp_config`. I would like to collect configurations to new structures for TOML configurations when we completely switch all configurations to TOML. 

I've been thinking about the naming, I think simply `tconfig` would be fine for the new structure.

metalefty:
Force-pushed. Commits will be squashed into several commits before merging.

matt335672:
tconfig looks fine to me!

xrdp/xrdp_tconfig.h
```diff
@@ -0,0 +1,69 @@
...
+ * @file xrdp_tconfig.c
+ * @brief TOML config loader and structures
+ * @author Koichiro Iwao
+ *
+ */
+#ifndef _XRDP_TCONFIG_H_
+#define _XRDP_TCONFIG_H_
+
+struct toml_table_t;
+typedef struct toml_table_t toml_table_t;
```
matt335672: Now that we've removed `tconfig_load_gfx_x264_ct` from the include file, I think these two lines can be removed entirely. Sorry - I failed to pick that up in my original review.

metalefty: Indeed, pushed.

metalefty:
@matt335672 Anything else?

matt335672:
Not from me - LGTM.

PR #3218, opened 2024-08-23
GFX: add config which to prefer H264 vs RFX

metalefty:
Not implemented to use loaded config yet.

I'd like to get reviewed whether the structure to save codec priority is appropriate.


tests/xrdp/gfx/gfx.toml
```diff
@@ -0,0 +1,40 @@
+[codec]
+order = [ "H264", "RFX" ]
```
matt335672: Since you're checking for H264 or H.264 in your string compare, how about making this one "H.264" rather than "H264"?

xrdp/xrdp_tconfig.c
```diff
@@ -199,6 +199,70 @@ tconfig_load_gfx_x264_ct(toml_table_t *tfile, const int connection_type,
...
+    TCLOG(LOG_LEVEL_TRACE, "[codec]");
+
+    int h264_found = 0;
+    int rfx_found = 0;
+
+    config->codec.h264_idx = -1;
+    config->codec.rfx_idx = -1;
+
+    toml_table_t *codec = toml_table_in(tfile, "codec");
+    toml_array_t *order = toml_array_in(codec, "order");
```
matt335672: Need to check codec against NULL.

xrdp/xrdp_tconfig.c
```diff
@@ -199,6 +199,70 @@ tconfig_load_gfx_x264_ct(toml_table_t *tfile, const int connection_type,
...
+    if (codec && order)
+    {
+        for (int i = 0; ; i++)
+        {
+            toml_datum_t datum = toml_string_at(order, i);
+
+            if (datum.ok)
+            {
+                if (g_strcasecmp(datum.u.s, "h264") == 0 ||
+                        g_strcasecmp(datum.u.s, "h.264") == 0)
```
matt335672: This is a slightly oddball case, but you might want to consider checking `h264_found` is zero here. Otherwise this string might not work how you want:-

```
order = [ "RFX", "H264" , "RFX" ]
```

xrdp/xrdp_tconfig.c
```diff
@@ -199,6 +199,70 @@ tconfig_load_gfx_x264_ct(toml_table_t *tfile, const int connection_type,
...
+                {
+                    h264_found = 1;
+                    config->codec.h264_idx = i;
+                }
+                if (g_strcasecmp(datum.u.s, "rfx") == 0)
```
matt335672: See above.

matt335672 (commented):
Few minor comments.

xrdp/xrdp_tconfig.c
metalefty: If either is NULL, I want to return default priority (jump to line 248) without using goto statement. My first code used goto but I've rewritten without goto.

I think it can be written like this. What do you think?

```C
toml_table_t *codec;
toml_array_t *order;

 if ((codec = toml_table_in(tfile, "codec")) &&
         (order = toml_array_in(codec, "order")))
```

xrdp/xrdp_tconfig.c
metalefty: You're right.

tests/xrdp/test_tconfig.c
```diff
@@ -31,6 +33,40 @@ START_TEST(test_tconfig_gfx_x264_load_basic)
...
+    ck_assert_int_gt(gfxconfig.codec.h264_idx, -1);
+    ck_assert_int_gt(gfxconfig.codec.rfx_idx, -1);
+    ck_assert_int_lt(gfxconfig.codec.rfx_idx, gfxconfig.codec.h264_idx);
+
+    /* RFX only */
+    tconfig_load_gfx(GFXCONF_STUBDIR "/gfx_codec_rfx_only.toml", &gfxconfig);
+    ck_assert_int_eq(gfxconfig.codec.h264_idx, -1);
+    ck_assert_int_gt(gfxconfig.codec.rfx_idx, -1);
+
+    /* RFX is preferred if order undefined */
```
metalefty: Incorrect comment: H264 is preferred actually

tests/xrdp/gfx/gfx.toml
metalefty: Agreed.

xrdp/xrdp_mm.c
```diff
@@ -1432,19 +1432,28 @@ xrdp_mm_egfx_caps_advertise(void *user, int caps_count,
                 break;
         }
     }
-    if (best_pro_index >= 0)
+#if defined(XRDP_H264)
+    struct xrdp_tconfig_gfx_codec_order co = self->wm->gfx_config->codec;
+    bool_t use_h264 = (best_h264_index >= 0 && (best_pro_index < 0 || (co.h264_idx >= 0 && co.h264_idx < co.rfx_idx)));
```
metalefty: We have 2 codecs for GFX now. It might be more complex when we support a third codec.

metalefty:
@matt335672 Added to use codec order.  Can you have a look at the part?

Commits will be squashed into 2 parts later, adding configuration and using configuration.

xrdp/xrdp_tconfig.c
matt335672: That's fine.

You can make the test explicit with :-

```c
 if ((codec = toml_table_in(tfile, "codec") != NULL) &&
         (order = toml_array_in(codec, "order") != NULL))
```

matt335672:
@metalefty - sorry it's taken me a while to get to this, but I wanted to review it properly.

I share your concern about working out the codec order, particularly if we add another codec.

I think there's a better way to do this by replacing the `_idx` values in `xrdp_tconfig.h` with an array which just contains the order of the codecs entered by the user, like this:-

```patch
--- a/xrdp/xrdp_tconfig.h
+++ b/xrdp/xrdp_tconfig.h
@@ -42,10 +42,17 @@ struct xrdp_tconfig_gfx_x264_param
     int fps_den;
 };
 
+enum xrdp_tconfig_codecs
+{
+    XTC_H264,
+    XTC_RFX
+};
+
 struct xrdp_tconfig_gfx_codec_order
 {
-    int h264_idx;
-    int rfx_idx;
+
+    enum xrdp_tconfig_codecs codecs[2];
+    unsigned int codec_count;
 };
 
 struct xrdp_tconfig_gfx
```

Then in xrdp_mm.c we can just iterate over the list and find the first match, ignoring any unsupported codecs. This scales for any number of codecs.

I've put together a complete patch (including tests) which illustrates this approach:-

[patch.txt](https://github.com/user-attachments/files/16760492/patch.txt)

Some of the variable names could maybe do with changing.

What do you think?

metalefty:
I like that idea. Could you attach a patch `git format-patch` when you have changed variable names? Then I'll add your changes to this pull request.

matt335672:
Will do.

I’ve also just realised the code I added to xrdp_mm.c is complete rubbish! I’ll fix that too, and test it. That will be tomorrow now.

matt335672:
How's this?

[0001-Rework-codec-order-in-tconfig.patch](https://github.com/user-attachments/files/16782422/0001-Rework-codec-order-in-tconfig.patch.txt)

After testing, it occurs to me that for some error conditions (e.g. missing file, badly formatted file) we can default to just "RFX" rather than nothing at all. Then at least we get a match on the codecs. What do you think?

metalefty:
Agreed.

matt335672:
Away from my PC ATM. I’ll fix the build errors and add a patch (and test) to default to RFX tomorrow, time permitting

matt335672:
Try this one:-

[0001-Further-changes-to-selectable-H.264-support.patch.txt](https://github.com/user-attachments/files/16798454/0001-Further-changes-to-selectable-H.264-support.patch.txt)

Two tests added:-
1) If the user tries to load a missing file, an RFX-only config is returned
2) If the config specified H.264, but there are no default for H.264, an RFX-only config is returned

CI should be fine now.

matt335672:
PS: Feel free to edit/squash as appropriate.

metalefty:
Thanks, wonderful! LGTM.

PR #3219, opened 2024-08-23
startwm.sh POSIX rewrite with better error handling.

rowingdude:
I was having trouble getting this to work on Gentoo using the boilerplate install, so I rewrote some scripts in hope of gaining functionality.

I cleaned this script up, I went for POSIX compatibility above all else - so we use `/bin/sh` and then use sh conventions. These are universal. 

Next, I added an error_exit function that will provide actual details if it fails (this was my primary motivation). Using `set -e` and `set -u` treats unset variables as errors so instead of the script and program starting, we find out what needs to change.

I have tested this on Gentoo, Arch, and Fedora. All seem to work fine.

### DIFF
--- a/sesman/startwm.sh
+++ b/sesman/startwm.sh
@@ -23,23 +23,32 @@
 #         END IF
 #     END IF
 # END IF
-pre_start()
-{
-  if [ -r /etc/profile ]; then
-    . /etc/profile
-  fi
-  if [ -r ~/.bash_profile ]; then
-    . ~/.bash_profile
-  else
-    if [ -r ~/.bash_login ]; then
-      . ~/.bash_login
-    else
-      if [ -r ~/.profile ]; then
-        . ~/.profile
-      fi
+
+error_exit() {
+    echo "Error: $1" >&2
+    exit 1
+}
+
+log() {
+    echo "[$(date +'%Y-%m-%d %H:%M:%S')] $1" >&2
+}
+
+source_if_exists() {
+    if [ -r "$1" ]; then
+        . "$1" || error_exit "Failed to source $1"
     fi
-  fi
-  return 0
+}
+
+
+pre_start() {
+    source_if_exists /etc/profile
+    if [ -r "$HOME/.bash_profile" ]; then
+        source_if_exists "$HOME/.bash_profile"
+    elif [ -r "$HOME/.bash_login" ]; then
+        source_if_exists "$HOME/.bash_login"
+    else
+        source_if_exists "$HOME/.profile"
+    fi
 }
 
 # When logging out from the interactive shell, the execution sequence is:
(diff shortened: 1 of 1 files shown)


sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+
+log() {
+    echo "[$(date +'%Y-%m-%d %H:%M:%S')] $1" >&2
+}
+
+source_if_exists() {
+    if [ -r "$1" ]; then
+        . "$1" || error_exit "Failed to source $1"
+    fi
+}
```
matt335672, line 40: The error condition above is triggered not only if the file can't be read, but if the source file returns an error from the last command executed. Unless `set -e` is in effect, this wouldn't normally cause an abort. So this looks like a change in behaviour.

As an example of what I mean, try creating temp.sh containing the single word `false`. Then:-

```
. ./temp.sh
echo $?
1
```

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-      . ~/.bash_login
+
+error_exit() {
+    echo "Error: $1" >&2
+    exit 1
+}
+
+log() {
+    echo "[$(date +'%Y-%m-%d %H:%M:%S')] $1" >&2
+}
```
matt335672, line 34: Suggest not doing this - the execution time of the file should be short, so a second-resolution timestamp will not be that useful (POSIX offers no better resolution than that). We're also invoking an external utility each time we log a line. By all means log the time at the start of the file.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+    if [ -r "$1" ]; then
+        . "$1" || error_exit "Failed to source $1"
+    fi
+}
+
+
+pre_start() {
+    source_if_exists /etc/profile
+    if [ -r "$HOME/.bash_profile" ]; then
+        source_if_exists "$HOME/.bash_profile"
```
matt335672, line 46: Following the `[ -r ...]` guard, use `.`, which any sh developer will understand and not need to look up.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-  if [ -r ~/.bash_profile ]; then
-    . ~/.bash_profile
-  else
-    if [ -r ~/.bash_login ]; then
-      . ~/.bash_login
+
+error_exit() {
+    echo "Error: $1" >&2
+    exit 1
+}
```
matt335672, line 30: The function namespace of this file is shared with files read with the dot command. You'll need to make sure that local function names are kept to a miniumum and reasonably unique, or they could be redefined within a sourced file resulting in general mayhem. For example, if a user defines `log` in their `.bashrc` it will overwrite the definition here. Maybe add a prefix to them?

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-    STARTUP=${STARTUP#Exec=*}
-    XDG_CURRENT_DESKTOP=$(grep ^DesktopNames= "/usr/share/xsessions/$1.desktop")
-    XDG_CURRENT_DESKTOP=${XDG_CURRENT_DESKTOP#DesktopNames=*}
-    XDG_CURRENT_DESKTOP=${XDG_CURRENT_DESKTOP//;/:}
-    export XDG_CURRENT_DESKTOP
-    export XDG_SESSION_DESKTOP="$DESKTOP_SESSION"
-  fi
+  
+get_xdg_session_startupcmd() {
+    if [ -n "$1" ] && [ -d /usr/share/xsessions ] && [ -f "/usr/share/xsessions/$1.desktop" ]; then
```
matt335672, line 69: Try to keep source lines within 80 characters (see C style guide). I think it is why the original was written like it is.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+        STARTUP=$(grep '^Exec=' "/usr/share/xsessions/$1.desktop" | cut -d= -f2-)
```
matt335672, line 70: Invokes two sub-processes rather than the 1 invoked by the original. Also (and this is really is an edge-case), fails to parse an executable name containing an '=' correctly. Consider using `sed -ne 's/^Exec=//p' /usr/share/xsessions/$1.desktop` which is only one process.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+        XDG_CURRENT_DESKTOP=$(grep '^DesktopNames=' "/usr/share/xsessions/$1.desktop" | cut -d= -f2- | tr ';' ':')
```
matt335672, line 71: See previous comment regarding process invocations.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-  if [ -r /etc/profile ]; then
-    . /etc/profile
-  fi
-  if [ -r ~/.bash_profile ]; then
-    . ~/.bash_profile
-  else
-    if [ -r ~/.bash_login ]; then
-      . ~/.bash_login
+
+error_exit() {
```
matt335672, line 27: In general, we try to keep sh scripts similar to our [C style guide](https://github.com/neutrinolabs/xrdp/wiki/Coding-Style). Please use separate lines for opening and closing braces in function definitions.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-    . /etc/default/locale
-    export LANG LANGUAGE
-  fi
-
-  # debian
-  if [ -r /etc/X11/Xsession ]; then
-    pre_start
+wm_start() {
+    if [ -r /etc/default/locale ]; then
+        . /etc/default/locale || error_exit "Failed to source /etc/default/locale"
```
matt335672, line 80: See comment above on `source_if_exists`. There's a few more of these which I won't comment on.

matt335672 (commented):
Hi @rowingdude

Thanks for looking at this. `sh` is a bit of a minefield.

I've had a good initial trawl through. My main concern is that there are potential changes in behaviour in some of your suggestions (in particular, the `. <file> || <error>` idiom) that could cause currently working production systems to fail. Those would need to be addressed before we could deliver this.

Most of the other comments are purely stylistic conventions, which you wouldn't have been aware of when you started this.

In any case, happy to chat. Some of this stuff is pretty esoteric, and I may not have explained myself very well on a first pass.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
 
 #start the window manager
-wm_start()
-{
-  if [ -r /etc/default/locale ]; then
-    . /etc/default/locale
-    export LANG LANGUAGE
-  fi
-
-  # debian
```
metalefty, line 83: Please do not remove comments for each Distro.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+        if [ -f /etc/X11/xinit/xinitrc ]; then
+            pre_start
+            /etc/X11/xinit/xinitrc || error_exit "Failed to execute /etc/X11/xinit/xinitrc"
+            post_start
+        else
+            error_exit "xinit package isn't installed"
+        fi
+    fi
 
-  # alpine
```
metalefty, line 109: Please do not remove comments for each Distro.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
+    if [ -r /etc/X11/xdm/Xsession ]; then
+        . /etc/X11/xdm/Xsession || error_exit "Failed to source /etc/X11/xdm/Xsession"
+        exit 0
+    elif [ -r /usr/etc/X11/xdm/Xsession ]; then
+        . /usr/etc/X11/xdm/Xsession || error_exit "Failed to source /usr/etc/X11/xdm/Xsession"
+        exit 0
     fi
-  fi
 
-  # el
```
metalefty, line 122: Please do not remove comments for each Distro.

sesman/startwm.sh
```diff
@@ -23,124 +23,102 @@
...
-  if [ -r /etc/X11/xinit/Xsession ]; then
     pre_start
-    . /etc/X11/xinit/Xsession
+    xterm || error_exit "Failed to start xterm"
     post_start
-    exit 0
-  fi
-
-  # suse
```
metalefty, line 130: Please do not remove comments for each Distro.

metalefty:
> I have tested this on Gentoo, Arch, and Fedora. All seem to work fine.

Refactoring must not break anything. Testing that newly added stuff is working as expected is important, however, it's much more important to test that your change doesn't break anything. Please focus on testing on Debian, Alpine, and SUSE. I would say that's more important.

rowingdude:
Holy cow, thank you so much for your feedback. I have never received such thorough feedback on any project I've worked on.

sesman/startwm.sh
rowingdude: This one is force of habit, apologies.

sesman/startwm.sh
rowingdude: The original block of logic was an inspiration for even undertaking this. I was trying to keep what you guys wrote but more posix-ize it. I suppose I'll need to retweak it tonight.

sesman/startwm.sh
rowingdude: I honestly hadn't thought of an '=' being within an executable filename. :-/

sesman/startwm.sh
rowingdude: >if a user defines log in their .bashrc

Thank you for this. Once again, I was too concerned with the semantic styling and breaking things down to consider the ramifications the bash initialization process.

I guess, tunnel vision applies here?

sesman/startwm.sh
rowingdude: I'm a little humbled now, I think I'll take my ideas and re-read the `sh` and `bash` manuals. Thank you again.

matt335672:
@rowingdude - thanks for taking the trouble to contribute. It's appreciated.

This area is a bit of a mess IMO. Personally, I think the biggest problem is we're assuming the user is using a compatible shell, and we're emulating stuff that a login shell can do perfectly well anyway. So this script is trying to do two different things. In the EPEL RPM, they've come up with a workaround for this, and they supply their own `startwm.sh`.

I think what we should probably be doing is running the user's chosen shell from the passwd entry as a login shell, and then calling a  simplified `startwm.sh` from that.

As a matter of interest, what were the problems you encountered with Gentoo?

PR #3239, opened 2024-09-09
Regression: Support fixed-size VNC sessions

matt335672:
This is a regression introduced in v0.10.x

This version introduced a state machine to handle resizes requested by the client and the server. Most configurations support resizeable sessions, but one that doesn't is xrdp connecting to x11vnc on (e.g.) a Raspberry PI.

If the session size requested by a client is differnt from the x11vnc size, an error is logged and the state machine fails to complete, resulting in a black screen.

This PR handles the problem by queueing a resize to the supported server size and then continuing with the state machine. It's not an optimal solution, but involves the least change to v0.10.x code.


metalefty (approved):
Thanks, let's fix this by the next v0.10 release.

PR #3273, opened 2024-10-14
[V0.10] Remove xrdp_sec_in_mcs_data() function

matt335672:
The function xrdp_sec_in_mcs_data() parses data within the TS_UD_CS_CORE struct which could just as easily be parsed when xrdp_sec_process_mcs_data_CS_CORE() is called.

Currently the contents of the MSC Connect Initial PDU are stored within the client_mcs_data member of the xrdp_sec struct. This stream is parsed once by xrdp_sec_process_mcs_data() and then separately by xrdp_sec_in_mcs_data(). There is no reason not to perform the parse in a single pass through the stream.

This commit folds the functionality in xrdp_sec_in_mcs_data() into xrdp_sec_process_mcs_data_CS_CORE() and removes xrdp_sec_in_mcs_data()

(cherry picked from commit 10cf8f2d1c156dc23048c14205d94dbd588eb1de)

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -1728,7 +1728,6 @@ xrdp_sec_process_mcs_data_CS_CORE(struct xrdp_sec *self, struct stream *s)
     int highColorDepth;
     int supportedColorDepths;
     int earlyCapabilityFlags;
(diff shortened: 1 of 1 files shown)


metalefty (approved):
Let's ship this to v0.10.

PR #3275, opened 2024-10-14
Move to fuse3

### LINKED ISSUE #3274: Move to FUSE3
See [Debian BTS #1048879](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1084479)

@Natureshadow / @alexmyczko - would one of you like to add a reference to this issue to the Debian bug tracker? Thanks.

matt335672:
Fixes #3274 

Tested on Ubuntu 22.04 (8 thread VM) by building an old version of xrdp on a Windows laptop over the network. This config is never going to win any speed awards. FWIW, timings seem very similar between FUSE2 and 3. Only one run was made with each config.

| Threads (i.e. make -j n) | FUSE3 | FUSE 2|
|-----|-----|-----|
| 1 | 6:41 | 6:47 |
| 2 | 4:30 | 4:29 |
| 4 | 3:39 | 3:43 |
| 8 | 3:26 | 3:34 |

This could be back-ported to v0.10 but that's TBD. I can' see any great advantage in doing so yet.

@bsmojver - are you aware of any plans in Fedora to deprecate FUSE2?


bsmojver:
Not aware of such plans, but I don't claim to be an authority on that. With each release of Fedora, a set of proposed and accepted changes is published, so we can keep an eye on this relatively easily.

PS. This seems like a pretty old thread, with no definitive resolution: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/UUOKLRLQHN2AC4SLAXERC7QAKW6FBSFX/

tsz8899:
make error
~~~
CC       chansrv_fuse.o
chansrv_fuse.c:899:21: error: ‘enum fuse_log_level’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 xfuse_log_func(enum fuse_log_level fuse_log_level, const char *fmt, va_list ap)
                     ^~~~~~~~~~~~~~
chansrv_fuse.c:899:36: error: parameter 1 (‘fuse_log_level’) has incomplete type
~~~

The compilation succeeded after canceling the function.
not notice any change in the transmission speed.

FUSE 3 support multi-threaded transmission, and can it improve transmission speed?
Thank you.

matt335672:
Thanks @tsz8899 

What platform are you compiling on? 

I've seen no improvement in speed here, but I'm just trying compilations, which exercise a lot of the redirected functions.

tsz8899:
fuse 3.4.2

b5772ce make ok

matt335672:
I'm going to merge this into devel so it gets a bit of use.

I'll set up a PR for a backport to V0.10, but whether we merge that or not is TBD. If we don't it at least gives distros an easy way to generate a patch for FUSE3.

Issue #3280, opened 2024-10-20
Graphic glitches when running targetting a windows RDP server from within an xrdp session


datiscum:
### xrdp version

0.10 and higher

### Detailed xrdp version, build options

~~~
xrdp 0.10.80
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-x264
      --enable-fuse
      --enable-jpeg
      --enable-rfxcodec
      --enable-mp3lame
      --enable-vsock
      --enable-painter
      --enable-pixman
      --enable-tjpeg
      --enable-fdkaac
      --enable-opus
      CFLAGS=-O3 -march=native
      CXXFLAGS=-O3 -march=native
      LDFLAGS=-flto

  Compiled with OpenSSL 1.1.1f  31 Mar 2020
~~~

### Operating system & version

Ubuntu 20.04

### Installation method

Doesn't matter

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

KDE

### Environment xrdp running on

Incus container

### Area(s) with issue?

Graphic glitches

### Steps to reproduce

Hello,
I have been using Xrpd/XorgXrdp for more than 5 years. I have always compiled it myself and the error described here is present from version 0.10 onward.

This is my main workstation open 14 hours a day continuously and only disconnected in the evening, the session always runs for several weeks. Client RaspberryPi5 resolution 2560x1440 does not play a role in the error.

I have compiled version 0.10.0 / 0.10.1 and the master branch for testing. The error described here does not occur in any earlier version before 0.10. Tested with 0.9.26 / 0.9.21. I have also tested with different FreeRdp versions (2.10/3.8), which showed no differences.

The error only occurs in special constellations. 

 If I open an RDP session in the Xrdp session or operate the screen of a VM with Virtmanager or Spiceviewer. The prerequisite is also that the mouse cursor is in the window of the RDP session. With a Windows login sequence in which the mouse is outside the remote desktop window, for example, there are no character problems. If the mouse is only in the window, even without movement, there are massive drawing problems. Which are also very similar to those in this bug report https://github.com/neutrinolabs/xrdp/issues/3211

Exactly the same black lines across the whole screen but also others. 

[...]

Issue #3211, opened 2024-08-17
xrdp 0.10 + x11vnc on Raspberry Pi OS Bookworm 64-bit + Jump Desktop for macOS: screen redraw and scaling problems


usa-:
### xrdp version

v0.10.1

### Detailed xrdp version, build options

~~~
xrdp 0.10.1
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      

  Compiled with OpenSSL 3.0.13 30 Jan 2024
~~~

### Operating system & version

Debian GNU/Linux 12 (bookworm)

### Installation method

git clone & make install

### Which backend do you use?

Xvnc (x11vnc: 0.9.16 lastmod: 2019-01-05)

### What desktop environment do you use?

LXDE

### Environment xrdp running on

Paspberry Pi 4b

### What's your client?

Jump Desktop 8.10.7 for macOS

### Area(s) with issue?

Graphic glitches

### Steps to reproduce

sudo apt install x11vnc
x11vnc -storepasswd
Enter and save the password.

wget https://github.com/neutrinolabs/xrdp/releases/download/v0.10.1/xrdp-0.10.1.tar.gz
tar xvzf xrdp-0.10.1.tar.gz
cd xrdp-0.10.1
./bootstrap
./configure
make
sudo make install
cd ..

wget https://github.com/neutrinolabs/xorgxrdp/releases/download/v0.10.2/xorgxrdp-0.10.2.tar.gz
tar xvzf xorgxrdp-0.10.2.tar.gz
cd xorgxrdp-0.10.2
./bootstrap
./configure
make
sudo make install

Edit /etc/xrdp/xrdp.ini.
Remove all connection (backend) definitions.
Add a connection (backend) definition.

```
[Xvnc]
name=Xvnc
lib=libvnc.so
username=ask
password=ask
ip=127.0.0.1
port=5900
```

sudo systemctl daemon-reload
sudo service xrdp restart
x11vnc -display :0 -forever -repeat -noxdamage -usepw
Connect with Jump Desktop.

### ✔️ Expected Behavior

Screen is being redrawn properly on every connection

### ❌ Actual Behavior

[...]

PR #3281, opened 2024-10-21
Ask for no videos on the bug form

matt335672:
I'm proposing we ask for no links to external websites for images or videos on the bug report form.

This has bothered me for a while. Not only are these not guaranteed to be available for future readers, but a link to an external website could be used to compromise a developer browser, or find out more information about the developer workstation configuration.

What do people think?

Header on updated bug report form looks like this:-

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -8,6 +8,7 @@ body:
       value: |
         Before submitting a bug, read the [FAQ](https://github.com/neutrinolabs/xrdp/wiki/Tips-and-FAQ).
         **In particular, on systemd-based systems. make sure you are not logged in on the console as the same user you are trying to use for xrdp**
+        Please do not include links to images or videos on external websites. These are not guaranteed to always be available, and could be used to compromise web browsers.
   - type: input
     attributes:
       label: xrdp version
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -6,9 +6,9 @@ body:
   - type: markdown
     attributes:
       value: |
-        Before submitting a bug, read the [FAQ](https://github.com/neutrinolabs/xrdp/wiki/Tips-and-FAQ).
-        **In particular, on systemd-based systems. make sure you are not logged in on the console as the same user you are trying to use for xrdp**
-        Please do not include links to images or videos on external websites. These are not guaranteed to always be available, and could be used to compromise web browsers.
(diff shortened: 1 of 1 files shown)


metalefty (approved):
I generally agree. BTW, Videos on YouTube will be acceptable for me if its file size is larger than GitHub accepts.

matt335672:
Thanks @metalefty 

Here's the scenario which bothers me. Some might accuse me of being paranoid, but we are developing software which runs with root privilege on users' machines.

Using the usual names, we have 'Alice' for an xrdp developer or long-time contributor, and 'Mallory' for an attacker.

1. Mallory raises issue/discussion
2. Alice attempts to address the issue
3. Mallory engages with Alice, explains the issue isn't resolved and posts helpful link to video in a reply. Video is hosted on website controlled by Mallory.
4. Alice clicks on link to watch video. At this point, from the connection, Mallory learns Alice's IP address, browser, operating system, etc. Mallory crafts second video containing an exploit targeted at Alice's machine.
5. Later in the conversation, Alice watches the second video and the exploit is delivered.
6. Mallory immediately replaces the video with a harmless version, or simply vanishes leaving no trail to follow.

We've had issues raised which follow this stages 1-3 of this pattern (e.g. #3211 / #3280). **To be absolutely clear, I'm not accusing any of our user community with malicious intent at this stage.** I'm simply saying that 1-3 are an expected path of events when we are triaging user issues, and so as developers we are less likely to be on our guard.

With that rather long-winded post out of the way, I think I'm OK with youtube postings too. It removes stage 4 from the attacker, and the attacker no longer has direct control over the video. It's still not perfectly safe, but it's a lot better:-

https://www.pandasecurity.com/en/mediacenter/youtube-virus-tips/

I'll reword the text to add an exception for youtube.

matt335672:
Updated form:-

image

metalefty:
Yup, your suggestion concern makes sense to me. Videos are helpful to see what's happening on the user side exactly but there are some risks like you mentioned.

BTW, I'm thinking of how can I ensure reporters let us know the result of `xrdp -v` in the issue form.
Some people paste nothing there, some people write a question there, some people removes `~~~` quotation regardless it is noted not to remove it. I'm not seriously suffering from that actually :) Most of reporters fill the issue form as intended.

Anyway, you changes to the issue form LGTM.

datiscum:
Please specify the size limit of Github with 10MByte in the text, otherwise it is like me, that everything is ready and a 20MByte video can not be uploaded. Unfortunately, I was only told this when I wanted to upload it to Github. 10 MByte is really little, but if I had known the limit, I could have kept to it.

matt335672:
@datiscum - the 10Mb limit is already mentioned in the text. Do you think it could be better worded?

datiscum:
The limit of 10Mbyte could be emphasized a bit more. I don't have a Youtube account and wouldn't set it up for that reason.

PR #3284, opened 2024-10-23
[V0.10] Move to fuse3

matt335672:
Backport of #3275 to V0.10

To summarise; this is a result of [Debian BTS #1048879](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1084479). The functionality is merged into devel, but it's to be decided whether we merge it into V0.10.

Before any merge, to get a patch which can be applied to v0.10 containing this functionality, use this git command:-

```
git diff --merge-base v0.10 v0_10_move_to_fuse3
```


metalefty:
@matt335672 What do you think about this one? This is somewhat of an incompatible change, but it is better than merging it much later, for example, in v0.10.10. We're making proactive changes in this release, such as H.264 and unprivileged daemon. I believe FUSE 3 is available in most supported versions of every distro. So I would say let's go with this.

matt335672:
I'm happy to - it seems pretty stable in `devel`.

PR #3312, opened 2024-11-15
Re-introduce -DHAVE_BOOST to cppcheck builds

matt335672:
A regression introduced in cppcheck 2.15.0 has been fixed in 2.16.0 - see @firewave's [comment](/neutrinolabs/xrdp/pull/3291#issuecomment-2477485731) in #3291

### DIFF
--- a/scripts/install_cppcheck.sh
+++ b/scripts/install_cppcheck.sh
@@ -149,8 +149,12 @@ fi
             # Cppcheck 2.8 added optional support for utilizing Boost
             make_args="$make_args CPPFLAGS=-DHAVE_BOOST"
             ;;
-        2.*)
+        2.15.*)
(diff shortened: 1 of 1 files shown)


firewave:
This did not rebuild the binary since your forgot to adjust the cache key.

matt335672:
That's a good point - thanks.

The binary did get rebuilt for my branch, as I used a different branch name:-

https://github.com/matt335672/xrdp/actions/runs/11858253988/job/33048444115

So I'm pretty sure the modification above is fine. 

I've just deleted the old cache at https://github.com/neutrinolabs/xrdp/actions/caches (see https://stackoverflow.com/questions/63521430), so I think we're all OK now, at least as far as this repository is concerned. Users with forks may also need to do this for branches they are working on.

PR #3317, opened 2024-11-18
[v0.10] Set different frame capture interval for H.264 and RFX and pass them to xorgxrdp

metalefty:
See [neutrinolabs/xorgxrdp#347](https://github.com/neutrinolabs/xorgxrdp/pull/347) for background.

To include this in the next v0.10 release, making a PR on `v0.10-h264` branch.

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -192,6 +192,11 @@ struct xrdp_client_info
     char variant[16];
     char options[256];
 
+    /* xorgxrdp: frame capture interval (milliseconds) */
+    int rfx_frame_interval;
+    int h264_frame_interval;
+    int normal_frame_interval;
+
     /* ==================================================================== */
     /* Private to xrdp below this line */
     /* ==================================================================== */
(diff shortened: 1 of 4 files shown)


common/xrdp_client_info.h
```diff
@@ -241,6 +241,11 @@ struct xrdp_client_info
 
     int pad1; /* unused; unicode_input_state */
     enum xrdp_capture_code capture_code;
+
+    /* xorgxrdp: frame capture interval (milliseconds) */
+    int rfx_frame_interval;
+    int h264_frame_interval;
+    int normal_frame_interval;
```
matt335672: These values need to be above the following comment in this file:-

```c
/* ==================================================================== */
    /* Private to xrdp below this line */
    /* ==================================================================== */

```

common/xrdp_constants.h
```diff
@@ -86,6 +86,11 @@
 #define MCS_SDRQ                       25 /* Send Data Request */
 #define MCS_SDIN                       26 /* Send Data Indication */
 
+/* xorgxrdp: frame capture interval (milliseconds) */
+#define RFX_FRAME_INTERVAL             32
+#define H264_FRAME_INTERVAL            16
+#define NORMAL_FRAME_INTERVAL          40
+
```
matt335672, line 93: Small point, but it might be worth calling these `DEFAULT_xxx_FRAME_INTERVAL` so that someone who is reading this code fragment realises they can be overridden.

metalefty: Makes sense.

common/xrdp_client_info.h
metalefty: Indeed.

matt335672 (approved):

PR #3318, opened 2024-11-18
Set different frame capture interval for H.264 and RFX

metalefty:
and pass them to xorgxrdp.

(cherry picked from commit c55694c51b570d84ad63bfb5dc7bdec75fee57da)

----
Forward port of #3317.


matt335672:
See #3317 for comments.

metalefty:
H.264 changes on devel branch will be noted separately, merging.

PR #3319, opened 2024-11-18
Colemak support

### LINKED ISSUE #2192: Feature Request:   Colemak Support?
Hello,

Xrdp currently supports US QWERTY and US Dvorak, but there aren't any colemak layouts available.  I realize that colemak is not supported by Microsoft Windows natively nor is it included in the standard RDP protocol, but it would be nice to be able to use colemak with xrdp.
This issue is to pose the question: would it be feasible to add a colemak keyboard layout in xrdp that freerdp and other clients could potentially be patched to use?

If this were to be implemented, it looks like something along the lines of the following would be required (with probably something else that I'm missing):

- `xrdp_keyboard.ini`  rdp_us_colemak (and perhaps rdp_us_colemak_dh) given unused IDs and added to `[default_rdp_layouts]`. 
- `dump-keymaps.sh` modified to dump colemak from `setxkbmap`
- whatever clients would need to be patched to allow selection of colemak with the xrdp ids

Is there any interest in implementing something like this in xrdp?  I realize that it is not included in the spec and would not be compatible with microsoft RDP, but it shouldn't break anything.  I'd be interested in getting thoughts and opinions on this

Thanks again for everyone's hard work on xrdp

matt335672:
Fixes #2192 

This would be simple, but Colemak on X11/UNIX dispenses with Caps Lock and replaces it with another backspace key.

Given the complexity of this, I can't see it being back-ported to v0.10

One outstanding (small) issue I am aware of is that the capslock backspace function doesn't auto-repeat. This may be a Windows problem, as the Windows Colemak mapping doesn't remap the caps lock key.

@prgreadonly - are you able to build from source?


thom-yorkes-brother:
I just tested with Windows 11 24H2 and its built-in colemak layout and everything seems to work!  I tested with the Xvnc backend.  Let me know if there's anything specific I can help test

matt335672:
@prgreadonly - the only way to give this a proper go is to build this branch from source. Are you able to do that?

The additional complexity is because of the need to remove caps lock and add an additional backspace. The existing v0.9.x and v0.10.x code won't let us do that.

thom-yorkes-brother:
> @prgreadonly - the only way to give this a proper go is to build this branch from source. Are you able to do that?
> 
> The additional complexity is because of the need to remove caps lock and add an additional backspace. The existing v0.9.x and v0.10.x code won't let us do that.

Yes - I've built your matt335672:colemak_support branch of xrdp to test and Windows 11's built-in colemak keyboard layout seemed to work. Thank you for the effort!

matt335672:
OK - that's great. Sorry for the misunderstanding.

@metalefty - are you happy to merge this into devel? It's too complex to backport to v0.10 I think.

metalefty:
@matt335672 Ah, I missed the notification. Sure. I think it's a good idea to merge this only into devel.

PR #3324, opened 2024-11-28
[v0.10] add OpenH264 support

metalefty:
- Add OpenH264 support (cherry-pick of #3311)
- Enable to select preferred H.264 encoder in `gfx.toml`
- Make OpenH264 encoding parameters configurable in `gfx.toml`
- Add tests for OpenH264 params and prefered H264 encoder
- Add `--enable-openh264` to CI

@matt335672 Could you review this? 

This works very well on my side. Comparing x264 and OpenH264 is easy. Just edit `h264_encoder` in `gfx.conf` and reconnect. It looks x264 is smoother and faster but OpenH264 can control bitrate stricter than x264 because OpenH264 can skip frames so as not to exceed the max bitrate. 

BTW, we forgot to add gfx_missing_h264.toml to `EXTRA_DIST` before. The test passed. The result is the same as expected, but the process is not what we expect.
```
ok 18 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_codec_order: Passed
PASS: test_xrdp 18 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_codec_order: Passed
[2024-11-28T14:35:36.509+0000] [ERROR] TConfig: Error loading GFX config file ../../../../tests/xrdp/gfx//no_such_file.toml (No such file or directory)
ok 19 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_missing_file: Passed
PASS: test_xrdp 19 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_missing_file: Passed
[2024-11-28T14:35:36.510+0000] [ERROR] TConfig: Error loading GFX config file ../../../../tests/xrdp/gfx//gfx_missing_h264.toml (No such file or directory)
ok 20 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_missing_h264: Passed
PASS: test_xrdp 20 - ../../../../tests/xrdp/test_tconfig.c:xrdp_tconfig_load_gfx:test_tconfig_gfx_missing_h264: Passed
[2024-11-28T14:35:36.511+0000] [INFO ] TConfig: Loading GFX config file ../../../../tests/xgg/uurdp/gfx//gfx.toml
[...]

PR #3327, opened 2024-12-04
add OpenH264

metalefty:
Forward of #3324


matt335672 (approved):
LGTM - this area of devel hasn't changed that much since we released v0.10.x

PR #3329, opened 2024-12-04
Merge v0.10-h264 into v0.10

metalefty:
I think the software H.264 functionality is complete so let's merge it to ship to the next release. 

The remaining task id documentation:
- Documentation updates
- Add man pages for gfx.toml

I started to write documentation on the wiki. It will be linked from the release note.
- https://github.com/neutrinolabs/xrdp/wiki/H.264-support

PR #3330, opened 2024-12-04
capture_code is not private to xrdp

metalefty:
@matt335672 pointed out this at https://github.com/neutrinolabs/xrdp/pull/3329#discussion_r1869803455


matt335672:
@metalefty - now we've just had a release it may be a good time to check all the client info and merge this along with (possibly) other changes.

I'm busy IRL for a couple of days, but I'll have a look at it when I've got some time.

metalefty:
Sounds good.

matt335672:
This has been already addressed by #3534, so we can close this one.

PR #3336, opened 2024-12-11
[regression] code to open local display in waitforx is broken by recent libxcb

matt335672:
Thanks to @derekschrock for raising this as a comment to #3297 

Commit 80fab03198854268fc2258d7bdea2013e5d109f7 introduced a way to prevent waitforx going to the network when trying to open a display, and hence potentially blocking.

This method turned out to be invalidated by libxcb version 1.16 and 1.17

This change adds an explicit check that the Unix socket for the display in `/tmp/.X11-unix/Xn` is open before trying to connect to display ':n'. This has the same effect.

In the original change I was keen to avoid a dependency on the string `"/tmp/.X11-unix/X%d"` for converting a display number to a unix socket. This was unjustified, as many other utilities assume this string, including ssh:-

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy
$  strings $(which ssh) | grep X11-unix
/tmp/.X11-unix/X%u
```

### DIFF
--- a/common/xrdp_sockets.h
+++ b/common/xrdp_sockets.h
@@ -61,4 +61,10 @@
 #define XRDP_X11RDP_STR       XRDP_SOCKET_PATH "/" XRDP_X11RDP_BASE_STR
 #define XRDP_DISCONNECT_STR   XRDP_SOCKET_PATH "/" XRDP_DISCONNECT_BASE_STR
 
+/* Where X11 stores its Unix Domain Sockets (unlikely to change) */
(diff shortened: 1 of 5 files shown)


matt335672:
I'm going to merge this now as it fixes an existing problem in the devel branch and it's been informally tested already by @derekschrock

Issue #3187, opened 2024-07-31
Release tarball contains wrong sesman.ini and xrdp.ini instead of sesman.ini.in and xrdp.ini.in


metalefty:
I found a long-lived problem occurred by #1036 similar to #3149.

Release tarball should contain `sesman.ini.in` and `xrdp.ini.in` for substitution on the builder's side. However, it contains substituted `sesman.ini` and `xrdp.ini` instead on the developer's side when making a release tarball by `make dist`.

```
% tar ztf xrdp-0.10.1.tar.gz | egrep "(xrdp|sesman).ini"
xrdp-0.10.1/xrdp/xrdp.ini
xrdp-0.10.1/docs/man/sesman.ini.5.in
xrdp-0.10.1/docs/man/xrdp.ini.5.in
xrdp-0.10.1/sesman/sesman.ini
```

This is not intended and, as a result, `sesman.ini` and `xrdp.ini` are not generated ones to suit the actual environment. This problem affects package maintainers, not end-users. The package maintainer may have had to write a patch that was not necessary.

I don't intend to make another release v0.10.1.1 for this problem at the moment because it mostly affects package maintainers. I think power-users who build and install xrdp by themselves would prefer  `git clone` rather than tarball so not affected.  It will be fixed in the regular v0.10.2 release. @matt335672 What do you think?

Here is the suggested patch.

```diff
diff --git a/sesman/Makefile.am b/sesman/Makefile.am
index 7b34470c..8d9d3102 100644
--- a/sesman/Makefile.am
+++ b/sesman/Makefile.am
@@ -27,6 +27,7 @@ xrdp_sesman_SOURCES = \
   scp_process.h \
   sesman.c \
   sesman.h \
+  sesman.ini.in \
   sesexec_control.c \
   sesexec_control.h \
   session_list.c \
@@ -52,7 +53,7 @@ SUFFIXES = .in
 .in:
 	$(subst_verbose)$(SUBST_VARS) $< > $@
 
-dist_sesmansysconf_DATA = \
+nodist_sesmansysconf_DATA = \
   sesman.ini
 
 dist_sesmansysconf_SCRIPTS = \
@@ -64,3 +65,5 @@ SUBDIRS = \
   sesexec \
   tools \
   chansrv
+
+CLEANFILES = $(nodist_sesmansysconf_DATA)
diff --git a/xrdp/Makefile.am b/xrdp/Makefile.am
index 3e11f4ee..4de3f23e 100644
[...]

PR #3340, opened 2024-12-12
make xrdp-chkpriv script fancy

metalefty:
This PR also includes the following changes:

- Fix for make dist (similar fix to #3187)
    - Do not include substituted xrdp-chkpriv into tarball
    - Dot not install xrdp-chkpriv.in
- Adjust permission of the *.c source file and scripts

----

Here are actual outputs of xrdp-chkpriv script. Now it's obvious if xrdp is unprivilegeable.

image

image

image


matt335672 (approved):
I like it!

I'd not come across the `[ -t 1 ]` trick before.

PR #3341, opened 2024-12-13
[v0.10] Merge unprivileged stuff to v0.10

metalefty:


matt335672 (approved):
This seems OK to me. As far as I can tell, it's unchanged from the changes we made for devel.

I ran it up, and it seems to work fine.

PR #3344, opened 2024-12-16
[V0.10] Allow a path to be specified for the chansrv log file

### LINKED ISSUE #3337: enable dynamic CHANSRV_LOG_PATH or/and possibility to change log name "xrdp-chansrv.%d.log"
### xrdp version

0.10.1

_No response_

Ubuntu 22.04

### Installation method

### Which backend do you use?

Xorg

Openbox

_No response_

### What's your client?

_No response_

### Area(s) with issue?

_No response_

### Steps to reproduce

Hi,

In our environment home dirs are NFS based, and I would like to avoid writing log files there to avoid race-condition issues if a user connects to several machines with xrdp and get same DISPLAY number.

There is a partial workaround to it by setting:

[SessionVariables]
CHANSRV_LOG_PATH=/tmp

But this is still not good enough, as it can create race-condition on the /tmp/xrdp-chansrv.%d.log file,
if user logged out, the log file is not being deleted, and new user get the old display number and /tmp/xrdp-chansrv.%d.log remined own by the last user, so it can't be overriding.

1. can somehow the CHANSRV_LOG_PATH variable be handled dynamically?
   (like for example the **FuseMountName=/run/user/%u/xrdp/drives**)
  we would like to set it as **/run/user/%u/xrdp/** so it can be resolved to **/run/user/%u/xrdp/xrdp-chansrv.%d.log**

2. can the logfile name be controlled from the config? so one can add username or pid to the filename for example?
    ex. **xrdp-chansrv.%d.%u.%p.log**

Currently we have a solution by renaming /usr/sbin/xrdp-chansrv to /usr/sbin/.xrdp-chansrv
and creating /usr/sbin/xrdp-chansrv as a wrapper for dynamic CHANSRV_LOG_PATH.

```
#!/bin/sh
export CHANSRV_LOG_PATH="/run/user/$(id -u)/xrdp"
exec -a "xrdp-chansrv" "/usr/sbin/.xrdp-chansrv" "$@"
```

Also, another best practice suggestion, move executables like xrdp-chansrv and xrdp-sesman that
are not supposed to be run outside of xrdp to be placed in **/usr/libexec/xrdp/** and not in **/usr/sbin** which is part of users PATH.

Thanks!

### ✔️ Expected Behavior

_No response_

### ❌ Actual Behavior

_No response_

### Anything else?

_No response_

matt335672:
Fixes #3337 

A parameter `LogFilePath` is added to the `[ChansrvLogging]` section in sesman.ini. Manpage excerpt:-

```
       LogFilePath=string
              Directory  for  storing  the chansrv log file. This setting only
              applies to chansrv. The sesman log file  is  always  created  in
              /var/log.

              Created  if  it doesn't exist.  If first character is not a '/',
              this is relative to $HOME, where chansrv is normally started.

              The following substitutions are made in this string:-
                  %U - Username
                  %u - Numeric UID
                  %% - Percent character

              This is most useful if you are using NFS-mounted  home  directo‐
              ries, and wish to move the chansrv log file to the local disk.

              If this parameter isn't specified, the log file is stored in one
              of the following locations :-
                  -   $CHANSRV_LOG_PATH
                  -   $XDG_DATA_HOME/xrdp
                  -   $HOME/.local/share/xrdp
```

### DIFF
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -98,7 +98,33 @@ stdout. Use for debugging only\fR
 
 It is ignored in the [ChansrvLogging] section
 since the channel server creates one log file per display and instead uses the
-following log file naming convention \fIxrdp-chansrv.${DISPLAY}.log\fR
+following log file naming convention \fIxrdp-chansrv.${DISPLAY}.log\fR. For
+details of the chansrv log file location, see \fBLogFilePath\fR.
+
+.TP
+\fBLogFilePath\fR=\fIstring\fR
+Directory for storing the chansrv log file. This setting only applies to
+chansrv. The sesman log file is always created in \fI@localstatedir@/log\fR.
+
+Created if it doesn't exist.
+If first character is not a '/', this is relative to $HOME, where
+chansrv is normally started.
+
+.RS
+The following substitutions are made in this string:-
+    %U - Username
+    %u - Numeric UID
+    %% - Percent character
+
+This is most useful if you are using NFS-mounted home directories, and
+wish to move the chansrv log file to the local disk.
+
+If this parameter isn't specified, the log file is stored in one of
+the following locations :-
+    -   $CHANSRV_LOG_PATH
+    -   $XDG_DATA_HOME/xrdp
+    -   $HOME/.local/share/xrdp
+.RE
 
 .TP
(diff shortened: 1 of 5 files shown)


metalefty (approved):
Approved. This one and Debian pam thing should be shipped to v0.10.2.

PR #3346, opened 2024-12-16
Add restartability to sesman

### LINKED ISSUE #800: Running sessions "lost" if xrdp-sesman restarts
If xrdp restarts then the new process is not aware of any sessions created by the old process, so you can't re-attach to them and they aren't listed by xrdp-sesadmin.  Would it be possible for the new process to "discover" them and populate g_sessions?

Test case:

1. Connect to a new xorgxrdp (Xorg) session.  
`xfreerdp /v:localhost /u:<user> /p:<password> &`

2. List sessions using xrdp-sesadmin.  The session created above will be listed.
`xrdp-sesadmin -u=<user> -p=<password> -c=list`

3. Restart xrdp.  Note that the existing sessions are not disconnected.
`sudo service xrdp restart`

4. List sessions using xrdp-sesadmin.
`xrdp-sesadmin -u=<user> -p=<password> -c=list`
Expected behaviour: the session created above should be listed.
Observed behaviour: "No sessions." (or, prior to #797, "Error getting session list.")

5. Disconnect the RDP client (without logging out) from the RDP session.

6. Attempt to reconnect to the existing session.
`xfreerdp /v:localhost /u:<user> /p:<password> &`

(Workaround: to reconnect to the "orphaned" sessions you can set up xvnc to view the appropriate X display.)

Note that the new xrdp would somehow have to obtain the session_item info from each session it discovers.  Also the new xrdp wouldn't be the parent of the sesman for each session so it wouldn't receive SIGCHLD when the session ends (i.e. it would have to poll).  I would suggest solving both of these using a shared mmap() object.

matt335672:
This is now ready for review.

Fixes #800 

Depends on #3328

See also #819 

This PR allows for sesman to be restarted without losing sessions.

A restart of xrdp will still drop connections, but these can easily be restarted.

In brief, the design is this (I can add more details around this):-

1) When sesexec starts a session, a listening socket is created in a directory (the 'restart directory') accessible by root only.
2) When sesman restarts, before starting its listening socket it iterates over the sockets in the restart directory, and attempts to connect to each one.
3) When sesexec receives the connection request, it validates the other end is running as root, and if so, re-announces the managed session.
4) sesman adds the announced sessions to its session list.
5) sesman restarts its listening socket.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -37,6 +37,7 @@
 #define ctid_t id_t
 #endif
 #include <unistd.h>
+#include <dirent.h>
 #include <errno.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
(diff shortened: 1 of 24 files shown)


metalefty:
I'll try this next.

matt335672:
I've added a minor update to this wiki page documenting the recovery directory used by this PR.

https://github.com/neutrinolabs/xrdp/wiki/The-socketdir-directory

PR #3358, opened 2024-12-23
[V0.10] add czech keyboard

matt335672:
Fixes #3348 

Not required for v0.10.2 which is at RC status. We can merge after v0.10.2 ships

PR #3359, opened 2024-12-23
Implement Czech keyboard

matt335672:
Forward port of #3358 to devel


metalefty (approved):
The keymap file is quite different from #3358 because it is TOML now but LGTM.

matt335672:
Thanks.

For anyone else reading this, there's a manpage for the new TOML keymap format `xrdp-km.toml(5)`

PR #3367, opened 2024-12-26
[v0.10] make x264 threads configurable

metalefty:
This may be my last PR in 2024.


metalefty:
I'll work on OpenH264 later, merging this.

PR #3368, opened 2024-12-27
Allow to change config file (sub)directory

metalefty:
This allows the `xrdp` part of the path `/etc/xrdp` where config files are placed to be customizable. This change is useful when trying the stable version and the devel version alternately.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -24,7 +24,7 @@ include_HEADERS = \
   xrdp_sockets.h
 
 AM_CPPFLAGS = \
-  -DXRDP_CFG_PATH=\"${sysconfdir}/xrdp\" \
+  -DXRDP_CFG_PATH=\"${sysconfdir}/${sysconfsubdir}\" \
   -DXRDP_SBIN_PATH=\"${sbindir}\" \
   -DXRDP_SHARE_PATH=\"${datadir}/xrdp\" \
   -DXRDP_PID_PATH=\"${localstatedir}/run\" \
--- a/configure.ac
+++ b/configure.ac
@@ -610,6 +610,12 @@ AC_ARG_WITH([pamconfdir],
                   [], [with_pamconfdir="$sysconfdir/pam.d"])
 AC_SUBST([pamconfdir], [$with_pamconfdir])
 
+AC_ARG_WITH([sysconfsubdir],
+  [AS_HELP_STRING([--with-sysconfsubdir=DIR],
+                  [Use subdirectory for config files (default: xrdp)]))],
+                  [], [with_sysconfsubdir="xrdp"])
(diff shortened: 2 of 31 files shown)


matt335672:
That's a great idea, and better than the solution I've hacked together on my development workstation.

Also there's a couple of /etc/xrdp/ strings embedded in C files. You may want to sweep these up if it's not too difficult. Here's some links to your repo:-

- https://github.com/metalefty/xrdp/blob/854d060917c207177205f5d9e086081889a8cd8d/keygen/keygen.c#L466
- https://github.com/metalefty/xrdp/blob/854d060917c207177205f5d9e086081889a8cd8d/genkeymap/genkeymap.c#L96

metalefty:
I have no idea why cppcheck complains to the former but doesn't to the latter. Anyway, I can pet cppcheck.

```c
fprintf(stderr, "Example: %s -r evdev -c \"en-US pc104 keyboard\" "
        XRDP_CFG_PATH "/km-00000409.toml\n", programname);
```

```c
fprintf(stderr, "Example: %s -r evdev -c \"en-US pc104 keyboard\" ", programname);
fprintf(stderr, XRDP_CFG_PATH "/km-00000409.toml\n");
```

matt335672:
cppcheck has some code to try to guess values of macros it can't find a definition for. Normally it uses `1` (without quotes) which breaks some uses of string macros.

 @firewave may have something useful to contribute on this example.

Personally, I'm happy with the change you've made for cppcheck. However, if you'd rather use your first `fprintf` line you could make use of a workaround that's present further up in the file for `cppcheck` and `PACKAGE_VERSION`. It will prevent cppcheck having to guess at all:-

```patch
-- a/genkeymap/genkeymap.c
+++ b/genkeymap/genkeymap.c
@@ -49,15 +49,21 @@
 #include "scancode.h"
 #include "xrdp_constants.h"
 
-// cppcheck doesn't always set this macro to something in double-quotes
+// cppcheck doesn't always set PACKAGE_VERSION and
+// XRDP_CFG__PATH to something in double-quotes
 #if defined(__cppcheck__)
 #undef PACKAGE_VERSION
+#undef XRDP_CFG_PATH
 #endif
 
 #if !defined(PACKAGE_VERSION)
 #define PACKAGE_VERSION "???"
 #endif
 
+#if !defined(XRDP_CFG_PATH)
+#define XRDP_CFG_PATH "/etc/xrdp"
+#endif
+
 #define NUM_STATES 9
 
 #define KEYMAP_FILE_FORMAT_VERSION "2"
```

Whatever you decide to do, I'm happy for this to be merged.

metalefty:
Let's merge this as-s and make further changes if cppcheck complains again.

PR #3377, opened 2025-01-04
Fix possible memory leak

peter15914:
First, I considered a single cleanup block at the end of the function with ```ret = 1; goto cleanup;``` instead of ```return 1```.
However, I decided to keep it simple because the function is relatively small and there are only two cases of returning without freeing memory.


matt335672:
Hi @peter15914 

We're on this one - see #3363 (which is slightly more complex, but has less exit points)

PR #3378, opened 2025-01-04
Fix possible memory leak

peter15914:
Memory referenced by ```fds``` was not freed at one exit point.


matt335672:
Hi again @peter15914 

Thanks for your contribution

PR #3380, opened 2025-01-06
[V0.10] Fix potential memory leaks

matt335672:
Backports of #3363 and #3378


metalefty:
Thanks, I'm happy with this.

PR #3382, opened 2025-01-06
Add sesman.ini FuseMountNameColonCharReplacement option

Bad-ptr:
The colon(`:`) in file paths is not supported on all OSes, on others need quoting/escaping, not all software treat colons in paths adequately.
The patch modifies default behaviour by stripping colon from redirected drive mount points(which emerges from windows clients with shared drive connected), but provides config option to customize the replacement(including reverting to "good old" colons in drive names).

### DIFF
--- a/sesman/chansrv/chansrv_config.c
+++ b/sesman/chansrv/chansrv_config.c
@@ -40,6 +40,7 @@
 #define DEFAULT_RESTRICT_INBOUND_CLIPBOARD  0
 #define DEFAULT_ENABLE_FUSE_MOUNT           1
 #define DEFAULT_FUSE_MOUNT_NAME             "xrdp-client"
+#define DEFAULT_FUSE_MOUNT_NAME_COLON_CHAR_REPLACEMENT ':'
 #define DEFAULT_FUSE_DIRECT_IO              0
 #define DEFAULT_FILE_UMASK                  077
 #define DEFAULT_USE_NAUTILUS3_FLIST_FORMAT  0
@@ -220,6 +221,25 @@ read_config_chansrv(log_func_t logmsg,
                 break;
             }
         }
+        else if (g_strcasecmp(name, "FuseMountNameColonCharReplacement") == 0)
+        {
+            size_t vallen = g_strlen(value);
+            if (vallen < 1)
+            {
+                cfg->fuse_mount_name_colon_char_replacement = '\0';
+            }
+            else
+            {
+                if (vallen > 1)
+                {
+                    logmsg(LOG_LEVEL_WARNING, "FuseMountNameColonCharReplacement "
+                           "must be 1 character length, now it is '%s'."
+                           "Only first char will be used!",
+                           value);
+                }
+                cfg->fuse_mount_name_colon_char_replacement = value[0];
+            }
+        }
         else if (g_strcasecmp(name, "FuseDirectIO") == 0)
         {
             cfg->fuse_direct_io = g_text2bool(value);
@@ -315,6 +335,7 @@ new_config(void)
         cfg->restrict_outbound_clipboard = DEFAULT_RESTRICT_OUTBOUND_CLIPBOARD;
         cfg->restrict_inbound_clipboard = DEFAULT_RESTRICT_INBOUND_CLIPBOARD;
         cfg->fuse_mount_name = fuse_mount_name;
+        cfg->fuse_mount_name_colon_char_replacement = DEFAULT_FUSE_MOUNT_NAME_COLON_CHAR_REPLACEMENT;
         cfg->fuse_direct_io = DEFAULT_FUSE_DIRECT_IO;
         cfg->file_umask = DEFAULT_FILE_UMASK;
         cfg->use_nautilus3_flist_format = DEFAULT_USE_NAUTILUS3_FLIST_FORMAT;
@@ -419,6 +440,7 @@ config_dump(struct config_chansrv *config)
     g_writeln("    EnableFuseMount            %s",
               g_bool2text(config->enable_fuse_mount));
     g_writeln("    FuseMountName:             %s", config->fuse_mount_name);
+    g_writeln("    FuseMountNameColonCharReplacement:             %c", config->fuse_mount_name_colon_char_replacement);
     g_writeln("    FuseDirectIO:              %s",
               g_bool2text(config->fuse_direct_io));
     g_writeln("    FileMask:                  0%o", config->file_umask);
--- a/sesman/chansrv/chansrv_fuse.c
+++ b/sesman/chansrv/chansrv_fuse.c
@@ -191,6 +191,36 @@ int xfuse_path_in_xfuse_fs(const char *path)
 #define XFUSE_ENTRY_TIMEOUT     5.0
 
 
+extern struct config_chansrv *g_cfg; /* in chansrv.c */
+
+
+/* Local utility functions */
+
+static inline char *
+_fuse_mount_name_colon_char_replace(const char *dirname)
+{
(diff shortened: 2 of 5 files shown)


matt335672:
Hi @Bad-ptr 

I can see what you're doing, and I think it's a good idea.

A couple of initial thoughts, prior to a full review.
1) If `FuseMountNameColonCharReplacement` isn't set at all, you *must* preserve existing behaviour. This is so users don't have to change their configs on a minor version upgrade.
2) You don't need a `char *` for the config value - just a  `char` will do. This will avoid messing about with memory allocation for this value.
3) Changing the signature of `xfuse_create_share()` violates the principle of least surprise, particularly if we later on modify the calling function to do something with the name we pass in. I suggest you copy this value locally if you need to.
4) `_fuse_mount_name_colon_char_replace()` doesn't consider array element 0 for replacement. There's nothing in the spec which suggests a client couldn't send a colon in this position.

Bad-ptr:
> If FuseMountNameColonCharReplacement isn't set at all, you must preserve existing behaviour. This is so users don't have to change their configs on a minor version upgrade.
You don't need a char * for the config value - just a char will do. This will avoid messing about with memory allocation for this value.
Changing the signature of xfuse_create_share() violates the principle of least surprise, particularly if we later on modify the calling function to do something with the name we pass in. I suggest you copy this value locally if you need to.
_fuse_mount_name_colon_char_replace() doesn't consider array element 0 for replacement. There's nothing in the spec which suggests a client couldn't send a colon in this position.

Done

Another idea is to make this option a string of pairs of chars -- the first char in pair is what to replace, the second is the replacement and the last unpaired character will be replaced by null character.

```c
static inline char* _fuse_mount_name_colon_char_replace(const char *dirname, size_t dirnamelen)
  {
     char *newdirname = (char*)dirname;
     if (g_cfg->fuse_mount_name_colon_char_replacement && dirnamelen > 1)
      {
         char *replacepair = NULL;
          char toreplace, replaceto;
  
          newdirname = g_strdup(dirname);

          for (size_t i = dirnamelen-1; i >= 0; --i) {
             replacepair = g_cfg->fuse_mount_name_colon_char_replacement;
              while('\0' != *replacepair) {
                toreplace = replacepair[0];
                replaceto = replacepair[1];
  
                if (toreplace == newdirname[i]) {
                  newdirname[i] = replaceto;
                  break;
                }
  
                replacepair += 2;
              }
          }
      }
      return newdirname;
  }
```

sesman/chansrv/chansrv_config.c
```diff
@@ -220,6 +221,22 @@ read_config_chansrv(log_func_t logmsg,
                 break;
             }
         }
+        else if (g_strcasecmp(name, "FuseMountNameColonCharReplacement") == 0)
+        {
+            size_t vallen = g_strlen(value);
+            if (vallen < 1) {
```
matt335672: You'll possibly have seen this from the CI, and I apologise for not mentioning i earlier, but this project uses opening-brace-on-new-line, e.g.:-

```c
    if (something)
    {
    }
```

See https://github.com/neutrinolabs/xrdp/wiki/Coding-Style

sesman/chansrv/chansrv_config.c
```diff
@@ -299,6 +316,7 @@ new_config(void)
     /* Do all the allocations at the beginning, then check them together */
     struct config_chansrv *cfg = g_new0(struct config_chansrv, 1);
     char *fuse_mount_name = g_strdup(DEFAULT_FUSE_MOUNT_NAME);
+    char fuse_mount_name_colon_char_replacement = DEFAULT_FUSE_MOUNT_NAME_COLON_CHAR_REPLACEMENT;
```
matt335672: Don't need this variable.

sesman/chansrv/chansrv_config.c
```diff
@@ -315,6 +333,7 @@ new_config(void)
         cfg->restrict_outbound_clipboard = DEFAULT_RESTRICT_OUTBOUND_CLIPBOARD;
         cfg->restrict_inbound_clipboard = DEFAULT_RESTRICT_INBOUND_CLIPBOARD;
         cfg->fuse_mount_name = fuse_mount_name;
+        cfg->fuse_mount_name_colon_char_replacement = fuse_mount_name_colon_char_replacement;
```
matt335672: See above comment. Just use `DEFAULT_FUSE_MOUNT_NAME_COLON_CHAR_REPLACEMENT` here.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -191,6 +191,31 @@ int xfuse_path_in_xfuse_fs(const char *path)
 #define XFUSE_ENTRY_TIMEOUT     5.0
 
 
+extern struct config_chansrv *g_cfg; /* in chansrv.c */
+
+
+/* Local utility functions */
+
+static inline char* _fuse_mount_name_colon_char_replace(const char *dirname,
```
matt335672: Coding style: Function name is on new-line (e.g.):-

```c
static inline char*
_fuse_mount_name_colon_char_replace(. . .
```

sesman/chansrv/chansrv_fuse.c
```diff
@@ -191,6 +191,31 @@ int xfuse_path_in_xfuse_fs(const char *path)
...
+        size_t dirnamelen)
+{
+    char *newdirname = (char*)dirname;
```
matt335672: Coding style: Replace `char*` with `char *`

sesman/chansrv/chansrv_fuse.c
```diff
@@ -191,6 +191,31 @@ int xfuse_path_in_xfuse_fs(const char *path)
...
+    if (g_cfg->fuse_mount_name_colon_char_replacement != ':' && dirnamelen > 0)
+    {
+        newdirname = g_strdup(dirname);
+        for (size_t i = dirnamelen-1; i >= 0; --i) {
```
matt335672: Awkward - this loop won't terminate as `size_t` is always unsigned. Also, `g_strdup()` may return null.

You could use `g_strrchr()` for this, which avoids having to pass the length in. The string is going to be pretty short, and this function is called a very few times so we're not gaining anything from passing the length in.

sesman/chansrv/chansrv_fuse.c
```diff
@@ -738,11 +762,17 @@ int xfuse_create_share(tui32 device_id, const char *dirname)
...
+    size_t dirnamelen = strlen(dirname);
 
-    if (dirname != NULL && strlen(dirname) > 0 &&
+    if (dirname != NULL && dirnamelen > 0 &&
             xfuse_init_xrdp_fs() == 0)
     {
-        xinode = xfs_add_entry(g_xfs, FUSE_ROOT_ID, dirname, (0777 | S_IFDIR));
+        char *newdirname = _fuse_mount_name_colon_char_replace(dirname, dirnamelen);
+        xinode = xfs_add_entry(g_xfs, FUSE_ROOT_ID, newdirname, (0777 | S_IFDIR));
+        if (newdirname != dirname)
```
matt335672, line 777: Nice way to add the free! Suggest a comment somewhere to explain this a bit more.

matt335672 (commented):
Thanks for that - looks pretty good. A few comments below. Feel free to come back to me with requests, or if you don't agree with anything I've said.

Bad-ptr:
Updated

matt335672:
Thanks for your contribution @Bad-ptr

matt335672:
@metalefty - worth a backport to v0.10?

metalefty:
@matt335672 Yes, let's backport.

PR #3394, opened 2025-01-13
Address some Coverity warnings

matt335672:
This PR addresses the following Coverity issues, which are not publically available:-

| CID | Description | File |
| --- | -------------- | ---- |
| 468106 | Buffer overrun in libipm_msg_in_wait_available() | libipm/libipm_recv.c |
| 468113 | Possible integer overflow in g_get_open_fds() | common/os_calls.c |
| 468121 | Possible integer overflow in g_set_wait_obj() | common/os_calls.c |
| 468142 | Possible string overrun in g_atoix() | common/string_calls.c |
| 468150 | Possible memory leak in start_x_server() | sesman/sesexec/session.c |

# 468106
In summary, `trans_get_wait_objs()` can write a maximum of two values to the array (for TLS connections), and there's only space for one. In practice this can't happen. Simplest fix is to make the buffer big enough for two values.

# 468113
Coverity can't be sure that subtracting the int parameters min and max passed to the function won't result in integer overflow when `max - min` is calculated. Extra checks are added to the parameters and the call to `sysconf(_SC_OPEN_MAX)` to mitigate this. In practice, this won't happen, as this function is only used in testing.

# 468121
Coverity can't be sure that adding the result of a `write()` to an `int` won't cause integer overflow. In practice it won't but a simple extra check should make this clear to Coverity.

# 468142
Coverity is complaining that if "0" (which is two bytes long) is passed to this code as a pointed-to value by `str` :-

```c
    if (*str == '0' && tolower(*(str + 1)) == 'x')
    {
        str += 2;
        base = 16;
    }
```

that `str` could be incremented beyond the end of the string. Coverity is unable to see that the second byte of "0" (i.e. a '\0' character) cannot pass the test `tolower(<byte>) == 'x'`. This is therefor a false-positive, and marked as such.

# 468150
A codepath in this function results in the `passwd_file` not being freed. This is a confirmed defect.

I can't currently run Coverity locally, so these are very much a best-shot at getting these addressed.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1805,7 +1805,7 @@ g_set_wait_obj(tintptr obj)
         return 0;
     }
     fd = obj >> 16;
-    to_write = 4;
+    to_write = sizeof(buf);
     written = 0;
     while (written < to_write)
     {
@@ -1823,12 +1823,13 @@ g_set_wait_obj(tintptr obj)
                 return 1;
             }
         }
-        else if (error > 0)
+        else if (error > 0 && error <= (int)sizeof(buf))
         {
             written += error;
         }
         else
         {
+            // Shouldn't get here.
             return 1;
         }
     }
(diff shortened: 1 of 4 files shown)


matt335672:
I'm going to merge this in the absence of any feedback, to see how the Coverity scan copes with this.

I'll then address a few more, and put a PR together for back-porting to v0.10.x

matt335672:
This merge has broken the CI - namely the 32-bit builds have stopped working with the following error:-

```
 shim-signed : Depends: grub-efi-amd64-signed (>= 1.191~) but it is not going to be installed or
                        grub-efi-arm64-signed (>= 1.191~) but it is not installable or
                        base-files (< 12.3)
               Depends: grub-efi-amd64-signed (>= 1.187.2~) but it is not going to be installed or
                        grub-efi-arm64-signed (>= 1.187.2~) but it is not installable
```

It was working fine yesterday, so something has changed on the Github end.

Issue #3399, opened 2025-01-17
Backport 'Add utmp/wtmp support' into v0.10 branch


mlewissmith:
Please consider backporting the PR #2745 'Add utmp/wtmp support' into the v0.10 development branch so that it can be included in the next stable release

matt335672:
Thanks @mlewissmith 

I'll close the issue - we don't need it in addition to #3400. We can continue discussions there.

PR #3400, opened 2025-01-17
[V0.10] Add utmp/wtmp support

mlewissmith:
Cherry-pick 2849dce1054c95d1fdba708ab75c26afd555768a..8cea9b03ab2d63283517151094824b5cfb63d862

See:
* PR #2745 "Add utmp/wtmp support" https://github.com/neutrinolabs/xrdp/pull/2745
* Discussion #2744 "utmp/wtmp/btmp/lastlog" https://github.com/neutrinolabs/xrdp/discussions/2744

Addresses issue #3399

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -3329,10 +3329,10 @@ g_set_allusercontext(int uid)
 /*****************************************************************************/
 /* does not work in win32
    returns pid of process that exits or zero if signal occurred
-   an exit_status struct can optionally be passed in to get the
+   a proc_exit_status struct can optionally be passed in to get the
    exit status of the child */
 int
-g_waitchild(struct exit_status *e)
+g_waitchild(struct proc_exit_status *e)
 {
 #if defined(_WIN32)
     return 0;
@@ -3340,14 +3340,14 @@ g_waitchild(struct exit_status *e)
     int wstat;
     int rv;
 
-    struct exit_status dummy;
+    struct proc_exit_status dummy;
 
     if (e == NULL)
     {
         e = &dummy;  // Set this, then throw it away
     }
 
-    e->reason = E_XR_UNEXPECTED;
+    e->reason = E_PXR_UNEXPECTED;
     e->val = 0;
 
     rv = waitpid(-1, &wstat, WNOHANG);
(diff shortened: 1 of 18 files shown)


matt335672:
Thanks for this @mlewissmith 

@metalefty - are you happy this to be backported?

If so, I'll retest before merging. The changes look fine to me.

metalefty:
@matt335672 The only thing I am concerned about is the known issue, the IP address updating issue. This may confuse users so I have negative (not strong, really) feelings about this option yet.  On the other hand, the feature is opt-in so I'm in favor of this feature, as long as it's explicitly stated as experimental.

matt335672:
OK - thanks.

To be clear, about the issue, what we're saying is that when the user reconnects, the IP address isn't updated in the relevant files. Is that right?

The feature is opt-in only in that it's a compile-time option. Do you want to add run-time control to that as well? Would that help?

mlewissmith:
If it helps, I am aware of the IP updating issue.
As the sysadmin for many machines used for work and research it's important I know _that_ the machines are in use remotely, less so from where.

I don't think you're asking me, but if the feature were compiled in by default and controlled by a flag in xrdp.ini that would be good

moobyfr:
Hello, IIRC, there is another issue on updating the IP: the entry is set when the session start, but a reconnect on xrdp is handled differently, as sesman wasn't aware of the new IP. I didn't look further on the first implementation I've made, but it could be probably improved

metalefty:
> To be clear, about the issue, what we're saying is that when the user reconnects, the IP address isn't updated in the relevant files. Is that right?

That's right.

> The feature is opt-in only in that it's a compile-time option. Do you want to add run-time control to that as well? Would that help?

compile-time opt-in is fine for me. 

If the IP updating issue can be resolved, let's improve it in the devel branch and then backport to v0.10.

matt335672:
@moobyfr - we're aware of that, and I intend to start updating that area soon. Thanks for the reminder though.

@mlewissmith - I've been a professional sysadmin too in the past, and I agree with your statement above.

@metalefty - suggest we merge this and update NEWS-v0.10 accordingly with the experimental warning, so that it goes in the release note for v0.10. Does that work for you?

metalefty:
Yes, that's fair enough.

metalefty (approved):

matt335672:
Thanks for your contribution @mlewissmith

matt335672:
Following text added to the release note for v0.10.3 in the 'General Announcements' section:-

Experimental support for utmp/wtmp file is provided in this release. If you use this, be aware that these files are only updated when an xrdp session is created or destroyed. Disconnections and reconnections to the same session are not tracked. In particular:-
* the FROM address for a client (as shown by the `w` command) reflects the IP address of the client at the time of creation, and not the address of the currently connected client.
* Sessions started by the `xrdp-sesrun` command do not have a FROM address.

PR #3402, opened 2025-01-19
xrdp.8.in: Fix "SEE ALSO" refs to `xrdp-sesman`, `xrdp-sesrun`

mvf:
`sesman` and `sesrun` got prefixed with `xrdp-` long ago, but the main `xrdp` manpage wasn't updated.


matt335672:
Good spot - thanks @mvf

Issue #3401, opened 2025-01-19
Stale file contents in redirected drive


fiddyschmitt:
Hi,

### **Environment**
Windows Client (mstsc) connecting to Ubuntu 22.04.5 (xrdp 0.9.17).

### To reproduce

I have shared a drive from Windows, and its content appear fine in `thinclient_drives`

1. On Linux open a terminal and run: `tail -f hello.txt`
2. On Windows, add a line to `hello.txt`
3. On Linux open another terminal and run `cat hello.txt`

### Expected results

- The `tail` command in Linux displays the added line
- The `cat` command shows the added line

### Actual result

- Neither the `tail` or `cat` commands show the new line

Only if `tail` is closed will `cat` see the new data. It seems that if the file is open for reading, its latest contents don't propagate from the client to the server.

matt335672:
Thanks for this, @fiddyschmitt 

An option to address the particular issue you raise was added by @bobcarroll in #3255 

To use it, you'll need version v0.10.2 of xrdp or later.

PR #3404, opened 2025-01-20
Add clarification to sesman.ini

matt335672:
When investigating #3402, it took me a while to find the correct option, and then I mis-typed it.

Small PR to add the commented out option to the config file.


matt335672:
Well this is embarrassing. Not only did I mean #3401, but the option is already there! Closing this.

PR #3408, opened 2025-01-23
[v0.10] Add '--enable-utmp' to .github/workflows/build.yml

mlewissmith:
Per commit 3b6c9bcba3c79d658415fb900504bd152a238e01 neglected by PR #3400


matt335672:
Good spot - thanks.

PR #3410, opened 2025-01-27
Fix more coverity warnings

matt335672:
This PR fixes the remainder of the high-impact Coverity warnings, plus a couple from the neutrinolabs directory which are not included in the automatic scan.

See #3394 for the precursor to this PR

|CID | Description | File |
|---| ------------| ----|
|468125 | Out-of-bounds reference parsing "OK" string in xrdp_wm_show_log() | xrdp_wm.c |
|468132 | Out-of-bounds reference parsing "OK" string in xrdp_wm_help_clicked() | xrdp_login_wnd.c |
| 468146 | Potential double-free of popup_wnd in xrdp_wm_key() | xrdp_wm.c |
| 468153 | Resource leak in scard_function_list_readers_return() | smartcard_pcsc.c|
| - | Access past end of neutrinolabs mod->pamusername in lxrdp_set_param() | xrdp-neutrinordp.c |
| - | Channels array overflow in freerdp->settings->channels in lfreerdp_pre_connect() | xrdp-neutrinordp.c |

## 468125 / 468132
These are both caused by an assumption made in utf8_get_next_char() that the character pointer can be advanced by 6 bytes for a credible end-stop for parsing. This isn't true for a short string such as (e.g.) "OK"

The fix is not to add 6 to the character pointer to get an end-stop, but just to assign NULL to the end-stop. The parsing logic for this function only needs an end-stop when parsing a short multi-byte character.

## 468146
This is caused by not assigning NULL to `self->popup_wnd` after it is freed. If this is done, a double-free is not possible.

## 468153
This code is currently unused. A memory leak on an error-return from a function is addressed to keep Coverity happy

## (neutrinordp)
The neutrinordp module is not currently scanned by the automated Coverity check, but this can be done manually.

These are two simple array overflows.

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -1191,7 +1191,7 @@ utf8_get_next_char(const char **utf8str_ref, unsigned int *len_ref)
     /*
      * Macro used to parse a continuation character
      * @param cp Character Pointer (incremented on success)
-     * @param end One character past end of input string
+     * @param end One character past end of input string, or NULL
      * @param value The value we're constructing
      * @param finish_label Where to go in the event of an error */
 #define PARSE_CONTINUATION_CHARACTER(cp, end, value, finish_label) \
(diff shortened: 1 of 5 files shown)


matt335672:
I'm going to merge this now, as all the fixes are pretty trivial. The commentary above should be sufficient to work out the reasoning behind all of these.

Issue #3405, opened 2025-01-22
AlmaLinux: Issue with H.264 when client request color depth 32 bit (xorgxrdp)


hauihau:
### xrdp version

0.10.2

### Detailed xrdp version, build options

~~~
xrdp 0.10.2
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --build=x86_64-redhat-linux-gnu
      --host=x86_64-redhat-linux-gnu
      --program-prefix=
      --disable-dependency-tracking
      --prefix=/usr
      --exec-prefix=/usr
      --bindir=/usr/bin
      --sbindir=/usr/sbin
      --sysconfdir=/etc
      --datadir=/usr/share
      --includedir=/usr/include
      --libdir=/usr/lib64
      --libexecdir=/usr/libexec
      --localstatedir=/var
      --sharedstatedir=/var/lib
      --mandir=/usr/share/man
      --infodir=/usr/share/info
      --enable-fuse
      --enable-pixman
      --enable-painter
      --enable-vsock
      --enable-ipv6
      --enable-openh264
      --with-socketdir=/run/xrdp
      --with-imlib2
      build_alias=x86_64-redhat-linux-gnu
      host_alias=x86_64-redhat-linux-gnu
      CC=gcc
      CFLAGS=-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection
      LDFLAGS=-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1
      CXX=g++
[...]

PR #3413, opened 2025-01-29
Cope with broken OpenH264 encoder

matt335672:
Addresses issues encountered in #3405 

Draft for now as I've run out of time today to test. More code comments are needed I think.

Any feedback is welcome.

**Edit:** Forgot to make this draft.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -15,6 +15,7 @@ jobs:
       CONF_FLAGS_amd64_max: "--enable-ipv6 --enable-jpeg --enable-fuse --enable-mp3lame
                   --enable-fdkaac --enable-opus --enable-rfxcodec --enable-painter
                   --enable-pixman --enable-utmp
+                  --enable-x264 --enable-openh264
                   --with-imlib2 --with-freetype2 --enable-tests"
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


metalefty:
LGTM at first glance.

matt335672:
Thanks.

I'm not happy about the proliferation of conditional macros in this code - it makes our testing and fault finding more complicated. I'll see if I can simplify the macros a bit, at the expense of meaning we ship slightly more code in the non-H.264 situation.

matt335672:
This is now ready for review.

I've moved the library check to the constructor for xrdp_mm, and set a member variable `libh264_loaded` which can be used to check the library has actually loaded.

The chunk of code in `xrdp_encoder_create()` which sets the software library methods has been moved to its own procedure.`set_h264_encoder_methods()`  This was to make the logging of the loaded library conditional on `libh264_loaded`.

I've not simplified the macros for now, so as to make this a minimal change.

Tested on Alma9 with both open264 and noopenh264 packages using these commands to switch between libraries. xrdp must be restarted after switching the library.

```
sudo dnf install openh264 --allowerasing
sudo dnf install noopenh264 --allowerasing
```

I've also added (I hope) the H.253 codepaths into the nightly Coverity scan.

metalefty:
LGTM, let me some time to test it on my local env.

metalefty (approved):
Let's merge and cherry-pick to v0.10.

PR #3422, opened 2025-02-04
Update TCP proxy to address Coverity errors

matt335672:
This is a follow-on from #3421 which appears to have triggered another warning.

Coverity seems to have some problems with the loop(s) copying data from one socket to another, in that it assume that eventually an integer overflow will occur. It's not obvious why this should be flagged, but this seems likely to be a false positive. There are guards around the integer operations in question.

This commit avoids the integer issue by using a simple pointer + count mechanism to copy the data.

The socket copy code has been placed in a separate function - before it was duplicated. Minor fixes have been made to error reporting around the connection code.


matt335672:
Merged to pick up the overnight Coverity scan

At some stage I'll batch up #3421 and #3422 and possibly others for a backport to v0.10

PR #3423, opened 2025-02-05
[V0.10] Coverity fixes

matt335672:
Backports of the following devel PRs:-

#3421
#3422
#3442

Draft for now, so a few more can be added. None of these look crucial at the moment.


matt335672:
2025-02-11: Needs rebasing - see #3427

PR #3427, opened 2025-02-11
[V0.10] : Bump FreeBSD CI to 14.2 (was 13.4)

matt335672:
FreeBSD 13.3 no longer appears to be available.

@metalefty - I've stuck with 13.x for the `V0.10` branch. Are you happy with that, or would you prefer 14.x? I've moved to 14.x for the `devel` branch.


metalefty:
Let's move to 14.

matt335672:
OK - I'll merge that.

Issue #3424, opened 2025-02-05
Please add Hungarian to the supported list of keymaps


domjant:
Hello,
As described at the end of the #2337  issue, would you add Hungarian language to the supported list of keymaps for devel and v0.10.x versions of Xrdp?

Actually just the xrdp-genkeymap km-0000040e.ini generated file need to be supplied in the /etc/xrdp folder. 
As well as insert following layouts to [default_rdp_layouts]; [default_layouts_map]; and [rdp_layouts_map_mac] in /etc/xrdp/xrdp_keyboard.ini file:

[default_rdp_layouts]
rdp_layout_hu=0x0000040E

[default_layouts_map]
rdp_layout_hu=hu

[rdp_layouts_map_mac]
rdp_layout_hu=hu

Thank you in advance!

ps.: kudos to @smokey2k (Dechev Daniel Ivanov) to share this in #2337

matt335672:
@domjant - can you confirm your keyboard looks like this:-

https://www.kbdlayout.info/kbdhu

Thanks.

domjant:
Yes, this is it (KLID 0000040e)

matt335672:
@domjant - can you test my keyboard file(s)?

Assuming you've got a v0.10.x system, can you copy these two files to your installation:-

- https://raw.githubusercontent.com/matt335672/xrdp/refs/heads/v0_10_add_hu_kbd/instfiles/km-0000040e.ini -> /etc/xrdp/km-0000040e.ini
- https://raw.githubusercontent.com/matt335672/xrdp/refs/heads/v0_10_add_hu_kbd/xrdp/xrdp_keyboard.ini -> /etc/xrdp/xrdp_keyboard.ini

If you're using devel, I've got a add_hu_kbd branch which is similar.

domjant:
@matt335672 - First of all apology for the delayed reply, I had busy days recently. 
I'm using Xrdp version 0.10.2. I tested the provided files and happy to report they are working fine on my RHEL 9 with Hungarian keyboard.

matt335672:
Thanks @domjant - that's great to know. Thanks for the testing.

No need to apologise. I think we all know how it is these days!

I'll get that merged now.

PR #3429, opened 2025-02-11
Add Hungarian keyboard

matt335672:
See #3424

I'm hoping for confirmation from @domjant that this works before I merge it.


domjant:
Confirmed, working fine. Thank you @matt335672 !

PR #3439, opened 2025-02-24
Add support for cppcheck 2.17.x

matt335672:
**Edit 2024-02-24: No longer draft**
**Edit 2024-02-26: cppcheck 2.17.1 released**

~~Draft for now, as I should probably add a test suite module for the list16 calls.~~

This is a relatively big change for us.

cppcheck 2.17.0 adds checks that a pointer returned from malloc() and calloc() is not used without a NULL-guard. We do this quite a lot.

I've addressed this by adding functions `g_malloc_nofail()` and `g_calloc_nofail()` which either allocate memory or abort. These functions are now called in places where we are not making these checks. The names are easy to search for, so I'm hoping over time we can remove many uses of these.

Many of the uses are in test programs or example programs.

I've modified the list16 module to handle out-of-memory conditions gracefully. A test suite is now included for list16.

@firewave - you asked to be notified about cppcheck-related PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
       CC: gcc
       # This is required to use a version of cppcheck other than that
       # supplied with the operating system
-      CPPCHECK_VER: "2.17.0"
+      CPPCHECK_VER: "2.17.1"
       CPPCHECK_REPO: https://github.com/danmar/cppcheck.git
     steps:
       # Set steps.os.outputs.image to the specific OS (e.g. 'ubuntu20')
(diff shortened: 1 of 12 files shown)


matt335672:
aargh - as usual I've forgotten to make this a draft PR.

matt335672:
Also for reference, here's the original error log from running cppcheck 2.17.0 on devel (~18K lines).

[cppcheck-error.txt](https://github.com/user-attachments/files/18941588/cppcheck-error.txt)

matt335672:
I've rebased the rfxcodec bit on neutrinolabs/librfxcodec@9ce01dfd6bdd1169bf5f2f6f8face677b814d353 rather than neutrinolabs/librfxcodec@26e40b29d05877e926f977c4b40b2a44bbb07216 used in devel. That pulls in neutrinolabs/librfxcodec#73, which we need anyway.

matt335672:
I'm going to merge this into devel so it gets a bit of use.

Later we can decide if we want to backport to v0.10.x

PR #3442, opened 2025-02-28
Fix more coverity warnings

matt335672:
Follows on from #3426 

These are all minor:-

|CID |Description |
|---|------|
|475385 | Result of `g_chmod_hex()` not checked |
|468102 | Possibility of `self->mod` being accessed when NULL |
|468103 | Result of `g_obj_wait()` ignored when other calls to it aren't |
|468108 | Repeated line in conditional (dead code) |
|468112 | Result of `libxrdp_orders_init()` ignored when other calls to it aren't | 
|468117 | Variables derived from 'tainted' values used in loop control |

475385 was reported as a new fault following the last merge of #3439 to devel. The merge had nothing to do with the area in question, so I don't understand why this is being reported now.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -1457,6 +1457,16 @@ g_sck_recv_fd_set(int sck, void *ptr, unsigned int len,
     if ((rv = recvmsg(sck, &msg, 0)) > 0)
     {
         struct cmsghdr *cmsg;
+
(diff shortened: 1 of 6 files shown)


matt335672:
I'm merging this now into `devel`. After the weekend I'll check all the included CIDs have been fixed.

PR #3452, opened 2025-03-06
Add new session type SCP_SESSION_TYPE_XVNC_UDS

### LINKED ISSUE #2518: VNC backend fails to work on FIPS-enabled machines
We are required to run FIPS enabled on all of our systems. However, we are unable to RDP into RHEL 9 systems when FIPS is enabled on RHEL 9 regardless of the FIPS setting on Windows. I believe this is due to RHEL 9 utilizing FIPS 140-3 standards while Windows uses 140-2. From the documentation I've found, when configured for FIPS, XRDP uses 140-1. 140-3 doesn't allow ciphers that are approved for 140-1 while 140-2 does. Are there any plans of making XRDP compliant with FIPS 140-2 or 140-3 even?

Thanks.

matt335672:
Fixes #2518 

~~Currently draft, as more testing required.~~

This PR adds the following
1) support for VNC servers which can be contacted by a VNC viewer over a Unix Domain Socket (UDS).
2) A new sesman session type, which is a VNC session using a UDS connection. This session type requires the Xvnc server to support the `-rfbunixpath`, `-rfbunixmode` and `-SecurityTypes` parameters. Currently known supported X servers are TigerVNC and TurboVNC.
3) FIPS detection and support (see below)

When sesman creates a VNC-over-UDS session, the same UDS socket name is used as would be used for an Xorg session. The socket protections are the same, in that the socket can only be accessed by the authenticated user and the xrdp daemon user.

## FIPS support
This PR provides a way for xrdp to support FIPS_based deployments using VNC - the classic VNC password algorithm is not supported by FIPS. This is primarily a problem with RHEL and derivatives with FIPS mode enabled, as these default to VNC as the xrdp backend.

FIPS-based systems are detected at run-time. The following changes are made:-
1) The default session type is changed from VNC-over-TCP to VNC-over-UDS.
2) The DES3-based encryption algorithm used by the VNC-over-TCP session is explicitly disabled to prevent authentication material protected by DES3 being written to disk. This should not be allowed if FIPS is enabled.

I've tested this on AlmaLinux 9.5 with FIPS mode enabled. If there's anyone out there willing to test this as well, I'd be grateful. I can build and sign test RPMs for anyone who can take me up on this offer.

@metalefty - this could be back-ported to v0.10.x fairly easily. For devel, I'm wondering if we should make VNC-over-UDS the default session type when a VNC backend is chosen.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -4282,6 +4282,29 @@ g_no_new_privs(void)
 #endif
 }
 
+
+/*****************************************************************************/
+int
+g_fips_mode_enabled(void)
+{
+    int rv = 0;
+#if defined (__linux)
+    char buff[16];
+    int fd = open("/proc/sys/crypto/fips_enabled", O_RDONLY);
+
+    if (fd >= 0)
+    {
+        if (read(fd, buff, sizeof(buff)) > 0)
+        {
+            rv = (buff[0] != '0');
+        }
+
+        close(fd);
+    }
+#endif
+    return rv;
+}
+
 /*****************************************************************************/
 void
 g_qsort(void *base, size_t nitems, size_t size,
--- a/common/os_calls.h
+++ b/common/os_calls.h
@@ -428,6 +428,14 @@ int      g_tcp4_bind_address(int sck, const char *port, const char *address);
 int      g_tcp6_socket(void);
 int      g_tcp6_bind_address(int sck, const char *port, const char *address);
 int      g_no_new_privs(void);
+/**
+ * Query whether FIPS mode is enabled
+ *
+ * In FIPS mode, some cryptographic algorithms are disabled
+ *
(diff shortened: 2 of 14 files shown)


metalefty:
Thank you. This would be useful to backport. Let's test and include in the next v0.10 release (in March).

matt335672:
Thanks.

I'll merge this into devel (I've done quite a bit of testing already), and work on a backport to v0.10.x. We can then make a separate decision whether to change the default Xvnc session type for the next major release.

Before I make the merge I'll try to run this through Coverity to see if it uncovers anything else unexpected.

**Edit:2025-03-10:** Scrub the Coverity scan. The queue appears to be blocked at the moment, and I have other things to do today. I'll pick up any scan issues when the queue is working again, and before we merge to v0.10.x

PR #3453, opened 2025-03-10
[V0.10] Add new session type SCP_SESSION_TYPE_XVNC_UDS

matt335672:
Backport of #3452 to V0.10.x

~~Draft for now while I retest, and wait for the results of a Coverity scan.~~

This was largely a clean cherry-pick, but xup.c required some additional change. In particular, the "some problem" message which could appear on the console with an Xorg session has been replaced with "Error connecting to Xorg - check log".

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -4260,6 +4260,29 @@ g_no_new_privs(void)
 #endif
 }
 
+
+/*****************************************************************************/
+int
+g_fips_mode_enabled(void)
+{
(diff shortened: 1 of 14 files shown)


matt335672:
@metalefty - are you happy for this to go into V0.10.3, following testing? This fixes #2518 which is applicable to AlmaLinux in FIPS mode.

metalefty:
Yes, this one is useful.

Issue #3381, opened 2025-01-06
FreeBSD : Collisions between redirected ssh X server and xrdp


matt335672:
**Edit : 2025-1-7 The problem symptoms have been edited. Please consult the issue history for where we started.**
### xrdp version

0.10.x / devel

### Detailed xrdp version, build options

~~~
xrdp 0.10.80
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-devel-all
      --enable-fuse
      --enable-pixman
      --enable-ipv6
      --enable-painter
      --enable-jpeg
      --with-imlib2
      --enable-vsock
      --with-freetype2
      --enable-utmp
      CC=clang
      CFLAGS=-g
      LDFLAGS=-L/usr/local/lib
      CPPFLAGS=-I/usr/local/include

  Compiled with OpenSSL 3.0.15 3 Sep 2024
~~~

### Operating system & version

FreeBSD 14.2-RELEASE

### Installation method

git clone & make install

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

XFCE

### Environment xrdp running on

VM

### What's your client?

Various

### Area(s) with issue?

Session manager (sesman)

### Steps to reproduce

1. Ensure sshd is configured to allow X11 forwarding
2. Log on to system from elsewhere using `ssh -X`:-

```
$ ssh -X freebsd14.test.lan
<snipped>
$ echo $DISPLAY
localhost:10
$ xauth list
freebsd14/unix:10  MIT-MAGIC-COOKIE-1  186a734d1d0ea2b8a2ae0c543a50c7c
```

3. Log in to system using xrdp and the same user

### ✔️ Expected Behavior

Following login, `xauth list` in ssh session shows same information

### ❌ Actual Behavior

```
$ xauth list
freebsd14/unix:10  MIT-MAGIC-COOKIE-1  00615f25c6bb6e0c71b20c32d4e9ea12
```

### Anything else?

Thanks to @derekschrock for raising this on Gitter.

An active X server is checked for by xrdp by using the following algorithm:-
1) Looking for the UNIX Domain Socket opened by the X server
2) trying to bind to `0.0.0.0:<tport>` and/or `[::]:<tport>` where `<tport>` is the TCP port which may be used by the X server.

[...]

PR #3456, opened 2025-03-12
[V0.10] Add g_sck_set_reuseaddr()

matt335672:
Back-port of #3390 to V0.10

Only set SO_REUSEADDR where it is actually required, which is before most (but not all) bind() calls.

(cherry picked from commit cad52028e0dc49ba0d37f8c335dfa1a887ebee6d)

This is primarily to prevent a problem on FreeBSD, where xrdp can start a session using the same number as a forwarded-X over ssh session.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -369,8 +369,6 @@ int
 g_tcp_socket(void)
 {
     int rv;
-    int option_value;
-    socklen_t option_len;
 
 #if defined(XRDP_ENABLE_IPV6)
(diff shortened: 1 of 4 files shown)


matt335672:
@metalefty - are you happy for this to go into V0.10.3, following testing?

metalefty:
@matt335672 Sure, let's go!

matt335672:
Thanks. I'll merge this in and update the NEWS v0.10 page. This will help some FreeBSD-based configurations raised by @derekschrock (see #3381)

PR #3476, opened 2025-03-25
Add support for 16-bit layout fallback for xorgxrdp

matt335672:
Keyboard layout values from the client such as 00010416 (Brazil ABNT2) may not have a matching entry in xrdp_keyboard.ini. For these clients, we map a US keyboard as a fallback.

Before falling back to US, we should first try to match on the lower 16-bits of the layout, which in this case is 0416 (ABNT). This is more likely to result in something usable.

We already implement this functionality for the keyboard files used by the login screen and VNC back-end.

See also [this comment](/neutrinolabs/xrdp/pull/3447#issuecomment-2750662893). I'd hoped to fix some of the string handling in this function, but now I think this can be merged into V0.10.x, and we can fix string handling in devel.

### DIFF
--- a/xrdp/lang.c
+++ b/xrdp/lang.c
@@ -705,6 +705,26 @@ xrdp_init_xkb_layout(struct xrdp_client_info *client_info)
                 break;
             }
         }
+        if (rdp_layout[0] == '\0' && (client_info->keylayout & ~0xffff) != 0)
+        {
+            // We failed to match the layout, but we may be able
+            // to match on the lower 16-bits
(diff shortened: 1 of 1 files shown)


metalefty:
I've been occupied lately and cannot spend much time but LGTM from a quick glance. Feel free to merge into v0.10 to include this in next v0.10 release.

matt335672:
Thanks @metalefty 

I'm merging this following [initial user acceptance testing](3447#issuecomment-2752563174).

PR #3477, opened 2025-03-25
Add support for strlcpy()

matt335672:
Too many places in xrdp use strncpy() to copy strings to fixed-length buffers, when this is not the correct function to use.

This PR makes sure strlcpy() from the BSDs is available as a saner alternative. This function is available by default on Linux and FreeBSD.

At the moment I don't intend for this to be back-ported to v0.10.x, but as it's a standalone function this can be done in the future if necessary.

**Edit:** On Linux this function was only added for glibc 2.38 (2024-07-31). When building on Ubuntu 24.04 and later, the glibc version of the function is used. When building on 22.04, the fallback function is used. The function is available in libbsd on the older platform, but the implementation is virtually identical:-

https://gitlab.freedesktop.org/libbsd/libbsd/-/blob/main/src/strlcpy.c

*This function is somewhat controversial* - here are some discussions about it:-

https://lwn.net/Articles/507319/
https://lwn.net/Articles/612244/
https://lwn.net/Articles/905777/

My own opinion is that given the style xrdp is written in `strlcpy()` is a better choice than `g_strncpy()` for most calls.

Given the controversy, comments would be very welcome on this!

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -15,6 +15,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
+ * The strlcpy implementation is taken from OpenBSD and reformatted. The
+ * original has the following notice attached:-
+ * |
+ * |   Copyright (c) 1998, 2015 Todd C. Miller <millert@openbsd.org>
+ * |
+ * |   Permission to use, copy, modify, and distribute this software for any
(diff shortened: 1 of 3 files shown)


metalefty:
+1 to use strlcpy().

metalefty (approved):

PR #3480, opened 2025-03-28
Add GitHub Actions to generate release tarball

metalefty:


matt335672:
This looks interesting!

I didn't realise you could use Copilot in this way. I'll try it too.

PR #3482, opened 2025-03-29
[V0.10] Remove SIGTERM race in chansrv

matt335672:
The signal handlers for SIGTERM are put in place before the sigterm object is created. If a SIGTERM is received between the two, it is ignored and chansrv will not exit.

(cherry picked from commit 343e84a76affa75e53a312f822e579dbf25be454)

@metalefty - this is a minor issue for V0.10.x and probably isn't a priority for v0.10.3 now. It can go into v0.10.4. I found it while I was testing some improvements to sesexec.

### DIFF
--- a/sesman/chansrv/chansrv.c
+++ b/sesman/chansrv/chansrv.c
@@ -1835,7 +1835,13 @@ main(int argc, char **argv)
         return 1;
     }
 
-    /*  set up signal handler  */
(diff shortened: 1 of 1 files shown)


metalefty:
I have no time to look into this before the v0.10.3 release, let's merge this after v0.10.3.

PR #3487, opened 2025-03-31
Rationalise g_htoi() / xrdp_wm_htoi()

matt335672:
This is a code quality PR.

xrdp contains two functions which do similar things:-
- g_htoi() converts a hex string to an integer, ignoring unrecognised characters
- xrdp_wm_htoi() converts a hex string to an integer, ignoring leading whitespace, but terminating on unrecognised characters

An analysis of the uses of g_htoi() shows that the only place where unrecognised characters might be encountered is parsing lines from xrdp_keyboard.ini, where all values have an '0x' prefix (i.e. the 'x' is unrecognised)

An analysis of xrdp_wm_htoi() shows that the functionality to ignore leading whitespace is not used.

Both functions are replaced with a re-written g_htoi() which is const- correct and provided with test cases. This function behaves in the same way as the atoi() library function, in that it terminates on an unexpected character.

The use of g_htoi() in parsing lines from xrdp_keyboard.ini is replaced with a call to g_atoix() which handles the '0x' prefix correctly.

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -502,82 +502,32 @@ size_t strlcpy(char *dst, const char *src, size_t dsize)
 #endif
 
 /*****************************************************************************/
-int
-g_htoi(char *str)
+unsigned int
+g_htoi(const char *str)
 {
-    int len;
-    int index;
-    int rv;
-    int val;
-    int shift;
-
-    rv = 0;
-    len = strlen(str);
-    index = len - 1;
-    shift = 0;
-
-    while (index >= 0)
+    unsigned int rv = 0;
+    while (*str != '\0')
     {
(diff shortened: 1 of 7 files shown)


matt335672:
I'll merge this now, as I've done a fair bit of testing:-
1) `xrdp_keyboard.ini` parsing works correctly (replacing `g_htoi()` with `g_atoix()`)
2) When using RDP security, the RSH keys can be read correctly (old use of `g_htoi()`)
3) Colour values are read correctly from xrdp.ini (old use of `xrdp_wm_htoi()`)

I haven't directly tested the call to `g_htoi()` in `clipboard_file.c`, but this is heavily checked anyway.

PR #3488, opened 2025-03-31
Use strlcpy() in xrdp_init_xkb_layout()

matt335672:
Replace uses of g_strncpy() in xrdp_init_xkb_layout() with the more correct strlcpy().

In addition, some values and pointers which do not need to be writeable have been made const.

~~Draft for now, as I'm out of time to test this today.~~

### DIFF
--- a/xrdp/lang.c
+++ b/xrdp/lang.c
@@ -700,9 +700,12 @@ xrdp_init_xkb_layout(struct xrdp_client_info *client_info)
 
         if (section_found == -1)
         {
-            g_memset(section_rdp_layouts, 0, sizeof(char) * 256);
-            g_memset(section_layouts_map, 0, sizeof(char) * 256);
             // read default section
+            strlcpy(section_rdp_layouts, "default_rdp_layouts",
+                    sizeof(section_rdp_layouts));
+            strlcpy(section_layouts_map, "default_layouts_map",
+                    sizeof(section_layouts_map));
+
             file_read_section(fd, "default", items, values);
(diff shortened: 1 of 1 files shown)


matt335672:
So far I'm not terrifically impressed with Copilot's review capabilities. It seems unable to determine that lang.c is a C language source file. When I asked it about this I got:-

> The pull request #3488 in the neutrinolabs/xrdp repository includes changes to the xrdp/lang.c file. The message "Language not supported" generated by Copilot might have been due to a temporary issue or misconfiguration. The file xrdp/lang.c is written in C, which is a supported language.

> If you continue to encounter this message, it might be useful to check the Copilot configuration or contact GitHub support for further assistance. If you need any specific help with the changes in this file, feel free to ask!

metalefty:
Changes LGTM. I'm fine with merging this at the moment.

As a little bit long-term thoughts,  let's proceed to TOML. Newly added configuration files should be TOML, so we implemented GFX configurations using TOML. 

The next step is to switch existing configuration files to TOML.  We already switched keymap files to TOML so now it is time for keyboard configuration. I suppose also `rsakeys.ini` can be switched to a more standard format such as PEM.  `xrdp.ini` and `sesman.ini` are big ones so they will be the last.

matt335672:
Agreed.

The keymap should be pretty straightforward. `rsakeys.ini` could be converted to PEM, but that would need a bit of work - it's not an area I'm familiar with.

The big two will probably need some migration utilities so people can talk old format files and produce toml outputs.

PR #3490, opened 2025-04-03
Refactor struct xrdp_client_info

matt335672:
Draft for comments.

The `xrdp_client_info` struct is refactored into items shared with xorgxrdp, and items which are private to xrdp.

The name `struct xrdp_client_info` is retained for xorgxrdp, for compatibility. The `struct xrdp_client_info_all` type is used within xrdp only.

@metalefty - what do you think of this approach? Does it help the maintainability or not? Any other suggestions?

I haven't tested this yet.

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -18,7 +18,6 @@
  * xrdp / xserver info / caps
  */
 
-#include "xrdp_constants.h"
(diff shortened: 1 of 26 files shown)


matt335672:
I've decided I'm not that happy with this approach. I'll close this one and try again.

PR #3492, opened 2025-04-04
Refactor struct xrdp_client_info (rework)

matt335672:
Rework of #3490 

The struct 'struct xorgxrdp_client_info' is factored out of 'struct xrdp_client_info'. This represents the items which are required to be passed to xorgxrdp.

This PR also requires a small change to xorgxrdp (neutrinolabs/xorgxrdp#384).

**Edit:** Because line lengths of some lines are increased because of the need for a `xup` member to access some data, I've added three commits to tidy up some source lines more than 80 characters in length. These commits should not change any functionality and could stand on their own.

**Edit:** `pub` member of client_info renamed to `xup`.

### DIFF
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -99,6 +99,51 @@ enum unicode_input_state
     UIS_SUPPORTED,       ///< Client supports Unicode, but it's not active
     UIS_ACTIVE           ///< Unicode input is active
 };
+
+/**
+ * Information about the xrdp client which is passed to xorgxrdp
+ *
+ * @note This structure is shared with xorgxrdp. If you change it,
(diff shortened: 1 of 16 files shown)


matt335672:
@metalefty - I've had another look at this, and I'm wondering if a simpler way to fix this problem is just to create a completely separate `struct xorgxrdp_client_info` which we copy things into before we send it to the X server. That reduces the amount of change, and makes it a lot simpler to edit `struct xrdp_client_info`. The copy only happens once when we connect to the X server.

What do you think?

matt335672:
I've implemented the separate struct, and it looks a lot simpler to me.

Superceded by #3534

PR #3500, opened 2025-04-15
Fix typo in CI build

matt335672:
--with-x264 should be --enable-x264


matt335672:
This doesn't need back-porting - V0.10.x seems to be correct already.

PR #3505, opened 2025-04-17
Coverity fixes

matt335672:
Address these Coverity CIDs:-
- 468122
- 468126
- 468127
- 468130
- 468134
- 468139
- 468148
- 468156

Comments in the individual commits


matt335672:
Fixed a few typos in comments and retested.

Issue #3498, opened 2025-04-14
xrdp sessions fail with Quest/OneIdentity Safeguard for Privileged Sessions


richlv:
### xrdp version

0.10.2

### Detailed xrdp version, build options

~~~
xrdp 0.10.2
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2024 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --build=x86_64-redhat-linux-gnu
      --host=x86_64-redhat-linux-gnu
      --program-prefix=
      --disable-dependency-tracking
      --prefix=/usr
      --exec-prefix=/usr
      --bindir=/usr/bin
      --sbindir=/usr/sbin
      --sysconfdir=/etc
      --datadir=/usr/share
      --includedir=/usr/include
      --libdir=/usr/lib64
      --libexecdir=/usr/libexec
      --localstatedir=/var
      --sharedstatedir=/var/lib
      --mandir=/usr/share/man
      --infodir=/usr/share/info
      --enable-fuse
      --enable-pixman
      --enable-painter
      --enable-vsock
      --enable-ipv6
      --enable-utmp
      --with-socketdir=/run/xrdp
      --with-imlib2
      build_alias=x86_64-redhat-linux-gnu
      host_alias=x86_64-redhat-linux-gnu
      CFLAGS=-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection  
      LDFLAGS=-Wl,-z,relro  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld
[...]

PR #3507, opened 2025-04-24
[V0.10] Refactor static channel name handling

matt335672:
~~This PR is currently based on v0.10.2. For the reasons for this, s~~See #3498

~~It will be rebased after user testing. Until it is rebased, it cannot be merged, and the CI won't work.~~

Changes are:-
1) Remove 'magic numbers' related to static channel name lengths, and replace with `CHANNEL_NAME_LEN`, or `CHANNEL_NAME_LEN + 1`, as appropriate.
2) Always add static channel definitions, even if they are malformed.
3) Log channels which the client sends, which aren't named in the `[Channels]` section of `xrdp.ini`.

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -1278,7 +1278,7 @@ libxrdp_query_channel(struct xrdp_session *session, int channel_id,
 
     if (channel_name != 0)
     {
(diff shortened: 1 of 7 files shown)


matt335672:
This branch is now rebased for retesting and merging.

PR #3518, opened 2025-05-05
Add Latvian keyboards

### LINKED ISSUE #3511: Support Latvian keyboard layout[s]
As a followup of #3498, creating this issue to track support for Latvian keyboard layout[s].

matt335672:
Fixes #3511 

Three layout IDs are supported:-
- 0x426 is `lv(ergonomic)` - see https://www.kbdlayout.info/kbdlv
- 0x10426 and 0x20426 map to `lv`.

We can't use 0x426 as a 16-bit fallback for these, as the layout is different. So we add a symlink to define the 0x20426 layout to be the same as the 0x10426 layout.


matt335672:
Rebased and fixed a couple of comment typos.

PR #3524, opened 2025-05-08
[V0.10] vmconnect mode: support all security modes when used in a Hyper-V environment

matt335672:
Backport of #3514 to V0.10

~~Draft for now, as testing required which I may not have time to do for several days.~~ The first commit in this PR makes xrdp_iso.c and xrdp_rdp.c from v0.10 more like that from devel so the cherry-picks work cleanly.

@metalefty - I'm assuming we want this in V0.10?


metalefty:
Thanks, let's merge this in v0.10.

metalefty (approved):

Issue #1443, opened 2019-11-14
Auto reconnect to xrdp


Serg-VL:
Preface: At work I have a buggy wi-fi, which causes regular reconnects.
After network cut-off rap connection to Windows Server restores automatically.
While connection to xrdp shows error The connection disconnected because network unavailable. Error code: 0x6000003
And I'm to accept error and reconnect manually.
Despite manual reconnect always successful, this is annoying to make some clicks every time...

tpacict:
I have a customer with the same problem. A brand new MacBook,  OS Catalina with latest Microsoft RDP application connected to a Cisco Meraki MR33 Wifi Accesspoint. Other MacOS users on this network don't have RDP errors. Wifi logs are clean.
![image](https://user-images.githubusercontent.com/2808025/73485552-94004f80-43a3-11ea-895b-efe6febd3ad3.png)

PR #3547, opened 2025-06-21
SCP updates : connect session message

matt335672:
This PR reworks the SCP protocol to split the 'create session' message into separate 'create session' and 'connect session' messages.

This has little user-visible changes, apart from:-
1)  `SessionSockdirGroup` in `sesman.ini`. For v0.10.2 and later, this is generally set to `xrdp` to allow the xrdp process running as `xrdp:xrdp` to connect directly to sockets. With this PR, this is no longer necessary, as sesman makes socket connections on behalf of xrdp. The value can now be set (or defaulted) to root for more secure configurations.
2) Running `xrdp-sesrun` for an already existing session now no longer causes a phantom run of the reconnectwm.sh script.

The point of the PR is to make a connection to a session a separate, identifiable action. This will allow us to implement the following functionality:-
1) A single connection to a session will be enforced. This fixes a number of PRs around chansrv not working on a reconnection.
2) We can provide dynamic information on the connected client (if any)
3) [Later] reconnection can be implemented (See #1443)

I've updated some wiki pages. New copies are on my wiki as follows:-

- https://github.com/matt335672/xrdp/wiki/The-socketdir-directory
- https://github.com/matt335672/xrdp/wiki/SCP-Sequence-Diagrams
- https://github.com/matt335672/xrdp/wiki/Session-management-architecture

At the source level, the following are worthy of note:-
1) A new module `set_int` which allows sets of integers to be manipulated with a bitmap. There's a test suite for this.
2) the `mod_connect()` call now allows a file descriptor to be passed in.  If this is done, the connect can be made immediately by assigning the file descriptor to the transport.
3) An extra state is added to the connection state machine in xrdp_mm.c to allow file descriptors to be acquired from sesman.

I'm making this draft for now while I perform some more testing. Please feel free to ask me any questions about the rationale, the design or the code.


tsz8899:
Thank you
Could you please describe on the plans listed below

~~~
SessionSockdirGroup in . For v0.10.2 and later, this is generally set to to allow the xrdp process running as to connect directly to sockets. With this PR, this is no longer necessary, as sesman makes socket connections on behalf of xrdp. The value can now be set (or defaulted) to root for more secure configurations.sesman.inixrdpxrdp:xrdp
~~~
Does it represent best practice to set runtime_user=xrdp and runtime_group=xrdp in xrdp.ini, and set SessionSockdirGroup=root in sesman.ini?
Or is it better to not set these values and simply use the defaults in the future?
Is it still necessary to create the xrdp user and group?



~~~
Running for an already existing session now no longer causes a phantom run of the reconnectwm.sh script.xrdp-sesrun
~~~
In my environment, each time a session is reconnected in the current version, reconnectwm.sh is executed. This script is used to write client IPs, rdp /app: remote application program info, and other forwarding information for xrdp.
Will it be possible to add a switch in the future to force or automatically run reconnectwm.sh?



~~~
We can provide dynamic information on the connected client (if any)
~~~
In the current version, I'm using a rough approach in to deliver information to the user's home directory.xrdp_mm.c (IP, domain, remote app, printer, LDAP,info) 
Will there be a reliable channel for this in the future?

matt335672:
Hi @tsz8899 

As usual, you raise some excellent points and questions - thank you.

> Does it represent best practice to set runtime_user=xrdp and runtime_group=xrdp in xrdp.ini, and set  SessionSockdirGroup=root in sesman.ini?
> Or is it better to not set these values and simply use the defaults in the future?
> Is it still necessary to create the xrdp user and group?

With this update, best practice for a normal configuration will be to set `runtime_user` and `runtime_group` to the username and group of an unprivileged account, and simply not set `SessionSockdirGroup`. Some rarer configurations may still require `SessionSockdirGroup` to be set. The only one I'm aware of is when running chansrv in a VNC session which has not been started by sesman.

You will still need to create the unprivileged account, or use an existing one.

> Will it be possible to add a switch in the future to force or automatically run reconnectwm.sh?

I'll add this - great suggestion.

> In the current version, I'm using a rough approach in to deliver information to the user's home directory.xrdp_mm.c (IP, domain, remote app, printer, LDAP,info)
> Will there be a reliable channel for this in the future?

I'm working on a PR based on this one which will give me a channel between the xrdp process and the sesman process. That should let you query the connected client using `xrdp-sesadmin`. The information available will change as different clients disconnect and reconnect.

PR #3572, opened 2025-07-21
Update to cppcheck 2.18.0

matt335672:
Update to cppcheck 2.18.0

This version of cppcheck is generating false positives on xrdp/xrdp_tconfig.c. This change works around that, but I've also sent a message to https://sourceforge.net/p/cppcheck/discussion/general/ to raise an issue. There's no issue yet, as are some hoops to jump through. However, I've sent this test program which illustrates the false positive:-

```c
#include <stdlib.h>

struct test
{
    int valid;
    char *mem;
};

static struct test get_alloc(void)
{
    struct test result = {0, NULL};

    if ((result.mem = malloc(10)) != NULL)
    {
        result.valid = 1;
    }

    return result;
}

int main()
{
    struct test t;

    t = get_alloc();
    if (t.valid)
    {
        free(t.mem);
    }

    t = get_alloc();
    if (t.valid)
    {
        free(t.mem);
    }
}
```

Then:-

```
$ ~/cppcheck.local/2.18.0/bin/cppcheck temp.c
Checking temp.c ...
temp.c:34:9: error: Memory pointed to by 'mem' is freed twice. [doubleFree]
        free(t.mem);
        ^
temp.c:28:9: note: Memory pointed to by 'mem' is freed twice.
        free(t.mem);
        ^
temp.c:34:9: note: Memory pointed to by 'mem' is freed twice.
        free(t.mem);
        ^
```
I'm marking this draft for now, pending a response from the cppcheck team.

@firewave - you asked to be notified of any cppcheck issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
       CC: gcc
       # This is required to use a version of cppcheck other than that
       # supplied with the operating system
-      CPPCHECK_VER: "2.17.1"
+      CPPCHECK_VER: "2.18.0"
       CPPCHECK_REPO: https://github.com/danmar/cppcheck.git
     steps:
       # Set steps.os.outputs.image to the specific OS (e.g. 'ubuntu20')
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
       CC: gcc
       # This is required to use a version of cppcheck other than that
       # supplied with the operating system
-      CPPCHECK_VER: "2.18.0"
+      CPPCHECK_VER: "2.18.3"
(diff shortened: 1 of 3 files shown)


matt335672:
cppcheck trac ticket created:-

https://trac.cppcheck.net/ticket/14039#ticket

firewave:
I think this should be merged - after a bump to `2.18.3` and a rebase.

matt335672:
Bumped to 2.18.3 and rebased.

@firewave - there appears to be no progress on the TRAC ticket, and it's not in a milestone. Do I take it this is unlikely to be fixed in the near future?

firewave:
> Do I take it this is unlikely to be fixed in the near future?

Yes. If you do not see any progress within a few days or hours that is usually the case.

PR #3519, opened 2025-05-05
[V0.10] Add Latvian keyboards

matt335672:
Backport of #3518 to v0.10

These files should work OK on v0.9.x systems.

The keyboard ini files are built on Ubuntu 20.04 to prevent a non-working capslock key (see #3022)

PR #3593, opened 2025-08-25
Add Korean keyboard file

lancard:
add 'km-e0010412.ini' file into instfiles.


metalefty (changes requested):
We no longer use ini for keymap files.

lancard:
Okay. Close this pr. Have a nice day.

matt335672:
@lancard - are you able to build from source? If so, we might be able to get a modern keyboard file sorted for you.

lancard:
It’s possible. I can help if needed.

Although it’s only been a few days, I’m creating a lightweight Docker for a Korean X Window environment using xrdp. Many Koreans have shown interest. If you’re interested, please share your thoughts.
https://github.com/lancard/xwindow-korean

matt335672:
We're always interested in getting xrdp into new environments.

I can easily generate a new-style toml file for you to try based on the `devel` stream.

These are easier to debug. Once this is done, we can backport the toml file to an ini file for xrdp v0.10.x. That's more complicated. We have to do it this way to make sure your keyboard is in the next major release of xrdp.

You can get a rough idea of the process by looking at #3518 and #3519 which was a Latvian keyboard we recently added.

If you're interested, let me know and I'll build a branch off devel with a Korean toml keyboard in it.

lancard:
Got it. I’ll take a look around. I’ll let you know when I’m ready.

lancard:
I've looked around, and I can do it. I'd appreciate it if you could make a branch.

matt335672:
Hi @lancard 

I've been looking into Korean keyboards, and I can't find a lot of information in a language I can read on the web. My favourite reference for keyboards (https://www.kbdlayout.info/) doesn't mention your keyboard variant at all.

Do you have a target keyboard working on a Linux machine somewhere? If so, what does this command tell you:-

```
setxkbmap -query
```

There are a number of options for Korean keyboards which can be set, and I don't know which ones correspond to the keyboard you are looking for.

Thanks.

lancard:
As i mentioned, https://github.com/lancard/xwindow-korean/blob/master/user-entrypoint.sh refers keyboard layout.
Plz check last line. :)

matt335672:
That's great - thanks!

matt335672:
See #3598 

Branch is at https://github.com/matt335672/xrdp/tree/korean_keyboard

I'm making the assumption that the 106 Key Korean keyboard reports itself as type 8 as specified in [[MS-RDPBCGR] 2.2.1.3.2 Client Core Data ](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/00f1da4a-ee9c-421a-852f-c19f92343d73). This should get the keyboard mappings working correctly for the xorgxrdp backend.

@lancard - can you try this out and let me know how it goes?

lancard:
sounds great. will write comment.

PR #3598, opened 2025-08-27
Add Korean Hangul keyboard

matt335672:
Add a mapping for the 106 key Korean Hangul keyboard.


metalefty:
I think @unstabler can review this. I'm not sure he follows recent xrdp though.

PR #3606, opened 2025-09-05
Fill in all fields for TS_DEACTIVATE_ALL_PDU

matt335672:
See Devolutions/IronRDP#314, and in particular [this comment](/Devolutions/IronRDP/issues/314#issuecomment-3254275590)

At the moment our TS_DEACTIVATE_ALL_PDU is not compliant with [MS-RDPBCGR], but rather emulates what Windows XP does.


metalefty:
LGTM. It appears we never cared about that.

matt335672:
Reading the linked comments, it seems we just did what Windows XP did.

I'll wait for a Coverity scan to run (it's very slow at the moment), then merge this and backport to V0.10, along with #3607

PR #3612, opened 2025-09-09
Code improvements

matt335672:
Here are a few changes I've wanted to do for a while:

- Another tranche of constants from [ms-rdpbcgr.h](/neutrinolabs/xrdp/blob/devel/common/ms-rdpbcgr.h) are renamed to match the defines in [MS-RDPBCGR]. This improves traceability back to the specification.
- A couple of `void *` pointers in `struct xrdp_session` are replaced with pointers to incomplete types. This allows us to remove a very large number of unnecessary C casts, primarily from libxrdp.

No functional changes are introduced by this PR. I'm not intending to back-port it to V0.10.x

### DIFF
--- a/common/ms-rdpbcgr.h
+++ b/common/ms-rdpbcgr.h
@@ -449,21 +449,21 @@
     )
 
 /* Share Data Header: pduType2 (2.2.8.1.1.1.2) */
-/* TODO: to be renamed */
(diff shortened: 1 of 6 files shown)


metalefty (approved):
Thank you!

PR #3613, opened 2025-09-10
Use fastpath for login screen

### LINKED ISSUE #3609: Login screen uses X224 PDUs even when fastpath enabled
Applies to all versions of xrdp.

The login screen uses X.224 (slowpath) PDUs for updates, even when fastpath is enabled.

Although this is perfectly fine as far as the specification goes, this causes incompatibilities for some RDP clients, such as IronRDP (Devolutions/IronRDP#314). The workaround for such client is to use autologin

matt335672:
Fixes #3609 

Draft for now, as there is unexplained screen corruption with IronRDP. ~~Also this PR depends on #3612 which has not yet been merged.~~

### DIFF
--- a/libxrdp/libxrdp.c
+++ b/libxrdp/libxrdp.c
@@ -440,6 +440,71 @@ libxrdp_send_bell(struct xrdp_session *session)
 }
 
 /*****************************************************************************/
+/**
+ * Initialises either a TS_UPDATE_BITMAP or a TS_FP_UPDATE_BITMAP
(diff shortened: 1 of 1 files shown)


jsorg71:
This looks good to me.  I did not see any corruption.

matt335672:
We got to the bottom of the corruption with IronRDP:-

https://github.com/Devolutions/IronRDP/issues/314#issuecomment-3311868571

I'll merge this into devel.

Issue #3614, opened 2025-09-10
Can't move / copy files in dolphin to redirected drives in KDE


vscfreire:
### xrdp version

0.10.3

### Detailed xrdp version, build options

~~~
~~~

### Operating system & version

Rocky Linux 9.6

### Installation method

dnf / apt / zypper / pkg / etc

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

KDE

### Environment xrdp running on

VM

### What's your client?

mstsc.exe (Windows 11)

### Area(s) with issue?

File transfer / drive redirection

### Steps to reproduce

1 - Create an example .txt file (test.txt) on your home directory
2 - Open dolphin and copy the file
3 - Navigate to a redirected folder (such as ~/thinclient_drives/C:/Users/<User>/Desktop/)
4 - Attempt to paste the file

### ✔️ Expected Behavior

The file should be copied to the desktop

### ❌ Actual Behavior

The file isn't copied to the desktop and an error appears "There is not enough space on the disk to write"

### Anything else?

Hi,

When copying files from a KDE (Plasma 5.27.12) xrdp session using Dolphin to the Windows desktop over xrdp drive redirection, Dolphin always reports:
 
`There is not enough space on the disk to write` 

Using either "cp" or using the "Save as" file dialogs works as expected.

I'm aware of a similar issue https://github.com/neutrinolabs/xrdp/issues/3300, which was closed by PR https://github.com/neutrinolabs/xrdp/pull/3304. From what I've gathered this fix is included in the version I am using. Running `df -h` as mentioned in that PR shows that the fix is working as intended:

```
df -h ~/thinclient_drives/  ~/thinclient_drives/C\:/
Filesystem      Size  Used Avail Use% Mounted on
xrdp-chansrv    0.0K  0.0K  0.0K    - /home/<user>/thinclient_drives
xrdp-chansrv    953G  782G  171G  83% /home/<user>/thinclient_drives
```

In a comment in PR https://github.com/neutrinolabs/xrdp/pull/3304 it suggests that the fix was tested in dolphin and that it was working as expected, which made me want  to take a close look at KIO's source code. Inspecting the code I found out the function responsible for getting the directory's available space when a copy is performed (https://invent.kde.org/frameworks/kio/-/blob/v5.116.0/src/ioslaves/file/file.cpp#:~:text=void%20FileProtocol%3A%3AfileSystemFreeSpace(const%20QUrl%20%26url):

[...]

PR #3632, opened 2025-10-06
Add FuseRootReportMaxFree option

matt335672:
Temporary fix for #3614 

Add an option to allow effectively disable file system space checks for some file managers before copying files to remote drives.

This is a temporary solution. A better solution is to provide each remote drive with its own mountpoint, so that the xrdp FUSE filesystem becomes POSIX compliant.

With this option set on a 64-bit Linux system:-

<pre>
$ <b>stat -f ~/thinclient_drives/</b>
  File: "/home/testuser/thinclient_drives/"
    ID: 0        Namelen: 1023    Type: fuseblk
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 2251799813685248 Free: 2139209823000986 Available: 2139209823000986
Inodes: Total: 0          Free: 0
$ <b>df -h ~/thinclient_drives/</b>
Filesystem      Size  Used Avail Use% Mounted on
xrdp-chansrv    8.0E  410P  7.7E   5% /home/testuser/thinclient_drives
</pre>

@metalefty, @vscfreire - Please comment / review if you have the time.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,7 @@ AM_CONDITIONAL(MACOS, [test "x$macos" = xyes])
 AC_CHECK_SIZEOF([int])
 AC_CHECK_SIZEOF([long])
 AC_CHECK_SIZEOF([void *])
+AC_CHECK_SIZEOF([fsblkcnt_t],,[#include <sys/statvfs.h>])
 
 # runstatedir not available for autoconf <= 2.69
 if test "x$runstatedir" = "x" ; then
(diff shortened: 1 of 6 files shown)


vscfreire:
Hi,

Sorry for the delay, I was only made aware of this when I checked the PR itself.

I think this PR does exactly what we are looking for. Nice!

Thanks.

matt335672:
Thanks for the comment - I'll merge this then, and backport to v0.10.

volkov0aa:
> A better solution is to provide each remote drive with its own mountpoint, so that the xrdp FUSE filesystem becomes POSIX compliant.

It's possible to return different values for different inodes from statfs.

matt335672:
Well it is, and that's what we're currently doing. However, it can confuse programs which aren't expecting it, and my reading of the POSIX spec. suggests that xrdp is in the wrong here.

volkov0aa:
Oh, I see, sorry!

PR #3637, opened 2025-10-13
Fix `PAM_TTY` item being set to uninitialized data in PAM userpass lib

AlexTMjugador:
The local variable it referenced was never initialized before use. I replaced it with a reference to the seemingly intended `SERVICE` preprocessor constant, which is also used in the `pam_start` call slightly above.


matt335672:
Nice catch - thanks.

I really don't know if anyone is still using `pam_userpass.so`. It seems unlikely, but I think it's worth fixing anyway.

PR #3640, opened 2025-10-16
xrdpapi: Add a way to monitor connect/disconnect events

### LINKED ISSUE #3622: Disconnecting custom static virtual channels
Hi,
 
I’ve been exploring ways to integrate custom channels into xrdp, and I came across your xrdpapi virtual channel implementation. I built a custom static channel using FreeRDP’s client, and so far everything is working as expected (read, write, open, close).
 
However, I noticed that when I disconnect from a session, my custom channel isn’t being closed properly. As a result, when I reconnect to the same session, the channel being used is still the old one rather than a new instance. To get this working properly, I need to reopen the channel.
 
Microsoft's official RDP documentation suggests using `WTSQuerySessionInformationA` ([https://learn.microsoft.com/en-us/windows/win32/api/wtsapi32/nf-wtsapi32-wtsquerysessioninformation…](https://learn.microsoft.com/en-us/windows/win32/api/wtsapi32/nf-wtsapi32-wtsquerysessioninformationa)) to get the connection's state. At a first glance the process I use to manage the channel could just call this function to get the connection's state, closing the channel if the state isn't active and opening it when it becomes active again. However xrdpapi only implements WTSVirtualChannelQuery.
 
Additionally, other channels, such as devredir or sound, have associated init() and deinit() functions in chansrv, which are called when the client connects / disconnects respectively. However, there doesn't seem to be a mechanism in place for externally managed channels.
 
Do you have any suggestions on how to approach this problem?
 
Thanks.

matt335672:
Fixes #3622

Demo program included.

@vscfreire - this is a draft for now. I'd very much appreciate any comments you have. The API is a little different from what we discussed, and you may have a different (and better) mental image of how this all should work.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -47,6 +47,7 @@ libcommon_la_SOURCES = \
   arch.h \
   base64.h \
   base64.c \
+  channel_defs.h \
   defines.h \
   fifo.c \
   fifo.h \
--- a/common/channel_defs.h
+++ b/common/channel_defs.h
@@ -1,0 +1,69 @@
+/**
+ * Copyright (C) 2024 Matt Burt, all xrdp contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @file    common/channel_defs.h
+ * @brief   Private xrdp channel definitions
+ */
+
+#if !defined(CHANNEL_DEFS_H)
+#define CHANNEL_DEFS_H
+
+#include <stdint.h>
+
+/**
+ * Channel IDs assigned to xrdp private channels. These are chosen to avoid conflicts with
+ * channel IDs returned by other servers.
+ */
+enum
+{
+    CHAN_ID_XRDP_BASE = (('x' << 24) | ('r' << 16) | ('d' << 8) | 'p'),
+    CHAN_ID_XRDP_SESSION_INFO = CHAN_ID_XRDP_BASE,
+    // Add further IDs here
+    CHAN_ID_XRDP_MAX
+};
+
+// Max length of a DVC name. This is taken from the specification of
+// WTSVirtualChannelOpenEx() which limits the length of a virtual channel
+// to Windows MAX_PATH
+#define MAX_DVC_NAME_LEN 260
+
+/**
+ * Information to connect to a channel using xrdpapi
+ */
+#define XRDPAPI_CONNECT_PDU_LEN 384 // Connect PDU is always this length
+#define XRDPAPI_CONNECT_PDU_VERSION 1 // Current connnect PDU version
+
+struct xrdp_chan_connect
+{
+    uint32_t version;
+    uint32_t private_chan; // 0 for a standard channel
+    uint32_t flags;
+    char name[MAX_DVC_NAME_LEN + 1]; // null-terminated
+};
+
+// Events received on the CHAN_ID_XRDP_SESSION_INFO channel
+
+// Update on all the session state data held by chansrv
+//
+// If more information is needed to be exported by xrdpapi, it can be
+// added to this event
+struct xrdp_chan_session_state
+{
+    uint32_t is_connected; // Is the session connected?
+};
+
+#endif /* CHANNEL_DEFS_H */
--- a/common/parse.h
+++ b/common/parse.h
@@ -529,6 +529,14 @@ in_utf16_le_terminated_as_utf8_length(struct stream *s);
     } while (0)
 
 /******************************************************************************/
+// Platform-endian out_uint32
+#if defined(B_ENDIAN)
+#define out_uint32_pe(s, v) out_uint32_be(s, v)
+#else
+#define out_uint32_pe(s, v) out_uint32_le(s, v)
+#endif
+
+/******************************************************************************/
 #define in_uint8p(s, v, n) do \
     { \
         S_CHECK_REM((s), (n)); \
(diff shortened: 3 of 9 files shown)


vscfreire:
Hi,

First of all, apologies for the delay.

I think this is a possible solution even if it is different from what we've discussed. My main issue, and please correct me if I am wrong, is that we might not be using virtual channels appropriately here. As I understand it, virtual channels are intended for sending and receiving data between the RDP client and server, however this implementation seems to go against that for two reasons:

1. This channel is not something that is requested by the RDP client
2. It is not possible to send data to that channel (in the chansrv changes there is a specific if clause to prevent that)

From my perspective, this should be something that is separate from virtual channel communication. The quickest solution that comes into my mind is using a separate socket for this such as `xrdpapi_$DISPLAY_management`, however I am not sure how great of an undertaking this is.

Please do let me know your thoughts.

Thanks.

matt335672:
@vscfreire - as ever, thanks for the feedback and an interesting discussion. No apologies are necessary.

Am I right in that your concerns are related to the implementation detail only? I don't think the choice of implementation would affect the API as presented to the user.

At an OS level, the sockets for each channel (including this one) are all separate anyway. I'd thought when I started looking at this that all the channels would be multiplexed over a single socket, but that's not the implementation that was chosen by the original developer. Each channel ends up with its own socket. The file system entry `xrdpapi_$DISPLAY` is simply used as an endpoint for the client to connect to.

It would be simple enough to modify the initial PDU sent to an open xrdpapi socket to include a value specifying whether or not this was a private channel of some sort. That would prevent the user being able to open `xrdp/SessInfo` directly, if this is a concern, and effectively separates all the sockets out from each other.

Let me know if that would help.

I've one other question;  having thought about this myself in the last few days, I was expecting you to come up with a question regarding how the user determines the initial state of the RDP client connection when running an xrdpapi client up. Is this a concern of yours? If so, what sort of mechanism would you like to see to get this information at an API level?

vscfreire:
Hi,

> Am I right in that your concerns are related to the implementation detail only? I don't think the choice of implementation would affect the API as presented to the user.

Yes. In terms of functionality this should work.

> It would be simple enough to modify the initial PDU sent to an open xrdpapi socket to include a value specifying whether or not this was a private channel of some sort. That would prevent the user being able to open xrdp/SessInfo directly, if this is a concern, and effectively separates all the sockets out from each other.

I think this is a great idea.

> I've one other question; having thought about this myself in the last few days, I was expecting you to come up with a question regarding how the user determines the initial state of the RDP client connection when running an xrdpapi client up. Is this a concern of yours? If so, what sort of mechanism would you like to see to get this information at an API level?

From my perspective, the xrdpapi client first tries to use `WTSRegisterSessionNotificationEx` to begin listening to session events. Afterwards, if successful, the client could use an implementation of `WTSQuerySessionInformationA` to get the initial state of the connection. Chansrv can probably store the connection state internally. Even if the client disconnected when `WTSQuerySessionInformationA` got an initial connect state the client would be aware of that disconnection thanks to `WTSRegisterSessionNotificationEx`. Let me know your thoughts on this.

Another question I have is how the client should handle a `WTSRegisterSessionNotificationEx` fail. For my use case, I think it would be perfectly acceptable that if the client starts up before chansrv has initialized the virtual channels (causing  `WTSRegisterSessionNotificationEx` to fail) the client can simply wait a bit before attempting again. However, since `WTSVirtualChannelOpenEx` can fail for different reasons (such as the display not being set) it might be helpful to have different error values as there are some situations where a retry might be useless (when the display is not set for instance).

Thanks.

matt335672:
OK - I'll add a private channels flag.

`WTSQuerySessionInformationA` should be easy enough to do. The simplest implementation might be to send an event immediately after the channel is opened to prime the client with the current state. This event would not be returned to the client and would be consumed before the `WTSRegisterSessionNotificationEx` call returns. Subsequent events would be cached before the client callback is issued.

I see the problem with `WTSRegisterSessionNotificationEx`. Since there's a race beween chansrv starting and the session starting, this could easily happen. 

We could implement a special function `WTSGetLastError()` maybe which would return a restricted subset of values for errors. Or we could return an error from the WTS* functions. The `WTSGetLastError()` function wouldn't be thread-safe, unlike the Windows `GetLastError()`. Is that a problem do you think?

vscfreire:
Hi,

I can't think of a reason as to why that would be a problem for this use case, but I also don't see the benefit of using that over returning an error from the WTS functions. Could you clarify why implementing WTSGetLastError() might be better?

Thanks

matt335672:
I'm not really fussed either way myself. The `WTSGetLastError()` would allow for a code flow more like the Windows API, but that's not a big consideration. Do you have a preference?

vscfreire:
Hi,

To simplify for now, I think it is best to just stick with returning an error error from the WTS functions

matt335672:
Updated as requested, with minimal testing so far!

The error return was trickier to do than I expected. Because Windows uses true for OK, I've only got the value 0 for an error return, so as to preserve some kind of compatibility. Consequently I've added an optional error code to some of the calls.

Let me know what you think. We're still in draft at the moment.

matt335672:
As a small note, I had a think about this last night, and I'm sure the interface between xrdpapi and chansrv can be improved somewhat.

That's not related to the API however. I'll save more updates until my next batch of changes.

vscfreire:
Hi,

Apologies for the delay.

It seems to me this is exactly what I am looking for. When I get the chance, I will test this code in my custom channel.

Thanks for all the hard work

matt335672:
That's great @vscfreire 

I'll currently looking at a couple of other issues, but when I've got those done I'll rework the internal API. That's not going to affect xrdpapi however, so we should be able to work in parallel.

matt335672:
Internal API reworked to remove 'Windows' definitions from chansrv.

vscfreire:
Hi,

I've finally managed to cherry pick the changes into my local  xrdp repository and test it against my custom channel and I'm glad to say that it seems to be working.

The only change I had to make was replacing ```strlcpy``` with ```strncpy``` (xrdpapi.c line 198).

Thanks

matt335672:
Thanks for the update @vscfreire 

Sorry about the `strlcpy` confusion - I'd forgotten about that.

I'll get this merged an back-ported when I can.

matt335672:
@jsorg71 - if I'm reading the git history correctly, you were the original author of the xrdpapi facility. If you get time to have a look at this before we merge, I'd be grateful.

I'll back-port to V0.10.x and re-test.

matt335672:
Retested, and regression fixed in xrdpapi/simple.c

PR #3642, opened 2025-10-17
session management: Allow for restricted shells

### LINKED ISSUE #3624: [Feature request] Allow for restricted alternate shells
This feature request has been prompted by discussion #3621.

At the moment, alternate shells can be specified by the user, to replace `/etc/xrdp/startwm.sh`, provided the configuration variable `AllowAlternateShell` is set to `true`. This allows users to run any script of their choosing, but also allows them to bypass any setup configurations which the system manager may have added to  `/etc/xrdp/startwm.sh`

The proposal is to add an additional config variable `PassShellAsEnv` to go with `AllowAlternateShell`:

```
; If this is set, alternate shells are not actioned directly, but passed in to the default window manager
; in the specified environment variable. 
#PassShellAsEnv=XRDP_ALTERNATE_SHELL
```

With that set, within  `/etc/xrdp/startwm.sh` the system manager can configure something like the following (as an example):-

```sh
   case "$XRDP_ALTERNATE_SHELL"
        '' | gnome)
            # Perform GNOME setup
            ;;
        kde)
            # Perform KDE setup
            ;;
        *)     zenity --error --text="Unrecognised shell option '$XRDP_ALTERNATE_SHELL'"
            exit 1
    esac
```

The advantage of this approach is everything remains within control of the system manager.

Comments? Is this a worthwhile addition to xrdp, or are there better ways to do this?

matt335672:
Fixes #3624 

Allows the `AlternateShell` specified by the user in the TS_INFO_PACKET to be passed to startwm.sh as an environment variable.

Draft PR for now, as I've not finished module testing yet; I'm going to be a bit busy for a few days IRL.

~~One thing I'd really like to test is using `PassShellAsEnv=DESKTOP_SESSION` to allow the user to specify which desktop they'd like to run on Debian/Ubuntu. I've added a comment about this to `startwm.sh`~~

**Update 2025-10-20:** Running on Ubuntu 24-04, I'm able to choose between XFCE and GNOME sessions by using `PassShellAsEnv=DESKTOP_SESSION`, and setting the startup shell to either `ubuntu or `xfce`.

@akarl10, @metalefty, @gumerov-r - any comments will be very welcome.

### DIFF
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -366,6 +366,13 @@ because the system is unable to check whether the user is an administrator.
 If set to \fB0\fR, \fBfalse\fR or \fBno\fR, prevent usage of alternate shells by users.
 
 .TP
+\fBPassShellAsEnv\fR=\fI<name-of-environment-variable>\fR
+If this is set, alternate shells are not actioned directly, but
+passed in to the default window manager in the specified environment
+variable. This allows the system manager more control over exactly
+what alternate shells are permitted.
+
+.TP
 \fBXorgNoNewPrivileges\fR=\fI[true|false]\fR
 Only applicable on Linux. If set to \fB0\fR, \fBfalse\fR or \fBno\fR, do
 not use the kernel's \fIno_new_privs\fR restriction when invoking the Xorg
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -372,6 +372,9 @@ passed in to the default window manager in the specified environment
 variable. This allows the system manager more control over exactly
 what alternate shells are permitted.
 
+This will override any setting of the same environment variable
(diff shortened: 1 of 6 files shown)


akarl10:
this looks very good. looking at the code the PassShellAsEnv should override `[SessionVariables]`. Even tough this is what people most probably expect, I think it should be documented.

The reason why I mention this is because currently the package in my ppa does set a DESKTOP_SESSION variable under the above mentioned key. This would be the default and the user might override this

matt335672:
Thanks @akarl10 

Does this change meet your requirement? If so, I'll merge this and backport to v0.10.

akarl10:
> Thanks @akarl10
> 
> Does this change meet your requirement? If so, I'll merge this and backport to v0.10.

of course it does 👍

Just checked this patch on my package and it works.

PR #3647, opened 2025-10-26
do not delete pid file unless this current pid matches the daemon pid

jsorg71:
current devel, the xrdp pid file gets deleted as when the first client connects.  This does a check to only delete the pid file if it's the daemon pid that is exiting.


matt335672 (approved):
Thanks Jay,

It looks like I introduce this in #2974, which has been merged in to v0.10.

If you're happy to merge this, I'll manage the backport.

PR #3648, opened 2025-10-26
pass frame id back to main thread in jpeg encode

jsorg71:
It seems jpeg has been broke since we did gfx, this simple fix re-enables jpeg.


matt335672 (commented):
LGTM. If you merge this I'll handle the backport.

matt335672 (approved):

PR #3657, opened 2025-10-31
enabled and fixed `missingInclude` Cppcheck warnings

firewave:
This enables the `missinInclude` warning to make sure all (local) includes are provided.

PR #3658, opened 2025-11-03
Code quality: Remove 'struct xrdp_process *' casts

matt335672:
The first argument to `libxrdp_init()` is an `intptr_t` / `tbus` value. This represents the xrdp instance which is using the library, but this value is always a `struct xrdp_process *` pointer.

This PR replaces the `intptr_t` with an incomplete type declaration at the interface between xrdp and libxrdp.

The original intention was probably to provide some separation from xrdp and the libxrdp code, but in practice this has turned out not to be useful.

This improves readability somewhat in the callbacks, and also unblocks the cppcheck 2.17.1 error in #3657, as cppcheck is no longer confused by the case of `self` to an `intptr_t`.

### DIFF
--- a/libxrdp/libxrdpinc.h
+++ b/libxrdp/libxrdpinc.h
@@ -65,12 +65,14 @@ struct xrdp_rect
 
 struct xrdp_rdp;
 struct xrdp_orders;
+struct xrdp_process;
 
 struct xrdp_session
 {
-    tintptr id;
+    struct xrdp_process *id;
     struct trans *trans;
-    int (*callback)(intptr_t id, int msg, intptr_t param1, intptr_t param2,
+    int (*callback)(struct xrdp_process *id, int msg,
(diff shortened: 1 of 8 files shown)


libxrdp/libxrdpinc.h
```diff
@@ -98,13 +102,14 @@ struct display_size_description;
 /***
  * Initialise the XRDP library
  *
- * @param id Channel ID (xrdp_process* as integer type)
+ * @param proc XRDP instance to use with this library
```
firewave, line 105:
```suggestion
 * @param id XRDP instance to use with this library
```

The parameter is still called `id`.

firewave:
LGTM. Using the actual type is much clearer and safer.

matt335672:
@firewave - could you try rebasing #3657?

PR #3660, opened 2025-11-03
configure.ac: enabled `-Wextra` compiler warning

firewave:


matt335672:
@firewave - I've taken the liberty of adding a couple of commits to your PR.

1) We use C zero initializers quite a bit in xrdp. These are well-defined, but obviously fail foul of -Wmissing-field-initializers. I've added a workaround for this. Let me know what you think.
2) The other warning related to a macro definition is pretty straightforward.

firewave:
> 1. We use C zero initializers quite a bit in xrdp. These are well-defined, but obviously fail foul of -Wmissing-field-initializers. I've added a workaround for this. Let me know what you think.

If you are fine with adding annotations so am I. I would have just disabled the warnings as to keep the code clean.

matt335672:
@firewave - you're quite right. It's simplest to just disable the warning. My apologies.

If you can clean up the mess I've made of your commit history, we'll merge this.

firewave:
> @firewave - you're quite right. It's simplest to just disable the warning. My apologies.

No problem. I think disabling specific warnings is fine even if they might find interesting things but it is more important to not amass additional technical debt because of that.

I also recently saw code which was addressing something similar via a macro which was used to assign as initial value. I will try to find it again.

> If you can clean up the mess I've made of your commit history, we'll merge this.

Not much of a mess. Resolved by a simple squash.

PR #3667, opened 2025-11-11
enabled and fixed `-Wmissing-variable-declarations` Clang warnings

firewave:


firewave:
Still some some cases to resolve which are not straight forward to fix.

There also appears to be at least one conflict which a global variable with the same name existing in different contexts (which could also be a bug). I will outline this in an additional comment if the changes are complete.

As a follow-up PR (or additional commit if desired) I would rename `g_*` to `s_*` to reflect their actual scope.

matt335672:
I've made a few changes, but I'm unable to push to your branch.

Do you want to take a look at https://github.com/matt335672/xrdp/tree/fw_missing-var-decl ? You should be able to merge it in to your branch and edit/rebase as appropriate.

The FUSE errors are down to incorrectly nested `#ifndef XRDP_FUSE` and `#ifdef XRDP_FUSE` directives. I take your point about the `g_*` prefix being poorly defined.

matt335672:
I updated my branch at 2025-11-14 10:28 approx to fix some more compilation errors. I'm not going to touch it now until I hear from you.

firewave:
You are fast. This was just posted to get a beat on on the prefix and if the duplicated instances might be a bug.

I will take a look later to align your changes into mine.

> I take your point about the `g_*` prefix being poorly defined.

Alright, I will add another commit which adjusts this.

matt335672:
Ready for review?

firewave:
> Ready for review?

Not yet. Still need to cleanup/squash some of the commits.

PR #3671, opened 2025-11-18
do not change port when not using sesman

shua21:
There is an issue where vnc-any does not work correctly when the port is set to 5901.
Even when sesman is not used, xrdp still references the display number and rewrites the port to 5900 + display. Since the display is always 0 in this mode, the connection is always forced to port 5900.

This patch modifies the behavior so that when sesman is not used, xrdp will no longer rewrite the VNC port, allowing vnc-any to work as intended.


matt335672:
Hi @shua21 

Thanks very much for this.

You'll need to reformat the braces in your code slightly before merging:-

https://github.com/neutrinolabs/xrdp/wiki/Coding-Style

PR #3684, opened 2025-12-03
CI: Bump FreeBSD to 14.3

matt335672:


metalefty:
Good catch. I've just done the same on [vm-bhyve](https://github.com/freebsd/vm-bhyve) repo today.

PR #3695, opened 2026-01-01
User shell environment fix

akarl10:
There is currently a typo in the check if the configuration for the environment variable is set.

Checking tough I saw that the environment variable value is set to an empty string if the user does not set a shell.
The second commit contains a rewrite that only sets the environment variable if the user actually set a shell.
This is to prevent the case where the default value of the environment variable is already set by `[SessionVariables] in sesman.ini

But @matt335672 may have wanted this to behave that way. This is the reason why the PR contains two commits and is marked as draft. Once we decide what to merge I will either drop the second or first commit

### DIFF
--- a/sesman/sesexec/session.c
+++ b/sesman/sesexec/session.c
@@ -255,38 +255,40 @@ start_window_manager(const struct login_info *login_info,
         }
     }
 
-    if (g_cfg->sec.allow_alternate_shell &&
-            g_cfg->sec.pass_shell_as_env != NULL &&
-            g_cfg->sec.pass_shell_as_env[0] != '0')
+    if (s->shell[0] != '\0')
     {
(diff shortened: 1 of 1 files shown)


matt335672:
Hi @akarl10 

Sorry about the delay in getting to this.

I hadn't considered the setting of the environment variable in `[SessionVariables]`. I agree that having a session manager default is best, so I think your second commit is the one we want.

matt335672 (approved):
As discussed, this change only sets the shell environment variable for `pass_shell_as_env`  if the user actually sets a shell. This allows for the system manager to set a global default if required, which my original code didn't.

matt335672:
Sorry - dropped this one as it happened while I had no write access. I'll get on to it now.

PR #3709, opened 2026-01-28
[V0.9] Update CI for security fix release

matt335672:
In order to make a late release of v0.9.27 we need the CI updating so that it compiles and tests successfully.


.github/workflows/build.yml
```diff
@@ -45,12 +45,6 @@ jobs:
             os: ubuntu-latest
             DISTCHECK: true
 
-          - CC: clang
-            feature_set: max
-            arch: amd64
-            os: ubuntu-latest
-            DISTCHECK: true
-
```
firewave, line 53: This looks like an unintentional change as it drops the `clang` + `amd64` combination for `max`.

On a side note there should probably also be a "Maximal debug 64-bit" build with `clang`.

matt335672: Thanks - I'd missed that.

matt335672: I'd removed that stanza during testing. The whole point of removing AC_C_CONST from autoconf was to fix tht build, so I'm very glad you'd spotted that!

PR #3719, opened 2026-02-04
Fix: Segfault at 0x18 due to race condition between UI destruction and event dispatching

tsz8899:
### Summary
High-load environments trigger a Segfault at offset `0x18` in `xrdp` due to a logical race condition between UI cleanup and stale event dispatching.

### 1. Symptom & Physical Interpretation
* **Error**: `kernel: xrdp[...] segfault at 18 ip ...`
* **Physical Offset**: The crash address `0x18` (24 bytes) corresponds to vital function pointers: `notify` in `struct xrdp_bitmap` and `mod_event` in `struct xrdp_mod`.

### 2. Root Cause: Asynchronous Event Queue Gap
The crash is caused by a logical race condition between the **UI Destruction Phase** and **Pending Message Dispatch**. When a user authenticates successfully, the login window is destroyed, but the event loop may still attempt to dispatch stale events via now-NULL pointers.

### 3. Proposed Solution: Local Variable Snapshotting
This PR implements a snapshotting strategy and enhanced NULL checks across 3 commits:
1. **UI Notify Safety**: Guarding all UI-level notification callbacks.
2. **Backend Robustness**: Multi-level validation for module event dispatching.
3. **Lifecycle Cleanup**: Ensuring global state pointers are cleared during object deletion.

This effectively turns potential fatal crashes into safe 'silent skips' during cleanup phases.

### DIFF
--- a/xrdp/xrdp_bitmap.c
+++ b/xrdp/xrdp_bitmap.c
@@ -708,34 +708,42 @@ xrdp_bitmap_invalidate(struct xrdp_bitmap *self, struct xrdp_rect *rect)
     }
     else if (self->type == WND_TYPE_SCREEN) /* 2 */
     {
-        if (self->wm->mm->mod != 0)
+        struct xrdp_wm *wm = self->wm;
+        if (wm == 0)
         {
-            if (self->wm->mm->mod->mod_event != 0)
+            return 0;
+        }
+        struct xrdp_mm *mm = wm->mm;
+        if (mm == 0)
+        {
+            return 0;
+        }
+        struct xrdp_mod *m = mm->mod;
+        if (m != 0 && m->mod_event != 0)
+        {
+            if (rect != 0)
             {
-                if (rect != 0)
-                {
-                    x = rect->left;
-                    y = rect->top;
-                    w = rect->right - rect->left;
-                    h = rect->bottom - rect->top;
+                x = rect->left;
+                y = rect->top;
+                w = rect->right - rect->left;
+                h = rect->bottom - rect->top;
 
-                    if (check_bounds(self->wm->screen, &x, &y, &w, &h))
-                    {
-                        self->wm->mm->mod->mod_event(self->wm->mm->mod, WM_INVALIDATE,
-                                                     MAKELONG(y, x), MAKELONG(h, w),
-                                                     0, 0);
-                    }
-                }
-                else
+                if (check_bounds(self->wm->screen, &x, &y, &w, &h))
                 {
-                    x = 0;
-                    y = 0;
-                    w = self->wm->screen->width;
-                    h = self->wm->screen->height;
-                    self->wm->mm->mod->mod_event(self->wm->mm->mod, WM_INVALIDATE,
-                                                 MAKELONG(y, x), MAKELONG(h, w),
-                                                 0, 0);
+                    m->mod_event(m, WM_INVALIDATE,
+                                 MAKELONG(y, x), MAKELONG(h, w),
+                                 0, 0);
                 }
+            }
+            else
+            {
+                x = 0;
+                y = 0;
+                w = self->wm->screen->width;
+                h = self->wm->screen->height;
+                m->mod_event(m, WM_INVALIDATE,
+                             MAKELONG(y, x), MAKELONG(h, w),
+                             0, 0);
             }
         }
         else
@@ -1144,7 +1152,11 @@ xrdp_bitmap_def_proc(struct xrdp_bitmap *self, int msg,
 
         if (self->focused_control != 0)
         {
-            xrdp_bitmap_def_proc(self->focused_control, msg, param1, param2);
+            struct xrdp_bitmap *fc = self->focused_control;
+            if (fc != 0 && fc->wm != 0)
+            {
+                xrdp_bitmap_def_proc(fc, msg, param1, param2);
+            }
         }
     }
     else if (self->type == WND_TYPE_EDIT)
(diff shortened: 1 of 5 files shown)


xrdp/xrdp_bitmap.c
```diff
@@ -708,35 +708,49 @@ xrdp_bitmap_invalidate(struct xrdp_bitmap *self, struct xrdp_rect *rect)
     }
     else if (self->type == WND_TYPE_SCREEN) /* 2 */
     {
-        if (self->wm->mm->mod != 0)
+        struct xrdp_wm *wm;
+        struct xrdp_mm *mm;
+        struct xrdp_mod *m;
```
matt335672: We're using C99 as a base now (https://github.com/neutrinolabs/xrdp/discussions/2441), so you can declare these variables inline. Some people prefer this. For example:

```
        struct xrdp_wm *wm = self->wm;
        if (wm == 0)
        {
            return 0;
        }

        struct xrdp_mm *mm = wm->mm;
        if (mm == 0)
        {
            return 0;
        }

        struct xrdp_mod *m = mm->mod;
. . .
 ```

This comment applies to a lot of your PR - I've not repeated it below.
I'll update the coding style with this information.

File: xrdp/xrdp_bitmap.c [hunk 2]
```diff
@@ -1120,7 +1134,7 @@ xrdp_bitmap_def_proc(struct xrdp_bitmap *self, int msg,
             {
                 if (self->default_button != 0)
                 {
-                    if (self->notify != 0)
+                    if (self != 0 && self->notify != 0)
```
matt335672: Don't need this check on `self` - `self->default_button` is checked at line 1135 so we know `self` is not NULL.

xrdp/xrdp_bitmap.c
```diff
@@ -1132,7 +1146,7 @@ xrdp_bitmap_def_proc(struct xrdp_bitmap *self, int msg,
             {
                 if (self->esc_button != 0)
                 {
-                    if (self->notify != 0)
+                    if (self != 0 && self->notify != 0)
```
matt335672: Don't need this check on `self` - `self->esc_button` is checked at line 1147 so we know `self` is not NULL.

xrdp/xrdp_bitmap.c
```diff
@@ -988,7 +1002,7 @@ xrdp_bitmap_invalidate(struct xrdp_bitmap *self, struct xrdp_rect *rect)
     }
 
     /* notify */
-    if (self->notify != 0)
+    if (self != 0 && self->notify != 0)
```
matt335672: Don't need this check on `self` - `self` is checked on entry to  `xrdp_bitmap_invalidate` so we know `self` is not NULL - there are no callbacks invoked before this line, so we know the check is still valid. Also, if it is by some chance, you will not call `notify`, but `self` is also dereferenced immediately on return, so we gain nothing.

matt335672:
Thanks @tsz8899 - this looks like a worthwhile addition to me. Comments above. Please feel free to discuss anything you don't agree with in what I've said.

tsz8899:
Hi @matt335672, thank you for the detailed feedback. I've updated the patch to adopt the C99 inline declaration style and removed the redundant NULL checks as you suggested. The core safety logic of using local snapshots is retained. Please take another look when you have a moment.

xrdp/xrdp_wm.c
```diff
@@ -1240,9 +1241,13 @@ xrdp_wm_mouse_move(struct xrdp_wm *self, int x, int y)
 
             if (self->button_down == 0)
             {
-                if (b->notify != 0)
+                if (b != 0 && b->notify != 0)
```
matt335672: Don't need `b != 0` here - `b` is tested at line 1228 before we get here.

xrdp/xrdp_wm.c
```diff
@@ -1630,8 +1636,9 @@ xrdp_wm_key(struct xrdp_wm *self, int keyboard_flags, int key_code)
 
     if (self->mm->mod != 0)
     {
+        struct xrdp_mod *m = self->mm->mod;
         // Backend module loaded...
-        if (self->mm->mod->mod_event != 0)
+        if (m != 0 && m->mod_event != 0)
```
matt335672: `m != 0` is not needed here. The same test is made 3 lines up (`self->mm->mod != 0`).

xrdp/xrdp_wm.c
```diff
@@ -1681,10 +1688,14 @@ xrdp_wm_key_sync(struct xrdp_wm *self, int device_flags, int key_flags)
         self->caps_lock = 1;
     }
 
-    if (self->mm->mod != 0 && self->mm->mod->mod_event != 0)
+    if (self->mm->mod != 0)
     {
-        self->mm->mod->mod_event(self->mm->mod, WM_KEYBRD_SYNC, key_flags,
-                                 device_flags, key_flags, device_flags);
+        struct xrdp_mod *m = self->mm->mod;
+        if (m != 0 && m->mod_event != 0)
```
matt335672: `m != 0` not needed here. The same test is made 3 lines up (`self->mm->mod != 0`)

xrdp/xrdp_wm.c
```diff
@@ -2088,10 +2099,11 @@ xrdp_wm_process_channel_data(struct xrdp_wm *self,
         }
         else
         {
-            if (self->mm->mod->mod_event != 0)
+            struct xrdp_mod *m = self->mm->mod;
+            if (m != 0 && m->mod_event != 0)
```
matt335672: `m != 0` is not needed here. The same test is made 10 lines up (`self->mm->mod != 0`)

matt335672 (commented):
Hi @tsz8899. Thanks for the swift response. A few minor comments below.

matt335672:
We'll need to run a Coverity scan on this, either before or after we merge it. At the moment however, Coverity is down. From https://scan.coverity.com/dashboard

image

The reason I mention this is that Coverity can get confused by some NULL checks, so this is the kind of PR that is more likely to trigger some false positives.

tsz8899:
Hi @matt335672, I've pushed the test update. This version incorporates all your feedback, including the removal of all redundant NULL checks while retaining the C99 inline style.  Thanks again for your guidance!

matt335672:
Thanks @tsz8899.

I'll merge this now. If Coverity (eventually) finds any further issues I'll get back in touch.

Issue #3628, opened 2025-10-03
user namespaces affects session management may be suspended after KillDisconnected


tsz8899:
### xrdp version

0.10

### Operating system & version

debian

### Installation method

git clone & make install

### Steps to reproduce

When KillDisconnected triggers termination of a session, it immediately releases that session’s X11DisplayNumber. However, the session may not have fully terminated at that time, or it may have failed to terminate properly. 
During this interval, a new session may connect and be assigned the same X11DisplayNumber, As a result, all subsequent new sessions are unable to enter the X11 session and instead display the XRDP login screen or a black screen.

If the user manually logout of the session, the X11DisplayNumber is released properly and this problem does not occur.

### ✔️ Expected Behavior

Request / Question
After KillDisconnected, is there a way to verify that an X11DisplayNumber is actually available before notifying the service to allow a new session to use it?
Or can a delay be introduced to ensure the old session has been fully cleaned up before the display number is reused?

matt335672:
Hi @tsz8899 

Thanks for posting this.

I'm having trouble understanding how this is happening to you.

When a session is started, an X display number is allocated by sesman. This is passed to sesexec to start the session. sesman keeps a record of the display number in the session list.
 
While the session is on the session list, the display is unavailable to be allocated to another session. This check is implemented in [session_list_get_available_display()](https://github.com/neutrinolabs/xrdp/blob/v0.10/sesman/session_list.c#L260)

The KillDisconnected is currently implemented in xorgxrdp. When this triggers, the X server will receive SIGTERM and exit. This is picked up by sesexec.

sesexec then kills the rest of the session and exits. sesman picks up on sesexec exiting and removes the session from the list. At that point the display is available for new sessions.

All the postings we get from you are high quality, so I'm sure you're encountering a problem somewhere, but based on the information I've got so far I'm unable to find it.

Have you got any logs for this?

Thanks,

tsz8899:
Thanks

Could you please check if the following analysis is correct?
I’m trying to share a file that both a service and users can access while /tmp user namespaces are enabled.

Failure trigger condition:
The system enables /tmp namespace isolation for regular users, but xrdp, sesman, and xorgxrdp still rely on the shared /tmp/.X11-unix/Xn and /tmp/.Xn-lock files to determine whether a display number is already in use by an existing session.
After the KillDisconnected timeout expires, xorgxrdp first sends a SIGTERM to the X server, and sesexec begins terminating the session.
[...]

PR #3720, opened 2026-02-04
Improve X server detection in sesman by checking /proc command lines

tsz8899:
### Summary
This PR improves the reliability of X server detection in `sesman` by verifying process command lines in `/proc`, addressing concerns raised in issue #3628.

### Problem
Previously, `sesman` relied solely on Unix domain sockets and TCP port checks to determine if a display was in use. This method is susceptible to:
1. **False Negatives**: During X server startup, there is a time window where the process is running but hasn't yet bound its ports or created socket files.
2. **False Positives**: Stale socket files from previous crashed sessions can lead `sesman` to believe a display is occupied when it is actually free.

### Solution
This patch introduces a robust process-level check:
- **New Function**: `check_process_running(int display)` scans the `/proc` filesystem.
- **Logic**: It looks for active processes named `Xorg`, `Xvnc`, or `Xxrdp` that have the matching `:display` parameter in their command line (`/proc/[pid]/cmdline`).
- **Integration**: This check is performed as the primary verification step in `x_server_running_check_ports`.
- **Improved Logging**: Updated logs to clearly state the display number being verified.

This ensures that `sesman` has an accurate view of active X sessions, preventing display allocation conflicts and improving overall session management stability.

### DIFF
--- a/sesman/display_utils.c
+++ b/sesman/display_utils.c
@@ -29,6 +29,8 @@
 #endif
 
 #include <stdio.h>
+#include <dirent.h>
+#include <ctype.h>
 
 #include "arch.h"
 #include "display_utils.h"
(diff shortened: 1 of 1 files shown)


matt335672:
Thanks @tsz8899 

I think we're unlikely to accept this as it's specific to Linux and also to the X server. At the moment we're looking to provide support for non-X11 desktops.

I'd much rather get to the bottom of why you're seeing the issues in #3628.  Your AI report has not been helpful here, as it is introducing other unrelated comments. Your original comment was:

> When KillDisconnected triggers termination of a session, it immediately releases that session’s X11DisplayNumber. However, the session may not have fully terminated at that time, or it may have failed to terminate properly.
During this interval, a new session may connect and be assigned the same X11DisplayNumber, As a result, all subsequent new sessions are unable to enter the X11 session and instead display the XRDP login screen or a black screen.

The subsequent analysis showed that the session’s X11DisplayNumber is not released until the X server has exited. The display number is only released when `sesexec` terminates, and `sesexec` only terminates after the X server terminates.

There are two possibilities here I think:
1) The design as outlined above has not been implemented correctly.
2) The X server is leaving the system in a state where it cannot be restarted. 

I'm not convinced we understand which of these possibilities is the one which is happening to you.

tsz8899:
Thanks @matt335672 

Thank you for your detailed and insightful feedback. I completely agree with your points.


   You are correct that the proposed patch is a platform-specific workaround aimed at immediate business availability rather than a proper architectural fix. I understand why it's not suitable for the upstream codebase.
   The main challenge we face is that this issue only manifests under extremely high concurrent load, occurring perhaps once every few weeks. In our environment,enabling DEBUG level logging significantly impacts performance, making it impossible to simulate the actual high-load conditions required to trigger the race condition. This is why we haven't beenable to provide a precise trace from sesexec.
   I am now shifting my focus to designing an automated stress-test script to reliably reproduce this "premature display release" state in a controlled lab environment. I want to verify if there's a specific edge case where sesexec perceives the X server as exited (perhaps due to a signal handling delay or a specific return code) while the process still lingers.

  I will close this PR for now and will return to the discussion once I have a reproducible test case or more concrete data from a successful lab reproduction.

matt335672:
Let me know how I can help. I've always found your contributions to xrdp to be most valuable.

A couple of things:
- Another possibility is that `sesexec` is crashing. This would leave an orphaned session. Because you're running in namespace mode, the active X server will not be visible to `sesman`.
- Local file logging in xrdp is slow, as the file is flushed after every line. You can achieve more logging performance by using syslog:
```
[Logging]
#LogFile=xrdp-sesman.log
#LogLevel=INFO
EnableSyslog=true
SyslogLevel=DEBUG
```

PR #3731, opened 2026-02-17
enabled and fixed cppcheck `portability` checks

firewave:


firewave:
```
sesman/libsesman/verify_user.c:341:20: style: Array index 'saltcnt' is used before limits check. [arrayIndexThenCheck]
        while ((pwd[saltcnt] != '$') && (saltcnt < 11))
                   ^
```

This is unexpected as that severity has not been enabled. I filed https://trac.cppcheck.net/ticket/14506 about it.

matt335672:
I'll take a look at these.

The code in `verify_user.c` is dreadful anyway and could do with a rewrite.

firewave:
I am not too familiar with the raised issues and have not looked them now. If you think they might be false positives - let's suppress them and file a ticket upstream.

matt335672:
I now understand this one - I've learned something new about C as a result:

```
portability: Passing NULL after the last typed argument to a variadic function leads to undefined behaviour. [varFuncNullUB]
```

From [C99 6.2.5](https://port70.net/%7Ensz/c/c99/n1256.html#6.2.5p27), there are no guarantees that all pointers have the same sizes. So for example. in this call:

```
nocase_matches(option, "--dump-config", NULL)
```

we need to replace `NULL` with `(char *)NULL` or `(char *)0`, so that the vararg parsing routine is guaranteed to see the correct size argument.

I'll look at the others tomorrow.

firewave:
> From [C99 6.2.5](https://port70.net/%7Ensz/c/c99/n1256.html#6.2.5p27), there are no guarantees that all pointers have the same sizes. So for example. in this call:
> 
> ```
> nocase_matches(option, "--dump-config", NULL)
> ```
> 
> we need to replace `NULL` with `(char *)NULL` or `(char *)0`, so that the vararg parsing routine is guaranteed to see the correct size argument.

I filed https://trac.cppcheck.net/ticket/14515 about improving the message as "undefined behavior" can mean anything.

matt335672:
@firewave - I've taken the liberty of implementing fixes for the portability warnings. Let me know what you think.

The dreadful code in `verify_user.c` has been replaced with something a little more robust. This codepath is in any case not currently used. If it was, a CVE might be in order.

firewave:
Rebased and LGTM.

PR #3732, opened 2026-02-17
enabled and fixed some cppcheck `style` checks

firewave:


firewave:
I think it makes sense to fix/enable these warnings incrementally and not address all of them to begin with in this PR. That's I am suppressing some IDs for now.

firewave:
Some of these should also be detected by Clang/GCC compiler warnings so it might make sense to enable those in unison when addressing the findings.

firewave:
This should be fine to merge as it sets a baseline and prevents further technical debt piling up. I wanted to get some additional low-hanging fruits in but some get messy as it clashes a bit with the coding convention and others require changes to the submodules. So let's keep it at that for now.

scripts/run_cppcheck.sh
```diff
@@ -84,9 +84,10 @@ fi
 # Supply default flags passed to cppcheck if necessary
 if [ -z "$CPPCHECK_FLAGS" ]; then
     CPPCHECK_FLAGS="--quiet --force --std=c11 --std=c++11 --inline-suppr \
-                    --enable=warning,missingInclude --error-exitcode=1 \
+                    --enable=warning,missingInclude,style --disable=portability --disable=performance --error-exitcode=1 \
```
firewave, line 87: `--enable=style` currently implies `warning`, `performance` and `portability` - see https://github.com/danmar/cppcheck/pull/4953.

PR #3730, opened 2026-02-14
Coverity: fix recent regressions

matt335672:
Coverity scan is back up, and has reported the following regressions on #3719:

```
** CID 899745:       Null pointer dereferences  (REVERSE_INULL)
/xrdp/xrdp_mm.c: 5534           in xrdp_mm_setup_mod2()


_____________________________________________________________________________________________
*** CID 899745:         Null pointer dereferences  (REVERSE_INULL)
/xrdp/xrdp_mm.c: 5534             in xrdp_mm_setup_mod2()
5528                 if (self->last_sync_saved)
5529                 {
5530                     int key_flags = self->last_sync_key_flags;
5531                     int device_flags = self->last_sync_device_flags;
5532                     self->last_sync_saved = 0;
5533                     struct xrdp_mod *m = self->mod;
>>>     CID 899745:         Null pointer dereferences  (REVERSE_INULL)
>>>     Null-checking "m" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
5534                     if (m != 0 && m->mod_event != 0)
5535                     {
5536                         m->mod_event(m, WM_KEYBRD_SYNC, key_flags,
5537                                      device_flags, key_flags, device_flags);
5538                     }
5539                 }

** CID 899744:       Resource leaks  (RESOURCE_LEAK)
/xrdp/xrdp_bitmap.c: 719           in xrdp_bitmap_invalidate()


_____________________________________________________________________________________________
*** CID 899744:         Resource leaks  (RESOURCE_LEAK)
/xrdp/xrdp_bitmap.c: 719             in xrdp_bitmap_invalidate()
713             {
714                 return 0;
715             }
716             struct xrdp_mm *mm = wm->mm;
717             if (mm == 0)
718             {
>>>     CID 899744:         Resource leaks  (RESOURCE_LEAK)
>>>     Variable "painter" going out of scope leaks the storage it points to.
719                 return 0;
720             }
721             struct xrdp_mod *m = mm->mod;
722             if (m != 0 && m->mod_event != 0)
723             {
724                 if (rect != 0)
```

I've put together this PR to address these reports.

The change to `xrdp_bitmap_invalidate()` simply makes sure the allocated painter is deallocated in the event of an error.

The change to `xrdp_mm_setup_mod2()` moves the null module check earlier, and allocates a local variable for readability.

@tsz8899 - please can you review?

PR #3734, opened 2026-02-18
coverity fix recent regressions 0x18

tsz8899:
@matt335672  thanks

Applied from current version bbd950d552809b812c3c65a5d970ff24fdb9d964:
- #3730 Coverity fix recommendations
- #3723 suggested hardening/refinements

Tested in 3 30+user environments; the 0x18 error was eliminated.
Please review.


tsz8899:
Updated based on the feedback @Firewave.
  In xrdp_mm_setup_mod2() the null check from if (self->mod == 0) to if (mod == 0) to consistently use the extracted local pointer.

matt335672:
I've just run this through Coverity, and it's all good. Thanks both.

PR #3739, opened 2026-02-23
fixed some `unreadVariable` Cppcheck warnings

firewave:


firewave:
```
neutrinordp/xrdp-neutrinordp.c:1584:28: style: Variable '((int*)(dst8))[0]' is assigned a value that is never used. [unreadVariable]
        ((int *)(dst8))[0] = pixel;
                           ^
```

This is a false positive and I filed https://trac.cppcheck.net/ticket/14535 about it.

firewave:
This does not fix all of the warnings but most of them. This only gets rid of unnecessary assignments as a start. This will also help with fixing some of other currently disabled warnings as it gets rid of some code which was triggering others as well.

Some of the remaining ones need to be suppressed (unmatched suppression warnings should be enabled before those will be introduced) and others will change the behavior.

matt335672:
I'll push this one through Coverity before I merge it

matt335672:
All good - thanks.

PR #3740, opened 2026-02-25
enabled and fixed `knownConditionTrueFalse` Cppcheck warnings

firewave:


firewave:
All my attempts in getting anywhere with any of the disabled ID has gotten messy so far - but this is a doozy. 

@matt335672 Better have a look as some of the required fixes (~~none included so far~~) might need some soak time before they hit any production.

firewave:
I will focus on the ones which do not change the behavior and back out all the others for now.

PR #3753, opened 2026-03-11
regression: Display number related issues

### LINKED ISSUE #3751: Regression: who/utmp only shows the last XRDP user after display-string migration (ut_id collisions)
### Summary

After recent display-string changes, `who` on the server only shows the most recently logged-in XRDP user, instead of all concurrent XRDP sessions.

This appears to be a regression in `sesman/sesexec/sessionrecord.c` utmp ID generation.

---

### Environment

- xrdp branch: `devel`
- Reproduced after updates including:
  - `656a125c` (`utmp: Remove display number from interface`)
  - display string changes merged via PR #3729
- Linux with utmp enabled (`who` output affected)

---

### Previous behavior (before update)

Multiple concurrent XRDP users were visible in `who`, e.g.:

- `xrdp:60`
- `xrdp:61`
- `xrdp:62`

---

### Current behavior (after update)

Only one XRDP session (typically the last one) remains visible in `who`, e.g.:

- `xrdp:X11-88`

Even with multiple users connected, earlier entries are overwritten.

---

### Steps to reproduce

1. Start xrdp/sesman with current `devel`.
2. Connect multiple concurrent XRDP users.
3. Run `who` on server.
4. Observe only one XRDP user line remains (or lines overwrite each other).

---

### Root-cause analysis

In `sesman/sesexec/sessionrecord.c`, `get_display_num()` derives a numeric ID from display string for `ut_id`.

Current logic extracts the first digit run in the display string.
For displays like `X11-88`, `X11-89`, `X11-90`, this yields `11` for all sessions.

So all sessions get the same `ut_id` (e.g. `:%x` => `:b`), and `pututxline()` overwrites previous entries.

---

### Impact

- In multi-user scenarios, server-side tools can no longer reliably obtain the full logged-in user list.
- Session-name dependent applications can break (example: IME/input-method integrations that rely on session identity).
- Incorrect operational visibility (`who`, utmp-based auditing/monitoring).

---

### Suggested fix

For display strings like `x11-<num>` / `wayland-<num>`, derive utmp ID from the numeric part after `-` to avoid collisions.

(Alternative acceptable approach: any stable collision-resistant mapping of display string to `ut_id` constrained by field size.)

matt335672:
Fixes #3751 

The move away from the X11 display number has introduced a couple of regressions
1) *XDG_SESSION_TYPE is not detected properly.*
   pam_systemd.so contains code to map a PAM_TTY of ':n' to an 'x11'  session type. This mapping is no longer done. We could re-introduce this code for X11, but there is no such code to detect a wayland display type. We try to fix this in a forward-looking way by setting XDG_SESSION_TYPE explicity before starting the PAM session.
2) *utmp is not being updated correctly.*
   The code for setting ut_id in the utmp[x] structure was setting the same value for all X11 displays, thus preventing utmp from being able to see more than one xrdp user

Also, an include is needed for sesman/eicp_process.c on some systems to get access to strlcpy()

@tsz8899 - As ever, if you have the time to look at this, I'd be extremely grateful.

### DIFF
--- a/sesman/eicp_process.c
+++ b/sesman/eicp_process.c
@@ -40,6 +40,7 @@
 #include "sesman.h"
 #include "sesman_access.h"
 #include "sesman_config.h"
+#include "string_calls.h"
 #include "guid.h"
 
 /******************************************************************************/
--- a/sesman/libsesman/verify_user_pam.c
+++ b/sesman/libsesman/verify_user_pam.c
@@ -400,9 +400,20 @@ auth_uds(const char *user, enum scp_login_status *errorcode)
 static int
 auth_start_session_private(struct auth_info *auth_info, const char *display)
 {
-    int error;
+    // For Linux and maybe other systems, pam_systemd needs to know the
+    // session type in order to set the session up correctly
+    const char *session_type;
+    if (g_get_x11_display_from_display_string(display) >= 0)
+    {
+        session_type = "x11";
+    }
+    else
+    {
+        session_type = "wayland";
+    }
+    g_setenv_log("XDG_SESSION_TYPE", session_type, 1);
 
-    error = pam_set_item(auth_info->ph, PAM_TTY, display);
+    int error = pam_set_item(auth_info->ph, PAM_TTY, display);
 
     if (error != PAM_SUCCESS)
     {
--- a/sesman/libsesman/verify_user_pam_userpass.c
+++ b/sesman/libsesman/verify_user_pam_userpass.c
@@ -209,9 +209,20 @@ auth_uds(const char *user, enum scp_login_status *errorcode)
 static int
 auth_start_session_private(struct auth_info *auth_info, const char *display)
 {
-    int error;
+    // For Linux and maybe other systems, pam_systemd needs to know the
+    // session type in order to set the session up correctly
+    const char *session_type;
+    if (g_get_x11_display_from_display_string(display) >= 0)
+    {
+        session_type = "x11";
+    }
+    else
+    {
+        session_type = "wayland";
+    }
+    g_setenv_log("XDG_SESSION_TYPE", session_type, 1);
 
-    error = pam_set_item(auth_info->ph, PAM_TTY, display);
+    int error = pam_set_item(auth_info->ph, PAM_TTY, display);
 
     if (error != PAM_SUCCESS)
     {
--- a/sesman/sesexec/sessionrecord.c
+++ b/sesman/sesexec/sessionrecord.c
@@ -71,9 +71,9 @@ typedef struct utmp _utmp;
 #include "string_calls.h"
 
 #define XRDP_LINE_FORMAT "xrdp:%s"
-// ut_id is a very small field on some platforms, so use the display
-// number in hex
-#define XRDP_ID_FORMAT ":%x"
+// ut_id is a very small field on some platforms, so use a discriminator
+// (e.g. ':' for x11) and a number in hex
+#define XRDP_ID_FORMAT "%c%x"
 
 /******************************************************************************/
 /**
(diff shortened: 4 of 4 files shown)


tsz8899:
thanks @matt335672 

The changes introduced in this PR have been 3 days tested in four X11 multi-user environment. The results confirm that the display number management issues reported in #3751 have been successfully resolved. The system now operates correctly and remains stable across multiple concurrent sessions.

After adjusting load_pa_modules.sh in pulseaudio-module-xrdp, the audio sessions are now associated with the correct session names, restoring the expected behavior.

Similar to:
~~~
load_pa_modules.sh 2026-03-13 04:54:20.005270244 +0800
+++load_pa_modules.sh.new       2026-03-13 05:03:31.750682369 +0800
@@ -2,18 +2,34 @@

 status=0

-if [ -n "$XRDP_SESSION" -a -n "$XRDP_SOCKET_PATH" ]; then
+if [ -n "$XRDP_SESSION" ]; then
+    # Resolve XRDP_SOCKET_PATH if not set (new versions use /var/run/xrdp/UID)
+    if [ -z "$XRDP_SOCKET_PATH" ]; then
+        USER_UID=$(id -u)
+        if [ -d "/var/run/xrdp/$USER_UID" ]; then
+            XRDP_SOCKET_PATH="/var/run/xrdp/$USER_UID/"
+        elif [ -d "/run/user/$USER_UID/xrdp" ]; then
+            XRDP_SOCKET_PATH="/run/user/$USER_UID/xrdp/"
+        else
+            XRDP_SOCKET_PATH="/tmp/.xrdp/"
+        fi
+    fi
+
     # These values are not present on xrdp versions before v0.9.8
     if [ -z "$XRDP_PULSE_SINK_SOCKET" -o \
          -z "$XRDP_PULSE_SOURCE_SOCKET" ]; then
         displaynum=${DISPLAY##*:}
         displaynum=${displaynum%.*}
-        XRDP_PULSE_SINK_SOCKET=xrdp_chansrv_audio_out_socket_$displaynum
-        XRDP_PULSE_SOURCE_SOCKET=xrdp_chansrv_audio_in_socket_$displaynum
-    fi

-    # Don't check for the presence of the sockets, as if the modules
-    # are loaded they won't be there
+        # Check for new X11- prefix style socket introduced in March 2026
+        if ls ${XRDP_SOCKET_PATH}xrdp_chansrv_audio_out_socket_X11-$displaynum >/dev/null 2>&1; then
+            XRDP_PULSE_SINK_SOCKET="xrdp_chansrv_audio_out_socket_X11-$displaynum"
+            XRDP_PULSE_SOURCE_SOCKET="xrdp_chansrv_audio_in_socket_X11-$displaynum"
+        else
+            XRDP_PULSE_SINK_SOCKET="xrdp_chansrv_audio_out_socket_$displaynum"
+            XRDP_PULSE_SOURCE_SOCKET="xrdp_chansrv_audio_in_socket_$displaynum"
+        fi
+    fi

     # Unload modules
     pactl unload-module module-xrdp-sink >/dev/null 2>&1
@@ -25,11 +41,7 @@
            xrdp_pulse_sink_socket=$XRDP_PULSE_SINK_SOCKET
     then
         echo "- pulseaudio xrdp-sink loaded"
-        if pacmd set-default-sink xrdp-sink; then
-            echo "- pulseaudio xrdp-sink set as default"
-        else
-            echo "? Can't set pulseaudio xrdp-sink as default"
-        fi
+        pactl set-default-sink xrdp-sink 2>/dev/null
     else
         echo "? Can't load pulseaudio xrdp-sink"
         status=1
@@ -40,11 +52,7 @@
            xrdp_pulse_source_socket=$XRDP_PULSE_SOURCE_SOCKET
     then
         echo "- pulseaudio xrdp-source loaded"
-        if pacmd set-default-source xrdp-source; then
-            echo "- pulseaudio xrdp-source set as default"
-        else
-            echo "? Can't set pulseaudio xrdp-source as default"
-        fi
+        pactl set-default-source xrdp-source 2>/dev/null
     else
         echo "? Can't load pulseaudio xrdp-source"
         status=1
~~~

matt335672:
Thanks @tsz8899. I'll merge this and look at both the audio modules.

matt335672:
@tsz8899 - the audio modules not working is a drop-off on my part. See #3757 which should make a change to `load_pa_modules.sh` unnecessary.

Issue #3749, opened 2026-03-05
RDP display channel triggers endless resize loop


akallabeth:
### xrdp version

0.10.4

### Detailed xrdp version, build options

Not sure if this is already knwon (did not find it in the issue tracker so far) but I had a report for fedora 41 (yes, a little outdated) that triggered an endless resize cycle due to `xrdp` doing some nasty stuff:

1. A display update is requested client side over the display channel
2. the remote answers it with both, `gfx` `ResetGraphics` as well as a `DesktopResize` order event. (bug 1)
3. if the procedure is repeated faster than `xrdp` can process it the `ResetGraphics` and `DesktopResize` messages get out of sync and report different resolutions triggering an endless loop of client resizes

did manage to trigger this with a default fedora 41 vm and a current FreeRDP client (3.23.0) but is fixed for later versions (just increased the delay between resize requests to >= 500ms)

see https://github.com/FreeRDP/FreeRDP/issues/12418 for the original report.

### Operating system & version

fedora 41

### Installation method

dnf / apt / zypper / pkg / etc

### Steps to reproduce

1. A display update is requested client side over the display channel
2. the remote answers it with both, `gfx` `ResetGraphics` as well as a `DesktopResize` order event. (bug 1)
3. if the procedure is repeated faster than `xrdp` can process it the `ResetGraphics` and `DesktopResize` messages get out of sync and report different resolutions triggering an endless loop of client resizes

### ✔️ Expected Behavior

resize once

### ❌ Actual Behavior

resize in endless loop

matt335672:
Thanks for raising this @akallabeth 

It's not known at the moment. I'll have a look when I get a mo'.

matt335672:
I've had a look into this.

The GFX support has been bolted on to the existing resize state machine in a way which means we always go through a deactivation-reactivation sequence when a resize occurs. This is not necessary for GFX, so I think we need a separate state machine for a GFX resize. I'll proceed along those lines.

akallabeth:
@matt335672 ok. second bug is that they get out of sync, maybe that is easier to fix as a first step?

matt335672:
@akallabeth - it's a good point, but I can't see how they do get out of sync, and I've had no luck reproducing it.

The code uses a resize queue. The queue is populated from resize requests from either the RDP client, or the X server. Requests are processed one at a time.

[...]

PR #3755, opened 2026-03-12
resizing: Simplify GFX resizing

matt335672:
See #3749 for the background on this.

For GFX, we currently resize the client with a deactivation-reactivation sequence. This is unnecessary, as the GFX RESET_GRAPHICS command does all the work for us, and avoids the need to teardown and re-establish the GFX channel. 

Draft for now - needs more testing.


matt335672:
@jsorg71, @Nexarian - I'd value your input on this.

Nexarian:
It's been a while since I've looked at this, but if I recall that while reset graphics was *supposed* to do the right thing, it didn't for ALL clients that supported resizing. I think this was especially a problem with the Mac OS client. Maybe they've since fixed it?

matt335672:
It wouldn't surprise me - there have been a lot of updates in the last two years looking at the App store history.

The logic is a bit simpler without deactivate-reactivate:
1) Stop encoder
2) Delete surfaces
3) Send reset graphics
4) Resize X server
5) Recreate surfaces
6) Start encoder

The GFX channel stays active the whole time, whereas with deactivate-reactivate we need to tear it down and restart it.

When this is no longer in draft, could you or Jay give it a spin from MacOS?

Nexarian:
> The GFX channel stays active the whole time, whereas with deactivate-reactivate we need to tear it down and restart it.

That is what the documentation said it was supposed to do, but in practice it never did :)

Would be blissful if it's fixed now! Happy to test it. I still use this feature regularly.

jsorg71:
I think we can simplify this.  We did a lot of fixes since we first brought up GFX.
Closing and reopening the channel seems unnecessary.
I think it was the Mac MS client that was the most picky.

jsorg71:
I grabbed this branch, rebased devel and running it now.  The MAC MS client seems to be working.  Let it run for a week.

jsorg71:
No problems over the last week.  I think this looks good.

matt335672:
Thanks @jsorg71. I'll merge this into `devel`, but keep it out of v0.10.x for now I think.

PR #3758, opened 2026-03-17
Disable smartcard code by default

matt335672:
The smartcard code is unfinished, and should not be employed in production due to the possibility that it may introduce security risks.

Rather than removing it, the code has been placed behind an `--enable-smartcard` switch. This allows developers to carry on working on the code, while removing any risk from production code.

The code continues to be build as part of CI, and any problems found by static analysis tools will be addressed.


metalefty (approved):
Let's go ahead.

Issue #2565, opened 2023-02-26
Resume existing session via MS RDP


KiLLeRRaT:
Hi,

Sorry that I have to create yet another "existing session" or "same session" issue.

I have tried to connect to existing desktop sessions from Windows to my Manjaro machine to no avail.

I have tried using clean installs of Manjaro i3, as well as Manjaro XFCE.

Is this at all possible somehow? I'm happy to start listing all my versions and configs, but first I just want to find out if this should be at all possible.  

I use my machine in such a way that I have all my applications and things open, then I leave the office to work remotely from a laptop (Windows, MacOS, OR Linux), and I want to be able to RDP using MSTSC from a Windows machine, and continue working on my machine at home with all my things still open.

I'm very hopeful, but I've tried a bunch of guides and such.  If someone can answer whether this is possible, then I can start providing versions and configs.

At a high level, it's a clean install, and then installing xrdp, and xorgxrdp, enabling and starting the services for both these, then doing things like removing `--exist-with-session` from the `~/.xinitrc` file, and a couple of other things I've read by following guides.  None of these things lets me use my existing desktop session.

Thanks for your time, I also appreciate all the effort that goes into this project.

Regards, 
Albert

matt335672:
Not only is it possible, but it should be default behaviour. Your use-case is one I used to use myself quite a few years back.

There are a couple of possibilities here. First of all, if you're at work and you disconnect from your session, can you reconnect to it immediately?

KiLLeRRaT:
Well it's great news that it's possible.  I can't for the life of me get it working though. What's the catch?

# Here are the steps I've taken

1. Clean install of Manjaro i3, with a non encrypted BTRFS setup.

2. Update system `pamac update`

3. Install xrdp via `pamac build xrdp`

4. Install xorgxrdp via `pamac build xorgxrdp`.  This fails with a PGP key problem `One or more PGP
  signatures could not be verified`, so I add the key using `sudo gpg --recv-keys
  61ECEABBF2BB40E3A35DF30A9F72CDBC01BF10EB` and without sudo: `gpg --recv-keys
  61ECEABBF2BB40E3A35DF30A9F72CDBC01BF10EB`.

5. This still didn't work, so then cloned the AUR package and ran makepkg:

  ```
  git clone https://aur.archlinux.org/xorgxrdp.git
  cd xorgxrdp
  makepkg -si
  ```

[...]

PR #3759, opened 2026-03-17
[V0.10] Disable smartcard code by default

matt335672:
The smartcard code is unfinished, and should not be employed in production due to the possibility that it may introduce security risks.

Rather than removing it, the code has been placed behind an --enable-smartcard switch. This allows developers to carry on working on the code, while removing any risk from production code.

The code continues to be build as part of CI, and any problems found by static analysis tools will be addressed.

Backport of #3758

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,11 +114,11 @@ jobs:
       CONF_FLAGS_amd64_max: "--enable-ipv6 --enable-jpeg --enable-fuse --enable-mp3lame
                   --enable-fdkaac --enable-opus --enable-rfxcodec --enable-painter
                   --enable-pixman --enable-utmp --with-imlib2 --with-freetype2 --enable-tests
-                  --enable-x264 --enable-openh264"
+                  --enable-x264 --enable-openh264 --enable-smartcard"
       CONF_FLAGS_i386_max: "--enable-ipv6 --enable-jpeg --enable-mp3lame
                   --enable-opus --enable-rfxcodec --enable-painter
                   --disable-pixman --with-imlib2 --with-freetype2
-                  --host=i686-linux --enable-tests"
+                  --host=i686-linux --enable-tests --enable-smartcard"
 
       PKG_CONFIG_PATH_i386: "/usr/lib/i386-linux-gnu/pkgconfig"
       CFLAGS_i386: "-m32"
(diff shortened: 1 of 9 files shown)


metalefty (approved):

matt335672:
Following note added to the NEWS v.10 file for the next release. I think that makes it clear that the live code shouldn't be shipped by distros by default.

---
## Changes for packagers or developers
. . .
-  The unfinished PIV smartcard support is now disabled by default, but can be re-enabled by adding `--enable-smartcard` to the configure command. Because of possible security issues with this code, this should only be done by developers working in non-production environments (#3759)

metalefty:
LGTM

rpiri1:
Hi folks,
I successfully debugged and modified the code and enabled user authorization via YubiKey/smartcard in a LDAPS environment, requiring PIN entry. Perhaps some of you would be interested in further testing and developing this code so that it can be incorporated into the official codebase. This is important to me because of security reasons, as I require smart card support in a production environment.

jsorg71:
@rpiri1 I'd be interest is your branch.

matt335672:
@rpiri1 - we've got some fundamental problems with the design of the existing smartcard code.

Have a read of #2565 which explains what the issues are.

rpiri1:
Hi matt335672,
I imagined the login via xrdp using a PIN so that both the password and the PIN can be entered in the same field as a hybrid structure

I only managed to get as far as entering the PIN at the console level. To also manage PIN entry via xrdp, structural changes would probably be necessary.

I am further testing.

metalefty:
@rpiri1 I think we should create a new issue for the discussion. It is no longer related to the PR itself.

rpiri1:
Hi metafty,
thanks for your hint.

rpiri1:
Hi guys,

my development has begun under 10.5. I spended much time and effort. I have noticed that the API has changed in 10.6.

matt335672:
@rpiri1 - we need to address the issues I've mentioned three weeks ago before we can do anything else with the smartcard code. Sadly, this is not a small piece of work.

I don't think working on the old codebase is a sensible thing to do, as we're using a private API for pcsclite. In the past we've had something working on Debian, which then breaks on latest Fedora. This makes a solution based on the old code undeliverable by us.

rpiri1:
After more than 6 weeks of hard and intensive work, countless hours, and tests in a productive environment, I've reached the point today where the existing smartcard code can't be sensibly fixed. There's a fundamental lack of consistent user-related systematics and process handling. On top of that, xrdp isn't really made for things like hybrid authentication, although I managed to get it working for my tests.

After I initially managed it with a user- and session-related environment file, I was too confident that I could replicate the whole thing in the existing code without an environment file. Unfortunately, that turned out to be impossible.

Additional, it seems that also modifications could be needed for the sssd code, when LDPS is included into the authenticitation method.

Issue #547, opened 2016-12-14
Disclaimer Banner before XRDP login screen


Criticize:
Is it possible to create a login/disclaimer banner with a message that the user has to be agree upon before the xrdp credentials prompt appears??

Thank you kindly for any assistance you can lend!

Natureshadow:
It would probably be best to just render `/etc/issue{,.net}` somewhere.

proski:
You can also replace the bitmap with a one containing the disclaimer text.

Criticize:
Currently I have the disclaimer placed in /etc/issue{,.net} and it works perfectly when logging into CentOS via VMware console. However when I use XRDP from a thin-client it skips the disclaimer completely and goes right to the desktop.

Is it possible to trigger a pop-up before the XRDP login screen? OR set XRDP up so that it connects to the normal blue gnome login screen (as if you physically accessing the computer)?

Kaydub00:
Yeah, I can't find a solution to this issue either, so here's what I'm doing. (which now that I just finished writing this, it's basically what @proski said)

I changed the login screen image and title. The image contains the message I need (in my case fedramp related message). The user has to enter their username and password and click okay, which is an implicit agreement to the terms in the image.

To make this work though, the user has to first fail the login or not pass login credentials... so yeah, still not a good solution but it's the only way I could get a message to display for XRDP on CentOS 7.2

kfiresmith:
+1 for a proper standard way to render /etc/issue w/o any BMP logo workarounds

taylorjtamu:
+1 for a way to display /etc/issue without image workarounds.

Nexarian:
I know of someone who would use XRDP if this feature were implemented. I know little about this part of the codebase, but adoption to XRDP is important, so this is now my next highest priority to get implemented... Working on CentOS

Nexarian:
One of the steps is to disable autologin even if the credentials are provided. I suspect we need to add a config for that: https://github.com/neutrinolabs/xrdp/blob/29ef7f896a831bb5e1053221e0837959599d4c34/libxrdp/xrdp_sec.c#L875

matt335672:
You can still do the autologin - you just need to display the message first in a dialog box. I think that's the hard bit.

matt335672:
@Nexarian - one thing I just remembered - there's code for a help dialog in the login screen module. The code is unused, but it might be useful to get you started maybe.

Nexarian:
[...]

PR #3760, opened 2026-03-17
interactive pam conv

jsorg71:
Add a prompt style dialog to the login screen.
When PAM_PROMPT_ECHO_ON or PAM_PROMPT_ECHO_OFF occur during the pam conversation, a question / answer message to sent from sesman to xrdp.  The question is displayed in the prompt dialog and a one line edit box is used for the answer.
When PAM_ERROR_MSG or PAM_TEXT_INFO occur during the pam conversation, text is sent from sesman to xrdp and displayed in the same dialog.

### DIFF
--- a/libipm/scp.c
+++ b/libipm/scp.c
@@ -922,3 +922,45 @@ scp_send_close_connection_request(struct trans *trans)
                (int)E_SCP_CLOSE_CONNECTION_REQUEST,
                NULL);
 }
+
+/*****************************************************************************/
+
+int
+scp_send_prompt_request(struct trans *trans, const char *prompt,
+                        enum scp_prompt_type prompt_type)
+{
+    return libipm_msg_out_simple_send(
+               trans,
+               (int)E_SCP_PROMPT_REQUEST,
+               "si", prompt, prompt_type);
+}
+
+/*****************************************************************************/
+
+int
+scp_get_prompt_request(struct trans *trans, char **prompt,
+                       enum scp_prompt_type *prompt_type)
+{
+    int rv = libipm_msg_in_parse(trans, "si", prompt, prompt_type);
+    return rv;
+}
+
+/*****************************************************************************/
+
+int
+scp_send_prompt_response(struct trans *trans, const char *response)
+{
+    return libipm_msg_out_simple_send(
+               trans,
+               (int)E_SCP_PROMPT_RESPONSE,
+               "s", response);
+}
+
+/*****************************************************************************/
+
+int
+scp_get_prompt_response(struct trans *trans, char **response)
+{
+    int rv = libipm_msg_in_parse(trans, "s", response);
+    return rv;
+}
--- a/sesman/libsesman/sesman_auth.h
+++ b/sesman/libsesman/sesman_auth.h
@@ -31,7 +31,9 @@
  * Opaque type used to represent an authentication handle
  */
 struct auth_info;
+
 #include "scp_application_types.h"
+#include "trans.h"
 
 /**
  *
@@ -45,7 +47,8 @@ struct auth_info;
  */
 struct auth_info *
 auth_userpass(const char *user, const char *pass,
-              const char *client_ip, enum scp_login_status *errorcode);
+              const char *client_ip, enum scp_login_status *errorcode,
+              struct trans *scp_trans);
 
 /**
  *
--- a/libipm/scp.h
+++ b/libipm/scp.h
@@ -65,8 +65,21 @@ enum scp_msg_code
     E_SCP_CREATE_SOCKDIR_REQUEST,
     E_SCP_CREATE_SOCKDIR_RESPONSE,
 
-    E_SCP_CLOSE_CONNECTION_REQUEST
-    // No E_SCP_CLOSE_CONNECTION_RESPONSE
+    E_SCP_CLOSE_CONNECTION_REQUEST,
+    // No E_SCP_CLOSE_CONNECTION_RESPONSE,
+
+    E_SCP_PROMPT_REQUEST,
+    E_SCP_PROMPT_RESPONSE,
+
+    E_SCP_END
+};
+
+enum scp_prompt_type
+{
+    SCP_PROMPT_ECHO_OFF = 1,
+    SCP_PROMPT_ECHO_ON,
+    SCP_PROMPT_ERROR_MSG,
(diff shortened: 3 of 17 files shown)


libipm/scp.c
```diff
@@ -185,7 +185,7 @@ scp_connect(const  char *port,
     struct trans *t;
 
     (void)scp_port_to_unix_domain_path(port, sock_path, sizeof(sock_path));
-    if ((t = trans_create(TRANS_MODE_UNIX, 128, 128)) != NULL)
+    if ((t = trans_create(TRANS_MODE_UNIX, 8192, 8192)) != NULL)
     {
```
matt335672, line 189: Shouldn't need this. The 128 is simply a placeholder. The transport is sized when `libipm_msg_out_init()` is called to `LIBIPM_MAX_MSG_SIZE` which is currently 8192.

libipm/scp.c
```diff
@@ -221,7 +221,7 @@ struct trans *
 scp_init_trans_from_fd(int fd, int trans_type, int (*term_func)(void))
 {
     struct trans *result;
-    if ((result = trans_create(TRANS_MODE_UNIX, 128, 128)) == NULL)
+    if ((result = trans_create(TRANS_MODE_UNIX, 8192, 8192)) == NULL)
```
matt335672: See previous comment.

libipm/scp.c
```diff
@@ -922,3 +922,53 @@ scp_send_close_connection_request(struct trans *trans)
...
+/*****************************************************************************/
+
+int
+scp_get_prompt_request(struct trans *trans, char *prompt, int prompt_max, int *flags)
+{
+    char *lprompt = NULL;
+    int rv = libipm_msg_in_parse(trans, "si", &lprompt, flags);
+    if (rv == 0)
+    {
+        g_snprintf(prompt, prompt_max, "%s", lprompt);
```
matt335672: The copy shouldn't be necessary. The string pointer returned from `libipm_msg_in_parse()` is valid until `scp_msg_in_reset()` is called for the next message. The string returned by `scp_get_prompt_request()` is processed by `xrdp_prompt_add_prompt()` which copies the string anyway.

libipm/scp.c
```diff
@@ -922,3 +922,53 @@ scp_send_close_connection_request(struct trans *trans)
...
+/*****************************************************************************/
+
+int
+scp_get_prompt_response(struct trans *trans, char *response, int response_max)
+{
+    char *lresponse = NULL;
+    int rv = libipm_msg_in_parse(trans, "s", &lresponse);
+    if (rv == 0)
+    {
+        g_snprintf(response, response_max, "%s", lresponse);
```
matt335672: Same comment about the copying applies here. The copied string here is immediately copied again using `g_strdup()` in `conv_echo_off()` so the intermediate buffer is not needed.

sesman/libsesman/sesman_auth.h
```diff
@@ -45,7 +47,8 @@ struct auth_info;
  */
 struct auth_info *
 auth_userpass(const char *user, const char *pass,
-              const char *client_ip, enum scp_login_status *errorcode);
+              const char *client_ip, enum scp_login_status *errorcode,
+              struct trans *scp_trans);
```
matt335672, line 51: The other authentication modules (verify_user_bsd.c, verify_user.c, verify_user_kerberos.c, verify_user_pam_userpass.c) will need updating too. We can do that later. Alternatively we could remove some of them in a separate PR - there seems to be no demand for  verify_user_kerberos.c and verify_user_pam_userpass.c. What do you think?

matt335672 (commented):
I've got a few minor comments at this stage. I haven't tried running it yet.

libipm/scp.c
jsorg71: ok, done, didn't see LIBIPM_MAX_MSG_SIZE.

libipm/scp.c
jsorg71: ok, extra copies removed

sesman/libsesman/sesman_auth.h
jsorg71: I added to them all.  I was thinking there may be a better way to pass the trans but it's kinda nice it's there so you know that the module can send messages.
I changed verify_user_pam_userpass and verify_user_kerberos for completeness but yea, we can remove them.

jsorg71:
Note on behavior, when you login with no password, you will get a prompt.  I don't know if we want it that way.
Another easy way to test this is to expire your password with sudo passwd -e <username> then try to reconnect or login.

matt335672:
That's looking good.

A couple of comments:

1) The flags arg for `scp_send_prompt_request()` needs some explanation.

2) When using text or errors, it's probably necessary to wait for a reply anyway. Otherwise the PAM stack continues to run, and the text or error disappears in short order.

I added the line `auth     optional  pam_echo.so This is the auth stage of PAM` to `/etc/pam.d/xrdp-sesman`. The following video was the result.

https://github.com/user-attachments/assets/94c63b43-ef01-4dac-99f7-77e2e0aa832f

We could maybe use the flags to specify when the edit box for a reply is not actually required.

This could also address #547. @Nexarian - you were interested in this one a while back(?)

matt335672:
I've also had some thoughts about implementing the SCP callback outside of `verify_user_pam.c` in a separate callback. This would allow us to use the SCP callback for the two remaining `verify_user_xxx.c` calls, and also to implement a separate text-based callback for use with the `xrdp-authtest` utility. I'd be happy to work on that bit.

jsorg71:
@matt335672 thanks, yea, scp_send_prompt_request can use some cleanup.  Flags is the wrong word there.  I can have every message stop and wait so we can see what that feels like.
When something like EntryID runs, you get one or more PAM_TEXT_INFO and then a PAM_PROMPT_ECHO_ON so having the user click after each PAM_TEXT_INFO might not feel smooth.  PAM_ERROR_MSG it makes more sense.  I'll try to see how difficult it to show / hide the edit.

Sure definitely, it's marked a work in progress so if you want to implement SCP fallback think that's good.
This is a tricky spot for security too, so if you have any suggestions let me know.

matt335672:
As a heads-up GHSA-3wr5-fwmh-qh34 in the security advisories may have a bearing on this. I'm still looking into it.

Issue #3391, opened 2025-01-10
Mixed IPv4/IPv6 code - address family mismatch


matt335672:
### xrdp version

All IPv6 

### Detailed xrdp version, build options

~~~
xrdp 0.10.80
  A Remote Desktop Protocol Server.
  Copyright (C) 2004-2025 Jay Sorg, Neutrino Labs, and all contributors.
  See https://github.com/neutrinolabs/xrdp for more information.

  Configure options:
      --enable-devel-all
      --enable-fuse
      --enable-pixman
      --enable-ipv6
      --enable-painter
      --enable-jpeg
      --with-imlib2
      --enable-vsock
      --with-freetype2
      --enable-ibus
      --enable-x264
      --enable-openh264
      --enable-utmp
      --with-libpcsclite
      --enable-neutrinordp
      CC=gcc
      CFLAGS=-g -fvar-tracking -Wl,-z,now -Wno-error=deprecated-declarations

  Compiled with OpenSSL 3.0.2 15 Mar 2022
~~~

### Operating system & version

All

### Installation method

Doesn't matter

### Which backend do you use?

All

### What desktop environment do you use?

any

### Environment xrdp running on

All

### Area(s) with issue?

Network

### Steps to reproduce

No steps as such - see below.

### ✔️ Expected Behavior

See below

### ❌ Actual Behavior

See below

### Anything else?

While testing #3390, I found some interesting messages in `xrdp-sesman.log`:-

## Linux
```
[2025-01-10T14:39:05.122+0000] [ERROR] [pid:401338 tid:125582510497920] [g_tcp_bind(os_calls.c:976)] g_tcp_bind(15, 6010) failed bind IPv6 (errno=98) and IPv4 (errno=22).
```

98 is `EADDRINUSE` and 22 is `EINVAL`

## FreeBSD
```
[2025-01-10T14:46:48.309+0000] [ERROR] [g_tcp_bind(os_calls.c:978)] g_tcp_bind(15, 6010) failed bind IP
v6 (errno=48) and IPv4 (errno=47).
```

48 is `EADDRINUSE`, and 47 is `EAFNOSUPPORT`.

In both cases, the IPv6 error was expected, and the IPv4 error wasn't.

The problem is caused by this code:-

```c
int
g_tcp_bind(int sck, const char *port)
{
    struct sockaddr_in6 sa;
    struct sockaddr_in s;
    int errno6;

    // First IPv6
[...]

PR #3766, opened 2026-03-19
g_tcp_connect address IPv4/IPv6 Mixing

Kropyls:
This addresses https://github.com/neutrinolabs/xrdp/issues/3391

The changes are as follows:
- add static addr_is_ipv4 function
- inside g_tcp_connect:
  - use getsockname to find family of provided sck
  - add switch case for sck family type to address ipv4 sockets failing
    - switch case addresses differences between ipv4/6 before handing control back
  - add comment about OSX to IPv6 function for clarity like exists in ipv4 version

This addresses the issue without a very significant rework of the existing code. Since this problem won't show up too often going forward as most sockets are local, it seems valuable to me to clear the issue without a more fundamental rework, as the initial issue talks about, and seems to not cause any issues. For anyone using remote vnc servers to connect on this will address the ip address mixing issue. I've tested on a remote vnc server and it can connect on both v4 and v6.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -849,6 +849,22 @@ connect_loopback(int sck, const char *port)
 #endif
 
 /*****************************************************************************/
+/* returns boolean, 1 if the address is IPv4, else 0 */
+static int
+addr_is_ipv4(const char *address)
+{
+    struct in_addr addr;
+
+    if (address == NULL || *address == '\0')
+    {
+        return 0;
+    }
+
+    /* inet_pton returns 1 on success */
+    return inet_pton(AF_INET, address, &addr) == 1;
+}
+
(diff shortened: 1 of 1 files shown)


Kropyls:
This is my first PR to an OSS project. Please let me know if anything is incorrect or needs to change in regards to it, and thanks for reading!

metalefty:
Why closed?

Kropyls:
There was some code that should have been cleaned up in the commit. I realized I had to clean that up and squash it and I didn't want anyone running with this one in the meantime. I reponed it under https://github.com/neutrinolabs/xrdp/pull/3767

In retrospect there is probably a way to re-attach the corrected code to this one but by the point I'd thought of that I'd already closed this.

PR #3767, opened 2026-03-19
g_tcp_connect ipv4/6 mixing fix

### LINKED ISSUE #3391: Mixed IPv4/IPv6 code - address family mismatch

Kropyls:
Addresses https://github.com/neutrinolabs/xrdp/issues/3391

- add static addr_is_ipv4 function
- use getsockname to find family of sck
- add switch case for sck family type to address ipv4 sockets failing
  - getaddrinfo and the connect logic left untouched
- add comment about OSX to IPv6 function for clarity

This addresses the issue without a very significant rework of the existing code. This likely won't show up very often due to local sockets being used in most cases, but some users may still need remote vnc connections. For anyone using remote vnc servers to connect on this will address the ip address mixing issue. I've tested on a remote vnc server and it can connect on both v4 and v6.

### DIFF
--- a/common/os_calls.c
+++ b/common/os_calls.c
@@ -791,60 +791,82 @@ g_sck_close(int sck)
 
 #if defined(XRDP_ENABLE_IPV6)
 /*****************************************************************************/
-/* Helper function for g_tcp_connect.                                        */
+/**
+ * Helper for g_tcp_connect
+ * Returns boolean, 1 if the address is IPv4, else 0
+ */
 static int
-connect_loopback(int sck, const char *port)
+addr_is_ipv4(const char *address)
 {
-    struct sockaddr_in6 sa;
-    struct sockaddr_in s;
-    int res;
+    struct in_addr addr;
 
-    // First IPv6
-    g_memset(&sa, 0, sizeof(sa));
-    sa.sin6_family = AF_INET6;
-    sa.sin6_addr = in6addr_loopback;             // IPv6 ::1
-    sa.sin6_port = htons((tui16)atoi(port));
-    res = connect(sck, (struct sockaddr *)&sa, sizeof(sa));
-    if (res == -1 && errno == EINPROGRESS)
-    {
-        return -1;
-    }
-    if (res == 0 || (res == -1 && errno == EISCONN))
+    if (address == NULL || *address == '\0')
     {
         return 0;
     }
 
-    // else IPv4
-    g_memset(&s, 0, sizeof(s));
-    s.sin_family = AF_INET;
-    s.sin_addr.s_addr = htonl(INADDR_LOOPBACK);  // IPv4 127.0.0.1
-    s.sin_port = htons((tui16)atoi(port));
-    res = connect(sck, (struct sockaddr *)&s, sizeof(s));
-    if (res == -1 && errno == EINPROGRESS)
-    {
-        return -1;
-    }
-    if (res == 0 || (res == -1 && errno == EISCONN))
+    /* inet_pton returns 1 on success */
+    return inet_pton(AF_INET, address, &addr) == 1;
+}
+
+/*****************************************************************************/
+/**
+ * Helper for g_tcp_connect
+ * Returns boolean, 1 if the address is IPv6, else 0
+*/
+static int
+addr_is_ipv6(const char *address)
+{
+    struct in6_addr addr;
+
+    if (address == NULL || *address == '\0')
     {
         return 0;
     }
 
-    // else IPv6 with IPv4 address
-    g_memset(&sa, 0, sizeof(sa));
-    sa.sin6_family = AF_INET6;
-    inet_pton(AF_INET6, "::FFFF:127.0.0.1", &sa.sin6_addr);
-    sa.sin6_port = htons((tui16)atoi(port));
-    res = connect(sck, (struct sockaddr *)&sa, sizeof(sa));
-    if (res == -1 && errno == EINPROGRESS)
-    {
-        return -1;
-    }
-    if (res == 0 || (res == -1 && errno == EISCONN))
-    {
-        return 0;
-    }
+    /* inet_pton returns 1 on success */
+    return inet_pton(AF_INET6, address, &addr) == 1;
+}
 
-    return -1;
+/*****************************************************************************/
+/**
+ * Another helper for g_tcp_connect
+ * Get the family/domain for a socket
+ *
+ * This is a little awkward, as POSIX does not yet offer a
+ * standardized way to do this
+ */
+static int
+get_socket_family(int sck, int *family)
+{
+    int error;
+#ifdef SO_DOMAIN
+    socklen_t len = sizeof(*family);
+    // SO_DOMAIN is widely used but not defined by POSIX
+    error = getsockopt(sck, SOL_SOCKET, SO_DOMAIN, family, &len);
+    if (error != 0)
+    {
+        LOG(LOG_LEVEL_WARNING, "getsockopt() failed on socket %d: %s",
+            sck, g_get_strerror());
+    }
+#else
+    union sock_info sock_info;
+    socklen_t sock_len = sizeof(sock_info);
+
+    /* Use getsockname() as a fallback. This is not guaranteed to work
+    on an unbound socket, according to POSIX */
+    error = getsockname(sck, &sock_info.sa, &sock_len);
+    if (error != 0)
+    {
+        LOG(LOG_LEVEL_WARNING, "getsockname() failed on socket %d: %s",
+            sck, g_get_strerror());
+    }
+    else
+    {
+        *family = sock_info.sa.sa_family;
+    }
+#endif
+    return error;
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


Kropyls:
This is my first PR to both open source and this project, so please let me know if anything is not in order or needs to change, and I'll happily fix it. Thanks!

matt335672:
Hi @Kropyls - thanks for raising this, and welcome to the world of open source!

I'm out of time today (UK), but I'll try to give this some significant attention tomorrow - speak then.

Kropyls:
Thank @matt335672 I got in one extra commit to hopefully resolve the failures with minimal options so hopefully that helps in the meantime!

matt335672:
Finally got round to looking at this @Kropyls.

Thanks again for taking the time to put this together, and to submit it.

Firstly, here's a test program `test.c` for `g_tcp_connect()` which you may find useful (I did):

```c
#include <stdio.h>

#include "config_ac.h"
#include "arch.h"
#include "os_calls.h"
#include "log.h"

int main(int argc, char *argv[])
{
    int rv = 1;
    if (argc < 3)
    {
        fprintf(stderr, "Args: host port\n");
    }
    else
    {
        const char *host = argv[1];
        const char *port = argv[2];
        int port_num = atoi(port);
        if (port_num <= 0 || port_num > 65535)
        {
            fprintf(stderr, "Port out of range\n");
        }
        else
        {
            struct log_config *logging;

            logging = log_config_init_for_console(LOG_LEVEL_DEBUG, NULL);
            log_start_from_param(logging);
            log_config_free(logging);

            int sck = g_tcp_socket();
            if (sck < 0)
            {
                perror("g_tcp_socket()");
            }
            else
            {
                rv = g_tcp_connect(sck, host, port);
                printf("tcp_connect(, %s, %s) -> %d\n", host, port, rv);
                if (rv == 0)
                {
                    g_file_close(sck);
                }
            }
            log_end();
        }
    }

    return rv;
}
```

If you copy it to the top level of the source tree you can build it with something like this on Linux:

```
cc -g -o test -Icommon  test.c -L common/.libs -Wl,-rpath,$(pwd)/common/.libs -lcommon
```

I've got two fairly significant comments on the PR:
## Issue 1 : IPv4 fallback
The `g_tcp_connect()` routine should be able to handle hostnames as well as addresses. I've got a mixed IPv4/v6 network here, and I can use the test program above to connect to both IPv4 and IPv6 hosts by name. For example, I can connect to TCP port 80 on my printer (IPv4) like this using the test program:

```
 ./test brother2.lan 80
tcp_connect(, brother2.lan, 80) -> 0
```

However, if I reboot my test machine, and add `ipv6.disable=1` to the kernel arguments in GRUB (to disable IPv6), the same test no longer works:

```
$ ./test brother2.lan 80
[2026-03-20T11:40:12.204+0000] [INFO ] [g_tcp_socket(os_calls.c:423)] IPv6 not supported, falling back to IPv4
[2026-03-20T11:40:12.204+0000] [ERROR] [g_tcp_connect(os_calls.c:937)] Address brother2.lan is not a valid IPv4 address
tcp_connect(, brother2.lan, 80) -> -1
```

## Issue 2: Code portability
Using `getsockname()` to get the address family of the socket is a nice trick, but isn't guaranteed to work on all platforms. [According to POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getsockname.html):

> If the socket has not been bound to a local name, the value stored in the object pointed to by address is unspecified.

Some platforms, including Linux and FreeBSD offer `getsockopt()` with `SO_DOMAIN`. This is guaranteed to work if available, but isn't part of the POSIX standard! So there seems to be no portable way to get the address family of a socket in POSIX. If there is one, I can't find it.

The best I could come up with is this:

```c
/**
 * Get the family/domain for a socket
 *
 * This is a little awkward, as POSIX does not yet offer a
 * standardized way to do this
 */
static int
get_socket_family(int sck, int *family)
{
    int error;
#ifdef SO_DOMAIN
    socklen_t len = sizeof(*family);
    // SO_DOMAIN is widely used but not defined by POSIX
    error = getsockopt(sck, SOL_SOCKET, SO_DOMAIN, family, &len);
    if (error != 0)
    {
        LOG(LOG_LEVEL_WARNING, "getsockopt() failed on socket %d: %s",
            sck, g_get_strerror());
    }
#else
    union sock_info sock_info;
    socklen_t sock_len = sizeof(sock_info);

    // Use getsockname() as a fallback. This is not guaranteed to work
    // on an unbound socket, according to POSIX
    error = getsockname(sck, &sock_info.sa, &sock_len);
    if (error != 0)
    {
        LOG(LOG_LEVEL_WARNING, "getsockname() failed on socket %d: %s",
            sck, g_get_strerror());
    }
    else
    {
        *family = sock_info.sa.sa_family;
    }
#endif
    return error;
}
```

Have a think and let me know how you'd like to proceed.

Kropyls:
thank for the feedback @matt335672 . I'll be using that program over the weekend to see what I can some up with, I had something similar but yours seems to be more flexible!

I didn't realize getsockname had that issue! getsockopt with so_domain seems preferable, and works on linux and bsd, so in terms of support osx is the real tricky one here. Do I have that right? If so I'll take a look on my old macbook and see if I can come up with something. Are there any other systems xrdp needs to support that I'd be missing with that approach?

matt335672:
I think getsockname() works OK on OSX for this:

https://stackoverflow.com/questions/13331610

Kropyls:
It seems to based on that post. After my own research on issue 2 I agree that your code is the best approach by far. Still crazy to me POSIX doesn't offer a way to do this as a general-case though. I'd have figured long ago in the advent of ipv6 that would've been implemented. But I digress...

Issue number 1 will be my focus now, I'll update when I have something!

Kropyls:
Hi again @matt335672 - I've done some thinking and reworking and I think we're at least getting a lot closer with this commit if we aren't all the way there.

The latest commit message goes into things a bit deeper, but big picture items:
- Covered the loopback logic directly in the switch-case, and got rid of it in favor of the addr_is_* helpers
- added your get_socket_family function directly as is
- in g_tcp_connect I added some prep work to ensure that if the address is a direct IP address we ensure it works against the socket family, and otherwise we set flags for getaddrinfo to ensure we have good options to attempt connections on. I also threw in some logging I found useful, but that can go if you don't agree
- line 942: seemed to be in slight conflict with the bsd man pages on what getaddrinfo could return, based on this: https://man.freebsd.org/cgi/man.cgi?query=getaddrinfo&sektion=3&manpath=FreeBSD+8.0-RELEASE it seemed like it only promises 0 on success, and a positive error could have caused some issues. I threw some additional source code in the commit for reference. Feel free to discard that part if you know something I don't about why it's that way though

I was able to test this with the ipv6.disable flag set and unset and it seems to function correctly, even when dns returns ipv4 and ipv6 addresses under the same hostname, I was able to establish a vnc session on a remote box.

One thing that threw me a bit was g_tcp_connect in the test binary you provided doesn't set nonblocking on it, so it worked through many address if a hostname had several. If some ips were offline, it would just try the next one. I sort of liked because it had such good redundancy but that but the behavior didn't carry over inside the larger configuration coming from trans_connect. I think I'm getting out of scope of this issue, but it could be a fun bit of follow-on work to see if I couldn't find a way to have that redundancy work asynchronously in the future...

Please let me know what you think of the changes, and if this addresses your concerns correctly or if more work is needed? Thanks,
-Matt

Kropyls:
sorry I missed the styling on that last one - it should be corrected now :(

matt335672:
This looks a lot better to me - thanks. I'm happy to merge this as it's an improvement. However, there's still a large problem with this architecture in general, which isn't your fault at all - it's just the way the code has evolved. I'd appreciate your thoughts on this.

The only consumer of `g_tcp_connect()` is `trans_connect()` in `common/trans.c`. The pseudo code for this routine is (roughly) as follows:-

While we don't have a connection, and the timeout hasn't happened.
- Create a socket
- Set the socket non-blocking
- Call `g_tcp_connect()`. Return if successful. If a connect is 'in progress' wait for completion and return
- Wait for a bit

The reason for the async stuff is so we can specify a timeout for `trans_connect()`. This design dates from a simpler time, where a name normally mapped to at most one IP address.

Within `g_tcp_connect()`, there's now ANOTHER connect loop with `getaddrinfo()` which works differently in that it re-uses the passed-in socket. All very well and good, apart from two things:
1) The additional hidden complexity
2) This statement from the [POSIX connect specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/connect.html)
> If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect.

Now you've worked on the code, you'll be aware we don't conform to this, at least within `g_tcp_connect()`.

This takes me back to something we touched on earlier; I think the reason POSIX doesn't offer a standardised way to get the address family for a socket is that you shouldn't need to do this.

I think the right thing to do at this point is merge your PR and raise an issue with the technical information above. I may be over-thinking this however, so let me know what you think.

PS: Before we merge you'll need to look into doing a git thing called a 'squash'.

Kropyls:
Hey again @matt335672  - thanks for the feedback, I'm glad this one is looking better to you! I don't think you're overthinking the wider architecture at all, that all seems very valid to me. I have an idea for how to correct that bigger picture issue too as you've described it, which I'm happy to get into or work on either here or in a separate issue based on your preference?

I went ahead and crunched it all down into one commit in the meantime. If you choose to go with the separate issue I'll continue the discussion there to try and keep things simple. Otherwise I can get into it in more detail here, but the sky-high POV is that I think we can offload a lot of the work to g_tcp_connect, which means it would "know" a lot more about the socket from the start and wouldn't have to do as much investigating before it starts working.

Thanks,
-Matt

matt335672:
Discussions can continue in #3772

PR #3768, opened 2026-03-23
sesman: Remove unused authentication methods

matt335672:
See [this comment](https://github.com/neutrinolabs/xrdp/pull/3760#discussion_r2953985424) for the background on this.

The Kerberos module and pam_userpass modules are now unused:

1) On all supported systems, PAM provides a far superior way to integrate Kerberos support.
2) The pam_userpass module (https://github.com/openwall/pam_userpass) (which is a lovely idea) is no longer maintained.


metalefty:
I'm fine with removing that.

jsorg71:
Looks good to me too.

matt335672:
I'll merge this then, and if Jay rebases #3760 it should give us a bit less work.

Issue #3769, opened 2026-03-23
FreeRDP proposed AV1 extension


akallabeth:
Hi!

Did create a pull request proposing a possible extension to the RDP protocol for use in open source RDP impmentations:
https://github.com/FreeRDP/FreeRDP/pull/12527

This would add `AV1` codec support, maybe you're interested in that as well.

metalefty:
Wow, that's interesting. We'd thought about that, but there were no clients that supported AV1 at the time.

akallabeth:
@metalefty that is why I thought it would be a good idea to start it 😁 
is an early draft, but with some feedback (and adoption to clients/servers) might grow into something.

metalefty:
When we thought about it before, we had a sponsor, but unfortunately the sponsorship was canceled. The priority is no longer high. What we will focus on next is Wayland.

However I'm still interested in a high-efficiency codec than H.264!

jsorg71:
I have some thoughts on this.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/da5c75f9-cd99-450c-98c4-014a496942b0
Just make up a new codecid that MS is not likely to use in 2.2.2.1.  Something like 0x003B.
Don't include anything like avc420MetaData in 2.2.4.4.
Everything else should be the same.

akallabeth:
@jsorg71 most of that is already done in the referenced pr.
not including anything like `avc420MetaData` is a bit of a nogo though, as the client then always needs to update the full surface/screen after each new decompressed frame. (or did I miss something with the AOM (or other AV1) decoder API that lets me query changed areas?
@jsorg71 btw, better continue the discussion over at the example pr, did also open issues for gnome-remote-desktop and krdp to get all on board 😁

jsorg71:
Oh sorry, yes regionRects is needed, it's quantQualityVals that I don't think makes since.
I think all newer codecs since H264 have color info in the SPS PPS type headers so I don't think you need many codec ids.
AV1 should have real YUV444 support so you don't need that ugly 444 method in GFX.  MS even hate that design.

akallabeth:
@jsorg71 I've merged the proposal for the time being (marked experimental, off by default) so we have something that we can test and have a little feedback from interested parties.
you're right about the `quantQualityVals`, but was too lazy to create a separate struct, but should add it to the list.

the `YUV444` is also supported by `H264` but hardware does not always support that :/

PR #3774, opened 2026-03-26
gfx/xup: add FFmpeg/dma-buf based VAAPI/Vulkan encoding support

### LINKED ISSUE #3119: Add VA-API adaption to enable video hardware acceleration in WSL2
[Here](https://devblogs.microsoft.com/commandline/d3d12-gpu-video-acceleration-in-the-windows-subsystem-for-linux-now-available/) the WSL2's GPU-PV has implemented VA-API API. I've noticed that our recent dev branch has implemented software x264 and nvenc hardware encodings. I hope libva can be added to increase x264 frame encoding speed. @jsorg71

FlyGoat:
This series adds an FFmpeg-based H.264 backend to xrdp and extends the
xup path so A2 updates from xorgxrdp can be encoded from imported dma-buf
surfaces instead of always going through the shmfd path.

The main goal is FFmpeg/dma-buf based VAAPI/Vulkan encoding support, but
the backend also has a software path for fallback and non-hardware use.

Highlights:

- add an optional FFmpeg H.264 backend (`--enable-ffmpeg`)
- add `gfx.toml` / `xrdp_tconfig` support for `path = software|vaapi|vulkan`,
  plus FFmpeg-specific encoder settings
- add parser coverage for the new FFmpeg config
- add `capture_transport_flags` to the xrdp/xorgxrdp client-info exchange
- add `XUP_ORDER_PAINT_RECT_DMABUF` and `XUP_ORDER_EGFX_DMABUF`
- teach `xup.c` to receive a dma-buf fd plus surface metadata
  (surface id, width, height, stride, fourcc, size)
- plumb that through `xrdp_mm` and the encoder thread as
  `struct xrdp_encoder_dmabuf_surface`
- add `xrdp_encoder_h264_encode_dmabuf()` support so the FFmpeg backend
  can encode directly from imported dma-buf surfaces

The FFmpeg backend works in three modes:

- `software`
  - encodes with `libx264`
  - shm input is encoded from NV12 system memory
  - dma-buf input is imported as DRM_PRIME, copied back to a software
    frame with `av_hwframe_transfer_data()`, then encoded with `libx264`

- `vaapi`
  - encodes with `h264_vaapi`
  - shm input uses `hwupload` + `scale_vaapi`
  - dma-buf input uses DRM_PRIME import plus
    `hwmap=derive_device=vaapi` + `scale_vaapi`

- `vulkan`
  - encodes with `h264_vulkan`
  - shm input uses `hwupload` + `scale_vulkan`
  - dma-buf input uses DRM_PRIME import plus
    `hwmap=derive_device=vulkan` + `scale_vulkan`

If a hardware path cannot be configured, the FFmpeg backend falls back to
the software path.

The first commit is a small cleanup which replaces remaining hard-coded
xup protocol ids with shared symbolic definitions in
`common/xup_client_info.h`. The second commit adds the FFmpeg backend and
the dma-buf transport support.

This is intended to pair with the matching xorgxrdp series that emits
the dma-buf orders and capture transport flag. With the matching
xorgxrdp changes, this is working for me on Intel xe.

Closes #3119


FlyGoat:
The pull request text is mostly Copilot-generated, so I hope it is not too bad :-)

This work is still in an early stage. I have not carried out extensive testing yet; I am posting it mainly to get feedback from the maintainers.

FlyGoat:
I'd like to add that it should be trivial to bring up AV1 (#3769) after this change.

jsorg71:
Hi @FlyGoat , interesting PR here.
Have you looked at the xrdp_accel_assist module in xrdp?  We are trying to keep all GLX, EGL, OpenGL, Cuda, OpenCL, Vulcan stuff in there because of the various limitations with each API.  It also provided a unified design for NVidia vs Glamor,
I didn't know Vulcan had an interface to encode, that is interesting too.  I guess that just hooks up the nvenc on NVidia?
I was wondering, what hardware are you testing with?

FlyGoat:
Hi @jsorg71,

Thanks for your insight! Yes I looked into  xrdp_accel_assist design but I really hate the pixmap handshake design which is a little bit fragile. Also it would be hard to enable other codecs like AV1 without three way communication between xrdp & xorg & xrdp_accel_assist.

Yes, nvidia have enabled vuklan video encoding in their 580 driver. I was actually thinking about get glamor work for nvidia (given that gbm support is now available) so we can get rid of the whole driver overlay thing.

This pr is tested on my AMD AI Max+ 395 iGPU as well as Intel Arc B570 eGPU. However I haven’t validated shm fallback path.

xrdp/xrdp_encoder_ffmpeg.c
```diff
@@ -0,0 +1,2138 @@
...
+static int
+ffmpeg_encoder_configure_hardware_codec(
+    struct ffmpeg_encoder *fe,
+    const struct xrdp_tconfig_gfx_ffmpeg_param *param,
+    const char *profile_name,
+    const char *entropy_coder)
+{
+    const char *vulkan_tune;
+
+    fe->codec_ctx->pix_fmt = ffmpeg_path_sink_format(fe->active_path);
```
hardening, line 1345: Would probably be a good idea to alias `fe->codec_ctx` and use that in the rest of the function.

jsorg71:
This PR is very similar to the way I first did hardware accel.
Hard to believe that was nearly 10 years ago.
It passed dmabufs and used libyami to encode.  It only worked with Intel as that was the only target.

Then the search for an NVidia solution.
NVidia is 90%+ of what people want to use xrdp with hardware acceleration.
Users wanted this for CUDA, self driving cars, AI, graphics development, etc.
Glamor for NVidia works with mesa drivers but no CUDA or NVidia's OpenGL / Vulcan
Glamor with NVidia's driver is not working and might not even be possible.
So I wanted a solution that would work with BSD, Intel, NVidia, AMD, and be flexible for whatever comes along.
That means dmabuf, although cool and works nice, could not be used.

The post capture processing issue.
MS uses a non standard color conversion matrix, standard color conversions may not match.
There is also the YUV444 algorithm in GFX that for sure needs custom GPU code.
OpenGL being the most common but OpenGL, Vulcan, Cuda, or OpenCL should be used for this step
to avoid transferring the image into CPU memory and back to video memory.

The xrdp encoder issue.
The nature of the main xrdp process.
Xrdp runs a different privilege than the user, can run in fork or thread me.
Many installs run xrdp as root today, not ideal for hardware access control.
Xrdp is not an ideal place to run OpenGL for the custom color conversion.
OpenGL needs a global context so it's limited to one per process.
So you can't do custom color conversion in Xorg or xrdp, another process is needed.

The ffmpeg issue.
I once tried to support ffmpeg in a project.  It was a nightmare.
The code had so many ifdefs because the API was always changing.
I don't think we every did get it working on all the platforms we wanted.
I've been soured by the experience and don't want to ever support ffmpeg again.
ffmpeg is really just a collection of other codecs so you can just use that lib directly.
ffmpeg -> x264
ffmpeg -> libva
ffmpeg -> h264_vulcan
So why not use x264 directly(wihich we already do)
why not use libva directly(which we've been working on)
as for h264_vulcan, I suspect that it's just a portal to nvenc on NVidia and libva on mesa.
so I suspect h264_vulcan won't be too useful.

I hope I've explained why the hardware acceleration is the way it is.
I wish we didn't have all these restriction but we do and I can't blame all of this on NVidia either.
There is a valid argument for separate process with Glamor too.

FlyGoat:
@jsorg71 Many thanks for your detailed explanation! I can feel your pain on dealing with graphics from those texts :-)

I think this is pretty valid, so I'll give up my work on XRDP side a and look into `xrdp_accel_assist` again. The good thing is nowadays most RGB2YUV programs (including libyuv / FFmpeg) do support custom LUT so it's less cumbersome.

I'll give up FFMpeg given that it's indeed a source of headache, but I'd like to keep Vulkan and DMA-BUF (Yes! NVIDIA do have proper DMA-BUF support nowadays) around. In our setup we are using Zink (OpenGL->Vulkan translation layer) and glamor for NVIDIA drivers which means raw GL interface and nvenc interwork is not an option but Vulkan acceleration is still quite possible. MESA is doing Vulkan video encoding natively (i.e. not routing over libva) so we can potentially enjoy lower overhead :-)

@jsorg71 Is it something acceptable for you to have DMA-BUF based Vulkan encoder in `xrdp_accel_assist`? For my own niche use this PR is good enough, but I'd like to bring the capability to everyone in long term :-)

jsorg71:
@FlyGoat Yes sure, the Vulcan h264 sounds useful.
There is some bits from this PR that seem useful too, like the defines.

lcniel:
I would just second that Vulkan-based encoding does sound very useful since Vulkan is in my experience typically very easy to get to work.

PR #3775, opened 2026-03-27
Add ScreenSleepTimeLimit for xorgxrdp sessions

tsz8899:
## Summary

This change adds a new sesman setting:

`ScreenSleepTimeLimit`

The setting allows xrdp to pass an idle screen-sleep timeout to xorgxrdp sessions.

In long-lived remote desktop sessions, the server can continue spending CPU time and network bandwidth on screen updates even when the user is idle.

This feature can save a significant amount of server CPU usage and network traffic in long-connection scenarios.

The effect is especially noticeable in these cases:

With H.264 encoding, the CPU savings are more obvious.
With RFX encoding, the network traffic savings are more obvious.
The benefit is even greater in multi-user deployments.

No change unless the feature is enabled by xrdp.
While idle, display updates are paused.
Keyboard or mouse activity wakes the session and forces a refresh.

In practice, users barely notice any difference in usage.

## Details

- Adds config parsing and validation for `ScreenSleepTimeLimit`
- Documents the new setting in `sesman.ini` and the man page
- Passes the configured value to xorgxrdp with `XRDP_SCREEN_SLEEP_TIME`

## Behavior

- `0` or `off` disables the feature
- Valid numeric range is `0..86400`
- This setting is intended for xorgxrdp sessions

## Dependency

This PR depends on the companion xorgxrdp PR which implements the actual screen-sleep logic.
https://github.com/neutrinolabs/xorgxrdp/pull/425

### DIFF
--- a/docs/man/sesman.ini.5.in
+++ b/docs/man/sesman.ini.5.in
@@ -213,6 +213,13 @@ If set to \fI0\fR, idle sessions will never be disconnected by timeout.
 This works only with xorgxrdp sessions. Moreover, xorgxrdp must be v0.2.9 or later.
 
 .TP
+\fBScreenSleepTimeLimit\fR=\fI[number|off]\fR
+Sets the idle time limit (in seconds) before xorgxrdp pauses display updates.
+Use \fBoff\fR or \fI0\fR to disable this feature. Allowed numeric range is
+\fI0\fR to \fI86400\fR.
+Paused output resumes when user activity is detected.
+
+.TP
 \fBPolicy\fR=\fI[Default|Separate|{UBDI}]\fR
 Session allocation policy. Used to decide when to allocate a
 new session. Set to one of the following values:
--- a/sesman/libsesman/sesman_config.c
+++ b/sesman/libsesman/sesman_config.c
@@ -28,6 +28,9 @@
 #include <config_ac.h>
 #endif
 
+#include <errno.h>
+#include <limits.h>
+
 #include "arch.h"
 #include "sesman_config.h"
 #include "sesman_clip_restrict.h"
@@ -84,6 +87,8 @@
 #define SESMAN_CFG_SESS_X11DISPLAYOFFSET "X11DisplayOffset"
 #define SESMAN_CFG_SESS_MAX_DISPLAY  "MaxDisplayNumber"
 #define SESMAN_CFG_SESS_STARTUP_WAIT_TIME "StartupWaitTime"
+#define SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT "ScreenSleepTimeLimit"
+#define SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT_MAX (24 * 60 * 60)
 
 #define SESMAN_CFG_SESS_POLICY_S "Policy"
 #define SESMAN_CFG_SESS_POLICY_DFLT_S "Default"
@@ -138,6 +143,43 @@ parse_policy_string(const char *value)
     }
 
     return rv;
+}
+
+/***************************************************************************//**
+ * Parse ScreenSleepTimeLimit value in seconds.
+ */
+static int
+parse_screen_sleep_time_limit(const char *value, int *time_limit)
+{
+    char *endptr;
+    long parsed;
+
+    if (value == NULL || value[0] == '\0')
+    {
+        return 1;
+    }
+
+    if (0 == g_strcasecmp(value, "off"))
+    {
+        *time_limit = 0;
+        return 0;
+    }
+
+    errno = 0;
+    parsed = strtol(value, &endptr, 10);
+    if (errno != 0 || endptr == value || *endptr != '\0')
+    {
+        return 1;
+    }
+
+    if (parsed < 0 || parsed > SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT_MAX ||
+            parsed > INT_MAX)
+    {
+        return 1;
+    }
+
+    *time_limit = (int) parsed;
+    return 0;
 }
 
 /******************************************************************************/
@@ -444,6 +486,7 @@ config_read_sessions(int file, struct config_sessions *se, struct list *param_n,
     se->kill_disconnected = 0;
     se->policy = SESMAN_CFG_SESS_POLICY_DEFAULT;
     se->startup_wait_time = 1500;
+    se->screen_sleep_time_limit = 0;
 
     file_read_section(file, SESMAN_CFG_SESSIONS, param_n, param_v);
 
@@ -510,6 +553,19 @@ config_read_sessions(int file, struct config_sessions *se, struct list *param_n,
             {
                 LOG(LOG_LEVEL_WARNING,
                     "Ignoring bad " SESMAN_CFG_SESS_STARTUP_WAIT_TIME " value");
+            }
+        }
+
+        else if (0 == g_strcasecmp(buf, SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT))
+        {
+            if (parse_screen_sleep_time_limit(value,
+                                              &se->screen_sleep_time_limit) != 0)
+            {
+                LOG(LOG_LEVEL_WARNING,
+                    "Ignoring bad " SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT
+                    " value '%s' (allowed: off or 0..%d seconds)",
+                    value,
+                    SESMAN_CFG_SESS_SCREEN_SLEEP_LIMIT_MAX);
             }
         }
     }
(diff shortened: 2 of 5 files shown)


metalefty:
Under what circumstances is it considered "idle" / "sleep"?

There is already a similar feature, [suppress output](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/79202daa-76b8-4a09-ac67-fc3cae7d895f). It is also a feature defined in the original RDP protocol. 

I don't see a valid reason to implement this feature.  I understand there is a difference between server-triggered and client-triggered triggers, though.

matt335672:
I think a similar result could be obtained by using the `BlankTime` setting for the X server. This might be a cleaner way to implement the same functionality.

metalefty:
Also,  I don't think it's a good idea to pause the display update while keeping the previous screen displayed.

Wouldn't this be possible by setting a "blank" screensaver that doesn't make any screen updates, rather than making changes on xrdp side?

metalefty:
Monitoring `BlankTime` and stopping screen updates after a certain period is essentially a blank screensaver.

tsz8899:
Thank @metalefty  @matt335672 

- The core of this feature is implemented in xorgxrdp; xrdp/sesman is only responsible for passing whether the feature is enabled and the `ScreenSleepTimeLimit` value to xorgxrdp. It is disabled by default and only takes effect when explicitly enabled by the user.

  - xorgxrdp currently registers `SaveScreen` in `xrdpdev/xrdpdev.c` (`pScreen->SaveScreen = rdpSaveScreen`), which is the entry point for the X Server to trigger `BlankTime` or the screensaver. However, this function in the source code only logs a message and returns `TRUE` — it does not perform any screen-clearing, drawing suspension, or notification to the RDP client to go black. See `xorgxrdp/xrdpdev/xrdpdev.c:360–372`. Therefore, even if the user enables `BlankTime` or `Option "DPMS"` in `xrdpdev/xorg.conf` (which itself appears only once in the config, at `xorgxrdp/xrdpdev/xorg.conf:42`), the virtual driver will not actually modify the framebuffer, nor will it stop sending updates to the client.

  - The virtual display adapter also has no power-switching logic implemented for DPMS: a search of the codebase reveals no DPMS state handler, so within this module `BlankTime` is effectively a placeholder — "the interface exists but is not implemented." The X session still runs the full render/encode pipeline. Even if OS-level or RDP-protocol-level Suppress Output is enabled, xorgxrdp will continue to call `FillSpans`, `Composite`, `PutImage`, and similar functions, generating encoding load with no actual power savings.

  - Our patch intercepts these drawing entry points directly inside the xorgxrdp module — `rdpFillSpans`, `rdpComposite`, `rdpCopyArea`, `rdpGlyphs`, `rdpCursor`, `rdpPutImage`, and `rdpClientConAddDirtyScreenReg` — and applies `rdpScreenSleepTimedOut` uniformly across all of them. Once the session has been idle for `XRDP_SCREEN_SLEEP_TIME`, frames are neither submitted to the screen nor encoded. When input arrives, `check_wake_up` forces a full frame refresh before resuming normal operation, truly achieving "stop updating, wake only on input" semantics. The xrdp/sesman side is only responsible for passing the user-configured `ScreenSleepTimeLimit` to xorgxrdp via the `XRDP_SCREEN_SLEEP_TIME` environment variable. The default value of `0` / off leaves existing behavior completely unchanged and is entirely opt-in.

  - This patch fills the gap described above: it intercepts frame generation directly inside xorgxrdp, truly stopping data delivery to the RDP client, rather than relying on the client to handle a "black screen." Compared to a system screensaver, pausing updates on the xorgxrdp side can bring CPU, encoding, and network activity to a complete halt during periods when the user is away — an effect that becomes especially significant when multiple idle sessions coexist.

  - This mechanism is designed for long-lived connections with brief periods of user absence. Users can configure a specific keyboard/mouse idle timeout after which the server immediately stops image encoding, CPU usage drops to near zero, and associated network traffic is substantially reduced. Unlike the OS-native `BlankTime` or screensaver, we preserve the last rendered frame unchanged, so when the user returns they immediately see the previous screen followed by a forced refresh — a far better experience than a black screen, and well suited to browser-based or monitoring-style use cases. *(Alternatively: refresh one frame per minute?)*

  - As a concrete example: in a multi-user environment where each xrdp session involves only 3–4 hours of actual interaction per day, with the remaining ~20 hours spent simply maintaining the connection, the existing `BlankTime` and RDP suppress logic leave xorgxrdp continuously performing H.264/RFX encoding for every session, consuming CPU, network bandwidth, and power throughout. With the new xorgxrdp sleep feature enabled, these sessions produce zero encoding and consume zero bandwidth during the ~20-hour idle window. Idle resource usage and energy consumption across the entire cluster are significantly reduced, while per-user resumption remains fast.

tsz8899:
For an X session, enabling the OS screensaver or setting BlankTime does not stop the X server from running, nor does it stop applications from drawing. It typically only switches the display state to a black screen or screensaver window, or triggers DPMS to power off the display. In other words, the display pipeline remains active, applications keep running, the event loop keeps spinning, and drawing requests continue to flow into the X server — so CPU usage generally continues. In xorgxrdp specifically, even with BlankTime or a screensaver enabled, the framebuffer → encoder computation pipeline keeps running as before.

Does CPU usage continue
This primarily depends on whether applications are still refreshing content. For example, a browser playing video, animated web content, a monitoring dashboard, or a UI with periodic refresh timers — even with the screen blanked, the application may continue decoding, laying out, compositing, and submitting draw requests, which the X server must still process. CPU usage therefore typically remains sustained.

Does GPU usage continue
If the desktop environment or applications have GPU acceleration enabled, the compositor is still running, or video is still being decoded, the GPU will generally not stop automatically due to BlankTime. A black screen does not mean rendering has stopped — applications continue rendering, with a black layer composited on top; the final display output simply appears black or the output is shut off.

What does DPMS actually do
DPMS primarily reduces the power consumption of the display output device itself — for example, a physical monitor's backlight or the output link. It does not equate to stopping application rendering. In most cases, only the monitor saves power while the host CPU and GPU remain fully busy.

The screensaver program itself
A dynamic screensaver may actually increase CPU/GPU consumption. Only a pure static black screensaver has any chance of reducing some redraw activity — and even then, only if no other application is continuously updating.

BlankTime, screensavers, and DPMS primarily alter the display output state. They provide no guarantee of stopping application drawing or halting X server graphics processing, so CPU and GPU usage typically continue unabated.

Testing confirms that compared to enabling the OS screensaver or configuring BlankTime, implementing the sleep feature directly inside xorgxrdp delivers significantly better results.

metalefty:
> Unlike the OS-native `BlankTime` or screensaver, we preserve the last rendered frame unchanged, so when the user returns they immediately see the previous screen followed by a forced refresh — a far better experience than a black screen, and well suited to browser-based or monitoring-style use cases. *(Alternatively: refresh one frame per minute?)*

Honestly, I strongly dislike this: _"preserve the last rendered frame unchanged"_, even if the feature doesn't change the default behaviour and opt-in.

If the session falls into the idle mode, it must display something that indicates idle mode, a black-blank screen, for example.  Not operating does not mean being away from the keyboard. It could also just mean they’re "just looking" at the monitoring dashboard or a stock chart, right? Displaying the last rendered frame as-is and pausing updates means that users may believe they are looking at up-to-date information, without realizing that the screen is actually paused and showing stale content. That sounds like a too-bad scenario for me.

It should display something that clearly indicates the screen is in idle state.

BTW, what about "lockscreen"? Do applications continue rendering under the lock screen layer?

tsz8899:
Thank @metalefty

Would it be better to add a switch with options such as black screen / logo / last screen?

In my environment AVC420, The Linux lock screen and screensaver are independent programs with no connection to other applications. When active, the browser still invokes graphics rendering as usual.
when playing 4K video in a browser, enabling the OS lock screen or screensaver does not reduce CPU usage — browser 50% xrdp sits at 65% and xorg at 35%. However, after enabling the patch, usage drops to browser 50% xrdp 1% and xorg 1%.

metalefty:
Thanks, I understand the benefits of the feature.  Let’s explore how we should implement it, and let's start looking at xorgxrdp side.

Since this change just adds an environment variable passed to xorgxrdp, and this feature essentially does not require changes to xrdp code.   I want to modify the xrdp code only for features that we've decided to keep for a long period of time.

I think it should work with the following configuration and doesn't require xrdp code change at the moment. 

```ini
[SessionVariables]
XRDP_SCREEN_SLEEP_TIME=60
```

metalefty:
I did a similar thing before, changing xorgxrdp behaviour without changing xrdp code. 
https://github.com/neutrinolabs/xorgxrdp/releases/tag/v0.9.19

metalefty (changes requested):
Not to be merged at this time since xorgxrdp behaviour can be changed without changing xrdp code.

neutrinolabs/xorgxrdp#425 actually does not depend on this PR.

metalefty:
Closing this to prevent accidental merging, but feel free to comment.

Let's proceed on https://github.com/neutrinolabs/xorgxrdp/pull/425.

PR #3776, opened 2026-03-29
Add support for vmconnect resizing

gpotter2:
Hi ! This PR is a followup to https://github.com/neutrinolabs/xrdp/pull/3514 to try and improve the experience on Hyper-V. It enables auto-resizing of the RDP client when used in vmconnect mode, which currently doesn't work.

I was trying to understand why vmconnect.exe (used in Hyper-V scenarios) would allow auto-resizing on Windows but not on Linux. The answer is that it's actually using the `OnLoginComplete` event to detect when resize is available (https://learn.microsoft.com/en-us/windows/win32/termserv/imstscaxevents-onlogincomplete), because resize isn't available on Windows on the Winlogon screen. This event is triggered when the client receives the `INFOTYPE_LOGON*` types, as described by [MS-RDPBCGR].

By implementing this message, I was able to have resize work with vmconnect as shown below, once the login screen is passed (this is the same behavior in a Windows VM):

https://github.com/user-attachments/assets/2bf99ee6-9f82-43c9-a17d-0d26f56baa80

As a side-effect, completely unrelated, it seems that MSRDC (Windows App / the "Remote Desktop" app) also uses `OnLoginComplete` to know if another connection is allowed (it only allows 1 login screen at a time). So I believe that this change will also help in that regard.

PR #3777, opened 2026-03-30
Rename RDP_INFO flags to spec names

gpotter2:
This is a part of https://github.com/neutrinolabs/xrdp/pull/3776

### DIFF
--- a/common/ms-rdpbcgr.h
+++ b/common/ms-rdpbcgr.h
@@ -152,13 +152,18 @@
 #define SEC_TAG_KEYSIG                 0x0008 /* BB_SIGNATURE_KEY_BLOB */
 
 /* Info Packet (TS_INFO_PACKET): flags (2.2.1.11.1.1) */
-/* TODO: to be renamed */
-#define RDP_LOGON_AUTO                 0x0008
-#define RDP_LOGON_NORMAL               0x0033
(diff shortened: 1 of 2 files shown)


matt335672 (commented):
I'm all for doing this - I've done a few myself, but is there a reason why we're not using the actual spec names here (i.e. without the `RDP_` prefix)?

The only reason I can think of not to use spec names is if it would cause a conflict in xrdp-neutrinordp.c. I've just run a quick compile check and this isn't the case.

I'm happy to be argued out of this position.

gpotter2:
I don't have an opinion. Do you want me to change it?

matt335672:
I think so. If we get a clash later we can fix it then. I think sticking as close to the spec as we can is a good thing to at least try to achieve.

gpotter2:
Fine by me, done ! In this case I also took the liberty to split the "normal". I think I kept the `RDP_` initially because I couldn't figure how to rename the "normal"

metalefty (changes requested):
LGTM

metalefty (approved):
Oops.

gpotter2:
Thanks ! this PR only exists to make https://github.com/neutrinolabs/xrdp/pull/3776 possible, feel free to take a look 😊

PR #3779, opened 2026-04-05
remove keymap file before creating sym link

jsorg71:
fix for error
ln: failed to create symbolic link 'km-00020426.toml': File exists

### DIFF
--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -110,4 +110,4 @@ if FREEBSD
 		$(DESTDIR)$(sysconfdir)/rc.d/xrdp-sesman
 endif
 	(cd $(DESTDIR)${startscriptdir} && \
-		$(LN_S) km-00010426.toml km-00020426.toml)
+		$(LN_S) -f km-00010426.toml km-00020426.toml)


instfiles/Makefile.am
```diff
@@ -110,4 +110,5 @@ if FREEBSD
 		$(DESTDIR)$(sysconfdir)/rc.d/xrdp-sesman
 endif
 	(cd $(DESTDIR)${startscriptdir} && \
+		$(RM) km-00020426.toml; \
```
metalefty: I think `${LN_S} -f` will work instead.

jsorg71: Ok, I was wondering about that.  If that is portable I'll change to ${LN_S} -f.

metalefty: It works both on GNU/Linux and FreeBSD.

Linux
```
       -f, --force
              remove existing destination files
```

FreeBSD
```
     -f    If the target file already exists, then unlink it so that the link
           may occur.  (The -f option overrides any previous -i and -w
           options.)
```

matt335672 (approved):
I'm happier with using `-f` too. It's in POSIX:

https://pubs.opengroup.org/onlinepubs/009695399/utilities/ln.html

PR #3782, opened 2026-04-14
Supress -Wunused-function warnings

metalefty:
sig64_to_uint64() is only called when devel logging is enabled. It will be unused unless configured with `--enable-logging-devel`.

@matt335672 Do you think this is a valid fix for this?

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -960,6 +960,8 @@ xrdp_sec_check_sig(struct xrdp_sec *self, const char *sig, int sig_len,
 
 /*****************************************************************************/
 /* Function to return a 64-bit int from an 64 bit signature */
+
+#ifdef USE_DEVEL_LOGGING
 static uint64_t
 sig64_to_uint64(const char sig[], int sig_len)
 {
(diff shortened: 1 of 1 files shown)


matt335672 (commented):
It will work.

`__attribute__((unused))` is GNU and CLANG specific I believe. A better approach might be to add this to `common/defines.h`:

```c
#ifdef __GNUC__
// Clang too
#define ATTR_UNUSED __attribute__ ((unused))
#else
#define ATTR_UNUSED
#endif
```

And then use `ATTR_UNUSED` instead of the raw attribute. This is the approach I took in neutrinolabs/pipewire-module-xrdp@5d3e2cf208d9ffef46389f72ab3facfbac63862a

metalefty:
I choose `MAYBE_UNUSED` inspired by `[[maybe_unused]]` in C++17.

matt335672:
LGTM, but Copilot makes a good point.

metalefty:
Let's go with the copilot, and remove the guard when we want to use the function without devel logging.

matt335672:
LGTM

PR #3787, opened 2026-04-17
Clarify handling of duplicate vulnerability reports

metalefty:
Then we can simply close duplicate vulnerability reports.


matt335672:
LGTM - great idea.

matt335672 (approved):

PR #3788, opened 2026-04-17
Release v0.10.6

metalefty:
@matt335672  Let me know if I missed something in the release note.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,49 @@
+# Release notes for xrdp v0.10.6 (2026/04/17)
+
+## General announcements
+
+If you like xrdp, please consider sponsoring or donating to the project. We accept financial contributions through [Open Collective](https://opencollective.com/xrdp-project), and direct donations to individual developers via GitHub Sponsors are also welcome.
+
+* [V0.10.3] Experimental support for utmp/wtmp file is provided in this release. If you use this, be aware that these files are only updated when an xrdp session is created or destroyed. Disconnections and reconnections to the same session are not tracked. In particular:
+  * the FROM address for a client (as shown by the `w` command) reflects the IP address of the client at the time of creation, and not the address of the currently connected client.
+  * Sessions started by the `xrdp-sesrun` command do not have a FROM address.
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,3 +1,49 @@
...
+Thanks to @explitintel, @smittix and @hessandrew!
+
+- [CVE-2026-32105](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-j2jm-c596-c5q3)
+- [CVE-2026-32107](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-p5m6-7m43-pjv9)
+- [CVE-2026-32623](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-phw3-qp59-x2v4)
+- [CVE-2026-32624](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-7q2g-6fjr-h6pp)
+- [CVE-2026-33145](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-rmvv-7633-fg7h)
+- [CVE-2026-33516](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-rvh9-9wm3-28c7)
+- [CVE-2026-33689](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-92mr-6wpp-27jj)
+- [CVE-2026-35512](https://github.com/neutrinolabs/xrdp/security/advisories/GHSA-jg6p-7fg8-9hh6)
```
matt335672, line 23: Are these publicly visible? For previous releases we've linked to the public CVE repository

matt335672 (commented):
One comment below

NEWS.md
metalefty: I'll publish them at the same time as the release, and it will be visible immediately.

Which do you think is better, GHSA or a public CVE repository? Either is fine for me, but GHSA has the minimum lag between disclosure and becoming visible.

matt335672: I don't really mind. If we ever migrate off of Github, which is possible given the Geopolitical situation, links to public repositories might be better, but we could fix that when necessary.

metalefty:
Thanks, let's ship it.

PR #3793, opened 2026-04-20
[V0.10] regression: Fix SEGV in xrdp when running over TLS

matt335672:
When not using classic RDP encryption, an uninitialsed pointer can be passed to sig64_to_uint64() in development mode.

(cherry picked from commit 5fa4dc02bc22b0bf54d301ef002dcaa0db48c390)

### DIFF
--- a/libxrdp/xrdp_sec.c
+++ b/libxrdp/xrdp_sec.c
@@ -1396,7 +1396,7 @@ xrdp_sec_recv_fastpath(struct xrdp_sec *self, struct stream *s)
     int ver;
     int len;
     int pad;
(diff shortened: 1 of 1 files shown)


matt335672:
My bad - I introduced this in 0d8cf57e9d12393bd452a2b6cb1af0e38887d8a2 as part of a security fix.

I don't think this is important enough to warrant v0.10.6.1 in its own right, as the call which causes the SEGV is not compiled in unless we're in development mode. I'm assuming we'll have a v0.10.6.1 however, and I've started a release note for it.

PR #3794, opened 2026-04-20
[v0.9] Update NEWS with EOL announcement

matt335672:
Due to the current use of AI-assisted vuln discoveries, and the age of this branch, we are no longer able to support it.

@metalefty - are you happy with this?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,10 @@
 
 ## General announcements
 
-_xrdp v0.9.x is end-of-life. New releases may happen when severe security vulnerabilities or critical bugs are found._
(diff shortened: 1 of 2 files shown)


metalefty:
I agree that v0.9 is effectively EoL'd. Personally, I would like to accept sponsorship for vulnerability fixes (extended support) for xrdp v0.9 if a sponsor comes forward, though this is unlikely. Let’s keep the door slightly open.

Maybe we could add a sentence like 'If you'd like us to fix vulnerabilities in xrdp v0.9 and are willing to sponsor the work, contact xrdp-core@googlegroups.com'.

matt335672:
Done. EOL announcement also copied to README.md

metalefty (approved):
LGTM

PR #3058, opened 2024-05-05
Add ibus unicode input support

seflerZ:
Close https://github.com/neutrinolabs/xrdp/issues/1990 . This PR was made by me and @matt335672. It introduced serveral unicode input interfaces and has the ibus interface implemented (That means you should have ibus configured in the server side first).

The unicode input is a kind of method of sending unicode codes directly from the client to the server. With it, you can input any characters or glyphs (like Emoji 😁) even if they are not present in the keymap of the keyboard. It is typically used by mobile devices like Android\iOS phones, especially in Asia languages.

Another benifit of the unicode input is you can use your device's voice input to generate a bunch of texts quickly. Although we only have the ibus implemented now, it is easy to do for other input services like fcitx.

_@matt335672 This is the latest rebased. Sorry I made merge before._

PR #3804, opened 2026-05-08
Support client-committed Unicode input in the Xorg backend

YTjungle666:
Summary
- Preserve RDP Unicode input values as unsigned UTF-16 code units so non-ASCII input is not corrupted by sign extension.
- Add WM_UNICODE_INPUT routing for Xorg sessions through the existing backend mod_event path, without adding a separate side channel.
- Keep the existing keyboard/scancode behavior as fallback for non-Xorg sessions or backend failures.
- Strip terminal bracketed-paste delimiters emitted by some RDP clients while preserving all user text.

This is intended to support Chinese text input and pasted Unicode text from Microsoft Remote Desktop mobile clients and Apache Guacamole browser sessions, while keeping existing keyboard behavior working.

Companion xorgxrdp PR: https://github.com/neutrinolabs/xorgxrdp/pull/429

Testing
- Built with the companion xorgxrdp branch and installed the resulting 0.10.80 local build on Debian.
- Ran make check for xrdp.
- Replayed captured Microsoft Remote Desktop mobile and Apache Guacamole input samples against the installed Xorg modules. The replay matched 83/83 expected text codepoints, covering direct Chinese input, pasted Chinese text, Chinese punctuation, lowercase ASCII, uppercase ASCII, and symbols.

### DIFF
--- a/common/xrdp_constants.h
+++ b/common/xrdp_constants.h
@@ -284,6 +284,7 @@
 #define WM_KEYDOWN     15
 #define WM_KEYUP       16
 #define WM_KEYBRD_SYNC 17
+#define WM_UNICODE_INPUT 19
 #define WM_MOUSEMOVE   100
 #define WM_LBUTTONUP   101
 #define WM_LBUTTONDOWN 102
--- a/libxrdp/xrdp_rdp.c
+++ b/libxrdp/xrdp_rdp.c
@@ -1023,8 +1023,10 @@ xrdp_rdp_process_data_input(struct xrdp_rdp *self, struct stream *s)
         in_uint32_le(s, time);
         in_uint16_le(s, msg_type);
         in_uint16_le(s, device_flags);
-        in_sint16_le(s, param1);
-        in_sint16_le(s, param2);
+        /* Unicode input carries UTF-16 code units, so keep the raw
+         * unsigned value. Sign-extension would corrupt non-ASCII input. */
+        in_uint16_le(s, param1);
+        in_uint16_le(s, param2);
         LOG_DEVEL(LOG_LEVEL_TRACE, "With field [MS-RDPBCGR] TS_INPUT_EVENT "
                   "eventTime %d, messageType 0x%4.4x", time, msg_type);
 
@@ -1697,4 +1699,3 @@ xrdp_rdp_send_session_info(struct xrdp_rdp *self, const char *data,
     free_stream(s);
     return 0;
 }
-
--- a/xrdp/xrdp_types.h
+++ b/xrdp/xrdp_types.h
@@ -568,6 +568,11 @@ struct xrdp_wm
     /* Unicode input */
     int last_high_surrogate_key_up;
     int last_high_surrogate_key_down;
+    /* Some clients wrap pasted Unicode text in terminal bracketed-paste
+     * delimiters. Keep a short pending buffer so these markers can be removed
+     * before committed text is forwarded to the backend. */
+    char32_t unicode_pending_chars[6];
+    unsigned int unicode_pending_chars_len;
     /* client info */
     struct xrdp_client_info *client_info;
     /* session log */
--- a/xrdp/xrdp_wm.c
+++ b/xrdp/xrdp_wm.c
@@ -1785,6 +1785,139 @@ fake_kbd_event_from_scancode_index(struct xrdp_wm *self, int device_flags,
 }
 
 /*****************************************************************************/
+#define BRACKETED_PASTE_SEQ_LEN 6
+
+/* Guacamole and mobile RDP clients may deliver pasted text as Unicode input
+ * events wrapped in terminal bracketed-paste delimiters. These markers are not
+ * user text, so strip them while preserving all other Unicode codepoints. */
+static const char32_t g_bracketed_paste_start[BRACKETED_PASTE_SEQ_LEN] =
+    { 0x1b, '[', '2', '0', '0', '~' };
+static const char32_t g_bracketed_paste_end[BRACKETED_PASTE_SEQ_LEN] =
+    { 0x1b, '[', '2', '0', '1', '~' };
+
+static int
+xrdp_wm_send_unicode_to_backend(struct xrdp_wm *self, char32_t unicode)
+{
+    if (self == NULL || self->mm == NULL || self->mm->mod == NULL ||
+            self->mm->mod->mod_event == NULL)
+    {
+        return 1;
+    }
+
+    /* Pass the codepoint over the normal Xorg backend input path. This avoids a
+     * separate channel and lets xorgxrdp inject text in the Xorg input driver. */
+    return self->mm->mod->mod_event(self->mm->mod, WM_UNICODE_INPUT,
+                                    unicode, 0, 0, 0);
+}
+
+/*****************************************************************************/
+static int
+unicode_pending_matches(struct xrdp_wm *self, const char32_t *sequence,
+                        unsigned int sequence_len)
+{
+    unsigned int index;
+
+    if (self->unicode_pending_chars_len > sequence_len)
+    {
+        return 0;
+    }
+
+    for (index = 0; index < self->unicode_pending_chars_len; ++index)
+    {
+        if (self->unicode_pending_chars[index] != sequence[index])
+        {
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
+/*****************************************************************************/
+static int
+unicode_pending_is_bracketed_paste_prefix(struct xrdp_wm *self)
+{
+    return unicode_pending_matches(self, g_bracketed_paste_start,
+                                   BRACKETED_PASTE_SEQ_LEN) ||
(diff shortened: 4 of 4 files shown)


matt335672:
Thanks @YTjungle666 

@seflerZ, @jsorg71 - I'd appreciate your thoughts on whether this approach is preferable to the current method using ibus support. I lack the experience to make a value judgement on this.

YTjungle666:
Thanks. My intention here is not to replace ibus support.

I see this as covering a different input path: cases where the RDP client has already committed Unicode text and sends it as RDP Unicode keyboard input. That includes mobile soft keyboards and browser-based clients such as Apache Guacamole, and also pasted Unicode text. In those cases, requiring a working server-side ibus session adds another dependency even though the text is already composed on the client side.

ibus is still the right layer for server-side IME composition, preedit text, candidate windows, and users who want to type using an IME inside the Linux desktop session. This PR is meant to provide a native fallback/base path for committed Unicode text through the existing Xorg backend input path, while leaving ibus-based composition intact.

matt335672:
I'm sorry @YTjungle666 - I don't understand the distinction here.

What is the use-case for this, i.e. what will a user see that's different when compared to using the ibus input path?

YTjungle666:
https://github.com/user-attachments/assets/ccdfba7c-1fc6-4630-b828-86bb38edd4e2

@matt335672  Sorry, I think my previous explanation was not clear enough. I described the implementation path too much, but did not explain the user-visible difference clearly.

The distinction I want to make is:

This PR is not intended to replace IBus or server-side IME composition.

IBus is still the right layer when the user wants to compose text inside the remote Linux desktop session. For example, the user types pinyin/key events into the remote session, IBus shows preedit text and candidate windows, and then IBus commits the final Chinese text in the Linux session.

This PR handles a different path: already-committed Unicode text from the RDP client.

A concrete example is Microsoft Remote Desktop / Windows App on a mobile device. The phone's Chinese keyboard can compose the text locally on the client side. After the user selects/commits the Chinese text on the phone keyboard, the RDP client sends Unicode input events to xrdp. At that point the text is already composed; the server does not need to run an IME to convert pinyin into Chinese again. xrdp only needs to deliver those committed Unicode code points to the Xorg session.

So the user-visible difference is:

- Before this PR: on mobile clients, Chinese input usually has to depend on a working server-side IME/IBus setup. The user types pinyin/English key events into the remote session and lets the remote Linux IME compose the Chinese text.
- After this PR: the user can use the phone's own Chinese keyboard directly. Once the text is committed on the phone, the committed Chinese characters are sent through RDP Unicode input and appear directly in the remote Xorg application.

The attached video demonstrates this difference. The important point in the video is not just that Chinese text appears, but where the composition happens:

- Before: composition is effectively done through the server-side input method.
- After: composition is done by the mobile/client-side IME, and xrdp/xorgxrdp only forwards/injects the committed Unicode text.

So I see this as complementary to the existing IBus path:

- IBus: server-side IME composition, preedit text, candidate windows, Linux desktop input method behavior.
- This PR: client-side committed Unicode text from RDP clients, especially mobile clients and browser-based clients such as Guacamole.

I hope this makes the use case clearer.

matt335672:
Thanks @YTjungle666 

I'm a bit snowed under IRL, but I hope to get your other PR reviewed very soon. It may take me a little while to get to this one.

seflerZ:
@YTjungle666 The client side unicode input is supported through iBus. There is reason why iBus should be introduced, it's the simulated keyboard only accepts chars on the keys and the keyboard has limits keys. You simply can't send it to the OS directly through that way.  There should be a translator.

Besides application frameworks like GTK or Qt haven different input channels for unicode, unless the xorg server side has fully implemented unicode input. A translator like iBus is a must.

You can have more tests on varies OS systems and Apps with iBus feature off. If it does work, the iBus layer can be removed. 

You can refer to this thread to get a background of it:  https://github.com/neutrinolabs/xrdp/pull/3058 and https://github.com/neutrinolabs/xrdp/issues/1990

YTjungle666:
@seflerZ Thanks, I agree that IBus / server-side IME support is still important.

I think the key point is that this PR is not intended to replace IBus. It is meant to coexist with it and cover a different input path.

IBus is still the right layer for cases where text composition happens inside the remote Linux desktop session, for example:

- the client sends normal keyboard/scancode input;
- the user types pinyin / kana / other composition keys in the remote session;
- the remote Linux IME needs to show preedit text, candidate windows, and then commit the final text;
- applications/toolkits rely on the Linux input method framework for composition behavior.

This PR is for a different case: the RDP client has already produced committed Unicode text before it reaches xrdp.

For example, on a mobile client, the phone keyboard or OS IME may already compose and commit Chinese text locally. In that case, the input is no longer “keyboard keys that need to be translated into text” — it is already committed Unicode text from the client. The server-side IME does not need to compose pinyin again; xrdp only needs to deliver those Unicode code points into the Xorg session.

So I see the two paths as complementary:

- IBus / server-side IME path:
  Used when composition should happen in the remote Linux session.
  This remains necessary for preedit, candidate selection, toolkit IME integration, and clients which only send normal keyboard events.

- This PR / RDP Unicode path:
  Used when the client has already committed Unicode text and sends that committed text through RDP Unicode input events.
  This is useful for mobile clients, OS-level client IMEs, and similar cases where composition happens on the client side.

So this PR should not remove or weaken the IBus layer. If the user wants server-side composition, or if an application/client path requires Linux IME behavior, IBus is still the correct solution. This PR only makes the already-committed Unicode input path work through the Xorg backend instead of forcing everything through keyboard simulation / server-side IME composition.

I can also update the PR description to make this distinction clearer.

seflerZ:
Sorry, didn't catch you. The keys are as you've described: 

> Before this PR: on mobile clients, Chinese input usually has to depend on a working server-side IME/IBus setup. The user types pinyin/English key events into the remote session and lets the remote Linux IME compose the Chinese text.
> After this PR: the user can use the phone's own Chinese keyboard directly. Once the text is committed on the phone, the committed Chinese characters are sent through RDP Unicode input and appear directly in the remote Xorg application.

The current implemention can do what you've expected. You type in Unicode chars like '你好' with your phones input method or voice recongition. iBus directly commit the '你好' onto the input field, no candidates appears popups, just as what you want. Below is my demo without the PR:

https://github.com/user-attachments/assets/e5ca2b11-7798-4ab6-9c83-e014de2ab699


So I really don't know what this PR resolves.

YTjungle666:
@seflerZ Oh, I understand what you mean. It's probably a bug I'm encountering. When I remotely access my Debian using the Windows app, I can't directly input Chinese characters using my phone's keyboard. I'll investigate why iBUS is having this problem in my environment.

YTjungle666:
@seflerZ 抱歉，请允许我直接用中文评论，因为我的英文不是特别好，导致我之前一直没完全看懂Ibus那个PR真正想表达的内容，我以为Ibus可能是实现了另外的场景？（之前的英文呢评论其实是让AI帮我代写的哈哈）我之前自己简单测试了Ibus，之前并没有实现我想要的功能，因为我的输入环境主要是fcitx5为主，简单配置Ibus导致和fcitx5可能有冲突，这次我修改了正确的配置文件，发现Ibus确实可以实现我想要的功能，那么我感觉我这次PR确实是要取代Ibus翻译层了，因为我也同时提交了xrdpxorg直接支持unicode的PR，我接下来会严格的测试一下是否可行，但是我的测试环境可能没有那么多，肯定是不能全面的测试到位的，所以如果这个PR真的通过了，那么希望和Ibus同时存在一段时间，看后续是否有其他的问题，我后续再测试修改一下这个方案，让他尽可能的全面实现Ibus的功能

Sorry, please allow me to comment directly in Chinese. My English isn't very good, which is why I didn't fully understand the Ibus PR's true intention before. I thought Ibus might be implementing something else entirely? (My previous English comments were actually written by AI, haha.) I did a simple test of Ibus before, and it didn't achieve the functionality I wanted because my input environment is mainly fcitx5, and the simple configuration of Ibus might have caused conflicts with fcitx5. This time, I modified the correct configuration file and found that Ibus can indeed achieve the functionality I want. So I feel that this PR is indeed going to replace the Ibus translator, because I also submitted a PR for xrdpxorg to directly support Unicode. I will rigorously test its feasibility, but I may not have enough test environments to fully test it. So if this PR is approved, I hope it can coexist with Ibus for a period of time to see if there are any other problems. I will then test and modify this solution to make it as comprehensive as possible in implementing the functionality of Ibus.

seflerZ:
@YTjungle666 如果确认系统已经广泛支持直接接受 unicode, 去掉 iBus 翻译层确实是更优解. Windows 系统无疑是支持的, 但是当时的 Ubuntu 22.04 待确认, 你可以试一下. 另外我向你推我的开源 RD 客户端: https://github.com/seflerZ/qidesk 

Well if you confirm the modern OS systems support handle unicode directly, it is a better way to remove iBus translator. Be sure to have widely tests, especially on old systems like Ubuntu 22.04. Besides, I recommend you my open source RD client: https://github.com/seflerZ/qidesk

PR #3810, opened 2026-06-06
Fixes compilation error for raspberry pi

Nexarian:
When building XRDP on Raspbian Trixie, we get this error. The compiler is noticing that jpeg_props[0] is unsigned, meaning this is a wasteful comparison.

Fix by removing the comparison.
```
  CC       xrdp_caps.lo
xrdp_caps.c: In function 'xrdp_caps_process_codecs':
xrdp_caps.c:650:48: error: comparison is always false due to limited range of data type [-Werror=type-limits]
  650 |             if (self->client_info.jpeg_prop[0] < 0 || self->client_info.jpeg_prop[0] > 100)
      |                                                ^
make[2]: *** [Makefile:561: xrdp_caps.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/tmp/tmp.L9ynfYnaji/xrdp/libxrdp'
make[1]: *** [Makefile:530: all-recursive] Error 1
make[1]: Leaving directory '/tmp/tmp.L9ynfYnaji/xrdp'
make: *** [Makefile:462: all] Error 2
```

Proof that it works:
image

### DIFF
--- a/libxrdp/xrdp_caps.c
+++ b/libxrdp/xrdp_caps.c
@@ -647,7 +647,7 @@ xrdp_caps_process_codecs(struct xrdp_rdp *self, struct stream *s, int len)
             g_memcpy(self->client_info.jpeg_prop, s->p, i1);
             self->client_info.jpeg_prop_len = i1;
             /* make sure that requested quality is  between 0 to 100 */
-            if (self->client_info.jpeg_prop[0] < 0 || self->client_info.jpeg_prop[0] > 100)
+            if (self->client_info.jpeg_prop[0] > 100)
             {
                 LOG(LOG_LEVEL_WARNING, "  Warning: the requested jpeg quality (%d) is invalid, "
                     "falling back to default", self->client_info.jpeg_prop[0]);


matt335672:
Hi Nex.

Sorry for the delay in getting to this.

One comment - the definition of jpeg_prop is:

```
char jpeg_prop[64];
```

The standard doesn't specify whether this is signed on unsigned. [From C99 6.2.5(3)](https://port70.net/%7Ensz/c/c99/n1256.html#6.2.5p3):

> An object declared as type char is large enough to store any member of the basic execution character set. If a member of the basic execution character set is stored in a char object, its value is guaranteed to be nonnegative. If any other character is stored in a char object, the resulting value is implementation-defined but shall be within the range of values that can be represented in that type. 

So it might be a good idea to make this change as well, to remove any ambiguity:

```diff
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -186,7 +186,7 @@ struct xrdp_client_info
     char ns_prop[64];
     int jpeg_codec_id;
     int jpeg_prop_len;
-    char jpeg_prop[64];
+    unsigned char jpeg_prop[64];
     int v3_codec_id;
     int rfx_min_pixel;
     char orders[XR_PRIMARY_ORDER_COUNT];
```

What do you think?

PR #3811, opened 2026-06-17
sesexec: Fix CVE-2026-42218 regression

matt335672:
cppcheck has picked up on the use of an unitialised variable in the implementation of the fix for CVE-2026-42218


metalefty:
Thanks!

PR #3815, opened 2026-06-23
Start Xorg sessions at the client monitor's DPI

### LINKED ISSUE #3473: X server is not starting with correct DPI
### xrdp version

0.10.2

~~~
~~~

Rocky Linux 9.5 

### Installation method

### Which backend do you use?

Xorgxrdp

KDE

VM

### What's your client?

mstsc.exe (Windows 11)

### Area(s) with issue?

Other

### Steps to reproduce

Connect to xrdp with the scaling factor set to 150%.

### ✔️ Expected Behavior

xrdp's display should have the same DPI as the display.

### ❌ Actual Behavior

xrdp's display defaults to 96 DPI.

### Anything else?

Hi,

I'm trying to use xrdp on a HiDPI display with a 3840x2160 resolution and a 150% scaling factor (the recommended scaling factor). The desktop environment I am running isn't automatically adjusting to the display's scaling factor, meaning everything looks too small. This problem is very similar to the problem described in issue https://github.com/neutrinolabs/xrdp/issues/3200, where the proposed solution is changing the DPI after xrdp has established the session and the display information has been passed to xrandr. However, I find this solution hard to work with for two reasons:

Firstly, it does not seem to account for different scaling factors, i.e., possible different DPIs. Indeed, the issue suggests using the resolution and physical size of the monitor passed to xrandr to calculate a "proper" DPI for the display, similarly to what is done in https://github.com/neutrinolabs/xrdp/wiki/Scaleable-login-screen, which might not correspond to the current DPI of the display. For example, starting a session with the scaling factor set to 150% outputs in the logs:

`Login screen monitor height is 2160 pixels over 336 mm (163 DPI)`

While starting a session with the scaling factor set to 200% also outputs in the logs:

`Login screen monitor height is 2160 pixels over 336 mm (163 DPI)`

What this indicates is that there is a disconnect between the DPI set by myself and this "proper" DPI that can be calculated through the sent resolution and physical size. This seems odd, given the fact that the RDP protocol communicates this desktop scale factor in the initial capabilities exchange https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/db6713ee-1c0e-4064-a3b3-0fac30b4037b, more specifically, in ClientCoreData https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/00f1da4a-ee9c-421a-852f-c19f92343d73.

Secondly, one has to change the DPI after the session has started, again, this is strange since the desktop scale factor is sent by the client. Is it not possible to launch the X server with the DPI defined by the received desktop scale factor? For example, if the received desktop scale factor is 150% why can't the X server be configured to start with a DPI of 144 (96 * 1.5)?

Thanks.

prwang:
# Start Xorg sessions at the client monitor's DPI

Fixes #3473

## Problem

xrdp already computes the client monitor's physical DPI and uses it to scale the
**login screen** (you can see it in the log: *"Login screen monitor height is N
pixels over M mm (D DPI)"*). But that value is never passed to xrdp-sesman, so
the **Xorg/xorgxrdp desktop** that follows always starts at the default 96 DPI.

The result is the inconsistency reported in #3473: on a HiDPI client the login
screen is scaled correctly, then the desktop appears with everything too small.

## What this changes

Propagate the already-computed client DPI from the xrdp front-end through
sesman and sesexec, and launch Xorg with `-dpi <client_dpi>` so the desktop
matches the login screen.

Data flow (no auth/PAM/identity changes):

```
xrdp_mm (compute+validate) → SCP → sesman → EICP → sesexec → Xorg argv "-dpi N"
```

- **`common/xrdp_client_info.{c,h}`** — new pure helper
  `xrdp_client_info_calculate_dpi(height_px, height_mm)` (the existing
  login-screen formula, extracted and made unit-testable, with overflow/
  underflow guards) plus `xrdp_client_info_dpi_valid_for_session()` enforcing a
  50–400 range. The login screen keeps using the **raw** value exactly as
  before — bounds are applied only on the new session path, so font scaling is
  unchanged.
- **`xrdp/xrdp_login_wnd.c`** — uses the helper; also guards an inverted-rectangle
  unsigned underflow in the height calculation.
- **`xrdp/xrdp_mm.c`** — for Xorg sessions, computes and range-checks the DPI and
  sends it in the SCP create-session request. Out-of-range/absent → sends 0.
- **SCP / EICP** (`libipm/scp.*`, `libipm/eicp.*`) — the create-session request
  gains one trailing DPI field; it is **re-validated on every receive** (sesman
  must not trust the value from xrdp; defense in depth). `0` means "no DPI".
- **`sesman/sesexec/session.c`** — appends `-dpi` `<n>` as two integer-only argv
  tokens (never a shell string) when the DPI is valid **and** the admin has not
  already set `-dpi` in `sesman.ini [Xorg]` (exact-token check; the admin value
  always wins, never a duplicate `-dpi`).
- Xvnc and other non-Xorg backends are untouched.

## Behavior / compatibility

- **No regression when DPI is absent/invalid/out-of-range:** no `-dpi` is
  appended and the session starts exactly as today (default 96 DPI). No
  `-dpi 0`, `-dpi 1`, or `-dpi 10000` can ever be emitted.
- **Admin override wins:** an existing `-dpi` in `[Xorg]` params is detected and
  the client value is skipped, so there is never a duplicate `-dpi`.
- **Wire-format change:** the SCP and EICP create-session messages
  gain a trailing field, so `LIBIPM_VERSION` is bumped **2 → 3**. libipm has no
  optional-field mechanism, so this is a breaking change by construction. xrdp,
  xrdp-sesman, xrdp-sesexec and sesrun ship and must be upgraded together; a
  mixed-version peer is rejected cleanly by the existing version check
  (fail-closed: failed session create, never a misparse or crash). The new field
  is appended at the end of the format strings to keep the change minimal.

## Scope / limitation: core DPI vs toolkit DPI

This sets the X server **core** DPI (via `-dpi`); desktops running in
**auto-DPI mode** (those that honor the core DPI, e.g. XFCE `Xft/DPI = -1`) pick
it up automatically. GTK/Qt take their font scale from `Xft.dpi` / XSETTINGS
`/Xft/DPI`, not the core DPI, so a desktop that pins a fixed toolkit DPI (e.g.
XFCE's default `96`) overrides `-dpi` and fonts stay at 96. Setting
`Xft.dpi`/XSETTINGS is per-user, desktop-specific session config and is
intentionally **out of scope** here. sesman logs a one-line reminder when it
applies a client DPI, so an admin whose fonts don't scale knows to switch the
desktop to auto-DPI.

## Design note: physical DPI

The DPI is derived from the client's **physical monitor size**
(TS_MONITOR_ATTRIBUTES, in mm) — the *same* computation xrdp already trusts for
the login screen. This deliberately makes the desktop consistent with the login
screen, which is the concrete defect in #3473. Using the RDP desktop *scale
factor* instead (e.g. 96×1.5) is a reasonable alternative policy; it's
intentionally out of scope here and could be added later behind a config knob
without changing this plumbing.

## Testing

- **Unit tests, pure helper** (`tests/common/test_xrdp_client_info.c`, 15 cases):
  the DPI formula incl. the truncation cases, zero/overflow/underflow → invalid,
  and the 50–400 range boundaries (e.g. 2160px/392mm → 139).
- **Semantic round-trip tests** (`tests/libipm/test_libipm_recv_calls.c`, 4
  cases): the SCP and EICP create-session messages are sent and received over
  the loopback link; a valid DPI survives intact (with the other fields
  unaffected by the appended field), and out-of-range/absent values are rejected
  to 0 on receive. These are value-based, not raw-byte assertions, so they don't
  become tripwires for future field additions.
- `make check` green across all suites (libcommon, libipm, libxrdp, daemon).
- astyle (pinned 3.4.14) and cppcheck clean.
- Built and smoke-tested against an Xorg/xorgxrdp session.

## Notes for reviewers

- Integer-only end to end: the value is a bounded `unsigned short`/`q` field all
  the way to the argv token, exec'd via `g_execvp_list` (no shell) — argument
  injection is not possible.
- The DPI is consumed **after** the setuid/setgid privilege drop in
  `start_x_server`; ordering is unchanged.
- Happy to split the `LIBIPM_VERSION` bump discussion out if you'd prefer a
  different compatibility strategy.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -81,6 +81,7 @@ libcommon_la_SOURCES = \
   trans.c \
   trans.h \
   unicode_defines.h \
+  xrdp_client_info.c \
   $(PIXMAN_SOURCES)
 
 libcommon_la_LIBADD = \
--- a/common/xrdp_client_info.c
+++ b/common/xrdp_client_info.c
@@ -1,0 +1,60 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
+ * Copyright (C) Jay Sorg 2004-2024
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @file common/xrdp_client_info.c
+ * @brief Pure helpers derived from client display information
+ */
+
+#if defined(HAVE_CONFIG_H)
+#include <config_ac.h>
+#endif
+
+#include <limits.h>
+
+#include "xrdp_client_info.h"
+
+/*****************************************************************************/
+unsigned int
+xrdp_client_info_calculate_dpi(unsigned int height_pixels,
+                               unsigned int height_mm)
+{
+    unsigned int dpi = 0;
+
+    /*
+     * DPI = height_pixels / (height_mm / 25.4)
+     *     = (height_pixels * 25.4) / height_mm
+     *     = (height_pixels * 127) / (height_mm * 5)
+     *
+     * Reject a missing physical size or pixel height, and guard the
+     * integer arithmetic against overflow before computing.
+     */
+    if (height_mm != 0 && height_pixels != 0 &&
+            height_pixels <= UINT_MAX / 127u &&
+            height_mm <= UINT_MAX / 5u)
+    {
+        dpi = (height_pixels * 127u) / (height_mm * 5u);
+    }
+
+    return dpi;
+}
+
+/*****************************************************************************/
+int
+xrdp_client_info_dpi_valid_for_session(unsigned int dpi)
+{
+    return (dpi >= XRDP_SESSION_DPI_MIN && dpi <= XRDP_SESSION_DPI_MAX);
+}
--- a/common/xrdp_client_info.h
+++ b/common/xrdp_client_info.h
@@ -287,4 +287,37 @@ struct xrdp_client_info
 #define OUTPUT_SUPPRESSED_FOR_REASON(ci,reason) \
     (((ci)->suppress_output_mask & (unsigned int)reason) != 0)
 
+/*
+ * Acceptable range for a session (desktop) DPI derived from client
+ * monitor metadata. The login screen uses the raw computed DPI; only the
+ * value propagated to the Xorg session is constrained to this range.
+ */
+#define XRDP_SESSION_DPI_MIN 50
+#define XRDP_SESSION_DPI_MAX 400
+
+/**
+ * Compute monitor DPI from a physical height.
+ *
+ * @param height_pixels Monitor/desktop height in pixels
+ * @param height_mm Physical monitor/desktop height in millimetres
+ * @return Computed DPI, or 0 if it cannot be determined (missing physical
+ *         size, zero pixels, or arithmetic overflow).
+ *
+ * This is the raw geometric DPI and is NOT clamped to any range. Callers
+ * needing a session-safe value must additionally check
+ * xrdp_client_info_dpi_valid_for_session().
+ */
+unsigned int
+xrdp_client_info_calculate_dpi(unsigned int height_pixels,
+                               unsigned int height_mm);
+
+/**
+ * Test whether a DPI value is within the acceptable session DPI range.
+ *
+ * @param dpi DPI value to test
+ * @return != 0 if XRDP_SESSION_DPI_MIN <= dpi <= XRDP_SESSION_DPI_MAX
+ */
+int
+xrdp_client_info_dpi_valid_for_session(unsigned int dpi);
+
 #endif
--- a/libipm/eicp.c
+++ b/libipm/eicp.c
@@ -32,6 +32,7 @@
 #include "guid.h"
 #include "os_calls.h"
 #include "trans.h"
+#include "xrdp_client_info.h"
 
 /*****************************************************************************/
 static const char *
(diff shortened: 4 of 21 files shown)


matt335672:
Hi@prwang

It's a nice idea, but it doesn't solve the general case, which is that a user can (and does) reconnect to the session from a different RDP client with a different resolution.

The DPI needs to be set on session connection rather than session creation. Sadly X11 isn't very good at handling this.

prwang:
Thanks @matt335672  — you're right, and digging into why changed my whole approach. Let me take your points in order.

**It doesn't solve the general case — confirmed.** I reproduced exactly the reconnect scenario you describe:

- Connect from a HiDPI client (fresh login) → correct. Log off, connect from a low‑DPI client → also correct.
- Connect from a HiDPI client (fresh login) → correct. **Disconnect** (leaving the session running), then reconnect from a low‑DPI client → **wrong**: the text stays large.

The split between those two is the tell. Log‑off destroys the session, so the next connection *creates* a new Xorg and the `-dpi` from this PR gets applied. Disconnect keeps the session alive, so reconnect *reuses* the running Xorg and nothing re‑applies the DPI — precisely your "on connection, not creation" point.

**The root issue is the layer this PR uses, not just the timing.** `-dpi` sets the X server's *core* DPI, and that value is fixed at server start — it can't be changed on a live server. So this approach is creation‑only by construction; there's no way to make it follow a reconnect short of restarting the session, which defeats the point of a persistent session. And the core DPI is the wrong target anyway: it's a single global value, and modern desktops largely ignore it — GTK/Qt/GNOME/KDE derive scaling from *per‑output* RandR physical size (plus a toolkit scale factor), and on Wayland there's no core DPI at all. So even when `-dpi` does take effect, it only moves desktops explicitly in auto‑DPI mode. That's the "X11 isn't very good at this" you were pointing at: there's no live, global DPI knob — the dynamic information lives per‑output, on RandR.

**So I'm going to close this PR** rather than leave a creation‑only half‑measure open where it could be merged by mistake.

**The route I'm now pursuing** mirrors how real hardware works. Instead of imposing a DPI number on the server, advertise the client monitor's true physical size (mm) on the X server's RandR output, and keep it in sync whenever the client's display changes — session create, reconnect, and mid‑session resize. The desktop then derives DPI from that the same way it does for an attached monitor, including across a hotplug/reconnect. Encouragingly, xrdp *already* forwards the client's physical monitor size to the backend on a mid‑session resize; the missing pieces are (a) carrying/refreshing that geometry on reconnect, and (b) having the Xorg backend (xorgxrdp) set the RandR output's physical size from it instead of reporting `0mm x 0mm`. It's a small cross‑repo change (xrdp + xorgxrdp), but on the standard, dynamic path rather than the legacy one.

One honest limitation, unchanged from this PR: a desktop that *pins* a fixed toolkit DPI (e.g. XFCE's default) overrides per‑output data and still won't scale — that's a DE‑config ceiling, not something xrdp should override. The gain is that this is the mechanism the rest of the desktop world already uses, and it handles reconnect and resize for the same reason a laptop handles a monitor hotplug.

Thanks again for the steer — it saved me from building on the wrong foundation. I'll open a fresh PR for the RandR‑based approach and link it back here; happy to sanity‑check the design with you first if you'd prefer.

PR #3821, opened 2026-07-02
Fix build after CVE-2026-41521

metalefty:


matt335672 (approved):
I was just looking at that - LGTM!

PR #3822, opened 2026-07-03
fixed #3820 - added initial fuzzing

### LINKED ISSUE #3820: add fuzzing
Taking a look at the latest batch of vulnerabilities it looks like most of these could have been caught by a fuzzer. Also given that this project is about handling a protocol that can easily be leveraged.

firewave:


firewave:
Obviously still a mess but the most basic stuff out of the way. Will elaborate what still needs to be done soon.

matt335672:
For command-line applications, look at using `log_config_init_for_console()`:

https://github.com/neutrinolabs/xrdp/blob/3af31df3fc18cb4f910524bbccfc90ff3189a757/common/log.h#L335-L351

Example: https://github.com/neutrinolabs/xrdp/blob/3af31df3fc18cb4f910524bbccfc90ff3189a757/sesman/tools/sesrun.c#L565-L571

Issue #3772, opened 2026-03-23
TCP connect : Re-use of socket when connecting.


matt335672:
See #3767 for the background on this.

In short, there's an issue with `g_tcp_connect()` potentially re-using a TCP socket which has failed to connect for a subsequent connect attempt.

The only consumer of `g_tcp_connect()` is `trans_connect()` in [common/trans.c](https://github.com/neutrinolabs/xrdp/blob/devel/common/trans.c). The pseudo code for this routine is (roughly) as follows:-

- While we don't have a connection, and the timeout hasn't happened.
  - Create a socket
  - Set the socket non-blocking
  - Call g_tcp_connect(). Return if successful. If a connect is 'in progress' wait for completion and return
  - Wait for a bit before going round again
- Return status to the caller.

The reason for the async stuff is so we can specify a timeout for `trans_connect()`. This design dates from a simpler time, where a name normally mapped to at most one IP address.

Within `g_tcp_connect()`, there's another connect loop with `getaddrinfo()` which works differently in that it re-uses the passed-in socket. All very well and good, apart from two things:

- The additional hidden complexity
- This statement from the [POSIX connect specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/connect.html)

> If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect.

When working on #3767, @Kropyls and I realised that POSIX doesn't offer a standardised way to get the address family for a socket. IMO, this is indicative that the current separation of responsibility between `trans_connect()` and `g_tcp_connect()` is flawed.

Any fix might want to consider bearing #2073 in mind. `getaddrinfo()` and `freeaddrinfo()` were first specified in IEEE Std 1003.1 (2004), and so we should be using these routines exclusively, and assuming that both IPv4 and IPv6 are available at compile time, although maybe both are not configured at runtime.

A possible solution might be to make g_tcp_connect() look a bit more like this. This is only a suggestion and has not been tested. This needs more discussion.

```c
enum tcp_conn_mode
{
    TCP_CONNECT_NO_PREF = 0,
    TCP_CONNECT_PREF_IPv4,
    TCP_CONNECT_PREF_IPv6
};

/***
 * Connect a TCP socket
 * @param host Hostname or address to connect to
 * @param port TCP port to connect to
 * @param async If non-zero, socket is to be marked non-blocking
 *                            before connect
 * @param mode Connection mode
 * @param[out] sck Socket number
[...]

PR #3826, opened 2026-07-09
Connect refactor for socket POSIX compliance

Kropyls:
addresses: #3772

Sorry I'm late on my time estimate I'd have this done by! There are some significant changes to connection attempts here, but it does get us to POSIX compliance I believe. I'm expecting at least some back and fourth before we all agree, and am always open to suggestions.

The basic idea was that the connect attempts need a way to recreate a socket gracfully when a connection fails, where the family stays consistent. In order to support this, I introduced an async_to (short for timeout) arg to *_connect(s) in os_calls. That allowed me to pull the async logic back into the os_calls.c. This gets me the ability for the *_connect(s) to attempt connections that have timeouts directly, which gives us the ability to recreate a socket on failure for reattempts, since it all happens inside the connect function now.

The sck_redo and mk_tcp_socket and 2 async functions are all just helper functions to enable this new behavior of connect. The vnc testing function and waitforx needed simple one-liner changes to comply with this new signature of the connect functions as well.

It had the added benefit of simplifying trans_connect a lot too, which I think is helpful because it basically just throws a timeout/is_term func into the os_calls code and gets the result. My thought is having the os calls manage the details of how socket-based connections are made makes sense.

I have tested this with a remote vnc server over IP6 and IPv4, with every iteration of IP address/compiled IP-versions I could do. I'm open to all positive and negative feedback on this, but hopefully this at least gets us started toward a POSIX-compliant solution.

More detail notes below (from the commit):
os_calls:
- add the following functions:
  - mk_tcp_socket:
    - creates socket and informs caller of socket family
  - sck_redo:
    - closes and reopens a socket
  - poll_for_async_connect:
    - pulled in from trans.c so tcp_connect can handle connection pass/fails itself for socket recreation
  - async_connect:
    - broken into new function from trans_connect in trans.c
- tcp_connect:
  - takes several new arguments so it can handle socket creation/connections/recreations as needed until connection is established (POSIX standard way)
  - async_to can be set to 0 to behave in sync manor
  - added some ipv6 only error handling
  - calls async_connect & sck_redo on connect failures to regen sockets
  - adjust formatting of ipv4 implementation to line-up with IPv6, both logic and structs used to navigate the code
  - adjust function signature in header file
- g_sck_local_connect:
  - adjust to match signature of tcp_connect
  - add async hooks inside trans:
- drop CONNECT_*_MS #defines and move the to os_calls
- local_connect_shim:
  - function signature to match tcp_connect
  - call g_tcp_local_socket inside function to make consistent with tcp_connect
- trans_connect:
  - drop all socket allocation logic (happens in connect functions now)
  - drop entire connection establishing loop because f_connect now has that logic internalized Also adjusted the tcp_proxy testing loop for changes and locl_connect call in waitforx.c


Kropyls:
sorry I missed the formatting on the first pass - thought I had it all :/

matt335672:
Thanks for this @Kropyls 

It'll take me a while to get to it I'm afraid - we're snowed under with security work at the moment.

Kropyls:
no worries @matt335672 I'll be here whenever you're ready!

Issue #1920, opened 2021-06-10
Consolidate channel chunking / dechunking code


matt335672:
Large static and dynamic virtual channel PDUs are split up into 'chunks' by the RDP protocol. This process is described for static channels in in [MS-RDPBCGR] [1.3.3](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/343e4888-4c48-4054-b0e3-4e0762d1993c) and [3.1.5.2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/528c0de3-922a-4c13-a450-4f41834fb5c9), and for dynamic channels in [[MS-RDPEDYC]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedyc/3bd53020-9b64-4c9a-97fc-90a79e7e1e06)

Code to chunk and de-chunk PDUs is scattered over the whole system. The following places have been found so far:-
1) The VNC module clipboard handler, used if chansrv isn't running.
2) Incoming chunks on the drdynvc static virtual channel are re-assembled in libxrdp/xrdp_channel.c before being passed to chansrv. Note that these chunks after reassembly are still small and may themselves require re-assembly as specified in [MS-RDPEDYC].
3) The routine send_channel_data() in chansrv.c is used to chunk messages to static virtual channels.
4) Various chansrv input routines (clipboard_data_in(), read_entire_packet(), devredir_data_in() dechunk data coming in.

There are almost certainly other places which are difficult to find due to the lack of use of symbolic names for the package message flags.

This issue suggests that a common chunking/dechunking facility is provided in a new or existing shared library. This facility can be exercised in the test suite

PR #3829, opened 2026-07-15
Fix for DYNVC Multi-Chunk Reassembly Logic Defects

the-deniss:
# DYNVC Multi-Chunk Reassembly Logic Defects

## Header

| Field | Value |
|-------|-------|
| Report ID | XRDP-2026-036 |
| Status | Informational |
| Date | 2026-06-29 |
| Audited Version | xrdp 0.10.6 |
| Component | xrdp |
| Function | `xrdp_channel_process_drdynvc()` |
| Lines | 611–651 |
| Vulnerability Type | CWE-670: Always-Incorrect Control Flow Implementation (2 findings) |
| Category | Pre-Auth Remote |
| CVSS 3.1 Score | N/A — Informational |
| CVSS 3.1 Vector | N/A — Informational |
| Reachability | Network-exposed, affects multi-chunk DYNVC messages |

---

## Summary

Two related pointer-state defects in the DYNVC multi-chunk reassembly path prevent proper parsing of multi-fragment messages. These defects act as accidental guards preventing deeper parsing vulnerabilities from being reached, but the root causes are independent logic bugs that must be corrected.

The two defects compound each other: together they prevent sub-parser dispatch from working for any multi-chunk DYNVC message. The bugs accidentally mask other vulnerabilities that would become reachable if either defect is fixed in isolation without fixing the other.

---

## Finding 1 — Missing `s_mark_end` and Pointer Reset Before Stream Reassignment

**Function:** `xrdp_channel_process_drdynvc()` | **Lines:** 611–613

### Vulnerable Code

```c
case 2: /* CHANNEL_FLAG_LAST */
    // ...
    out_uint8a(self->s, s->p, length);  // writes to reassembly buffer
    // ...
    break;
// (later, after all chunks received)
case 3: /* CHANNEL_FLAG_FIRST | CHANNEL_FLAG_LAST */
case 2:
    ls = self->s;   // line 613
```

### Impact

`init_stream(s, n)` initializes both `s->p = s->data` (read cursor) and `s->end = s->data` (end marker), both pointing to the beginning. When `out_uint8a()` writes chunks to the reassembly buffer, it advances `s->p` but never updates `s->end`.

After all fragments are written:

```c
// self->s->p   = self->s->data + total_written   ← points past all written data
// self->s->end = self->s->data                   ← never moved; points to beginning
```

Setting `ls = self->s` then causes all subsequent stream reads to fail bounds checks (`s_check_rem()` compares `s->p + n <= s->end`, which evaluates false since `s->p > s->end`). All multi-PDU DYNVC messages are silently discarded after reassembly, preventing the protocol from functioning correctly.

### Recommended Fix

Before setting `ls`:

```c
s_mark_end(self->s);        // sets self->s->end = self->s->p (marks all data readable)
self->s->p = self->s->data; // resets read cursor to beginning
ls = self->s;
```

---

## Finding 2 — Sub-Parsers Receive Drained Original Stream Instead of Reassembly Buffer

**Function:** `xrdp_channel_process_drdynvc()` | **Lines:** 637–651

### Vulnerable Code

```c
case CMD_DVC_DATA_FIRST:
    rv = drdynvc_process_data_first(self, cmd, s);  // ← s, not ls
    break;
case CMD_DVC_DATA:
    rv = drdynvc_process_data(self, cmd, s);       // ← s, not ls
    break;
case CMD_DVC_CLOSE_REQUEST:
    rv = drdynvc_process_close_request(self, cmd, s);  // ← s, not ls
    break;
case CMD_DVC_CREATE_REQUEST:
    rv = drdynvc_process_create_request(self, cmd, s); // ← s, not ls
    break;
case CMD_DVC_CREATE_RESPONSE:
    rv = drdynvc_process_create_response(self, cmd, s); // ← s, not ls
    break;
```

### Impact

All five dispatch arms pass `s` (the original input stream) to sub-parsers instead of `ls` (the reassembly buffer or original stream, depending on the case).

**For case 3 (`CHANNEL_FLAG_FIRST | CHANNEL_FLAG_LAST` — single-chunk):** Both are the same pointer; harmless.

**For case 2 (`CHANNEL_FLAG_LAST` — last chunk of multi-chunk message):** At line 612, the original stream `s` is fully consumed by `in_uint8s(s, length)`, leaving `s->p == s->end`. Sub-functions then receive an empty stream and immediately fail their first `s_check_rem()` call, returning error (1). As a result, all multi-PDU DYNVC messages are silently discarded after reassembly completes.

Combined with Finding 1, the two defects prevent multi-chunk DYNVC parsing from reaching any sub-parser. This is an accidental guard that masks other bugs, but is fragile: fixing one defect without the other may expose deeper vulnerabilities.

### Recommended Fix

All dispatch arms should pass `ls`:

```c
case CMD_DVC_DATA_FIRST:
    rv = drdynvc_process_data_first(self, cmd, ls);
    break;
// ... etc for all five cases
```

---

### Defect Summary

| Defect | Location | Symptom | Fix |
|--------|----------|---------|-----|
| 1 | Lines 611–613 | `self->s->end` never updated after writes | Add `s_mark_end(self->s)` and reset `self->s->p` |
| 2 | Lines 637–651 | Sub-parsers receive drained `s` instead of `ls` | Change all dispatch arms to pass `ls` |

---

**Disclaimer:** Generated using an LLM-assisted vulnerability research pipeline. Reviewed by Denis Skvortsov

### DIFF
--- a/libxrdp/xrdp_channel.c
+++ b/libxrdp/xrdp_channel.c
@@ -18,6 +18,7 @@
  * channel layer
  */
 
+#include "parse.h"
 #if defined(HAVE_CONFIG_H)
 #include <config_ac.h>
 #endif
@@ -610,6 +611,8 @@ xrdp_channel_process_drdynvc(struct xrdp_channel *self,
             }
             out_uint8a(self->s, s->p, length); /* append data to chunk buffer */
             in_uint8s(s, length);              /* virtualChannelData */
+            s_mark_end(self->s);
+            self->s->p = self->s->data;
             ls = self->s;
             break;
         case 3: /* CHANNEL_FLAG_FIRST and CHANNEL_FLAG_LAST */
@@ -635,19 +638,19 @@ xrdp_channel_process_drdynvc(struct xrdp_channel *self,
     switch (cmd & 0xf0)
     {
         case CMD_DVC_CAPABILITY:
-            rv = drdynvc_process_capability_response(self, cmd, s);
+            rv = drdynvc_process_capability_response(self, cmd, ls);
             break;
         case CMD_DVC_OPEN_CHANNEL:
-            rv = drdynvc_process_open_channel_response(self, cmd, s);
+            rv = drdynvc_process_open_channel_response(self, cmd, ls);
             break;
         case CMD_DVC_CLOSE_CHANNEL:
-            rv = drdynvc_process_close_channel_response(self, cmd, s);
+            rv = drdynvc_process_close_channel_response(self, cmd, ls);
             break;
         case CMD_DVC_DATA_FIRST:
-            rv = drdynvc_process_data_first(self, cmd, s);
+            rv = drdynvc_process_data_first(self, cmd, ls);
             break;
         case CMD_DVC_DATA:
-            rv = drdynvc_process_data(self, cmd, s);
+            rv = drdynvc_process_data(self, cmd, ls);
             break;
         default:
             LOG(LOG_LEVEL_ERROR, "Received header [MS-RDPEDYC] with "
--- a/libxrdp/xrdp_channel.c
+++ b/libxrdp/xrdp_channel.c
@@ -18,12 +18,12 @@
  * channel layer
  */
 
-#include "parse.h"
 #if defined(HAVE_CONFIG_H)
 #include <config_ac.h>
 #endif
 
 #include "libxrdp.h"
+#include "parse.h"
 #include "string_calls.h"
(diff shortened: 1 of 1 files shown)


matt335672:
Nice find @the-deniss - I'll take this.

You won't be surprised to learn I'm looking at chunk reassembly in general at the moment (see also #1920), so this area will be reworked shortly. I'll ask you to look at the changes when I've got something worth posting.

PR #3832, opened 2026-07-17
[V0.10] Add support for strlcpy()

matt335672:
Too many places in xrdp use strncpy() to copy strings to fixed-length buffers, when this is not the correct function to use.

This PR makes sure strlcpy() from the BSDs is available as a saner alternative. This function is available by default on Linux and FreeBSD.

This is cherry picked from commit 960ea7ce0521dc417709ff3deabf865a45030a4e, wjich has been running in `devel` for some time now.

### DIFF
--- a/common/string_calls.c
+++ b/common/string_calls.c
@@ -15,6 +15,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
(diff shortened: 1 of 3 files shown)


metalefty:
Good idea.

PR #3837, opened 2026-07-22
Fix for trans_force_read_s Bounds Check Uses Wrong Pointer

the-deniss:
# trans_force_read_s Bounds Check Uses Wrong Pointer (Write-Origin vs Check-Origin)

## Header

| Field | Value |
|-------|-------|
| Report ID | XRDP-2026-015 |
| Status | Informational |
| Date | 2026-06-29 |
| Audited Version | xrdp 0.10.6 |
| Component | xrdp |
| Function | `trans_force_read_s()` |
| Lines | 468–484 |
| Vulnerability Type | CWE-119: Improper Restriction of Operations within Bounds of Memory Buffer |
| Category | Pre-Auth Remote (via any caller without compensating `bytes <= s->size` guard) |
| CVSS 3.1 Score | N/A — Informational |
| CVSS 3.1 Vector | N/A — Informational |
| Reachability | Pre-auth via `xrdp_mcs_incoming`; not exploitable in that specific call chain due to compensating guard. Audited every other caller in the current tree (VNC backend, chansrv audio/PCSC handlers, clipboard) — each independently maintains the `p == end` invariant (via `init_stream` reset or full `in_uint*` consumption of the prior read), so no exploitable call chain currently exists. See "Caller Audit" below. |

---

### Root Cause

`trans_force_read_s` writes incoming data starting at `in_s->end` but its bounds check is computed from `in_s->p`:

```
check:  in_s->p + size <= in_s->data + in_s->size
write:  in_s->end, in_s->end + rcvd, ...
```

The effective safe condition should be `in_s->end + size <= in_s->data + in_s->size`. The discrepancy is `end - p` bytes: in any caller that first reads a header (advancing `end`) and then calls `trans_force_read_s` for the body, the body write will extend `end - p` bytes beyond what the check allowed — into memory past the stream buffer.

### Vulnerable Code

```c
/* common/trans.c:468 */
int
trans_force_read_s(struct trans *self, struct stream *in_s, int size)
{
    if (self->status != TRANS_STATUS_UP ||
            size < 0 || !s_check_rem_out(in_s, size))  /* checks: p + size <= data + s->size */
    {
        return 1;
    }
    while (size > 0)
    {
        rcvd = self->trans_recv(self, in_s->end, size); /* writes at end, NOT p! */
        if (rcvd <= 0)
        {
            self->status = TRANS_STATUS_DOWN;
            return 1;
        }
        in_s->end += rcvd;
        size      -= rcvd;
    }
    return 0;
}
```

`s_check_rem_out(in_s, size)` expands to `(in_s->p + size <= in_s->data + in_s->size)`. After any `trans_force_read_s` call that advanced `end > p`, the next call's check is off by `end - p` bytes.

### Impact

- **Heap buffer overflow (conditional):** A caller that reads a protocol header first (advancing `in_s->end`) without consuming it via `in_uint*` (leaving `p` behind `end`), and then reads the body with `trans_force_read_s`, would write `end − p` bytes past the end of the stream buffer, without any error. No caller currently in the tree does this — see "Caller Audit" below — so there is no live overflow today.
- **Latent footgun, not a hardened API:** `trans_force_read_s` trusts every caller to keep `p == end` at call time; it does not enforce this itself. Any new caller (or future refactor of an existing one) that reads a header without immediately consuming it via `in_uint*`, then issues a second `trans_force_read_s` call, would reintroduce the overflow with no compiler or runtime check catching it. The fix below removes this trust requirement entirely by checking from the actual write origin (`end`), and should be applied defensively even though no exploitable path exists today.

### Exploitation / Attack Path

**In `xrdp_mcs_incoming` (compensated, not exploitable):**
- `libxrdp_force_read` has a compensating guard: `if (bytes < 4 || bytes > s->size) return NULL`.
- After `trans_force_read(trans, 4)` sets `end = data + 4`, the body call `trans_force_read(trans, bytes-4)` reaches at most `data + bytes ≤ data + s->size`. Not exploitable here.

**Hypothetical exploitation pattern (not currently reachable by any caller in the tree — see Caller Audit):**
1. A caller allocates a stream with `init_stream(s, HEADER_SIZE + BODY_SIZE)`.
2. Calls `trans_force_read_s(t, s, HEADER_SIZE)` → `end = data + HEADER_SIZE`, `p` still at `data` (i.e. the caller does *not* consume the header via `in_uint*` before the next call).
3. Calls `trans_force_read_s(t, s, BODY_SIZE)` → check: `p + BODY_SIZE <= data + s->size` → passes.
4. Write: `end + BODY_SIZE = data + HEADER_SIZE + BODY_SIZE` — potentially past `data + s->size` if `s->size < HEADER_SIZE + BODY_SIZE`.

This requires step 2's precondition (`p` left behind `end`), which no current caller exhibits.

### Caller Audit

The safety of every `trans_force_read_s`/`trans_force_read` call site depends on one prerequisite holding at call time:

> **`in_s->p == in_s->end`** — guaranteed either by a fresh `init_stream()` call (which resets both `p` and `end` to `data`) immediately beforehand, or by the previous read's bytes having been fully consumed via `in_uint*` macros (which advance `p`) so `p` catches back up to `end`.

Every caller in the current tree was traced and satisfies this:

| Caller | Guarantee |
|---|---|
| `libxrdp/libxrdp.c` `libxrdp_force_read` | `init_stream` reset + explicit `bytes > s->size` guard |
| `vnc/vnc.c` `read_extended_desktop_size_rect` | Header fully consumed (`in_uint8`+`in_uint8s(3)`) before 2nd read |
| `vnc/vnc.c` rect-matching loop | Fresh `init_stream` every loop iteration |
| `vnc/vnc.c` `skip_trans_bytes` | Fresh `init_stream` every loop iteration |
| `vnc/vnc.c` `lib_framebuffer_update` (RAW/COPY_RECT/CURSOR) | Fresh `init_stream` per sub-read (often a distinct stream, e.g. `pixel_s`) |
| `vnc/vnc.c` `lib_palette_update` | Header fully consumed; explicit `init_stream` again before body read |
| `vnc/vnc.c` `get_reason_string` | Header (4 bytes) fully consumed via `in_uint32_be` |
| `vnc/vnc.c` `negotiate_protocol_version` | Each read fully consumed via `in_uint*` before the next `init_stream`+read |
| `vnc/vnc.c` `choose_security_type` | Single read only |
| `vnc/vnc.c` VNC-auth challenge / SecurityResult | Separate `init_stream` per read |
| `vnc/vnc.c` `receive_server_init` | Header (24 bytes) fully consumed, then explicit `init_stream` again before name read |
| `vnc/vnc_clip.c` `vnc_clip_process_rfb_data` | Header (7 bytes) fully consumed; 2nd read targets a different, freshly-`init_stream`'d stream |
| `sesman/chansrv/chansrv.c` `my_trans_data_in` | `header_size=8` consumed exactly (`in_uint8s(4)+in_uint32_le(4)`) |
| `sesman/chansrv/smartcard_pcsc.c` `my_pcsc_trans_data_in` | `header_size=8` consumed exactly (`in_uint32_le`×2) |
| `sesman/chansrv/sound.c` `sound_sndsrvr_sink_data_in` | `header_size=8` consumed exactly (`in_uint32_le`×2) |
| `sesman/chansrv/sound.c` `sound_sndsrvr_source_data_in` | Only one `trans_force_read` call; further data goes to a separate, freshly-allocated stream never re-read via `trans_force_read` |

For the three chansrv-family callers, the guarantee additionally depends on `header_size` (set once at connection setup, always `8` for these) matching the number of bytes the callback consumes from the header via `in_uint*` before the body read — a source-level convention enforced by nothing but code review.

---

**Disclaimer:** Generated using an LLM-assisted vulnerability research pipeline. Reviewed by Denis Skvortsov


matt335672:
Code formatting check has located a trailing space ' ' on common/trans.c:474

Agreed the current test is wrong, but suggest the following instead:

```c
size > in_s->size - (in_s->end - in_s->data)
```

I believe the existing definition can possibly invoke UB if `in_s->end + size` points beyond the allocated area ([C99 6.5.6(8)](https://port70.net/%7Ensz/c/c99/n1256.html#6.5.6p8)). This is avoided by starting with the pointer subtraction - `in_s->end` should always be >= `in_s->data`.

Similarly we should have (IMO):

```c
#define s_check_rem(s, n) ((n) <= ((s)->end - (s)->p))
. . .
#define s_check_rem_out(s,n) ((n) <= ((s)->size - ((s)->p - (s)->data)))
```

What do you think?

the-deniss:
@matt335672 Yes, I agree - I've pushed hardened version of the bounds check

matt335672:
Great - thanks.

One other thing related to e4f4364c9b2e569053325353254c8e2c2c53a033 which I must admit I hadn't fully twigged. Rather than the fix you've added, how about:

```c
#define s_check_rem(s, n) ((unsigned int)(n) <= (unsigned int)((s)->end - (s)->p))
.
.
.
#define s_check_rem_out(s, n) ((unsigned int)(n) <= (unsigned int )((s)->size - ((s)->p - (s)->data)))
```

That would guard against a computed `n` which for some reason is < 0 from passing the test unexpectedly.

the-deniss:
Yes, you're right again - casting to an unsigned type avoids negative offset issues. I considered it earlier, but a quick cast to `size_t` introduced a warning in `xrdp_mcs.c`, so I kept it simple out of laziness. Since it matters to you, now I went ahead with the cast, tracked down the warning, and removed the dead check that was causing both the warning and a compilation error

matt335672:
Thanks - I'll merge this now.

PR #3838, opened 2026-07-23
[V0.10] Add dechunker module

matt335672:
Prompted by (among other things) #1920 

[MS-RDPBCGR] section [3.1.5.2](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/528c0de3-922a-4c13-a450-4f41834fb5c9) contains a description of how large PDUs on virtual channels are split into smaller PDUs or 'chunks'.

We have a lot of code in xrdp to handle dechunking, but this is spread all over the place. It also contains some common deficiencies.

This PR adds a virtual channel dechunker to the common library. A test suite is also provided.

The dechunker is in libcommon rather than libxrdp as it is required in places which do not currently link to libxrdp (e.g. the VNC module and chansrv).

As an example, the dechunker is retro-fitted to the `drdynvc` channel in xdp in which deficiencies were recently discovered by @the-deniss (#3829 #3831). I've informally tested the dechunker on other virtual channels, and PRs for these will be added in due course.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -43,6 +43,8 @@ libcommon_la_SOURCES = \
   base64.c \
   channel_defs.h \
   defines.h \
+  dechunker.c \
+  dechunker.h \
   fifo.c \
   fifo.h \
   file.c \
--- a/common/dechunker.c
+++ b/common/dechunker.c
@@ -1,0 +1,371 @@
+/**
+ * xrdp: A Remote Desktop Protocol server.
+ *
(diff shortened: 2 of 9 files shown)


matt335672:
Just found some code quality issues when forward porting to `devel`, largely around the added compiler checks @firewave has been working on. I've rebased the fixes for these issues into this PR.

I should have really started with `devel` and back-ported to `v0.10`.

PR #3851, opened 2026-08-19
chansrv: handle the PRIMARY selection, not just CLIPBOARD

smooker:
X has two selections: CLIPBOARD, set by an explicit copy, and PRIMARY, set by dragging the mouse over text and pasted with the middle button or shift+insert. RDP has a single clipboard, and chansrv mapped it onto CLIPBOARD only -- g_primary_atom was interned at startup and then never used again, which is still true upstream (checked a02ed62, 2026-08-14).

The effect in a session: selecting text with the mouse produced nothing the client could see, and text copied on the client could only be pasted with ctrl+v. Middle-click and shift+insert did nothing at all, which reads as a broken session rather than an unimplemented feature.

Four changes, all on paths CLIPBOARD already used:

- XFixesSelectSelectionInput also subscribes to PRIMARY, so ownership changes on it are delivered.
- clipboard_event_selection_owner_notify converts from lxevent->selection rather than a hardcoded g_clipboard_atom. This is the one that made mouse selection invisible: the event arrived and the code then asked CLIPBOARD, which had not changed.
- clipboard_set_selection_owner takes PRIMARY as well as CLIPBOARD, so data arriving from the client can be pasted with the middle button. Failing to take it is a warning, not an error: CLIPBOARD is already held and behaviour is unchanged from before.
- clip_s2c gains a `selection` field recording which selection advertised the data, so the follow-up XConvertSelection asks that same owner. Asking CLIPBOARD for something PRIMARY offered returns whatever CLIPBOARD happens to hold, or nothing.

clipboard_event_selection_request needed no change: it already answers whatever selection is asked of it.

Tested on Gentoo with xrdp built from devel and the xorgxrdp backend: both directions work -- text selected with the mouse in the session reaches the client, and ctrl+c on the client pastes with the middle button in the session.

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -412,6 +412,18 @@ clipboard_init(void)
                                    XFixesSetSelectionOwnerNotifyMask |
                                    XFixesSelectionWindowDestroyNotifyMask |
                                    XFixesSelectionClientCloseNotifyMask);
+        /* Also watch PRIMARY, the selection filled by dragging the mouse over
+         * text and pasted with the middle button or shift+insert. RDP has a
+         * single clipboard, so it maps onto CLIPBOARD and PRIMARY was left
+         * unhandled -- g_primary_atom was interned and then never used. A
+         * user selecting text in an RDP session had nothing to paste with the
+         * middle button, and text copied on the client could not be pasted
+         * that way either. */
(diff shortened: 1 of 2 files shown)


matt335672:
@smooker - I can see wher you're coming from, but I think there's a lot of scope for confusion if this is merged. The existing behaviour is by design and will work the same when we (eventually) get round to getting Wayland support to xrdp.

PR #3852, opened 2026-08-21
oscalls: vsock errors should be logged as errors

gpotter2:
2 errors in `g_sck_vsock_socket` are marked using "DEBUG" logs instead of "ERROR". Took me a second to debug 😅


matt335672:
Thanks @gpotter2

gpotter2:
You're welcome. Have a good weekend

PR #3854, opened 2026-08-22
gfx bulk compressor

jsorg71:
Bulk compression from MS-RDPEGFX 3.1.9.1.1

### DIFF
--- a/xrdp/xrdp_tconfig.c
+++ b/xrdp/xrdp_tconfig.c
@@ -546,6 +546,11 @@ tconfig_load_gfx(const char *filename, struct xrdp_tconfig_gfx *config)
     toml_table_t *tfile;
     int rv = 0;
 
+    if (filename == NULL || config == NULL)
+    {
+        TCLOG(LOG_LEVEL_ERROR, "Error loading GFX config, bad params");
+        return 1;
+    }
     /* Default to just RFX support. in case we can't load anything */
     config->codec.codec_count = 1;
     config->codec.codecs[0] = XTC_RFX;
--- a/xrdp/xrdp_tconfig.c
+++ b/xrdp/xrdp_tconfig.c
@@ -482,6 +482,19 @@ static int tconfig_load_gfx_order(toml_table_t *tfile, struct xrdp_tconfig_gfx *
     return 0;
 }
 
+static int tconfig_load_gfx_bulk(toml_table_t *tfile, struct xrdp_tconfig_gfx *config)
+{
+    toml_table_t *codec;
+    toml_datum_t datum;
+
+    if ((codec = toml_table_in(tfile, "codec")) != NULL)
+    {
+        datum = toml_bool_in(codec, "bulk");
+        config->bulk = datum.ok ? datum.u.b : 0;
+    }
+    return 0;
+}
+
 /**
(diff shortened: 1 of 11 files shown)


xrdp/xrdp_tconfig.c
```diff
@@ -482,6 +482,19 @@ static int tconfig_load_gfx_order(toml_table_t *tfile, struct xrdp_tconfig_gfx *
...
 
+static int tconfig_load_gfx_bulk(toml_table_t *tfile, struct xrdp_tconfig_gfx *config)
+{
+    toml_table_t *codec;
+    toml_datum_t datum;
+
+    if ((codec = toml_table_in(tfile, "codec")) != NULL)
+    {
+        datum = toml_bool_in(codec, "bulk");
+        config->bulk = datum.ok ? datum.u.b : 0;
```
matt335672, line 493: Small point - might want to check for config being NULL before the assignment.

matt335672:
Nice job.

Another small point - I haven't looked at libbulk in any detail yet, but can you make the main branch `devel` like everything else?

xrdp/xrdp_tconfig.c
jsorg71: I added a check a function up, the other functions didn't have a check either.  Now they should all be protected.

jsorg71:
oh this reminds me, @metalefty in the file gfx.toml, I added a way to turn off the gfx bulk compressor.
Now the codec section looks like.
```
[codec]
order = [ "H.264", "RFX" ]
bulk = true
```
Does this look ok?
If so, I can edit the man page.

metalefty:
Let me check. I'm still not sure if it is reasonable to put it in `codec` section.

jsorg71:
> Let me check. I'm still not sure if it is reasonable to put it in `codec` section.

I just put it in the easiest place.  Maybe a "options" section or "global" is better. 
Let me know what you think.

metalefty:
As far as I understand, bulk compression compresses all data in `Microsoft::Windows::RDS::Graphics` dynamic virual channel, including H.264 stream. So bulk compresison is performed at a layer under the codec. That sounds like compression at the layer that transports the codec to me.

@matt335672 @jsorg71 What about like this?

```toml
[transport]
bulk_compression = true

[codec]
order = [ "H.264", "RFX" ]
```

matt335672:
Agreed about the bulk compression, but is supporting bulk compression of H.264 data a good idea?

metalefty:
I think we can mix compressed data and uncompressed data in the virtual channel so we should skip compression for H.264 stream. 

BTW, metadata, header, or control commands such as `RDPGFX_START_FRAME_PDU` or `RFX_AVC420_METABLOCK` in GFX channel would be effective to compress.

jsorg71:
To my surprise, x264 data is compressing 2x.  nvenc I see about 1.2x.  To be fair with x264 we do set it to be easy on the cpu.
These are just some early numbers, I'll get more numbers with just the h264 part.

jsorg71:
If we want to allow fine grain control over the bulk, maybe we should have a bulk section.
```
[bulk]
enabled = true
skip_codec = [ "H.264", "RFX" ]

[codec]
order = [ "H.264", "RFX" ]
```

PR #3567, opened 2025-07-14
Close on reconnect

matt335672:
Fixes #3543 
Fixes #1188 
Helps with #2765 

The draft PR builds on #3547. It uses a new link between sesexec and xrdp to allow sesexec to request xrdp to exit if another client wishes to connect to the same session.

So, if I start a session with:-

```
xfreerdp /v:<host> /u:<user> /p:<password> /tune:FreeRDP_SupportErrorInfoPdu:true /drive:C,/tmp
```

and then connect to the same session from elsewhere with the same command, the first command is terminated with these errors (on FreeRDP version 3.16.0):-

```
[19:54:02:528] [16973:0000424f] [INFO][com.freerdp.core] - [rdp_print_errinfo]: ERRINFO_DISCONNECTED_BY_OTHER_CONNECTION (0x00000005):Another user connected to the server, forcing the disconnection of the current connection.
[19:54:02:529] [16973:0000424f] [ERROR][com.freerdp.core] - [rdp_set_error_info]: ERRINFO_DISCONNECTED_BY_OTHER_CONNECTION [0x00010005]
[19:54:02:529] [16973:0000424f] [ERROR][com.freerdp.core.rdp] - [rdp_recv_callback_int][0x6404001dc8c0]: CONNECTION_STATE_CAPABILITIES_EXCHANGE_DEMAND_ACTIVE status STATE_RUN_QUIT_SESSION [-2]
[19:54:02:529] [16973:0000424f] [ERROR][com.freerdp.core.transport] - [transport_check_fds]: transport_check_fds: transport->ReceiveCallback() - STATE_RUN_QUIT_SESSION [-2]
[19:54:02:531] [16973:00004287] [WARN][com.freerdp.channels.drdynvc.client] - [check_open_close_receive]: {Microsoft::Windows::RDS::DisplayControl:2} OnOpen=(nil), OnClose=0x733a5bca83eb
```

The session which is now connected can reconnect to chansrv and so copy-paste and drive redirections work as expected.

I've kept this draft for now, as I'm still testing, and also:-
1) I want to get some info regarding the connection available in the xrdp-sesadmin output
2) I need to document the new messages

When this is done, I'm likely to merge #3547, followed by this one.

PR #3856, opened 2026-08-25
Support for Windows like env CLIENTNAME, Session Reconnect Policy Flag H

pp-sepag:
Extended xrdp sesman so the client hostname reported by the RDP client is exposed as a CLIENTNAME environment variable in the session, and added a new session-reconnect policy flag H (e.g. Policy=UBH in sesman.ini) that matches on that client hostname ('H') in addition to user and bits-per-pixel.

This is an alternate implementation for #2099. It uses the client information introduced in #3567, especially in commit https://github.com/neutrinolabs/xrdp/commit/cd98b01

Please feel free to comment. Thank you!


matt335672:
Hi @pp-sepag 

Thanks for this.

The biggest reservation I have about the CLIENTNAME environment variable is the one I made in [this comment](https://github.com/neutrinolabs/xrdp/pull/2099#issuecomment-1002973465). Once the session is created, the client name is fixed. There may also be no client name if the session was created with `xrdp-sesrun`. In other words, we need a dynamic solution, not one based on environment variable.

pp-sepag:
Hi @matt335672 

I'm not familiar with `xrdp-sesrun`. Is this tool used only for testing or productive use?
The idea of the `CLIENTNAME` environment variable in combination with this exact reconnect mechanism is indeed to spawn a single session for each client by its name. So it is intended, that the environment variable is fixed at the client's first connect.
All major Desktop systems (Windows, macOS) pass their local name to the RDP server. But you are right, one should think about how to handle empty names. Maybe a DNS lookup of the client's IP address?
With dynamic solution, do you mean to access the client name via a tool like `xrdp-sesadmin`?

metalefty:
> With dynamic solution, do you mean to access the client name via a tool like xrdp-sesadmin?

I assume something like D-Bus.

matt335672:
`xrdp-sesrun` starts a session without an active connection. It has some productive use-cases.

The plan I have is to update `xrdp-sesadmin` so that the user (or sysadmin) can query the system to obtain information about the session, including the currently connected client (if any) and the connection status.

I'm still not happy with `CLIENTNAME` being set by default when the session is started, as the scope for confusion is too great - I accept that this wouldn't be a problem in your use-case. However, I'll be the one having to triage the inevitable series of issues we get raised about this!

Another alternative might be to set `CLIENTNAME` for the reconnection script. This would always be valid, and we could broaden the mechanism out into something more generic to allow disconnects to be handled too.

pp-sepag:
I picked up you suggestion and replaced the static `CLIENTNAME` env variable by the helper tool `xrdp-sesadmin`. The syntax would be `xrdp-sesadmin -c=clientname:<sid>` where the current sid can be taken from `$XRDP_SESSION`.
Are you happy with this solution? Maybe this PR need still some cleanup (some leftover comments).
Do you want me to extend `xrdp-sesrun` with an additional additional argument to pass a specific clientname?

pp-sepag:
If desired, the `CLIENTNAME` env variable can be set via a system-wide mechanism for a new (login/non-login interactive) shell.

```
sudo tee /etc/profile.d/xrdp-clientname.sh <<'EOF'
# xrdp: expose the RDP client's computer name as $CLIENTNAME.
#
# Queried fresh on every new shell rather than set once at session
# start, so it reflects whichever client is *currently* connected.
# Already-open shells keep whatever value they captured when they
# started - nothing can update the environment of a running process.
#
# Sourced here for login shells; for non-login interactive shells
# (most GUI terminals) see the matching hook in /etc/bash.bashrc.

if [ -n "$XRDP_SESSION" ]; then
    XRDP_SESADMIN=$(command -v xrdp-sesadmin 2>/dev/null)
    if [ -z "$XRDP_SESADMIN" ]; then
        # Not on PATH (common for /sbin dirs) - check typical install prefixes
        for candidate in \
            /usr/local/bin/xrdp-sesadmin \
            /usr/bin/xrdp-sesadmin \
            /usr/local/sbin/xrdp-sesadmin \
            /usr/sbin/xrdp-sesadmin \
            /opt/xrdp/bin/xrdp-sesadmin
        do
            if [ -x "$candidate" ]; then
                XRDP_SESADMIN="$candidate"
                break
            fi
        done
    fi

    if [ -n "$XRDP_SESADMIN" ]; then
        export CLIENTNAME=$("$XRDP_SESADMIN" -c=clientname:"$XRDP_SESSION")
    fi
fi
EOF
```
and
```
sudo tee -a /etc/bash.bashrc <<'EOF'

# xrdp: source CLIENTNAME lookup for non-login interactive shells
# (GUI terminals typically don't start a login shell, so /etc/profile.d/
# is never reached without this). See /etc/profile.d/xrdp-clientname.sh
[ -r /etc/profile.d/xrdp-clientname.sh ] && . /etc/profile.d/xrdp-clientname.sh
EOF
```

@matt335672 What do you think about this solution?

matt335672:
Hi @pp-sepag 

Thanks for this.

It will take me a while to get to review your code, as we're buried under security advisories at the moment (of varying quality), and those have to take a priority over new functionality. I've had a quick look though. I could be wrong, but there doesn't seem to be any code to handle a disconnect, so that the reported  `CLIENTNAME` is not longer set. Implementing this will however cause other complications. I can see we may need to record the `CLIENTNAME` which started the session in addition to a dynamic `CLIENTNAME`. This would be needed to make your script above work without a race condition. We'll talk more later.

pp-sepag:
I'm not aware if it is possible to clear an environment variable in a running shell, when the session gets disconnected. Neither to update the env. I dont't think so.
Though the dynamic query using `xrdp-sesadmin` should no longer return a client name, since the associated variable gets cleared on disconnect event: https://github.com/neutrinolabs/xrdp/blame/cd98b01/sesman/ercp_process.c#L129
In my test so far, I didn't notice any issues with race conditions, since I only startet an interactive shell after connecting to the rdp session. But it is a good point one should rule out.
Take your time!

PR #3857, opened 2026-08-26
CVE-2026-79928: Remove rdpsnd_rec extension

matt335672:
This commit removes the implementation of rdpsnd_rec extension of rdpsnd protocol. The reason for this is that the required driver (which only works on Windows 2003) provided by Cendio is deprecated.

See also the commit rdesktop/rdesktop@d4df8c9f20cc85bbc0493410a9cb5c607a84f255 in the rdesktop project, dated 2016-12-05

The extension is not enabled by default, and NOT documented in [MS-RDPEA]. Instead, sound input is supposed to use the protocol in [MS-RDPEAI], which is supported in xrdp. This has been tested by the author.

Despite the extension not being enabled by default, some of its code is still live in a standard installation, and could contain vulnerabilities.

@metalefty, @jsorg71 - are you happy to remove this code?


metalefty (approved):
I'm fine with the removal.

jsorg71:
Sorry, a little late.  Yea, this was the way we did microphone support before we had rdpaudioin(MS-RDPEAI).

PR #3862, opened 2026-09-04
login screen clipboard paste

### LINKED ISSUE #816: Cannot paste password into Xrdp login screen
Ubuntu 16.04 / xrdp latest built from GitHub

If I provide user/password in rdp client, logs in fine.  If I provide user but not password, I get the "XRDP login screen"

I can't paste my password into the password text box -- I do CTRL-V and it shows only one * character.  I have to type the password out.  This seems like a bug?

thuvh:
fix #816


matt335672:
Thanks @thuvh. This would indeed be a useful feature.

You'll need some patience I'm afraid. We're a bit snowed under with security advisories at the moment so it may be a while before we can accept new contributions.

PR #3692, opened 2025-12-28
Add Apple Remote Desktop (ARD) authentication support for macOS

Cyclic:
## Summary

- Implements Apple Remote Desktop (ARD) authentication (Security Type 30) for VNC connections to macOS Screen Sharing
- Adds Diffie-Hellman key exchange with AES-128-ECB credential encryption
- Includes macOS installer packaging (.pkg) with launchd auto-start configuration
- Provides documentation and helper scripts for TCC permissions setup

### ARD Authentication
- Full DH-1024 key exchange implementation in `ssl_calls.c`
- Supports both OpenSSL 1.x and 3.x APIs
- Handles proper credential encoding (64-byte username + 64-byte password blocks)

### macOS Packaging
- Creates ARM and Intel installer packages via `build-pkg.sh`
- launchd plist files for automatic service startup
- Pre/post install scripts for configuration

### Documentation
- Complete setup guide in `docs/macos/README.md`
- TCC permission fix script for Screen Recording access
- VNC test script for connection diagnostics

## Test Plan

- [ ] Build on macOS ARM (Apple Silicon)
- [ ] Build on macOS Intel
- [ ] Test ARD authentication against macOS Screen Sharing
- [ ] Verify screen capture works after TCC permissions granted
- [ ] Test installer package installation and service auto-start

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3864, opened 2026-09-04
chansrv: Always answer a client clipboard data request

### LINKED ISSUE #3863: chansrv leaves a CB_FORMAT_DATA_REQUEST unanswered when the X selection owner does not reply
### xrdp version

0.10.80

~~~
xrdp 0.10.80
  Copyright (C) 2004-2025 Jay Sorg, Neutrino Labs, and all contributors.

  Configure options:
      --with-systemdsystemunitdir=/usr/lib/systemd/system
      --enable-ibus
      --enable-jpeg
      --enable-fuse
      --enable-mp3lame
      --enable-fdkaac
      --enable-opus
      --enable-rfxcodec
      --enable-painter
      --enable-pixman
      --enable-utmp
      -with-imlib2
      --with-freetype2
      --enable-vsock

  Compiled with OpenSSL 3.0.13 30 Jan 2024
~~~

Ubuntu 24.04.4 LTS (Kubuntu). Also reproduced on Ubuntu 26.04.

### Installation method

git clone & make install

### Which backend do you use?

xorgxrdp

KDE Plasma

Physical machine for the main setup; a VirtualBox VM for the second one

### What's your client?

Remmina 1.4.43 with libfreerdp 3.30.0, on Manjaro

### Area(s) with issue?

Clipboard

### Steps to reproduce

1. Connect to the session with a client that waits for a `CB_FORMAT_DATA_RESPONSE` and does not recover if it never arrives. Remmina 1.4.43 is one such client.
2. Arrange for the X selection owner in the session not to answer a conversion. Any of these does it:
   - the owning application exits between the format list and the paste. On Plasma, Klipper takes over the selection and masks this, so a session without a clipboard manager is needed;
   - the owner is itself a remote desktop client fetching the data over its own connection, i.e. RDP inside RDP;
   - a process that still owns the selection but is stopped:
     ~~~
     echo test | xclip -selection clipboard
     sleep 1
     kill -STOP "$(pgrep -nx xclip)"
     ~~~
3. Paste on the client. Nothing arrives, which is expected — the conversion genuinely failed.
4. Restore a healthy clipboard: copy something in an ordinary application in the session, then paste on the client again.

### ✔️ Expected Behavior

Step 4 pastes the new text. A single failed conversion should cost one paste, not the whole direction.

### ❌ Actual Behavior

Step 4 pastes nothing, and neither does any later attempt. Server to client clipboard stays dead for the rest of the connection, while client to server keeps working. Reconnecting is the only recovery.

On the client side this shows up as one timeout followed by an unbounded run of "Cannot paste now":

~~~
14:10:07  Clipboard data from the server is not available in 6 seconds.
14:10:07  Cannot paste now, I'm already transferring clipboard data from server.
14:17:37  Cannot paste now, ...
...
22:50:26  Cannot paste now, ...
~~~

One timeout at 14:10, then 8.5 hours of a dead direction on the same connection and the same PID, with no further request ever being sent.

### Anything else?

`chansrv` answers a client `CB_FORMAT_DATA_REQUEST` asynchronously: it calls `XConvertSelection()` on whoever owns the X `CLIPBOARD` selection, and the matching `CB_FORMAT_DATA_RESPONSE` is only sent later, from `clipboard_event_selection_notify()`.

If that conversion never completes, no response is ever sent and the client is left waiting indefinitely. `clipboard_send_data_response_failed()` already exists and is used for an unknown format id, so answering a failure is clearly the intended behaviour — these paths just miss it.

#### Paths that produce no response

In `sesman/chansrv/clipboard.c`, `clipboard_event_selection_notify()`:

| Condition | Result |
|---|---|
| `property == None` (owner refused) | `rv = 1`, falls through, nothing sent |
| `clipboard_get_window_property()` fails | early `return 0` |
| target `XA_STRING`, `data_size == 0` | inner guard skipped |
| target `image/bmp`, `data_size <= 14` | inner guard skipped |
| `restrict_outbound_clipboard` set | logged, nothing sent |
| unhandled target | "unknown target", nothing sent |
| unhandled selection | "unknown selection", nothing sent |

Plus: an owner that never answers at all produces no `SelectionNotify`, and there is no timeout anywhere in this path.

That last case is easy to hit in practice: the owning window has already gone away, or the owner is itself a remote-desktop client fetching the data over its own network connection (RDP inside RDP).

#### Impact

Clients do not necessarily recover. Remmina 1.4.43 gives up after 6 seconds but leaves `rfClipboard.srv_clip_data_wait` at `SCDW_BUSY_WAIT`; the reset to `SCDW_NONE` only happens when a `CB_FORMAT_DATA_RESPONSE` actually arrives (`plugins/rdp/rdp_cliprdr.c:607`). Every later paste then returns early at "Cannot paste now, I'm already transferring clipboard data from server", so **server to client clipboard stays dead for the rest of the connection** while client to server keeps working.

That is a Remmina bug too, and I intend to report it there. But xrdp is the side that stops replying, and fixing it in xrdp helps every client.

#### Suggested fix

Track the pending request and always answer it: fail it when the conversion cannot complete, and arm a timeout so an owner that never replies cannot leave the client waiting. The timeout has to be shorter than the client's own, so the client sees a real `CB_RESPONSE_FAIL` instead of hitting its own timeout.

I have a patch and will open a PR.

The symptom pattern reported in #2596 looks like the same thing, although that report is with mstsc and I have not checked what that client does with an unanswered request. #2763 is the same family on the X side.

richardwellgood-byte:
Fixes #3863

`chansrv` answers a `CB_FORMAT_DATA_REQUEST` asynchronously and, on several paths, never sends the `CB_FORMAT_DATA_RESPONSE` at all. The client is then left waiting forever, and some clients never recover: they disable the server to client clipboard for the rest of the connection. Details in the issue.

### What this does

- adds `in_request` / `request_serial` / `incr_rearm_count` to `struct clip_s2c`
- `clipboard_arm_request()` marks a request pending and arms a timeout using the existing `add_timeout()` machinery from `chansrv.c`
- `clipboard_fail_pending_request()` sends `CB_RESPONSE_FAIL` and logs the specific reason at `LOG_LEVEL_WARNING`
- the pending flag is cleared wherever a response actually goes out
- a catch-all at the end of `clipboard_event_selection_notify()` covers the refused / no-usable-data / unsupported-target / restricted cases
- an INCR transfer that is still making progress extends the deadline instead of being cut short; the extension is bounded, so a stalled INCR still fails

`CLIP_S2C_REQUEST_TIMEOUT_MS` is 3000. It has to stay below the client's own timeout; Remmina uses 6 seconds. I am happy to make it a `sesman.ini` setting if you would prefer that.

Target names in the log messages come from a small local table rather than `get_atom_text()`, because that function rejects atoms above 512 and most interned atoms are above that.

### Field results

Kubuntu 24.04.4 server, xrdp 0.10.80 devel, Xorg backend, Remmina 1.4.43 client. One full working day of normal use.

Two failed conversions all day, both recovered:

| Time | Side | Event |
|---|---|---|
| ~14:38:03 | client | request 1 starts |
| 14:38:06 | server | `owner refused to convert the selection to UTF8_STRING`, FAIL sent immediately |
| 14:38:06 | client | 4x "Cannot paste now" (overlapping paste attempts) |
| 14:38:11 | client | **request 2 starts** — proves the state was reset by the FAIL |
| 14:38:14 | server | `the X selection owner did not respond`, timeout fired at +3s, FAIL sent |
| 14:38:17 | client | its own 6 second timeout |
| after | — | quiet, clipboard keeps working |

The important line is 14:38:11: before this patch a new request was never sent after a failure. Compare with the 8.5 hour dead period quoted in the issue.

Cost for the day: one lost paste. The direction never died once.

Note that in the second case the 3 second timeout did not beat the client's 6 second one — the client took about 3 seconds to process the response. It recovers regardless, because Remmina resets its state unconditionally when a response arrives, whether or not it already gave up.

### Testing

| Role | System | xrdp |
|---|---|---|
| client | Manjaro, Remmina 1.4.43, libfreerdp 3.30.0 | — |
| server | Kubuntu 24.04.4, daily driver, several days of use | 0.10.80 devel, patched |
| server | Kubuntu 26.04 VM, Plasma 6 | 0.10.80 devel, patched |
| nested server | Debian 10 terminal server, reached through the above | 0.10.2, unpatched |

The nested case is the interesting one: the X selection owner on the middle machine is itself an RDP client fetching data over its own connection, with its own 6 second timeout. That is exactly the situation this patch has to handle, and it now behaves.

Also:

- reproduced the original failure and confirmed the direction no longer dies
- `astyle 3.4.14` with `astyle_config.as` reformats nothing
- `make check` passes; note it does not cover `sesman/chansrv`, so it is a regression guard here rather than a test of this change

### Relationship to #3805

#3805 touches the same function and would conflict textually. The two changes
are orthogonal: #3805 improves which text targets are negotiated, this one guarantees that a request is answered at all. #3805 adds no response guarantee and no timeout. If #3805 lands first I am happy to rebase; note that it adds new response emitters (`_for_text_bytes`, `_for_locale`) which would also need to clear the pending flag.

### Disclosure

This fix was developed with the help of Claude Code: the analysis, the patch and this description were produced with it. The diagnosis was then checked by hand against the sources of xrdp, Remmina and FreeRDP, and the result has been running on my own machines for several days in the configurations above.

I noticed that @matt335672 raised the question of a policy for AI-assisted contributions in #3692 and that there is not one yet. I would rather say this up front than have it surface later. If you would prefer not to take contributions produced this way, or would rather wait until you have a policy, please just say so and I will close the PR — the issue stands on its own either way.

### Thank you

Separately from all of the above: thank you for xrdp! I use it every day and it has been dependable for years. So I wanted to say that the work you put into keeping this project going is appreciated for me.

### DIFF
--- a/sesman/chansrv/clipboard.c
+++ b/sesman/chansrv/clipboard.c
@@ -224,6 +224,16 @@ static XSelectionRequestEvent g_saved_selection_req_event;
 /* xserver maximum request size in bytes */
 static int g_incr_max_req_size = 0;
 
+/* How long to wait for the X selection owner to answer a conversion started
+   for a client CB_FORMAT_DATA_REQUEST. This has to stay well below the
+   timeout used by the client, otherwise the client gives up first. Remmina
+   waits 6 seconds and does not recover from its own timeout. */
+#define CLIP_S2C_REQUEST_TIMEOUT_MS 3000
+
+/* How many times that timeout may be extended while an INCR transfer is
+   still running, so a large paste is not cut short. */
+#define CLIP_S2C_MAX_INCR_REARM 10
+
 /* server to client, pasting from linux app to mstsc */
 struct clip_s2c g_clip_s2c;
 /* client to server, pasting from mstsc to linux app */
@@ -726,6 +736,7 @@ clipboard_send_data_response_for_image(const char *data, int data_size)
 
     LOG_DEVEL(LOG_LEVEL_DEBUG, "clipboard_send_data_response_for_image: data_size %d",
               data_size);
+    g_clip_s2c.in_request = 0;
     make_stream(s);
     init_stream(s, 64 + data_size);
     out_uint16_le(s, CB_FORMAT_DATA_RESPONSE); /* 5 CLIPRDR_DATA_RESPONSE */
@@ -752,6 +763,7 @@ clipboard_send_data_response_for_text(const char *data, int data_size)
     LOG_DEVEL(LOG_LEVEL_DEBUG, "clipboard_send_data_response_for_text: data_size %d",
               data_size);
     LOG_DEVEL_HEXDUMP(LOG_LEVEL_TRACE, "clipboard send data response:", data, data_size);
+    g_clip_s2c.in_request = 0;
     num_words = utf8_as_utf16_word_count(data, data_size);
     LOG_DEVEL(LOG_LEVEL_DEBUG, "clipboard_send_data_response_for_text: data_size %d "
               "num_words %d", data_size, num_words);
@@ -1027,6 +1039,7 @@ clipboard_send_data_response_failed(void)
     int rv;
 
     LOG_DEVEL(LOG_LEVEL_ERROR, "clipboard_send_data_response_failed:");
+    g_clip_s2c.in_request = 0;
     make_stream(s);
     init_stream(s, 64);
     out_uint16_le(s, CB_FORMAT_DATA_RESPONSE); /* 5 CLIPRDR_DATA_RESPONSE */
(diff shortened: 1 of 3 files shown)


matt335672:
Thanks @richardwellgood-byte 

You'll need some patience I'm afraid. We're a bit snowed under with security advisories at the moment so it may be a while before we can accept new contributions.

richardwellgood-byte:
Hello, matt335672!
It's ok, I'm willing to wait as long as it takes for you to take the time to review this and provide feedback!

I've encountered this problem too often, and now I'm just happy that this PR has solved it!

PR #3873, opened 2026-09-08
instfiles: refine FreeBSD specific RC script

metalefty:
- Generate snakeoil certificate on xrdp daemon start
- Set the proper ownership and permissions after generation

This change address the duplicate host key issue when creating a VM image with xrdp installed. The keys are be generated on the daemon's first startup.

Requested by:   Colin Percival <cperciva@FreeBSD.org>

### DIFF
--- a/instfiles/Makefile.am
+++ b/instfiles/Makefile.am
@@ -105,6 +105,7 @@ endif
 if FREEBSD
 	sed -e 's|%%PREFIX%%|$(prefix)|g' \
 	    -e 's|%%LOCALSTATEDIR%%|$(localstatedir)|g' \
+	    -e 's|%%SYSCONFSUBDIR%%|$(sysconfsubdir)|g' \
 		-i '' \
 		$(DESTDIR)$(sysconfdir)/rc.d/xrdp \
 		$(DESTDIR)$(sysconfdir)/rc.d/xrdp-sesman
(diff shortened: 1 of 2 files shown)


matt335672:
LGTM

I noticed the failing CI tests last night. At the moment it's not clear to me what's going on. I'll investigate further.

metalefty:
BTW, the CI failure looks identical to the issue as a guy reported on the Google Groups.

We also need to bump to FreeBSD 15.1 as 15.0 will reach EoL.

matt335672:
I can't reproduce the fail on my local FreeBSD 15.1 machine.

However, I've noted that the version of imlib2 is different, and imlib2 is performing the scaling for the failing tests.

My version is 1.12.6_1,2 (works)
The CI version for both platforms is 1.12.7,2 (fails)

I don't know why my local VM is using an older version. Different mirror maybe?

metalefty:
I think it is caused quarterly pkg repo vs latest repo. I remember CI image refers latest.

metalefty:
Let's address CI issue separately, merging.

matt335672:
I'll try running the CI on 15.1-STABLE and see what happens.

matt335672:
This happens on 15.1-STABLE too. I'll see what I can find.

matt335672:
See #3880 for a fix.

PR #3878, opened 2026-09-14
Invalidate stale pointer bookkeeping on cache reset

### LINKED ISSUE #3877: Client disconnect after dynamic-resolution resize: cached-pointer update references slot discarded by xrdp_cache_reset()
### xrdp version

0.10.6.1 (also present in devel)

Debian 13 package 0.10.6.1; also reproduced with a local build of the v0.10.6.1 release tarball (`./configure --sysconfdir=/etc`). Code paths cited below verified identical in current `devel`.

Debian 13 (server), Arch Linux client

### Connection type

TCP (FreeRDP 3.31.1 client, `/dynamic-resolution /gfx:AVC444`)

### What happened?

After a dynamic-resolution resize, xrdp can send a fastpath **Cached Pointer** update (`TS_FP_CACHEDPOINTERATTRIBUTE`) referencing a pointer-cache slot that xrdp itself has just discarded in `xrdp_cache_reset()` — a slot for which no pointer data has been re-sent since the reset. Clients that drop their pointer cache on the deactivation-reactivation sequence treat the resulting empty-slot lookup as fatal and disconnect. FreeRDP does exactly this since 3.31.0 (FreeRDP/FreeRDP#13196 recreates all caches on reactivation), so every xrdp resize became a client-disconnect roulette:

```
[ERROR][com.freerdp.core.fastpath] - fastpath_recv_update: Fastpath update Cached Pointer [a] failed
[INFO][com.freerdp.client.common] - client_auto_reconnect_ex: Network disconnect!
```

FreeRDP 3.32 will make the client tolerant again (FreeRDP/FreeRDP#13317), but xrdp's underlying behaviour is a protocol violation regardless of client: it references a cache entry it has itself already forgotten and never re-uploaded.

### Mechanism (verified in source, v0.10.6.1 and devel)

On a dynamic resize via the deactivation-reactivation path (`process_display_control_monitor_layout_data`, `xrdp_mm.c`):

1. `WMRZ_XRDP_CORE_RESET` → `libxrdp_reset()` + `xrdp_cache_reset()`. The latter `g_memset`s the whole `struct xrdp_cache` to zero — all `pointer_items[]` are discarded (`xrdp_cache.c`).
2. **But nothing resets `wm->current_pointer` or `wm->screen->pointer`** — they keep pre-reset slot indices (only ever assigned in `xrdp_cache.c` add-pointer paths and `xrdp_wm.c` mouse handling).
3. `WMRZ_XRDP_CORE_RESET_PROCESSED` → `xrdp_wm_load_static_pointers()` re-uploads only slots 0/1 (and leaves `wm->current_pointer == 1`).
4. First mouse move after reactivation (`xrdp_wm_mouse_move`, `xrdp_wm.c`): `screen->pointer` (stale, ≥2) `!= current_pointer` → `xrdp_wm_set_pointer()` → `libxrdp_set_pointer()` → **cached-pointer update for a slot that is empty on both sides**.

Whether a given resize crashes the client is a race: if the backend module happens to re-announce its cursor (`server_set_pointer` → `xrdp_cache_add_pointer`, which re-uploads and re-syncs the bookkeeping) before the next mouse move, that resize survives. Cursor-rich desktops (KDE Plasma) hit the bad ordering far more often than static-cursor ones (XFCE).

Note devel's #3755 removes the deactivation-reactivation from the **GFX** resize path, so there the trigger is gone — but the stale bookkeeping remains reachable via the non-GFX resize path, and 0.10.x uses the reactivation path for all resizes.

### Repro

FreeRDP 3.31.x client, `/dynamic-resolution /gfx:AVC444`, X11 backend. Connect to a **cursor-rich desktop (KDE Plasma)** — it reproduces within ~2s of connect, as the initial dynamic-resolution resize (deactivation-reactivation) is immediately followed by pointer activity referencing pre-reset per-window slots. A sparse-cursor desktop (XFCE) survives hundreds of scripted resizes, which is the tell that this is about *which* cursor slots are live at reset time, not the resize itself. Windows RDP servers never trigger it (they resize without reactivation), which is what makes affected clients blame the crash on the network.

### Proposed fix

PR attached: invalidate the WM pointer bookkeeping (`wm->current_pointer`, `wm->screen->pointer`) at the end of `xrdp_cache_reset()`, so no cached-pointer update can reference a slot the reset just discarded; the next module pointer update re-adds and re-uploads the cursor as a new pointer. Tested with a local build of 0.10.6.1 against a KDE guest: the unpatched binary disconnects FreeRDP 3.31.1 within seconds; the patched binary survives a continuous resize storm. Minor cosmetic trade-off on the non-GFX resize path: the cursor may briefly show the default arrow until the module next announces a cursor.

jayizcoding:
Fixes #3877.

## Problem

`xrdp_cache_reset()` (called on the deactivation-reactivation resize path, `WMRZ_XRDP_CORE_RESET`) `g_memset`s the whole `struct xrdp_cache` to zero, discarding every cached pointer. But it does **not** reset `wm->current_pointer` or `wm->screen->pointer`, which retain their pre-reset slot indices. After the reactivation, the next pointer change (`xrdp_wm_mouse_move` → `xrdp_wm_set_pointer` → `libxrdp_set_pointer`) can therefore emit a `TS_FP_CACHEDPOINTERATTRIBUTE` for a slot that is now empty on both sides and was never re-sent.

Clients that drop their own pointer cache on reactivation treat the empty-slot lookup as fatal. FreeRDP does this since 3.31.0 (FreeRDP/FreeRDP#13196 recreates all caches on reactivation), so the client aborts:

```
[ERROR] fastpath_recv_update: Fastpath update Cached Pointer [a] failed, status 0
[ERROR] fastpath_recv_update_data: fastpath_recv_update() - -1
```

Windows RDP servers resize without a deactivation-reactivation, so they never hit this — which is why the crash looks client- or network-side until you diff the two servers. The failure is a race (it needs a pointer change referencing a pre-reset slot before the module re-announces its cursor), so it is intermittent and far more frequent on cursor-rich desktops (KDE Plasma) than sparse ones (XFCE).

## Fix

Reset `wm->current_pointer` and `wm->screen->pointer` to 0 at the end of `xrdp_cache_reset()`. Slot 0 is always repopulated immediately afterwards by `xrdp_wm_load_static_pointers()`, so it is a safe, always-valid target; the next module pointer update re-adds and re-uploads the real cursor as a new pointer. Minor cosmetic effect on the non-GFX resize path only: the cursor may briefly show the default arrow until the module next announces a cursor.

## Testing

Local build of the 0.10.6.1 release against a KDE Plasma guest, FreeRDP 3.31.1 client with `/dynamic-resolution /gfx:AVC444`, driven through a continuous resize/move storm:

| | fatal `fastpath_recv_update` failures | outcome |
|---|---|---|
| unpatched | 2 | client disconnects, dies after ~135 resize events |
| patched | 0 | survives 184+ resize events |

Same guest, same client, same harness — the xrdp binary is the only variable.

## Note on `devel`

devel's #3755 removes the deactivation-reactivation from the **GFX** resize path, so the trigger is gone there for GFX sessions. This fix still applies: the stale bookkeeping remains reachable via the non-GFX resize path, and it is the correct invariant for `xrdp_cache_reset()` to maintain regardless of which resize path calls it. 0.10.x uses the reactivation path for all resizes, so a backport would resolve the disconnect there too.


matt335672:
Thanks for your contribution, @jayizcoding 

I'll manage the backport here. We're looking into the CI fail.

PR #3883, opened 2026-09-21
Send the Set Keyboard Indicators PDU when the session's lock keys change

MAButz:
xrdp learns the client's lock key state from the Synchronize event and applies it to the session, but never tells the client when the session's own state moves. The keyboard in front of the user then shows one thing and the session does another, and the only way back is to press the key twice.

This adds `libxrdp_send_set_keyboard_indicators()` ([MS-RDPBCGR] 2.2.8.2.1.1), a `server_set_keyboard_indicators` callback, and order 67 in xup, so a module can report the session's lock state and have it reach the client. It is the other half of neutrinolabs/xorgxrdp#436 — on its own this changes nothing, since nothing calls it yet.

Two details cost me an afternoon and are worth recording:

**The PDU is sent uncompressed.** In a GFX session the whole picture travels over the graphics channel, so a slow path data PDU like this one may well be the first thing the MPPC compressor is ever asked to handle. Compressed, mstsc.exe lit the right lamp and *then* dropped the connection; uncompressed it is happy. The monitor layout PDU is sent uncompressed for a related reason.

**Nothing is sent while a resize is queued or running.** A session hand-over renegotiates the whole connection, and a data PDU sent into that window makes mstsc.exe hang up. It is dropped rather than queued: the client reported its lock state when it connected, so its lamps are already right, and the next change carries the state again.

Tested on Debian 13 against mstsc.exe (Windows 11) and xfreerdp 3.31, both over H.264/GFX.


matt335672:
Thanks @MAButz  - that looks very useful.

I'm currently trying to merge quite a few security fixes before the next release. I'm going to hold off on this one until I've managed that, as we've had to change a large number of files to do so.

